Issue #23, opened 2026-08-22
prd43 w4: every count a document states is derived from the thing it counts


ciaran-slow:
## Fence (may touch ONLY)
- `README.md`
- `SECURITY.md`
- `packages/server/src/server/mutation-guard.ts`
- `packages/server/src/api/route-class-law.test.ts`
- `.github/workflows/ci.yml`

## Blocked by
**Blocked on prd-29's ruling** about the four permanently-tokenless `/api` reads. `SECURITY.md`'s sentence about them cannot be written until that is decided — filing the wording now would queue work on a decision nobody has made.

**This is the ONLY issue in wave 4 that may claim `README.md`** (see `.swarm/coupling.txt`). The install identity and the CLI list are owned by prd43 w3; the lab-session location by prd43 w5.

**Model:** sonnet. **Wave:** 4.

## Definition of done
- `route-class-law.test.ts` greps `SECURITY.md`, `README.md` and `mutation-guard.ts` for the route counts they state and compares them to `ROUTE_CLASSES`; a mismatch is a red build.
- `README.md:99`'s "three hands" becomes four — ADR-0019 added the concierge.
- `SECURITY.md`'s "these two classes" becomes four: ADR-0024 added `gated-read` and prd-29 wave 1 moved seven reads behind the token. `POST /api/retarget` stops being uncounted.
- The concierge's permitted directories (`~/rhizomorph/repos`, `~/.claude/projects`) are named in `SECURITY.md`.
- `README.md:297`'s support matrix says what CI actually proves about macOS, and a law asserts the matrix and `.github/workflows/ci.yml` agree in BOTH directions.
- `README.md:288-289`'s `fetch(` recipe states the real count (8 across 6 modules) with the loopback argument, and the suite runs the recipe and asserts the stated result.

KelliherL:
**The prd-29 blocker is resolved** — ruled 2026-08-24, landed on main in #28 ([the amendment](https://github.com/launchpad-26/rhizomorph/blob/main/docs/prds/prd-29-the-identity-seam.md#amendment--the-seam-is-ruled-operator-2026-08-24)).

What this issue's `SECURITY.md` sentence can now say, honestly and in stages:

[...]

Issue #18, opened 2026-08-22
prd43 w2: the roadmap says the lab has a UI


ciaran-slow:
## Fence (may touch ONLY)
- `docs/roadmap.md`

## Blocked by
Blocked by prd43 w1.

Sibling lanes own `docs/architecture.md`, `docs/design-notes/` and the judge sources.

**Model:** sonnet. **Wave:** 2.

## Definition of done
- `:88`'s claim — *"Engine only; nothing in `packages/web/` reaches it yet"* — is corrected: `packages/web/src/lab/` holds 32 tracked files including `LabPage.tsx`.
- The prd-14 pointer stays accurate about what the console still owes.

ciaran-slow:
Done by 062057b on #25. Both DoD bullets are satisfied: docs/roadmap.md:88's claim ("Engine only; nothing in packages/web/ reaches it yet") is corrected to state that the engine is what prd-12 shipped and its UI is prd14's and is live — packages/web/src/lab/ is 32 tracked files, reached from App.tsx:129 — and the prd14 entry eight lines below is untouched, so it still says what the console owes (comparison artefacts not saved into prd16's library; timeline, spend-cap and fold-order rulings open).

Fixed there rather than here because #25 rewrites 125 lines of the same file under the heading of reconciling the paper against the codebase; leaving a known-false claim inside the file being reconciled was the wrong place to draw that boundary. The claim was also contradicting the prd14 entry in the same document.

Not a path claim, so prd43 w1's citation law (#16) would never have caught it — this was always a manual prose fix, which is why closing it does not weaken the wave-1 keystone. prd-43's Evidence bullet was updated in the same commit to record the correction and retain the shape.

PR #25, opened 2026-08-22
docs(prds): bless PRD 39-43 and reconcile the backlog

ciaran-slow:
## Summary

- bless PRD 39 through PRD 43 and add them to the roadmap
- reconcile the older PRD corpus against the current codebase and open GitHub backlog
- move nine completed PRDs to `done/` and two gated PRDs to `parked/`
- document code-backed recommendations for the eleven retained PRDs
- align the PRD index, roadmap, UI progress notes, and the direct PRD-32 test reader

This is a paper and backlog reconciliation PR. It does not close the implementation issues created from PRD 39 through PRD 43.

## Verification

- `npm run typecheck`
- `npm run lint`
- `VITEST_MAX_WORKERS=6 npm test` — 6,356 passed; 3 expected failures
- staged diff whitespace check
- all 35 local links in the retained-PRD review resolve

## What is the sibling case?

The sibling case is any index, roadmap entry, design note, test, or other code that refers to a PRD by its old root path. The reconciliation searched those consumers together; the PRD-32 window-frame reader was the direct code sibling that needed updating when the document moved to `done/`.

The PRD-32 reader can survive wording changes outside its S5 acceptance section, which is intentional. It will not survive deleting or moving the document without updating the reader, removing S5, or changing the declared floor while leaving the implementation expectation stale. The link audit likewise detects a moved document whose local references were not updated.


gabriel-canaan (changes requested):
Reviewed commit-by-commit at `eabf9ed`, in a clean worktree, against `origin/main` (`03df141`). Findings labelled **EXECUTED** (I ran it and read the exit status) or **REASONED** (argued from the code, nothing run).

## What verified in this PR's favour

Nearly everything, and several claims are unusually precise. All **EXECUTED**:

- **The suite reproduces the body's numbers exactly.** `VITEST_MAX_WORKERS=6 npm test` → `375 passed (375)` files, `Tests 6356 passed | 3 expected fail (6359)`, exit 0. The "3 expected failures" are vitest's own `expected fail` category, not unexplained reds. `npm run typecheck` → 0. `npm run lint` → 0, 841 files.
- **"All 35 local links in the retained-PRD review resolve"** — exactly 35 local links, 0 broken.
- **The 11 file moves left nothing dangling.** A repo-wide sweep for every moved basename found no reference to an old `docs/prds/prd-NN-*.md` path outside `done/`+`parked/`. A link-check over all 291 local markdown links found 0 genuine breaks (the 38 hits are pre-existing GitHub-relative `../../issues/N` links that resolve on github.com).
- **The one code change is load-bearing, proven by mutation.** Reverting `window-frame.test.ts:43` to the pre-move path makes the file fail to load (ENOENT at module scope). The sibling was real and the PR caught it.
- **`docs/roadmap.md:88` is byte-identical to `main`.** The rewrite begins at line 93, so issue #18's `:88` citation survives this PR intact. That is careful scoping, not luck.
- **Board claims exact:** 24 open issues, prd39=2 / prd40=3 / prd41=5 / prd42=5 / prd43=9. Issue #23 does carry an explicit "Blocked on prd-29's ruling".
- **Counts spot-checked and correct:** 7 `gated-read` routes; 7 tokenless API reads (with `GET /*` named as the deliberate exclusion); 5 `provided` + 1 `partial` pi signals; 32 tracked files under `packages/web/src/lab`; 4 source files citing the deleted judge research note; 4 links to the non-existent `docs/decisions/` across 3 design-notes; `export-otlp` in the CLI and absent from `README.md`.
- **prd-39's premise is real.** `gate.sh:174` prints `!! BUILD BROKEN ON MAIN` and `gate.sh:181` runs `git push origin main` unconditionally; `gate.sh:62`'s NUL guard is `python3 … || echo 0`, so a missing `python3` yields a clean verdict. Every line citation in prd-39 (`:10`, `:25`, `:62`, `:104`, `:169`, `:170-175`, `:181`) resolves, and the PRD already notes that `AGENTS.md:279-280`'s `gate.sh:172`/`:177` are stale — they are.
- **prd-42's evidence is exact**, including its negative claim: `server/static.ts:119` does compare `startsWith(root + path.sep)` and does not import `paths/containment.ts`.
- **`scripts/dev/coupling.test.sh` passes with the two new entries** — 11 passed, 12 entries parsed, every path exists. (This one is not in the vitest suite and not in CI, so nobody would otherwise have run it.)
- No home paths, no trailing whitespace in any changed file.

## Findings

### 1. `prd-33`'s raised cap is an event cap, not a lane cap — **EXECUTED**

The amendment this PR adds to `docs/prds/prd-33-the-living-scene.md` is exact: ruling 10 supersedes criterion 1's blanket *"no older cap has moved"* clause by raising `EVENT.maxConcurrent` from 5 to 7. `packages/web/src/scene/motion.test.ts:77` and `:215` both assert `EVENT.maxConcurrent === 7`.

`docs/prds/retained-prds-review-2026-08-22.md` transcribes that correctly as *"the later seven-event cap"*. Its two summary siblings did not — both `docs/roadmap.md:196-197` and `docs/prds/reconciliation-2026-08-22.md:53` say **"seven-lane cap"** superseding a **"five-lane"** criterion.

No lane cap exists. prd-33 measures 3 colonies × 30 lanes at 60 fps (`prd-33:264`), and criterion 1 states no number at all (`grep -n 'five-lane\|five lane\|5 lane' docs/prds/prd-33-the-living-scene.md` → nothing). A reader of the roadmap — the repo's headline record — would conclude the observatory caps at seven lanes.

This is the sibling shape the runbook names: the fact was transcribed correctly into one document and wrongly into its two structurally identical neighbours.

### 2. Three citations in the newly blessed evidence point at the wrong lines — **EXECUTED**

Every substantive claim survives; only the pointers are wrong. Worth fixing before merge because prd-43 ruling 1, blessed in this same PR, is the law that a cited path must resolve.

- **prd-41, containment.** The file cites `lab/namespace-law.test.ts:519-527` as calling `forkThreeArms({ install: false })`. Lines 519–527 are a path-classification helper. `forkThreeArms` is defined at `:440-462`, takes **no arguments**, and hard-codes `install: false` at `:459`; five callers inherit it (`:468`, `:477`, `:489`, `:535`, `:559`).
- **prd-41, "both stub `npm`".** The two `install: true` sites are `restore.test.ts:365` and `:387`. The first stubs `exec`; the second uses `exec: realExec` but deletes `package.json` at `:378`, so `installed` returns `false`. Neither runs npm — which is the point — but only one of them stubs it.
- **prd-41, timeouts.** *"`withTimeout` is wired at exactly one site in the tree"* — there are three: `server/poll-loop.ts:91`, `api/retarget.ts:179`, `api/doctor.ts:211`. `grep -rn 'withTimeout(' --exclude-dir=node_modules --include='*.ts' packages`. None is a lab path, which is the bullet's actual point.
- **prd-40, buffer.** `session-recorder.ts:139` is `eventsSoFar()`'s return. The buffer is emptied by `openSession` at `:125`.

Every other citation in prd-40 and prd-41 checks out (`cli/index.ts:225-227`, `compare.ts:79`, `checkpoint.ts:57`, `fork.ts:256`, `restore.ts:273`, `api/lab.ts:390-392` and `:509`, `poll-loop.ts:164`, `:216-218`, `session-recorder.ts:80-84` and `:104-108`, `api/meta.ts:238`, `api/index.ts:132`).

### 3. `prd-36` moving to `done/` has one undeclared gap — **EXECUTED**, and this one is a disposition question, not a typo

prd-36 success criterion 3: *"Switching representations is instant and lossless. **Not met while** selection, scroll or filter state is dropped by a toggle, or the switch costs a re-fetch."*

Selection and re-fetch are genuinely handled and tested — state is lifted to `SelectionProvider`/`FleetProvider` above the toggle, and `TwoRepresentations.tsx`'s own doc explains why that is structural. Scroll is not:

- The list's scroll container is **inside** the representation: `packages/web/src/panels/fleet/index.tsx:150-153`, `className="min-h-0 flex-1 overflow-auto …"`. `TwoRepresentations` renders only the active view's `render()`, so list → organism → list unmounts and remounts it.
- No scroll preservation exists anywhere in `packages/web/src` outside the drawer's autoscroll (`grep -rn 'scrollTop\|scrollRestoration\|scroll position' packages`).
- No test anywhere asserts scroll survives a toggle. The single place the guarantee is quoted — `packages/web/src/recordings/historyAxis.test.tsx:166-170` — quotes *"preserves selection, scroll position intent and any filter"* and then tests the **re-fetch** clause only, by its own title.
- The fleet instance's declared `gap` (`TwoRepresentations.tsx:93-100`) names only the unremembered representation choice, not scroll.

Contrast prd-35, which I also checked and which is **fine**: its watched-repo preference cannot be changed from the UI, but that is the PRD's own *unavailable* state (`prd-35:172-174` — "renders disabled with its reason, never hidden"), it carries its reason at `settings/registry.ts:212-219`, and `settings/SettingsPage.test.tsx:251` pins it. That is a declared gap. prd-36's scroll gap is not declared anywhere.

Smallest honest resolution, in the same shape as every other Outcome line this PR writes:

```diff
-> **Outcome:** shipped.
+> **Outcome:** shipped, with one declared gap — S3's scroll clause. Selection and
+> the no-re-fetch guarantee hold structurally (state lifted above
+> `TwoRepresentations`); scroll does not, because the list's `overflow-auto`
+> container lives inside the representation (`panels/fleet/index.tsx:150-153`)
+> and a toggle remounts it. No test covers it. Reconciled 2026-08-22 at `03df141`.
```

Whether that is enough to keep prd-36 in `done/`, or whether it wants an issue first, is yours and the operator's call — not mine.

## Suggested patch for findings 1 and 2

Two commits, docs-only. `typecheck`, `lint`, `window-frame.test.ts` and the link check all still green after them. I have deliberately **not** pushed to your branch.

```diff
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -193,8 +193,9 @@
   switch whose chrome never impersonates status.
 - **prd33 — the living scene** (`docs/prds/prd-33-the-living-scene.md`, BLESSED 2026-08-08,
   acceptance pending): the scene work and recorded before/after frame measurements ship. The
-  remaining gate is the first-glance operator act. Success criterion 1's five-lane cap is
-  historical and superseded by ruling 10's seven-lane cap.
+  remaining gate is the first-glance operator act. Success criterion 1's blanket "no older cap
+  has moved" clause is superseded by ruling 10's measured `EVENT.maxConcurrent` rise from 5 to 7
+  — an event cap, not a lane cap; the scene is measured at 30 lanes.
```

`docs/prds/reconciliation-2026-08-22.md:53`, same substance:

> Success criterion 1's blanket "no older cap has moved" clause is superseded by ruling 10's measured `EVENT.maxConcurrent` rise from 5 to 7 — an event cap, not a lane cap.

`docs/prds/prd-41-the-laboratory-is-confined-in-fact.md`:

```diff
-- **The containment law cannot see it.** `lab/namespace-law.test.ts:519-527` calls
-  `forkThreeArms({ install: false })`; only two call sites exercise `install: true`, and both stub
-  `npm`. The law asserts a fence the default path never crosses under test.
+- **The containment law cannot see it.** `lab/namespace-law.test.ts:440-462`'s `forkThreeArms`
+  helper hard-codes `install: false` at `:459`, and every one of its five callers inherits it
+  (`:468`, `:477`, `:489`, `:535`, `:559`); the other `install: false` is `:652`. Only two call
+  sites in the tree pass `install: true`, both in `restore.test.ts` — `:365` stubs `exec` so
+  `npm` never really runs, and `:387` deletes `package.json` first (`:378`) so `installed` comes
+  back `false`. The law asserts a fence the default path never crosses under test.
@@
-  `lab/fork.ts:256`, `lab/restore.ts:273` all take `options.exec ?? realExec` raw. `withTimeout`
-  is wired at exactly one site in the tree: `server/poll-loop.ts:91`. `#236`'s fix landed for
-  collectors and stopped there.
+  `lab/fork.ts:256`, `lab/restore.ts:273` all take `options.exec ?? realExec` raw. `withTimeout`
+  is wired at three sites in the tree — `server/poll-loop.ts:91`, `api/retarget.ts:179` and
+  `api/doctor.ts:211` — and not one of them is a lab path. `#236`'s fix landed for collectors and
+  stopped there.
```

`docs/prds/prd-40-the-record-survives-the-write.md`: `session-recorder.ts:139` → `:125`.

## Claims I generated and then threw away

A model seat proposed two further findings; both collapsed on verification, and I am recording them so nobody re-derives them:

- *"prd-23 criterion 1 unmet — mutations proven only by mutual fakes."* **Refuted.** `packages/contract/` is the seam harness: it boots the real `buildApp`, serves the real stamped shell, and drives the real web client, with method/URL/headers/body coming from the web module. All five mutating routes have contract tests, and `contract-coverage-law.test.ts` fails if one is missing — enumerated from `mutating-calls-law.test.ts`'s own source so the two cannot drift. `npx vitest run packages/contract` → 6 files, 26 tests, passed. The seat cited `api/label.test.ts:52`, which is the historical gap the harness's own header names as the thing it exists to close.
- *"prd-35 criterion 1 unmet — the watched repo is CLI-only."* **Refuted**, as above: that is the PRD's declared *unavailable* state, with its reason on screen and a test pinning it.

## What I did not run

`scripts/gate.sh` (the operator's landing step, and it merges + pushes), the macOS and Windows legs, and any browser. CI's own 7 checks are green and I confirmed each job actually executed via `gh pr checks 25`, rather than trusting the bar.

ciaran-slow:
## Review answered — six findings, six fixed, none rebutted

Every finding was re-verified against the tree at `eabf9ed` before being fixed. Five held exactly
as reported; one (4b) is real but pre-existing and untouched by this PR, and is fixed here anyway
for the reason given below. Each fix is its own commit, so each can be reviewed against its
finding independently.

| # | Verdict | Commit | What changed |
|---|---|---|---|
| 1 (scope) | fixed | `c143b72` | Ruling 1 excludes the whole PRD tree, matching prd-43's own non-goal; Success 1 names the in-scope set |
| 1 (citation) | fixed | `1e3fc01` | All five PRDs anchor on `03df141`; the `docs/audit/` `.gitignore` hunk is committed |
| 2 | fixed | `84b895c` | Ruling 1 enumerates all three publishing sites; wave 2 fences both files as one issue |
| 3 | fixed | `7a8e03b` | The coupling record describes the derived timing set and the count ratchet |
| 4a | fixed | `f12a074` | `static.ts` listed as the fifth call site, not a worked example |
| 4b | fixed | `062057b` | `roadmap.md:88` corrected; prd-43's Evidence bullet updated in the same commit |
| 5 | fixed | `1a15d47` | Success 5 falsifies the untruth rather than the print order |
| 6 | fixed | `94b2950` | Ruling 4's sidecar carries a SHA-256 of the image bytes; `visit.mjs` becomes the sole capture path |

---

### 1 — prd-43's citation-law scope, and the missing audit

**EXECUTED, confirmed, both halves.**

The contradiction is real. Non-goals said the law excludes `docs/prds/` "by directory,
deliberately"; Ruling 1 excluded only `docs/prds/done/`. The narrower scope is unshippable, and
the proof is in prd-43's own Evidence section — all four paths it names as dead are in fact
absent:

```
packages/web/src/fleet/buildFleet.ts              MISSING
packages/web/src/panels/spend/format.ts           MISSING
docs/research/2026-08-04-semantic-judge-spike.md  MISSING
docs/decisions                                    MISSING
```

Under Ruling 1 as written, the first run of the law prd-43 commissions would have gone red on
prd-43 and on the four PRDs beside it. That is the "a law that must be suppressed to pass teaches
its readers to suppress it" case prd-43's own rejected-alternatives section already names.
Resolved toward the wider exclusion in `c143b72`. The exclusion-honesty check is unaffected and
still bites: `docs/prds/` holds four paths that would otherwise trip the law, so it cannot go
stale silently. Success 1's falsifier was also over-broad — written across all of `docs/**` it
would have been permanently unmet by construction — and now names the in-scope set.

The citation half was worse than reported. `docs/audit/REFINED-ISSUES.md` exists on the authoring
machine, but `git ls-files docs/audit/` is empty **and** the `.gitignore` hunk excluding it was
sitting uncommitted in a working tree. So on the branch as pushed the directory was not even
ignored, merely untracked — either way the opening evidence line of all five PRDs resolved nowhere
for any other reader.

Fixed by re-anchoring rather than by committing the audit, in `1e3fc01`. The `.gitignore` comment
already made the argument ("a findings list goes stale the moment the findings are fixed, and a
stale one tracked in `docs/` reads like current state"), so that hunk is now committed and the
citations honour it. The finding numbers stay because they index the reasoning; the path goes
because it promises a file. Safe to do because every Evidence section was already self-contained —
each restates what the audit found with its own paths and line numbers, so none of them actually
depended on the file.

### 2 — prd-40's durable-before-live ruling missed both recorder paths

**EXECUTED, confirmed.** `session-recorder.ts:78-84` is `buffer.push` → `emitter.emit('event')` →
`await this.writer.append`. `closeWith` at `:97-111` has the identical order one screen down.

Success 1 is *"an event that reaches a subscriber has reached the log"*, and Ruling 1 moved only
the poll loop's snapshot advance behind the append — but the poll loop does not publish events,
the recorder does. Awaiting `record()` in the loop changes nothing about either site: the
subscriber already holds an event whose append then rejects. Fixing the loop alone would have met
the snapshot half of Success 1 and left the subscriber half broken, which is exactly the half-fix
the criterion exists to forbid.

`84b895c` rewrites Ruling 1 to enumerate all three publishing sites and their order. `closeWith`
still takes its seal **first**, so prd17 ruling 1's structural guarantee is untouched — only the
publish moves, and the existing `catch` that releases the seal is reused as-is. Wave 2 is now
declared as one issue fencing `recorder/session-recorder.ts` *and* `server/poll-loop.ts`, because
two issues could land one half. Two supporting corrections went in with it: Success 1's falsifier
now names publishing rather than only collector state, and the "do not rebuild" line claiming *"the
gap is above it, not in it"* is corrected — the seal handling is right, the publish ordering around
it is half the defect. Worth noting the Evidence bullet had `session-recorder.ts:80-84` pinned
correctly all along; the ruling simply did not follow where its own evidence pointed.

### 3 — the timing coupling record describes a deleted mechanism

**EXECUTED, confirmed.** `gate.sh:100-101` derives the set: every `packages/**` test carrying a
`// @gate-timing` marker, wherever it lives, plus every `*.bench.test.ts` by name. Six files
today, and only three are in `packages/web`:

```
packages/core/src/reduce.bench.test.ts
packages/core/src/selectors/spend-cursor.bench.test.ts
packages/web/src/replay/replayFold.bench.test.ts
packages/web/src/scene/SceneView.test.tsx
packages/web/src/scene/marks.test.ts
packages/web/src/scene/perf.test.ts
```

The entry named two of those six — the literal path list `#209` deleted, and deleted for the
reason the entry then went on to warn about. A coupling record naming two of six is worse than
none: a lane reading it would conclude `packages/core` is safe ground.

It also missed the half that actually costs a landing. `gate.sh:104` fails when the set is
**empty**, which is the case the old text described; `gate.sh:107-110` ratchets against
`.swarm/timing-count` and fails when the count merely **drops**. So removing one marker out of six
breaks the landing while five timing tests still run and pass, and only a human clearing that file
releases it. Corrected in `7a8e03b`, with the line references re-derived from the script rather
than carried over.

### 4a — prd-42's `static.ts` claim

**EXECUTED, confirmed.** `static.ts:119` hand-rolls `requested.startsWith(root + path.sep)` on an
uncanonicalised path and imports nothing from `paths/containment.ts`.

The "what already exists" line contradicted prd-42's own Evidence four sections above, which
states the opposite correctly and in the same words (*"`static.ts` does not import it"*), and
contradicted Ruling 2, which exists precisely to make it import it. Of the three statements the
tree agreed with two; the reassuring one was the outlier. `f12a074` corrects it and names the four
real importers so the claim is checkable rather than trusted: `collectors/pi/collector.ts`,
`collectors/sessionlog/process-probe.ts`, `concierge/paths.ts`, `concierge/migrate.ts`.

### 4b — the roadmap's lab-UI claim

**EXECUTED, confirmed — with one correction to the finding's framing.** `git ls-files
packages/web/src/lab` returns 32 files and `App.tsx:129` renders `LabPage` behind a lazy import,
so the claim is false. But that line is byte-identical on `main` and predates this PR — it is not
a defect this PR introduced, and prd-43's Evidence already books it (citing `:88`, correctly).

Fixed anyway in `062057b`, because this PR rewrites 125 lines of the roadmap under the heading of
reconciling the paper against the codebase, and leaving a known-false claim inside the file being
reconciled is the wrong place to draw that boundary. The line also contradicted the prd14 entry
eight lines below it in the same file, which already describes that console as live.

**Sibling case for this fix:** fixing the claim without touching prd-43 would leave prd-43
asserting a false current state about a document it quotes — the exact failure mode prd-43 exists
to prevent. So the Evidence bullet is updated in the same commit: past tense, marked corrected
here, and retained, because the shape is the finding. Two entries in one file disagreeing about
whether a feature exists is what nothing was checking.

### 5 — prd-41 Success 5 was reversed

**EXECUTED, confirmed.** `cli/index.ts:225-227` prints prd-12 ruling 1's three namespaces at the
end of the no-`--launch` path, after `forkFromCheckpoint` has already run the checkpointed tree's
install hooks outside all three. The criterion read *"not met while any surface states its
confinement **before** the thing it confines has finished running"* — so the current CLI satisfies
it exactly, and would keep satisfying it while lying to the operator.

This is AGENTS.md's mutation question answered in the negative before any test existed: no
mutation of the CLI's print position makes that falsifier bite, because print position was never
the fault. `1a15d47` rewrites it to falsify the claim's truth — not met while any surface names a
fence the work it reports has already crossed — with the reversal called out so nobody restores
it. **No new ruling was needed:** Ruling 1 is what delivers it, since once `--ignore-scripts` stops
the restore executing the tree, the sentence the CLI prints becomes true where it already stands.
The Evidence bullet was reframed for the same reason — it was factually right but led with "after
the hook has already run", which reads as a sequencing complaint and is what the falsifier then
encoded.

### 6 — prd-43's PNG sidecar did not bind a manifest to image bytes

**Accepted as reasoned; not disputed.** Nothing in the ruling tied the manifest to the image, so a
sidecar declaring `/Users/operator` could sit beside a PNG rendered against a real home directory:
the law reads the sidecar, goes green, and the leak ships — because no grep can read a path drawn
into pixels. The manifest was an assertion the law believed, not evidence the law checked.

`94b2950` makes two changes. The sidecar now carries the SHA-256 of the bytes it describes and the
law recomputes it, so a manifest cannot be paired with a different image. That is necessary and
not sufficient, so the second half carries the rest: `scripts/dev/visit.mjs` becomes the only
sanctioned way to produce a tracked PNG, substituting the synthetic root *before* the screenshot
and emitting the sidecar itself. A hand-made capture cannot produce a valid pair, which is the
point — provenance is enforced by the tool that holds the root, not by a human remembering to
declare one. The ruling already said the alternative was "a law that will be true until the first
time it matters"; unbound, it was that law.

---

### Validation of the fixes

Run on the final committed tree (`062057b`), all EXECUTED:

- `npm run typecheck` — pass
- `npm run lint` — pass, 841 files
- `VITEST_MAX_WORKERS=6 npm test` — **375 files passed, 6,356 passed, 3 expected failures**,
  matching this PR's stated baseline exactly
- `git diff --check` on every commit — no whitespace damage
- `grep -rn "docs/audit" docs/prds/ docs/roadmap.md .swarm/` — no matches, so no citation was
  missed in the re-anchor
- Each of the eight commits was reassembled from per-hunk patches and diffed against the verified
  tree to prove the split is lossless

Also re-checked, since a move is the sibling case for this PR's whole subject: no tracked file
still refers to any of the eleven relocated PRDs by its old root path.

### Loose end, not fixed

`.swarm/timing-count` does not exist yet, so the ratchet documented in `7a8e03b` arms on the first
gated landing rather than now. That reads as the intended design from `gate.sh:138`, but whether it
was meant to be seeded is an operator call and no commit here assumes one.

PR #40, opened 2026-08-24
prd39 w1: the gate that holds — a check that prints a fault holds the landing

### LINKED ISSUE #1: prd39 w1: a gate check that prints a fault holds the landing
## Fence (may touch ONLY)
- `scripts/gate.sh`

## Blocked by
Nothing — this is the wave-1 keystone. Wave 0 is the operator's own act: they land this by
hand, without using `gate.sh`, because the gate cannot be trusted to land its own fix.

Sibling lane owns `AGENTS.md` (the runbook's landing section). The `@gate-timing` markers in
`packages/web/src/scene/perf.test.ts` and `SceneView.test.tsx` are owned by prd-24 — `gate.sh:104`
greps them and fails the landing on a zero count, so that marker contract must survive untouched
(now recorded in `.swarm/coupling.txt`).

## Definition of done
- `npm install` after merge routes through `fail()` instead of `|| echo` — a broken install holds the landing.
- The `npm run build` else-branch calls `fail()` after its `tail -6`, so a build-broken merge cannot reach the push.
- Every other branch in the script that prints a fault also exits through `fail()` — the defect is a class, and closing only the two found leaves the next one open.
- The offline-push exemption at the `git push origin main` line survives, moved out of an inline comment into a named block that scopes it to that one line.
- The NUL-byte guard asserts `python3` exists once, up front, and `fail`s if it does not — the same posture the timing pass takes toward its own file count.
- The NUL guard's `|| echo 0` fallback is gone, so an unreadable file fails rather than reporting a clean verdict.
- A rigged failing build, run against a scratch clone, does not reach `git push` — shown, not argued.
- `bash -n scripts/gate.sh` passes and the script's header still describes what it does.

### LINKED ISSUE #2: prd39 w1: the runbook's landing section survives the script moving
## Fence (may touch ONLY)
- `AGENTS.md`

## Blocked by
Nothing. Sibling lane owns `scripts/gate.sh`; this lane never reads its line numbers, which is
the point of the change.

## Definition of done
- The landing section names which single check in `gate.sh` is deliberately non-fatal (the push) and why (an offline operator must still be able to land), so a reader cannot mistake one line's exemption for the script's posture.
- The two stale citations `gate.sh:172` and `gate.sh:177` are replaced by behaviour-and-section references. Verified stale at `03df141`: `:172` is now `else` and the push is at `:181`.
- No line-number citation into `scripts/gate.sh` survives anywhere in the file — the runbook must not rot the next time the script moves.
- The section still tells a lane it must not run the gate, and still explains that running it is the landing.

gabriel-canaan:
prd-39, wave 1. `scripts/gate.sh` is not a pre-push check — running it **is** the
landing: it rebases, audits the fence, runs the suite, merges into local `main`,
rebuilds, and pushes to `origin/main`. Its header promised *"every check exiting
non-zero on failure."* Two checks did not, and a guard reported success it had
not earned. This makes them true.

Closes #1
Closes #2

Two issues, one wave, one PR — the working agreement's bundle. Fence-disjoint by
construction: #1 owns `scripts/gate.sh`, #2 owns `AGENTS.md`, and `scripts/dev/fence-lint.sh 1 2` passed before dispatch.

## What changed

**Ruling 1 — a check that prints a fault exits through `fail()`.** The post-merge
`npm install` and the `npm run build` else-branch now route through the existing
`fail()`. The sweep covers the whole class, including `workmux rebase` at the top
of the script — see the sibling case below. The push remains the one deliberate
exception, moved out of an inline comment into a named `push_or_warn()` block so
a reader can see it is one line's exemption and not the script's posture.

**Ruling 2 — a guard probes its own dependency before claiming to have run.**
`python3` is asserted once, up front. The NUL guard's `|| echo 0` is gone, so an
unreadable file is a hold rather than a clean verdict.

**`fail()` is now state-aware.** A `MERGED` flag flips after the merge
postcondition, so the recovery line is true on both sides of the merge: before
it, *"HOLDING — not merged"*; after it, *"MERGED to local main, NOT pushed — fix
forward on main immediately, then push"*. Both post-merge checks still route
through the **existing** `fail()`, exactly as ruling 1 required, so nothing new
is introduced and no ruling amendment is owed.

**The runbook stops rotting.** `AGENTS.md` cited `gate.sh:172` and `gate.sh:177`;
both were already stale (`:172` was `else`, the push had moved to `:181`). Line
citations into the script are gone entirely — the section now cites behaviour,
and says which single check is deliberately non-fatal and why.

## What is the sibling case?

This is where the PR earned its review. The first version of #1 fixed the two
checks the PRD named and left `workmux rebase "$H" ... || echo "(rebase reported
an issue — continuing to checks)"` untouched, arguing its real failure mode — a
conflict — is verified two lines later by the mid-rebase directory check.

**That argument was half right, and half is the defect ruling 1 forbids.** It
covers the conflict. It does not cover the rebase *never running* (workmux
absent, unknown handle) or aborting cleanly. In those modes no `rebase-merge`
directory exists, the mid-rebase check passes, and the gate runs the fence audit
and the suite against a branch that was never rebased onto current `main` —
measuring a tree that is not the one landing — then merges and pushes it. That
is #408's shape (green on a stale base) reached through a fault-printing branch.

The fix does **not** convert `workmux rebase`'s exit code to `fail()`. That would
reintroduce the false-hold class `db05cf2`'s postmortems record, which is why
this script prefers checking git state over trusting a tool's exit status. It
asserts the **state** instead:

```sh
git -C "$W" merge-base --is-ancestor main HEAD 2>/dev/null || fail "branch $BRANCH is not on top of main (the rebase did not take) — rebase it, then re-run"
```

That closes both uncovered modes, depends on no external tool's contract, and —
the payoff — makes `gate.sh`'s *"every other check above exits through `fail()`"*
and `AGENTS.md`'s *"no other step in the script inherits it"* true **as already
written**. Fixing the code let the runbook keep its stronger claim instead of
hedging it.

Two remaining `|| echo` sites are deliberate and were judged individually: the
push (the named exemption) and `GD=$(... || echo "$W/.git")`, which is a fallback
default, not a fault print — filed as a follow-up rather than changed here.

**All of them — there is no test.** `grep -rln "gate.sh" packages` finds only
`// @gate-timing` marker comments, so nothing in the suite asserts this script's
behaviour, and a green suite is not evidence about it. Said plainly because the
merge would otherwise imply coverage that does not exist. A law test and a
`shellcheck` job are filed as follow-ups; the PRD itself names shellcheck's
absence as unfiled work.

So every claim was proved by running the logic against rigged inputs in scratch
directories — never this worktree, never the real `origin`, and **the gate itself
was never run**, because running it is the landing. All `EXECUTED`, exit statuses
read directly:

| Probe | Before | After |
|---|---|---|
| Rigged failing build | `!! BUILD BROKEN`, `MERGED`, **`PUSHED`**, exit 0 | `GATE FAILED`, exit 1, push never reached |
| `python3` absent | prints `no NUL bytes`, exit 0 | `GATE FAILED`, exit 1 |
| Unreadable file | `pass (clean)` | holds |
| File with a real NUL byte | holds | holds (no vacuity traded) |
| Stale branch + `workmux` exiting 127 | reached the checks, exit 0 | `GATE FAILED`, exit 1 |
| `fail()` after the merge | *"not merged"* (false) | *"MERGED to local main, NOT pushed"* |

The state assertion false-holds nothing: a properly-based branch, a post-rebase
branch, `HEAD == main`, and a fresh zero-commit branch all pass.

## Gates

Both lane worktrees, and again on this merged branch against current `main`
(`25d8647`): `npm ci` 0, `bash -n scripts/gate.sh` 0, typecheck 0, lint 0,
**build** 0, suite 0 — 375 files, 6356 passed, 3 expected fail. Tree clean before
and after every run. Node 22.22.2 per `.nvmrc`.

## Fence widenings

**None.** #1 touched `scripts/gate.sh` only (+38/−11); #2 touched `AGENTS.md`
only (+20/−3). Nothing else, no new check joined the gate, and
`.github/workflows/` was not entered — that is prd-24's.

## Review provenance

Lanes built by Claude Sonnet 5. Reviewed by a multi-strategy pass — Opus and
Codex `gpt-5.6-terra` initial reviews at xhigh (neither seeing the other), both
cross-reviews, reconciled on Fable — then fixed by Claude Opus 5. Full findings,
including the ones that collapsed on verification, are on #1 and #2.

Worth knowing for the landing: the review also surfaced **eleven pre-existing
defects** in `gate.sh` that are deliberately **not** in this PR, because the
PRD's non-goal is explicitly *"not a rewrite of the gate."* Eight of them touch
`scripts/gate.sh`, so they are fenced behind this PR and cannot start until it
lands. Titles are on #1. The most operationally reachable is an invalid fence
regex: `FENCE='['` makes `grep` error, `|| true` swallows it, and the audit
prints `fence OK` while out-of-fence work lands — reachable by a typo in an
argument that is hand-typed on every landing.


ciaran-slow (approved):
## Approve — landing held for the operator's manual merge

Read commit-by-commit (`786f7d1`, `f0ca8ef`) against `#1` and `#2`. Every done-when on both
issues is met, and I re-derived the load-bearing claims myself rather than taking the body's
table on trust. No blocking findings. Two precision notes and one grooming gap, below.

Method, so the evidence is checkable: I built a synthetic repo (bare `origin` + root clone +
a `wt-lane` worktree) with a stub `workmux` and ran **the actual `scripts/gate.sh`** end-to-end
against it, `HOME` and `PATH` redirected into a scratch tree so the real `workmux`, the real
`origin` and this checkout were never touched. Every row below is the real script's stdout and
exit status, old (`origin/main:scripts/gate.sh`) versus new, same rig.

### What verified in the author's favour — all EXECUTED

| Probe | old gate | this PR |
|---|---|---|
| Rigged failing post-merge build | `!! BUILD BROKEN`, **pushed to origin**, exit 0 | exit 1, `origin/main` untouched, local main holds the merge |
| `python3` absent from `PATH` | `no NUL bytes`, **pushed**, exit 0 | `GATE FAILED: python3 not found`, exit 1, before the loop |
| Stale branch + `workmux` exiting 127 | **merged and pushed a never-rebased tree**, exit 0 | `GATE FAILED: branch lane is not on top of main`, exit 1 |
| File with real NUL bytes | holds | holds — no vacuity traded |

The build row is the one that matters most, and it reproduces exactly as claimed: on the old
script `origin/main` moved to `Merged 'lane'` behind a broken build; on this one `origin/main`
stayed at its prior commit and `fail()` printed `>>> MERGED to local main, NOT pushed — fix
forward on main immediately, then push`. The state-aware `MERGED` flag is doing real work — the
message is true on both sides of the merge.

**The sibling case is real, and the self-criticism in the body is correct.** The
`staleworkmux` row is the whole argument: the old script reached the fence audit and the suite,
merged, and pushed a branch that had never been rebased onto current `main`. `#408`'s shape
through a fault-printing branch, exactly as described.

**I went looking for a false-hold in the new state assertion and there isn't one.** My worry was
`merge-base --is-ancestor main HEAD` resolving `main` locally: if `workmux rebase` rebased onto
`origin/main` while local `main` sat ahead — the state `push_or_warn()` leaves behind after an
offline landing — the assertion would hold a perfectly good branch with a misleading message.
It doesn't, because that is not what workmux does:

```
$ workmux rebase --help
Rebase a worktree branch onto its base branch
$ strings "$(command -v workmux)" | grep -i 'do NOT'
IMPORTANT: Do NOT run `git fetch`. Do NOT rebase onto `origin/<branch>`. Only rebase onto
the local branch name (e.g., `git rebase main`, not `git rebase origin/main`).
```

The assertion matches workmux's own stated contract and is consistent with the `main...HEAD`
the fence audit and the stranded-work check already use. All four false-hold cases the body
claims pass, and the stale one holds:

```
  PASS  <- branch ahead of main, properly based
  PASS  <- HEAD == main
  HOLD  <- main moved ahead: stale base (correctly held)
  PASS  <- after rebase onto main
  PASS  <- fresh zero-commit branch
```

`push_or_warn()`'s exemption also works as documented, which is not free given the pipeline —
`set -uo pipefail` on line 13 carries git's status past `| tail -1`, so the `|| echo` is
reachable and "loud but non-fatal" is accurate:

```
$ git remote add origin /nonexistent/nowhere.git
$ bash -c 'set -uo pipefail; git push origin main 2>&1 | tail -1 || echo "  ! push FAILED ..."'
and the repository exists.
  ! push FAILED — main is LOCAL-ONLY until pushed by hand
```

**`#2`:** `grep -nE 'gate\.sh:[0-9]+' AGENTS.md` exits 1, and so does the wider
`grep -nEi '\.sh:[0-9]+|line [0-9]+'` — no line citation into any script survives anywhere in
the file. `lane never does` (`:306`) and `#408` (`:303`) both preserved.

**Gates, on the assembled branch at `f0ca8ef` against `main` `25d8647`:** `bash -n
scripts/gate.sh` 0, `npm run typecheck` 0, `npm run lint` 0 (841 files).
`runbook-delivery-law` + `no-personal-paths-law` — 20 passed, including the substantive-runbook
law's three anchors (`## Review`, `Waves and the bundle unit`, `One commit per issue`). One
failure in that run was **my own rig artifact**, not this PR: I symlinked `node_modules` from
the primary checkout, and `.gitignore`'s `node_modules/` is a directory-only pattern, so
`check-ignore` reports it unignored. Proved it: exit 1 as a symlink, exit 0 as a real
directory. **I did not run the full 375-file suite locally** — with a borrowed `node_modules`
it would have generated more artifacts than signal, and CI ran it on the real tree instead.

**CI: all 7 expected jobs executed and passed.** The absent `build-test-boot (macos-latest,
min)` is the workflow's own deliberate `exclude`, not a skipped leg. Per-step confirmation on
the ubuntu/current leg — Install, Build, Test, Typecheck, Lint, Packaging guard, Boot smoke all
`success`; the single `skipped` is the min-node setup, correctly gated by its `if:`. Nothing
silently unrun.

**Fence:** `git diff origin/main...HEAD --name-only` is exactly `AGENTS.md` and
`scripts/gate.sh`. No widening, as claimed. The `// @gate-timing` markers survive in all three
files, so prd-24's coupling contract is intact.

---

### Finding 1 — the "unreadable file" probe row isn't reachable end-to-end (non-blocking)

`EXECUTED` at the loop level, `EXECUTED` for the two end-to-end paths I tried.

The row `Unreadable file | pass (clean) | holds` is true of the loop line in isolation, and I
reproduced it:

```
old:  c=$(python3 -c "...count(b'\x00')" u.txt 2>/dev/null || echo 0)  ->  c=[0], "PASS (clean)"
new:  c=$(python3 -c "..."             u.txt 2>/tmp/...) || fail ...   ->  "GATE FAILED: u.txt
      unreadable — cannot verify NUL bytes", exit 1
```

But I could not reach it through the whole gate. Both paths I could construct are intercepted
before the loop:

- `chmod 000` on a lane file — `git status --porcelain` reports it modified, so the
  stranded-work check (`:70`) fires first. Old and new fail identically there, for a different
  reason.
- an unreadable `package-lock.json`, the one path exempted from that dirty check — `clean()`
  at `:45` does `git checkout -- package-lock.json` and restores mode 644 before the loop runs.
  Both merge green.

So the residual trigger is narrower than the table reads: a `MemoryError` on a very large file,
or a stat/read race. **The change is still right** — removing a `|| echo 0` that manufactures a
clean verdict is unambiguously safer, and `#1`'s done-when asks for exactly that. This is a note
on the body's framing, not on the code: one sentence saying the row is a unit-level result would
stop a later reader treating it as an end-to-end behavioural difference. Nothing to fix in the
diff.

### Finding 2 — the strengthened doc claim reads wider than criterion 1 (your call, offered not asserted)

`EXECUTED`. `f0ca8ef`'s message says *"Neither sentence needed narrowing in the end: the code
earned the stronger claim."* For criterion 1 — every branch that **prints a fault** — that is
true, and I confirmed it. But `AGENTS.md`'s *"Every earlier check exits non-zero on failure"*
and `gate.sh:4`'s *"Every check exits 1 on failure"* will also be read as covering checks that
fail **silently**, and one of those is reachable by an operator typo today:

```
### CONTROL: off-fence file, valid fence '^f_'
  outside fence:
    OUTSIDE.txt
GATE FAILED: fence violated ...            exit 1,  origin/main unchanged

### same off-fence file, invalid fence '['
  fence OK: OUTSIDE.txt f_lane.txt         exit 0,  origin/main -> Merged 'lane'
$ git -C origin.git ls-tree -r --name-only HEAD
  OUTSIDE.txt        <-- landed
```

`grep` errors on the bad regex, `|| true` at `:59` swallows it, the audit prints `fence OK`, and
the off-fence file merges and reaches `origin`.

**This is pre-existing and correctly out of scope** — it is `|| true` on a line this PR never
touches, the PRD's non-goal is explicit, and you already filed it as deferred item 1, the
highest-value follow-up. I raise it only because the sentence this PR *authored* is the one a
reader checks the script against. PRD-39 splits criterion 1 (prints a fault → exits) from
criterion 2 (cannot run → must not report success), and under that split your wording is
defensible. If you want it airtight, *"every earlier check that reports a fault exits
non-zero"* costs three words and survives the fence-regex defect. Entirely your call; I have
not changed it, because you chose the wording deliberately and defended it in the commit
message.

### Grooming gap — PRD-39 criterion 2 is not met by this wave, and nothing tracks that

Not a defect in this PR, and **not a reason to hold it**: `#1`'s done-when asks for the
`python3` assertion and the removal of `|| echo 0`, and both are delivered. Criterion 2 in its
general form — *"Not met while any check's failure path and its 'nothing found' path produce
the same output"* — was never groomed into a wave-1 issue, and holding this PR to it would be
the groomer's gap, not the author's.

But it should not vanish either. Your own note on `#1` says deferred item 2 *"is what keeps
criterion 2 unmet even after this PR"*, and I confirmed it independently, `EXECUTED`:

```
$ printf 'x\000y' > "café.txt" && git add -A && git commit -qm nul
$ git diff main...HEAD --name-only
"caf\303\251.txt"                     <-- quoted under default core.quotePath
$ [ -f '"caf\303\251.txt"' ] && echo runs || echo "SKIPPED silently"
SKIPPED silently
$ python3 -c "print(open('café.txt','rb').read().count(b'\x00'))"
1                                     <-- a real NUL byte, and the guard prints "no NUL bytes"
```

Failure path and nothing-found path, byte-identical output — criterion 2's falsifier, verbatim.
Deferred item 3 (a malformed `lanes.json` printing `lane manifest pruned` at exit 0) is a second
instance; I spotted that one reading `:181` before I found your list.

So after this lands, PRD-39's Success section will read as satisfied by wave 1 when criterion 2
demonstrably is not. Suggest filing deferred items 1, 2 and 3 as a **prd39 w2** bundle before
this PR is closed out, so the criterion has a home — they are fenced behind this PR and can
start the moment it lands. Not a change to this diff.

---

`#1` and `#2` are both fully delivered, the sibling-case fix closes a defect I reproduced
landing broken work on `origin`, and the gates and CI are genuinely green rather than green-
looking. **Approving.** Not merging — merges on this repo are the operator's manual act on
GitHub, and `gate.sh` is the landing, so that step is deliberately left alone.

PR #41, opened 2026-08-24
docs(prd): prd-45 — the earned verdict

gabriel-canaan:
prd-45, blessed in session 2026-08-24. Document only — one new file under
`docs/prds/`, no code.

**The thesis.** prd-39's Success criterion 2 — *"a guard cannot report success
when it could not run"* — is still **unmet**, because the criterion was broader
than the ruling that served it. Ruling 2 named the NUL guard's two remedies and
both shipped. The same defect shape survives in seven other places in
`scripts/gate.sh`, and that file is the landing tool: it merges to local `main`
and pushes to `origin/main`.

All seven were executed against fixtures during the prd-39 verify pass — the
gate itself was never run, because running it is the landing. Every instance is
latent rather than firing, which is exactly why they survived a PRD written
about them.

**It adopts prd-24's orphans.** prd-24 retired to `done/` on 2026-08-24 with two
residuals "described for the next groom" and no owner named. Both are the same
thesis, so this discharges them rather than leaving them homeless. It does
**not** re-rule prd-24 ruling 3 — that ruling still governs and its citations
keep resolving.

**Sequencing was amended before any issue existed**, by running the grooming
against the draft. Two things were wrong first and are recorded as such:

- the gate fix and its law were two issues in one wave. The law can only be
  green *after* the fix — a stack wearing a bundle's clothes. Now one issue,
  with the law landing shown red against the pre-fix tree.
- the law's fence read `packages/**`, which strictly contains the recordings
  fence the same PRD claims. That is a hard fence-lint failure, caught before it
  reached the tracker. Now a single named file.

A third correction is an honesty fix rather than a structural one: the
recordings evidence implied a firing defect. Measured, the directory has **zero
subdirectories**, so the flat walk sees all 13 source files today and hides
nothing. Real under ruling 3, fires the day a subdirectory appears — which is
how the `lab/` sibling hid `lab/branching/geometry.ts:1` — but the bullet now
says so.

**What is the sibling case?** The PRD itself is the answer: prd-39 fixed the two
instances its ruling named and left the class open, and ruling 1 here refuses
that explicitly — *"a fix that closes the enumerated set and not the class earns
this PRD a successor."* For this PR specifically, the sibling risk is a PRD that
re-rules territory another PRD owns; checked by hand — prd-24 ruling 3 is linked
rather than restated, and `packages/contract/` is named as prd-29 wave 3's and
refused.

**What mutation would this test survive?** No test covers a PRD. What stands in
for one: every `gate.sh` line cited was verified to exist at `51e971c` after PR
#40 moved them all, the four fences were checked for pairwise prefix containment
in both directions, and the two `ci.yml`-reading tests were read to confirm
`if: always()` breaks neither (`node-floor-law.test.ts:51,55` asserts
`node-version:` strings and a count; `boot-line.test.ts:72` asserts the boot-line
marker — both text `toContain` on strings this change does not touch).

**Fence widenings.** None. One added file, `docs/prds/prd-45-the-earned-verdict.md`.

Milestone `prd45` and the three wave-1 issues follow this; the PRD lands first,
because a PRD that is not on the default branch does not exist — prd-38 sat
finished on a side branch and was reported never written.


ciaran-slow (approved):
## Approved — with two Evidence citations fixed in `b0a1dba`

Reviewed commit-by-commit (one commit, one file, 176 lines). Every finding below is labelled
**EXECUTED** (I ran it and read the exit status) or **REASONED** (argued from the code, nothing
run). I ran the shell reproductions in disposable throwaway repos — **never `scripts/gate.sh`
itself**, because running it *is* the landing.

This PRD's whole value is that its citations are checkable, so I checked all of them. They hold.

**All eight `gate.sh` instances are real.** Six reproduced by execution:

| cited | reproduced |
|---|---|
| `:59` invalid fence regex | **EXECUTED** — `FENCE='['` → grep errors → `\|\| true` → `viol` empty → `fence OK`, off-fence file proceeds. And the `\|\| true` genuinely cannot just be deleted: `grep -v` exits **1** on the clean no-match path, as the bullet says. |
| `:69` failed `git status` | **EXECUTED** — producer exit 128 → `dirty=[       0]` → stranded-work check passes. `set -uo pipefail` is on at `:13` but there is no `set -e`, so the assignment's status is discarded exactly as claimed. |
| `:77`+`:83-84` NUL guard | **EXECUTED** — real repo, `café.ts` containing NUL bytes, default `core.quotePath`: git emits `"caf\303\251.ts"`, `[ -f ]` fails on the literal quotes, loop prints `no NUL bytes`, **exit 0**. Precisely what the guard exists to catch. |
| `:83` process substitution | **EXECUTED** — producer exits non-zero, loop finishes with status 0, clean verdict would print. `pipefail` does not reach it. |
| `:49` worktree fallback | **EXECUTED** — linked worktree, `.git` is `ASCII text`, real conflicted rebase in progress: correct resolution **DETECTED**, `$W/.git` fallback **MISSED**. Matches the bullet word for word. |
| `:127-129` garbage `timing-count` | **REASONED** from `case "$PREV" in ('' \| *[!0-9]*) PREV=0` — `TCOUNT -lt 0` is unreachable, so the shrink check cannot fire. |

`:166` (`rev-parse --verify refs/heads/$BRANCH` as merge proof) and `:172-181` (`catch {}` then
`&& echo "lane manifest pruned"`) read exactly as described — **REASONED**; and the comment at
`:171` does record *"three landed lanes read OFF-FENCE in a live UI"*, as quoted.

**`ci.yml:66,69,72,97` are exact** — Typecheck, Lint, packaging guard, boot smoke. **EXECUTED:**
the only four `if:` lines in the entire file are matrix gates on the two setup-node steps
(`:43,49,178,184`). Ruling 4's scope is also *complete*, which is worth saying: `pack-smoke` is
the only other job and it has no `Test` step, so "every step after `Test`" is unambiguously
those four.

**The recordings measurement is precise.** **EXECUTED:** `:31` is a flat `readdirSync`, `:39` is
a hardcoded `toBeGreaterThan(3)`, the directory has **0** subdirectories, and the walk's filters
(`.ts|.tsx`, minus `.test.tsx?`) select **exactly 13** files. The honesty correction the PR body
describes is the right call, and the number is right. The `lab/` sibling's own header at
`lab/no-live-fleet-law.test.ts:22-24` independently records the story cited — 5 of 17 files seen,
`toBeGreaterThan(3)` passing vacuously — and `lab/branching/geometry.ts:1` is indeed the
`../../scene/palette.js` import a shallow walk hid. `walkSourceFiles` is at that file's `:130`.

**Cross-PRD claims all resolve.** prd-39 Success 2 is quoted verbatim; ruling 2's two remedies
both shipped (`gate.sh:39` python3 probe, `:81` `|| fail` where `|| echo 0` was); ruling 1's push
exemption is real, so the `set -e` rejection is sound. prd-24 is in `done/`, superseded
2026-08-24, and its amendment numbers the residuals **exactly** as this PRD cites them — 1 is the
recordings walk at that same `:31`, 2 is the CI leg naming the same four steps, 3 is prd-29 wave
3 / `packages/contract/` (confirmed at `prd-29:179`). Rulings 3 and 4 are characterised
accurately and neither is re-ruled.

**The thesis survives the obvious objection.** prd-39 ruling 1 *also* carries an "extends to every
branch" clause, so I checked whether criterion 2 was really left open. It was: ruling 1's class is
branches that **print a fault**, ruling 2's is guards that **claim success they did not earn**.
The eight instances here are all the second shape. The distinction is correctly drawn.

**No territory collision.** I specifically checked prd-44, since it is named "the flat instrument"
and this PRD carries a flat-walk residual — different subject (cost curves, not directory walks),
and its declared fences (`log/`, `recorder/`, `scene/gl/`, `collectors/`) never reach
`packages/web/src/recordings/`.

**Parallelism is proven by the repo's own tool, not by assertion.** **EXECUTED:**
`scripts/fence-lint.sh 42 43 44` → *overlaps: none*, `scripts/gate.sh` shows `ok` as an owned
coupling point, **fence lint PASSED (exit 0)**. Pairwise prefix containment across all four
fence paths, both directions: 0 collisions. `packages/server/src/gate-honesty-law.test.ts` is
free, and `packages/server/src/` already holds four repo-reading laws, so the placement is right.
The `packages/**` counterexample the grooming caught is a genuine containment.

**Coverage of the new file is mutation-proven.** `no-personal-paths-law` reads every tracked file
via `git ls-files`, so it covers this PRD. **EXECUTED mutation:** injecting a home path into the
new file turns it red (`prd-45-the-earned-verdict.md: home-directory name "realperson"`);
restored, green. 7 law files / 57 tests pass; `typecheck` and `lint` both PASS.

**CI genuinely ran.** 7/7 legs pass, and the 3 `build-test-boot` legs are exactly the declared
matrix (2×2 minus the `macos-latest`/`min` exclude). I did not take the green bar for it —
step-level inspection of the macOS leg shows Test, Typecheck, Lint, packaging guard and boot
smoke all `completed/success`, with only the min-node setup correctly `skipped`. Fitting, for
this PRD.

No `Closes` line is correct here, not an omission — #25, #28 and #29 are all paper PRs and none
carries one. Heading levels match prd-39/43/44 (`## Rulings` then `## Ruling N`).

### Fixed in `b0a1dba` (Evidence only — no ruling, criterion, fence or wave moved)

**1. `:70` → `:57`.** **EXECUTED.** The parenthetical bounding the `:83` instance said an
unresolvable `main` "dies at `:70` first". It dies at **`:57`** — `git merge-base --is-ancestor
main HEAD` exits 128 and routes through `fail()`. Line 70 is the stranded-work check, and it does
not fire at all in that state: `git status` succeeds whether or not `main` resolves, so
`dirty=[       0]` and it passes. The *argument* was right — the script does die before reaching
`:83` — only the line was wrong.

**2. The suite-coverage grep is restated.** **EXECUTED.** As written, `grep -rln "gate.sh"` is a
regex whose `.` also matched *"the coalescer's gate shut"*
(`web/src/replay/useReplaySession.test.ts:298`), and "finds only `@gate-timing` marker comments"
is not what it returns. `grep -rlnF` gives three marker files, three **prose** comments in
`core/src/fleet/` (`buildFleet.ts:371`, `buildFleet.test.ts:180`, `fixtures.ts:323`), and three
captured codex fixtures that happen to quote AGENTS.md's own *"do not run `scripts/gate.sh`"*
line. None asserts anything about the script, so the bullet's claim stands untouched — and it now
leads with the stronger fact I verified separately: **no test file reads `scripts/` from disk at
all.**

Both are the shape prd-43 rulings 1 and 2 exist to catch, which is why I fixed rather than
flagged them. Revert either freely if you'd rather word it yourself.

**A. "closed two and left seven" vs "the eight found", in one sentence** (Ruling 1). Evidence
enumerates **eight** `gate.sh` instances; the ruling says prd-39 "left seven". Both reconcile
only under the gloss *seven **other than** the NUL guard* — which the Problem does supply, 100
lines earlier. A reader auditing "did wave 1 close the seven?" counts eight. I did not touch it:
it is a blessed ruling, and rulings are yours. If you want it airtight, "closed two paths and
left the eight above" costs one clause.

**B. Two more instances of the same shape, not enumerated.** Ruling 1's class clause already
covers these, so nothing here is falsified — but the PRD's value is its enumeration, and wave 1
is measured against criterion 1's *"every check"*:

- **`clean()` at `:26-27`** — **EXECUTED.** With `package-lock.json` in a conflicted state (`UU`),
  `git checkout -- package-lock.json` exits **1**; `2>/dev/null || true` swallows it, `clean()`
  returns 0 and prints nothing — byte-identical to the success case — and the conflict markers
  survive. This is the exact shape, and it is *reachable at the worst moment*: `clean` runs at
  `:164`, immediately before `workmux merge`. Nothing downstream notices, because `:69`'s dirty
  check greps `package-lock.json` out by design.
- **`:23-24` worktree resolution** — **EXECUTED**, low reachability, closer to ruling 2's proxy
  shape. When `workmux path` fails, `W` is empty and `:24` substitutes a *constructed* path —
  which is what the comment at `:20-22` forbids outright (*"never by string construction… a
  constructed path then silently misses a live lane"*). If that path exists and has `main` as an
  ancestor, `:41` and `:57` both pass and the gate audits a tree nothing proved is the handle's
  lane. In practice the fallback usually resolves correctly, which is precisely why the
  tool-said/I-guessed distinction is never drawn.

If either belongs in Evidence, adding them before #42 is dispatched is cheaper than a successor
PRD — and #42's fence already covers the lines.

### What I did not run

The full suite locally (CI ran it on 3 legs; I ran the 7 repo-reading law files plus typecheck
and lint against the post-fix tree). Not `scripts/gate.sh` — running it is the landing, and it is
this PRD's own subject. `scripts/land.sh` does not exist in this repo.

**Not merging** — merges here are the operator's, by hand. Ready when you are.

Issue #48, opened 2026-08-24
prd45 w2: an accidental timing enrolment cannot raise the ratchet unseen


ciaran-slow:
## Fence (may touch ONLY)
- `scripts/gate.sh`
- `packages/server/src/gate-honesty-law.test.ts`

## Blocked by
- **#42** (prd45 w1) — it holds both files above as its live claim and its DoD already reaches the same block (`:127-129`, a corrupt `timing-count` must be a hold rather than `PREV=0`). **Nothing here may be dispatched while #42 is open.** If prd-45's owner would rather absorb this into #42, close this as superseded — that is the cheaper outcome and this issue exists only so the finding has a home either way.

Filed by a passer-by, not by prd-45's author: found while verifying prd44 #33, whose whole subject is not enrolling a file in the timing set. Offered rather than claimed.

**Model:** sonnet. **Wave:** 2.

## Definition of done
- A timing-set count that **rises** is distinguishable from one that holds steady. The shrink check at `:129` fires on a drop; nothing observes a rise, and `:157` records the new higher number unconditionally.
- A rise that was not intended fails the landing, or is at minimum reported, rather than being written into the floor.
- The existing contract is preserved exactly, per #42's own instruction not to disturb it: the set stays DERIVED (`// @gate-`+`timing` marker or `*.bench.test.ts` name), an empty set still fails at `:123`, and a genuine drop still fails at `:129`.
- Whatever tolerance is chosen is **declared data in the law**, not a condition buried in a regex — the same requirement #42 puts on the push exemption.

## Evidence (EXECUTED, at `4140f6b`)

The ratchet is one-directional, and the consequence compounds:

```
:129   [ "$TCOUNT" -lt "$PREV" ] && fail …     # only a DROP fails
:157   printf '%s\n' "$TCOUNT" >"$COUNT_FILE"  # any count, including a risen one, becomes the new floor
```

So a file that enrols **by accident** is not merely unnoticed — on the next successful landing its inflated count becomes the floor. A later, entirely correct removal of a genuinely-timing file then fails the gate, and the failure names the wrong change. The operator is told a timing test "silently fell out" when what actually happened is that an unrelated file wrongly joined earlier.

**This is not hypothetical.** prd44 #33 came within one doc comment of it: a sentence explaining that the file carried no such marker *spelled the marker out*, and `:119` greps it with `grep -rlF`, which does not care that the occurrence is prose. Caught in review; nothing in the tool would have caught it.

[...]

Issue #46, opened 2026-08-24
prd44 w2: a failed append releases its descriptor


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/recorder/session-log-writer.ts`
- `packages/server/src/recorder/session-log-writer.test.ts`

## Blocked by
- **#31** (prd44 w1) — this is a consequence of the held descriptor that issue introduces, and there is nothing to fix until it has landed.

Sibling lanes hold the three collectors and `packages/server/src/recorder/session-recorder.ts`. **Do NOT touch** `packages/server/src/recorder/session-recorder.ts` — it is prd-40 #3's live claim, and the fix belongs in the writer regardless.

**Model:** sonnet. **Wave:** 2.

## Definition of done
- A write that fails after the descriptor is already open leaves **no** descriptor held. A law asserts `closes` reaches 1 after a rejected append, not 0.
- The next append after a failed one opens a **fresh** descriptor rather than reusing the handle whose last operation failed.
- The existing four writer laws stay green, unmodified in intent: appends land in issue order (`session-log-writer.test.ts:46`), `sync()` leaves every issued append on disk (`:68`), `sync()` on an un-appended writer is a no-op creating no file (`:81`), and a write failure still reaches the caller that issued it (`:89`). **That last one is the one to watch** — releasing the handle must not swallow the rejection.
- #31's own three count laws stay green: opens=1 across many appends, opens=closes=5 across five cycles, and a post-`sync()` append reopens.
- **Assert a count, never a wall clock.** Do not name any new law `*.bench.test.ts`, nor mark it `// @gate-timing` — `scripts/gate.sh` derives its 4x-load timing set from exactly those and ratchets a count only a human can clear.

## Evidence (from #31's verify pass, EXECUTED)

A probe constructed a writer, let its `ensureHandle()` open a real descriptor, then forced the
subsequent write to reject — simulating `ENOSPC` mid-write — while counting real open/close calls
through a mock of `node:fs/promises`:

```
opens  === 1
closes === 0        and still unclosed 50ms later
```

Nothing in the class closes it except `sync()`, which is never reached: on an append failure the
recorder's `closeWith()` re-throws without calling it.

**This is a genuine regression from the pre-#31 behaviour.** The free-standing `appendFile(path, …)`
it replaced opened, wrote and closed its own descriptor inside Node's implementation whether or not
the write succeeded, so a failed append held zero descriptors. After #31 it holds one, indefinitely,
[...]

Issue #42, opened 2026-08-24
prd45 w1: a gate check that cannot run says so


gabriel-canaan:
## Fence (may touch ONLY)
- `scripts/gate.sh`
- `packages/server/src/gate-honesty-law.test.ts` (new file)

## Blocked by
Nothing. **Wave 0 is the operator's own act**: this is landed by hand, without
`gate.sh`, because the gate still cannot be trusted to land its own fix. prd-39
booked the same wave 0 for the same reason and it held.

Sibling lanes hold other surfaces: `.github/workflows/ci.yml` and
`packages/web/src/recordings/` belong to the two other wave-1 issues — do not
enter either. `packages/contract/` is prd-29 wave 3's territory.

**You own the `scripts/gate.sh` coupling entry** (`.swarm/coupling.txt`): the
script DERIVES its timing set from every `packages/**` test carrying a
`// @gate-timing` marker plus every `*.bench.test.ts`, and fails the landing both
when that set is empty (`:110`) and when its count drops below the ratchet in
`.swarm/timing-count`. Do not disturb that contract.

**Model:** sonnet. **Wave:** 1.

## Definition of done

Rulings 1–3. The fix and the law are ONE issue because the law can only be green
after the fix exists — a stack, which the wave contract allows in a wave only as
a single issue. Land the law **shown red against the pre-fix tree**, in commit
order, inside this lane.

- Every check in `gate.sh` produces output distinguishing "could not run" from "ran and found nothing". Each of these, verified individually:
  - `:59` an invalid fence regex is a hold, not `fence OK` — note `|| true` cannot simply be deleted, because `grep -v` exits 1 on the clean no-match path, so the producer's status is checked separately from the consumer's.
  - `:69` a failed `git status` is a hold, not `worktree clean`.
  - `:77`+`:83` paths are read `-z`-delimited with `read -r -d ''`, so a quoted or otherwise hostile filename is checked rather than skipped.
  - `:83` the file-list producer's failure is a hold, not the clean verdict.
  - `:172-181` a failed manifest prune says so instead of printing `lane manifest pruned`.
  - `:49` the git-dir probe failing is a hold, not a fallback to a path that makes the mid-rebase guard structurally dead.
  - `:127-129` a corrupt `timing-count` is a hold, not `PREV=0`.
- `:166` proves the lane commit is contained in `main`, not that a ref is absent — branch deletion is `workmux merge`'s side effect, not the landing's meaning.
[...]

Issue #32, opened 2026-08-24
prd44 w2: a settled retired lane is tessellated once


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/scene/gl/frame.ts`
- `packages/web/src/scene/gl/batch.ts`
- `packages/web/src/scene/gl/frame.test.ts`

## Blocked by
Nothing.

Sibling lanes hold other surfaces: `log/`, `recorder/session-log-writer.ts`, `server/concurrency.ts`.

**Scope, and it is narrow on purpose.** This is the *tessellation* stage on the *retired* axis. prd-33 ruling 13 already states the model-stage ceiling — 90 threads at 60 fps, 180 at 30 fps — and says outright that "180 threads is not a bug to be closed by relaxing this table". **Move no cell in that table**, and do not touch `scene/marks/` or the model stage, which are prd-33's.

**Do NOT touch** `packages/web/src/scene/palette.test.ts`, `packages/web/src/scene/palette.ts`, `packages/web/src/scene/marks/` or `packages/web/src/theme/theme.css`. The first two are coupling points owned by nobody in this wave (`.swarm/coupling.txt`): palette.test.ts parses theme.css off disk and imports across scene/, so a change here can redden it from outside your boundary. If that happens, STOP and record a widening on the PR before touching it. marks/ and the model stage are prd-33's.

**Model:** sonnet. **Wave:** 2. (Was wave 1 in the five-to-three amendment; wave 1 landed as #30/#31/#33 without it, so it dispatches with wave 2. Title and board already say w2.)

## Definition of done
- A retired lane past `dissolve >= 1` contributes no per-frame tessellation. A law asserts the vertex-write count for a settled lane is zero on a frame where none of its inputs changed.
- Cached vertices are **byte-equal** to a fresh build. This is the same contract `livingSpineCache` already holds (`scene/geometry/layout.ts:200`).
- The cache is keyed on the same world signature the geometry cache already uses — see `docs/design-notes/geometry-cache-audit-178.md` for the key discipline; reuse its reasoning, do not re-derive it.
- A resize, a camera change, or a lane leaving the settled state invalidates that lane alone (prd-44 open question 4: whether a persistent range survives a camera change without a full rebuild is unanswered — answer it here).
- `buildFrame`'s existing `into` parameter and `Batch` are reused; `vertices.reset()`'s all-or-nothing scope (`frame.ts:141`) is what changes.
[...]

PR #50, opened 2026-08-24
prd44 w1: the instrument stops paying twice for what it already has

### LINKED ISSUE #30: prd44 w1: a finished recording is parsed once, not once per request
## Fence (may touch ONLY)
- `packages/server/src/log/lane-index.ts`
- `packages/server/src/log/listing.ts`
- `packages/server/src/log/lane-index.test.ts`
- `packages/server/src/log/listing.test.ts`

## Blocked by
Nothing.

Sibling lanes hold other surfaces: `recorder/session-log-writer.ts`, `scene/gl/`, and `server/concurrency.ts`. **Do NOT touch `packages/server/src/api/lane-index.ts`** — prd-29 ruling 7 (landed `5f45252`) is adding a capability gate to that file, and this lane would collide with it. The cache lives in `log/`, which is why ruling 1 is written that way.

## Definition of done
- A session file that is not the live one is parsed at most once, however many requests arrive. A law asserts the parse count is constant as request count grows.
- The cache is keyed on the file's identity plus `mtime` and `size`; a file that changes on disk is re-parsed.
- The live session is still read from the recorder's buffer, never from disk (`lane-index.ts:481-484` today).
- The three per-session reads (`readSessionLog`, `readSessionLabel`, `readTranscriptCaptureManifest`, `lane-index.ts:484-490`) issue concurrently; they are three different files.
- The route's answer is byte-identical to today's for the same directory. This caches *parsing*, never answers — nothing it reports may go stale.
- The cache has a stated bound and eviction rule (prd-44 open question 1).
- **Assert a count, never a wall clock.** A timing assertion measures the box, not the code — this repo reports times and asserts counts. And do NOT name any new law `*.bench.test.ts`, nor mark it `// @gate-timing`: `scripts/gate.sh` derives its timing set from exactly those, runs it under 4x load, and ratchets `.swarm/timing-count`, which only a human can clear.

## Provenance — why this route reads every recording

It is not an oversight, and the cache must not undo it. **prd-31 ruling 5** (shipped) created this
surface, and made the index part of the ruling rather than an optimisation of it:

> `workmux merge` deletes a worktree the moment work lands, which is precisely when a person most
> wants to read what happened. [...] a lane that ran across three sessions has its life scattered
> across three recordings with no index tying them together. **A lane index is therefore part of
> this ruling, not an optimisation of it** [...] so `/lane/519-migrate` opens a week later without
> the reader knowing which recording to find first.

So reading across every recording is the feature. What this lane changes is only that the reading
is not *repeated* — a finished recording is immutable, so parsing it twice is waste, while parsing
it once is the ruling working as designed. Two consequences for the design:

- The cache must stay **cross-session**. A cache that only remembers the live session, or that
  scopes itself per request, defeats ruling 5's whole point.
- A recording that has been pruned must degrade the way the index already degrades for one it
  cannot read (`unreadableSessionIds`), never by silently answering as if the lane never existed.
  This matters because prd44 #38 will start deleting old recordings.

**Measured today:** 281.3 / 280.0 / 252.2 ms per request on 4 files / 20 MB / 51,690 lines, page cache warm; a 20 ms heartbeat beside it ran 92.5 ms late and fired 3 of ~14 ticks.

### LINKED ISSUE #31: prd44 w1: the log is opened once per session, not once per event
## Fence (may touch ONLY)
- `packages/server/src/recorder/session-log-writer.ts`
- `packages/server/src/recorder/session-log-writer.test.ts`
- `packages/server/src/recorder/namespace-law.test.ts` (minimal reconciliation only)

## Blocked by
Nothing — but **this lane should land before or with prd-40 #3**. prd-40 ruling 1 puts the append on the critical path ("the append is awaited before the event is anyone's"); today the write races behind publication. After it, the 61.4 us of open/close per event becomes operator-visible latency. Landing this first is what stops prd-40 shipping that regression.

Sibling lanes hold other surfaces: `log/`, `scene/gl/`, `server/concurrency.ts`. **Do NOT touch `packages/server/src/recorder/session-recorder.ts`** — it is prd-40 #3's live fence. The close does not go in `closeWith`; see below.

## Definition of done
- One `open` per session, one `write` per event. A law asserts the `open` count does not grow with the event count.
- **The handle is released on `sync()`**, not by `openSession`. `closeWith` already calls `sync()`, and `sync()` already opens a descriptor of its own (`r+`) today — so nothing outside this file changes. An `append` issued after a `sync()` reopens; prove that interleaving with a test, it is not an assumption.
- No descriptor survives a rotation. A law asserts repeated close/open cycles leak none.
- The three existing writer laws stay green, unmodified in intent: appends land in the order they were ISSUED even when nobody awaits them (`session-log-writer.test.ts:46`), `sync()` leaves every issued append on disk (`:68`), and a write failure reaches the caller that issued it (`:89`).
- prd17 ruling 1's "a final `session.closed`" still holds structurally — the serialised `tail` chain is what makes it so (`session-log-writer.ts:61`) and is untouched.
- The symlink refusals in `prepare()` still run before the first write.
- `namespace-law.test.ts` may be reconciled ONLY as far as a held handle requires: its regex at `:157`/`:191` matches `open(`, and the law that every filesystem write in the module lives in the log writer must still bite.
- Bytes on disk are unchanged for the same event stream.
- **Assert a count, never a wall clock.** A timing assertion measures the box, not the code. And do NOT name any new law `*.bench.test.ts`, nor mark it `// @gate-timing`: `scripts/gate.sh` derives its timing set from exactly those, runs it under 4x load, and ratchets `.swarm/timing-count`, which only a human can clear.

**Measured today:** 306.9 ms for 5,000 events (61.4 us/event) against 74.1 ms for one held handle — 4.1x, with no semantic movement.

### LINKED ISSUE #33: prd44 w1: a bounded fan-out runs its subprocesses at a stated ceiling
## Fence (may touch ONLY)
- `packages/server/src/server/concurrency.ts` (new file)
- `packages/server/src/server/concurrency.test.ts` (new file)

## Blocked by
Nothing — this is the wave's keystone: two new files, additive, claimed by no other lane. Verified unclaimed against the open fences of #7, #8 and #14.

Sibling lanes hold other surfaces: `log/`, `recorder/session-log-writer.ts`, `scene/gl/`. **Do NOT touch `packages/server/src/server/exec.ts`, `server/poll-loop.ts`, or any collector** — this lane ships the helper alone; prd44 wave 2 is what consumes it.

**Model:** opus. **Wave:** 1.

> **Correction (2026-08-24):** this issue was groomed as `sonnet`, but the build agent was dispatched without an explicit model and inherited `opus` from its parent session — the commit trailer reads `claude-opus-5`. The first verify pass then also ran on opus, so it carried no model independence; a second, genuinely independent pass was run on sonnet and found two coverage gaps the same-model pair had missed. The field is corrected to match the artefact rather than the artefact re-made to match the field: the work is verified twice and sound. Recorded because prd-45 ruling 2 is about postconditions asserting the fact they claim, and a Model field nobody honours is the same defect in paperwork.

## Definition of done
- A helper that runs N async tasks with at most K in flight, preserving **result order** regardless of completion order. Result order is the load-bearing half: the record is hash-chained (ADR-0009), so wave 2's collectors gather concurrently and must append in their original sequence or a replay stops being byte-identical.
- A law asserts the in-flight count never exceeds K, at K=1 and K>1, including when a task rejects.
- A rejecting task does not cancel its siblings, and its rejection reaches its own caller — the shape `session-log-writer.ts`'s `tail` chain already uses ("one failed append must not poison every later one").
- K=1 is exactly today's sequential behaviour, provably: same order, same results.
- It composes with `withTimeout` (`server/exec.ts`) rather than growing a second timeout story. It does not import or re-implement it.
- The ceiling is a stated constant with a named reason, not a magic number. The reference box is an 8 GB machine that already thrashes at four concurrent lanes, and six collectors each fanning out per entity could otherwise start 40+ processes at once.
- No dependency is added (`package-lock.json` is untouched — at most one dependency-adding issue per wave, and this wave spends that on nothing).
- **Assert a count, never a wall clock.** A timing assertion measures the box, not the code. And do NOT name any new law `*.bench.test.ts`, nor mark it `// @gate-timing`: `scripts/gate.sh` derives its timing set from exactly those, runs it under 4x load, and ratchets `.swarm/timing-count`, which only a human can clear.

**Why it exists:** wave 2's three collector lanes each need this, and a shared file inside one wave would make them a stack rather than a bundle. Shipping it here keeps those three genuinely parallel.

ciaran-slow:
Closes #30
Closes #31
Closes #33

prd-44 wave 1. Three costs that grew with how long the instrument had been used, and no longer do. One commit per issue, cherry-picked from each lane; the fourth commit is the wave's own changelog entry.

**#32 is not in this PR.** It was built, gated fully green, measured at a net regression at its own named configuration, and refused rather than landed. Re-scoped to wave 2 by operator decision so three verified lanes were not held behind a redesign. Its work is preserved on `32-spike-per-mark-cache` (`6ad2777`), explicitly not a merge candidate.

---

## `8ef41c2` — #30 a finished recording is parsed once, not once per request

`GET /api/lane-index` re-read and re-parsed every recording on disk on every request, with no cache. Measured on this repo's own session directory (4 files, 61,157 lines, 23 MB), page cache warm: **344.4 ms → 130.1 ms median, 2.6x.**

Only the *parse* is cached, never the answer — `listSessions()`'s directory walk stays live, which is what makes wave-3 pruning degrade through the existing `unreadableSessionIds` path rather than reporting a lane that never existed. [ADR-0028](docs/adr/0028-bounded-lru-cache-for-parsed-session-logs.md) records the 128 MB byte-bound, LRU eviction, and the rejected alternatives (TTL, unbounded, per-instance).

**Fence widenings, both pre-authorised before the change:** `docs/adr/0028-*.md` and its index row in `docs/adr/README.md`. The row is not optional — `adr-log-law.test.ts` asserts every record has one and every link resolves.

**Sibling case.** The cache is keyed on path + mtime + size, and the sibling of "a stale entry served after the file changed" is "a stale entry served after the file was *deleted*". Both are laws: a mutation dropping either half of the key is caught, and a deleted file served from cache is caught.

**What mutation would this test survive?** It survived one, and that was a must-fix from verify. The size-check law did `utimes(f, new Date(first.mtimeMs))` — whole milliseconds, while APFS stores sub-millisecond mtimes, so the "restored" mtime never matched and the size check never ran. Dropping `size` from the validity key left **all 58 tests green while the cache served stale content**. Fixed in `0b0ac3c`: the mtime is pinned to a whole second *before* the first read and the precondition is asserted. Under the same mutation it now fails `expected 1 to be 2`.

**Known shortfall, not a defect in this commit.** The stall improves 2.6x but does not vanish: profiled, `buildLaneIndex` is 107.1 ms of the remaining 116.5 ms and `reduceAll` alone is 88.9 ms. So **prd-44 ruling 1's claim that steady-state work falls "to the one recording still moving" is measurably false**, and the issue's own criterion under-specified its PRD — Success 1 reads "**Fold**, read and parse work" and the grooming dropped "Fold". Worse, the residual freeze is now *less interruptible*: `buildLaneIndex` is fully synchronous, so this change removed `main`'s yield points and left one uninterruptible ~112 ms block. Verify's judgement, which I agree with: three defects, three owners — the ruling's claim, the issue's criterion, and not the build. Caching the fold under the same key is legitimate and would take it to ~25 ms; that needs a ruling amendment, which is an operator act.

## `77f910b` — #31 the log is opened once per session, not once per event

`appendFile` per event meant an open and a close per event: **306.9 ms per 5,000 events (61.4 µs each) → 74.1 ms, 4.1x.** One `write(2)` per event either way, bytes on disk unchanged, ordering unchanged.

The handle is released on `sync()`, not by `openSession`. That placement is the load-bearing detail: `session-recorder.ts` is prd-40 issue #3's live fence, so putting the close in `closeWith` would have reached into another PRD's territory *and* broken ruling 2's own "before or with prd-40 wave 1" timing. `sync()` already opened a descriptor of its own, so the change stays entirely inside `session-log-writer.ts`.

prd17 ruling 1's "a final `session.closed`" is untouched — the serialised `tail` chain is what makes it structural, and it is unchanged.

**Sibling case.** A descriptor released on the normal path has a sibling in the *error* path. Verify probed it: a write that fails after the handle is open leaves `opens=1, closes=0`, still unclosed 50 ms later, where the old free function held none. Bounded to one fd per affected session and self-healing on the next `sync()`, so it was raised as non-blocking and **filed as #46** rather than fixed here — and this PR corrects the build report, which had reasoned that case was unchanged. It is not.

**What mutation would this test survive?** Three count laws, all mutation-proven: reverting `append()` to open-per-event fails all three; removing `sync()`'s release leaves opens at 1 across five cycles instead of 5/5; a post-`sync()` append reusing a stale handle is caught by content on disk. The `namespace-law` reconciliation was also a must-fix — the first version's blanket `(?<!\.)` excluded *any* dot-preceded call and provably weakened the law (an injected namespaced `.open('/tmp/attacker-controlled-path')` passed 19/19). Narrowed to `(?<!handle\.)` in `c469cf2`; the same mutation now fails.

## `502eaa5` — #33 a bounded fan-out runs its subprocesses at a stated ceiling

Two new files, no caller yet — prd-44 wave 2's collectors are what consume it. `runBounded` over thunks plus a `mapBounded` wrapper, at `COLLECTOR_FANOUT_LIMIT = 4`, with the reason in the module: the reference box is an 8 GB machine that thrashes past four lanes, and six collectors each fanning out per entity could otherwise start 40+ processes at once.

**Order is preserved by construction**, not by test: a shared `tasks.entries()` iterator pulled by K worker loops, each writing `outcomes[index]` positionally, never `push`. That is ADR-0009's hash chain — wave 2 gathers concurrently and appends in original order, so a replay stays byte-identical. Results are an ordered settled union rather than `Promise.all`, so a rejecting task never cancels its siblings and its reason reaches its own index with identity preserved.

**Sibling case.** A synchronous `throw` from a thunk and an async rejection are the same failure by two mechanisms; both are laws. Verify found the deeper sibling: the same order bug reachable by a *different* mechanism — a slot claimed from a counter after the await — which fails 7 laws.

**What mutation would this test survive?** Verified across two independent passes and my own: `outcomes[index] =` → `push` fails 13 laws; `Math.min(limit, n)` → `n` fails 12. Two gaps were found and are recorded as non-blocking follow-ups: the statelessness law only tests *sequential* batches (a genuine module-level shared budget passed all 26 laws), and the zero-imports law misses an *inert* dynamic import. The shipped file genuinely has zero imports and zero state — confirmed by reading it, not by trusting its laws.

**This commit's own near-miss is worth reading.** Its first-draft doc comment explained that the file carries no gate-timing marker — by spelling the marker out. `scripts/gate.sh` greps that fixed string with `grep -rlF` and does not care that the occurrence is prose, so the comment alone would have enrolled these laws in the 4x-load timing pass and ratcheted `.swarm/timing-count`, which only a human can clear. Caught, and a law now asserts the file is not in the timing set with every needle assembled from pieces so it cannot trip its own guard. Filed as **#48** (blocked by #42): the ratchet is one-directional, so an accidental enrolment raises the floor silently and permanently.

## `dafd525` — the wave's changelog entry

All three lanes deliberately declined an entry of their own, because four wave-1 lanes editing one paragraph is exactly the fence collision `AGENTS.md` warns about. Written once here, and it states both limits the wave did not close.

---

## Verification

Every issue: built, verified by a **different model** than its build, must-fix findings answered with a named sha. #33 was verified twice — the first pass shared a model with its build, which is a real process defect I introduced by dispatching without an explicit model; the sonnet re-run found two coverage gaps the same-model pair had missed.

**Assembly gated here, the first time the three coexist** — node 22.23.2:

| gate | result |
|---|---|
| `npm run typecheck` | clean, 5 workspaces |
| `npm run lint` | clean, 843 files |
| `VITEST_MAX_WORKERS=6 npm test` | **376 files, 6402 passed**, 3 expected-fail |
| `npm run build` | clean — server, web, app |
| timing-set derivation (read-only) | **6 files, unchanged from main** — no ratchet movement |

All three cherry-picks applied **without conflict**, which is the wave contract holding: within a wave nothing depends on anything else in it, so a conflict would have been a fence violation rather than a merge chore.

`scripts/gate.sh` was **not** run. In this repo running it *is* the landing — past `:167` it merges to local `main`, then `npm install`, `npm run build`, and `git push origin main`. `AGENTS.md` says so explicitly and this skill's own step 5 says do not merge, so the assembly was gated with the lane-level checks instead. **Not merged — that is the operator's step.**


gabriel-canaan (approved):
Reviewed commit-by-commit against #30 / #31 / #33, on `dafd525`, in a clean worktree on node 22.22.2. Everything below is labelled **EXECUTED** (I ran it and read the exit status) or **REASONED** (argued from the code, nothing run).

## What verified

**Every number in the PR body reproduces exactly.** EXECUTED:

| claim | measured here |
|---|---|
| `VITEST_MAX_WORKERS=6 npm test` → 376 files, 6402 passed, 3 expected fail | **identical** (`main` at 4140f6b: 375 / 6356 / 3) |
| `npm run typecheck` clean, 5 workspaces | clean |
| `npm run lint` clean, 843 files | clean, 843 files |
| timing set 6 files, unchanged from `main`, no ratchet movement | **identical 6-file set** on both trees, derived read-only with `gate.sh:119-120`'s own two greps. `concurrency.test.ts` is correctly absent, so #48's trap really was avoided — and `.swarm/timing-count` does not exist yet, so there was no floor to raise |

**The mutations reproduce to the exact count.** EXECUTED, all four:

| mutation | PR claims | measured |
|---|---|---|
| `outcomes[index] =` → `outcomes.push(...)` (`concurrency.ts:92,94`) | 13 laws | **13 failed / 13 passed** |
| `Math.min(limit, tasks.length)` → `tasks.length` (`concurrency.ts:101`) | 12 laws | **12 failed / 14 passed** |
| `ensureHandle()` reopens every call (`session-log-writer.ts:134`) | 3 count laws | **3 failed** — `opens the file once no matter how many events`, `closes the held descriptor on sync()…`, `an append issued after sync() reopens…` |
| `sync()` no longer nulls `this.handle` (`session-log-writer.ts:116`) | "opens at 1 across five cycles instead of 5/5" | **4 failed** — including two rotation laws in `session-recorder.test.ts`. Stronger than claimed |

**The must-fix from verify on #30 is genuine, and both halves of the key are independently load-bearing.** EXECUTED — I mutated each half separately at `lane-index.ts:563`:

- drop `size` → `× re-parses on a size change with mtime provably identical — the size half of the key, on its own`
- drop `mtimeMs` → `× re-parses on a same-size edit if mtime moved` **and** `× a refreshed entry counts as freshly used, not still the oldest`

So `0b0ac3c` fixed a law that provably could not fail before it, and neither half is decoration.

**#30's load-bearing invariant holds.** REASONED, by reading: the cache's `stat === null` branch returns `EMPTY_SESSION_LOG`, which is byte-identical to what `readSessionLog`'s own `catch` returned (`session-log.ts:44`) — so an unreadable file degrades exactly as before. And the pruned-recording path that produces `unreadableSessionIds` runs off `events: null` from `listCapturedSessionIds` (`lane-index.ts:681-690`), which the cache never touches. The issue's "must degrade through the existing `unreadableSessionIds` path" is satisfied structurally, not by luck.

**The single-flight guard is sound, and for a simpler reason than the comment gives.** REASONED. The comment at `lane-index.ts:509-515` argues from Node draining one I/O completion's continuation before the next. It doesn't need to: between `await statOrNull(...)` resuming and `this.pending.set(filePath, load)` there is **no await at all** — `populate()` runs synchronously to its own first await and returns. The property is straight-line, not scheduler-dependent. Not a defect; the justification is just heavier than the code needs.

**Two findings I generated and then killed** — recording them so they don't get re-raised:

1. *The narrowed `(?<!handle\.)` lookbehind (`namespace-law.test.ts:197`) excludes any identifier literally named `handle`, so `const handle = fsp; await handle.writeFile(arbitraryPath, …)` slips past the regex.* True of the regex — and **irrelevant**, because I injected exactly that and **8 behavioural laws failed**, including the one named `that fence bites — a write anywhere else in the tree would be caught`. For comparison, injecting `await fsp.open('/tmp/attacker-controlled-path')` fails **9**. The narrowing is not a hole. EXECUTED both.

2. *A `sync()` racing an in-flight `append()` could close the handle between `ensureHandle()` resolving and `handle.appendFile()` being called → `EBADF`, a silently lost event.* **Not reachable.** EXECUTED: 200 appends interleaved with 67 `sync()` calls, plus 50 append→sync→append triples — zero rejections, all lines on disk, three probe laws green. The reason is worth stating because it is load-bearing and undocumented: `sync()` reads `this.handle` and nulls it with **no await between the two statements** (`:115-116`), and any append registered on the same `tail` necessarily runs after `sync()`'s own reaction to it. `closeWith` also takes the seal before the append, so the recorder never issues one concurrently.

The three disclosed shortfalls are all honest, and I have nothing to add to them: the residual `buildLaneIndex` fold, #32's refusal, and #46/#48 as filed follow-ups. The report of "#33's first verify pass shared a model with its build" is the kind of thing most reports would have quietly dropped.

## One finding — filed, not blocking: #55

**A clean stop never releases the session log's descriptor.** EXECUTED.

`runServerCommand`'s `stop()` (`packages/server/src/cli/run.ts`) is `clearInterval` → `pollLoop.stop()` → `app.close()` → `removeSessionLock()`. Nothing syncs or closes the recorder. Since this PR the writer *holds* the descriptor and releases it only in `sync()`, whose only other caller is rotation — so the session open at shutdown is closed by **garbage collection**:

```
Warning: Closing file descriptor 22 on garbage collection
[DEP0137] DeprecationWarning: Closing a FileHandle object on garbage collection is
deprecated. … In the future, an error will be thrown if a file descriptor is closed
during garbage collection.
```

Per-directory counts of that warning, one vitest process each — `main` emits none:

| suite | `main` (4140f6b) | this PR (dafd525) | with `await recorder.sync()` in `stop()` |
|---|---|---|---|
| `src/api/` | 0 | 4 | 4 |
| `src/cli/` | 0 | **1** | **0** |
| `src/log/` | 0 | 1 | 1 |

(Full-suite counts move with GC timing — 2 on one run, 8 on another — so the per-directory numbers are the measurement.) The `api/` and `log/` warnings are test-local: `api/doctor.test.ts`, `api/retarget.test.ts` and `api/stream.test.ts` each build a recorder and abandon it. Not a product defect.

**This is the sibling of #46, not a duplicate.** #46 is scoped to *a write that fails* after the handle is open. This is the normal path, on every clean stop. Rotation and retarget are both clean — `retargetSession` (`recorder/rotate.ts:250-272`) goes through `closeCurrentSession` → `closeWith` → `sync()` before `openSession` replaces the writer.

**Why I filed it instead of pushing it.** I did write and measure the fix — `await recorder.sync()` in a `try`, with `removeSessionLock` moved into a `finally` so a failing fsync can neither be swallowed nor strand the lock. It took `cli/` to 0 with typecheck, lint and the full suite still green. But `cli/run.ts` is outside all three of #30/#31/#33's fences, it has no test file of its own (the path is exercised by `cli/index.test.ts` and merely unasserted), and #46 already set the precedent that this family goes to wave 2. Adding a substantive shutdown-path change to a bundle two independent verify passes had already cleared is what "a bundle is not a grab-bag" is about — and this repo's own rule is that nobody reviews their own fix into `main`. So it is **#55**: fenced to `cli/run.ts` + `cli/index.test.ts`, blocked by #31, with the measurement and the patch in the body.

## Not run

`scripts/gate.sh` — correctly, and not by me either: in this repo running it *is* the landing. I derived the timing set with its own two greps, read-only.

## Verdict

Approving on `dafd525`. Three issues, one commit each, every claim in the body independently reproduced, and the only new finding is a disclosed-family sibling that by the repo's own fence rules cannot land here. Landing is the operator's step.

gabriel-canaan:
## Fence widening — recorded before the change

Reviewer, taking the fix mandate on the finding in my review above (the clean-stop descriptor release). Two files, neither claimed by any open issue's fence nor by PR #54:

- `packages/server/src/cli/run.ts` — the fix. Outside #30/#31/#33's fences, and outside `session-recorder.ts` (prd-40 #3's live claim), which is not needed: `SessionRecorder.sync()` is already public at `session-recorder.ts:133`.
- `packages/server/src/cli/run-descriptor-law.test.ts` — **new file**, for the law.

A new law file rather than an edit to `cli/index.test.ts`: the law needs a module-level `vi.mock('node:fs/promises')` to count opens and closes, and `cli/index.test.ts` is 1,846 lines with no mock in it today. Scoping that mock to a file of its own keeps it off 1,846 lines of unrelated tests.

This supersedes **#55**, which I filed before deciding to fix it here. I will close it against the commit rather than leave a duplicate open.

My approval above stands for `dafd525`. The new commit is **not** covered by it — per this repo's own rule I do not review my own fix into `main`, so @ciaran-slow, this one needs your confirmation before it lands, and the operator's gate after that.

gabriel-canaan:
## Fixed in `ddf7e41` — and a retraction

The finding from my review is fixed on this branch: `fix(cli): a clean stop releases the session log's descriptor`. Two files, exactly as recorded above — `cli/run.ts` and the new `cli/run-descriptor-law.test.ts`.

### Retraction: the DEP0137 table in my review is withdrawn

I reported the Node "closing a FileHandle on garbage collection" warning as a **count that moved with the fix** — 0 on `main`, 1 on this PR's `src/cli/`, 0 with the patch. **That count is not a measurement.** Whether GC runs before a vitest worker exits varies between otherwise identical runs, in both directions. Isolating it in one worktree, changing only the patch:

| | `src/cli/` DEP0137 count |
|---|---|
| this PR + fix, no law file | **1** |
| this PR, **no** fix, no law file | **0** |
| this PR + fix + law file | **0** |

That is the opposite of what I posted, from the same code. I had also measured each test file individually and found none emitting it while the directory reliably did — which should have told me the metric was about GC timing and not about descriptors. It did not, until I isolated it. The table, and the "`main` emits none" line beside it, are both withdrawn.

### What the evidence actually is

The defect is real and the code fact was never in doubt: `stop()` was `clearInterval` → `pollLoop.stop()` → `app.close()` → `removeSessionLock()`, and nothing in it touched the recorder, while since #31 the writer releases its descriptor only in `sync()`. What was missing was a *counted* proof, which is what the repo asks for and what the warning was standing in for.

`run-descriptor-law.test.ts` counts `.jsonl` opens against closes through a `node:fs/promises` mock — the technique `session-log-writer.test.ts:16-33` already uses. Deterministic in both directions, 5 consecutive runs each:

- with the fix — **3 passed**, 5/5
- without it — **3 failed**, 5/5

And the two mutations:

| mutation | result |
|---|---|
| drop `await recorder.sync()` (the pre-fix behaviour) | all **3** laws fail |
| `try`/`finally` → a plain two-statement sequence | exactly **1** fails — the fsync/lock one; the other two stay green |

That second one is the sibling case, and it is why this is a `finally` rather than a third statement: `removeSessionLock` was the last thing `stop()` did, so an `await recorder.sync()` in front of it would let a failing fsync strand the lock for the next boot. The `finally` keeps both promises — a real fsync error still reaches the caller, and it is never the reason a stale lock outlives the process.

Gated on node 22.22.2: `npm run typecheck` clean, `npm run lint` clean (844 files), `VITEST_MAX_WORKERS=6 npm test` **377 files / 6405 passed / 3 expected fail** (376 / 6402 / 3 before — the three new laws).

### Where that leaves the PR

My approval covers `dafd525` and **not** `ddf7e41` — I do not review my own fix. @ciaran-slow, this needs your confirmation before it lands; everything in the bundle itself verified and I have nothing outstanding against it. #55 is closed as superseded.

Issue #53, opened 2026-08-24
prd42 w6: four tidy-ups left by wave 2


gabriel-canaan:
Four small things the round-2 verification of prd42 w2 found. None blocks anything; all are cheap, and each is the kind of thing that reads as deliberate once it has survived a few readings.

## 1. Two Windows-illegal fixture names, in a test with no platform guard

**RE-GROOMED 2026-08-27, after prd42 w5 landed (PR #121).** This item used to
name a `colon:segment` directory in collectors/sessionlog/worktree-slug.test.ts.
That file no longer contains it — w5 moved the round-trip law into the concierge
and inverted the colon case from an honest refusal to a real round trip. The
defect survived the move; only its address and its size changed.

Where it lives now, concierge/repos.test.ts around line 263, inside the
generator the round-trip law iterates:

```
const ROUND_TRIP_SEGMENTS = [
  'plain-word', 'dotted.segment', 'under_score',
  'a space here',
  'a:colon here',        <- illegal filename on Windows
  'a\\backslash here',   <- also illegal, as a path separator
]
```

The law `mkdir`s a real directory per segment. **A colon cannot appear in a
Windows filename, and a backslash inside a segment is a path separator there**, so
both `mkdir`s fail outright — two illegal fixtures now, where this item
originally described one.

It is harmless today: there is no `windows-latest` leg in ci.yml. But AGENTS.md
names that leg as the real coverage gap (prd-25), and a built clone could not
boot on Windows at all until #281, so the leg is wanted. A guard now is cheaper
than a red leg for whoever adds it, and cheaper than the confusion of a fixture
failing for a reason unrelated to what it tests.

**The guard already exists in that file and is declared too late to use.**
`const isPosix = platform() !== 'win32'` sits at repos.test.ts:590, while the
test needing it is the `it()` at :248. Hoisting the declaration is part of this
item, not a separate tidy-up.

**Do not "fix" this by deleting the two segments.** They are the whole point of
w5 — the law asserts they round-trip, and the non-vacuity floors beside it assert
the generated set actually contains them. Guard the test, or skip those two
segments on win32 while keeping the floors honest about what was generated.

## 2. `RetargetInFlightError` is not in the `recorder/index.ts` barrel

So `api/retarget.ts` now imports from **both** the barrel and `../recorder/rotate.js` for the same module. The barrel enumerates every other public symbol including types. One line.

## 3. The refusal prose is written twice

[...]

Issue #52, opened 2026-08-24
prd42 w6: a test reaches the concierge without computing a specifier


gabriel-canaan:
## The defect

`concierge/namespace-law.test.ts` clause 1 sweeps every file for imports reaching into `concierge/`, with a declared-importer set of `api/concierge.ts` alone and **no test-file exclusion**. Clause 2 exempts test files, but only for its own "no blind spot" rule.

So a test that legitimately needs `reverseProjectSlug` cannot import it. prd42 w2's round-trip law (`collectors/sessionlog/worktree-slug.test.ts`) reaches it through a **computed specifier** — `['..','..','concierge','repos.js'].join('/')` — precisely so clause 1's text scan cannot see the edge.

Both round-1 and round-2 verification passes checked both clauses and agree the lane's reading is **accurate** and the workaround is **necessary as the law is currently written**. This issue is not about that lane being wrong. It is about the law now having an undocumented escape hatch, and prd42 w2 being its first user.

## Why it matters

The law's own doc names this as its worst failure mode:

> *"a fully computed specifier (`import(modulePath)`) is invisible to any regex at all… an unanalysable edge would let the reachability check pass while knowing nothing"*

The precedent is now set: any test can leave the reachability graph by computing a string, and the only record that this was a considered decision is a comment inside one test file. The next person to hit clause 1 will find the workaround before they find the reasoning.

**Second cost, concrete:** a computed specifier is typed `any`, so the law had to hand-write `ReverseProjectSlugResult` as a second copy of `reverseProjectSlug`'s return type — unchecked by `npm run typecheck`. That is a second spelling of one fact, inside the PRD whose title is *one path, one spelling*. If the real return type changes, the law keeps compiling against the stale copy.

## The cheaper lawful route, for whoever takes this

Clause 1 skips the concierge directory outright (`if (isInConcierge(file)) continue`), so a round-trip law living in `concierge/repos.test.ts` needs no dodge, no cast and no hand-copied type. Either move it, or amend clause 1 to carry a declared **test**-importer set the way it already carries a production one.

## Also: a coupling point nobody had recorded

This is a file that a change **elsewhere** forces an edit to, which is exactly what `.swarm/coupling.txt` is for, and it is not in there. Importing the concierge module from anywhere new reddens a law in a directory the lane never entered. It could not be added while prd42 w2 was in flight (editing a tracked file in the main checkout while a wave runs is what `AGENTS.md` forbids outright), so it is recorded here instead.

[...]

Issue #51, opened 2026-08-24
prd42 w4: the served page's symlink verdict is pinned


gabriel-canaan:
## The defect

prd42 w2 (#13) replaced `static.ts`'s prefix comparison with the shared containment primitive. That closed a real hole — a symlink inside `dist` resolving outside it — and also **widened** the route in the opposite direction, which nobody asked for and nothing tests.

A path that is textually **outside** the dist root but resolves **inside** it, via a symlink pointing into `dist`, went from **403 to 200**.

Found by the round-2 verification pass while enumerating every input where the old and new checks disagree. Reproduced there via `/..%2f<basename>-sibling/assets/app.js`.

## Why it is arguably fine, and why it still needs a decision

It is the primitive's honest semantics: `isInside` canonicalizes both sides, and after canonicalization the path *is* inside the root, so nothing outside `dist` is ever read. On that reading the new answer is more correct than the old one, and the old 403 was an artefact of comparing unresolved strings.

But:

- **It is a permissive change to a permanently tokenless route** (`GET /*`, ADR-0012, prd-29 ruling 1), and permissive changes to that route deserve to be deliberate rather than incidental.
- **Nothing asserts it either way.** `static.test.ts:152` asserts the *sibling-directory* case is refused; there is no test for sibling-symlink-into-dist. So the behaviour is currently whatever the primitive happens to do, and a future change to `canonicalize` could flip it silently in either direction.
- The commit that introduced it does not mention it, because it was not the change being made.

## Definition of done

- A test pins the outcome explicitly, whichever outcome is chosen, with the reasoning in a comment — so the next reader knows it was decided rather than inherited.
- If the answer is "refuse", the refusal is expressed through the containment layer rather than by re-adding a string comparison (ruling 2: containment has one implementation).
- The PR states it as a behaviour change, the way #13's 200→403 change was stated.

## Fence (may touch ONLY)
- `packages/server/src/server/static.ts`
- `packages/server/src/server/static.test.ts`

packages/server/src/paths/containment.ts and packages/server/src/log/transcript-attribution.ts are consumed, never edited — EXPLICITLY OUT of this claim.

## Blocked by
Nothing — prd42 w2 landed (#13 in PR #54). This is about the code that wave introduced.
Sibling lane holds recorder/rotate.ts, api/rotate.ts and their tests (#49); do not enter them.
**Model:** sonnet. **Wave:** 4.
[...]

Issue #49, opened 2026-08-24
prd42 w4: a rotation refuses a retarget's boundary


gabriel-canaan:
## The defect

`POST /api/rotate` issued while a retarget is in flight returns **200 carrying the retarget's boundary** — a `Rotation` whose `closed.reason` is `'retargeted'` and whose `opened` session is in a **different repo** than the caller asked about.

This is the argument prd42 w2 (#14) makes against coalescing, pointed the other way. #14's commit body refuses coalescing for retarget because *"asked to move to B, told 'moved to A', status 200"*. The same input class now exists for rotation: asked to rotate A, handed a boundary into B, status 200.

## Why it exists

#14 made `rotateSession` and `retargetSession` share one in-flight `WeakMap` keyed by recorder (`recorder/rotate.ts:219`) — correctly, since that is what stops a retarget interleaving with a rotation. The two differ in what the *second* caller gets: retarget refuses, rotation coalesces. Rotation's coalescing is a considered decision with its rationale in its own doc — but that rationale (`rotate.ts:203-207`) is about **two rotations**: *"two operators asking at once is ONE boundary, not two — the second caller awaits the first's answer instead of closing a session that is a millisecond old."* It does not carry to a rotation coalescing onto a **repo move**.

`api/rotate.ts` cannot tell the difference: it receives a `Rotation` and has no way to know the boundary it coalesced onto was a retarget.

## What it does and does not cost — adjudicated, not assumed

Reviewed by two independent model passes plus both cross-reviews. One pass claimed durable corruption; the other refuted it, and the refutation was upheld against the code:

- **Does NOT happen:** durable snapshot-store or boot-metadata corruption. Boot metadata is a `WeakMap` (`api/meta.ts:80`) that the retarget route overwrites with `'retargeted'`, and the poll loop is stopped for the duration, so the transiently wrong snapshot directory is never written before the retarget's `reset` replaces it.
- **Does happen:** the response body is a lie to its caller, and `api/rotate.ts` runs its own reset and `lastBootReason: 'rotated'` path against a boundary that was not a rotation. The coalescer returns the **raw** inner promise (`rotate.ts:220`) while the retarget route awaits the `.finally()`-derived one, so rotate's continuation deterministically resumes *before* the retarget re-points context — this is the determined ordering, not a possibility.

[...]

PR #54, opened 2026-08-24
prd42 w2: one path, one spelling — the space, the served page, and the retarget boundary

### LINKED ISSUE #12: prd42 w2: a path with a space finds its transcripts
## Fence (may touch ONLY)
- `packages/server/src/collectors/sessionlog/worktree-slug.ts`
- `packages/server/src/collectors/sessionlog/worktree-slug.test.ts`

The last two are **comment text only** — no logic, no assertions, no exported behaviour. See the widening note below.

## Blocked by
Nothing. #11 was absorbed into this issue at grooming (see below) — this issue now carries both the law and the fix.

Sibling lanes hold other surfaces: the served-page containment lane (#13) and the retarget pair (#14). All three are the same wave and land as ONE PR.

## The defect
The encoder maps a path to a project slug with one character class; the reverse walk decodes with another. They differ by a single character — a space. For any repo whose path contains a space, **every transcript is invisible, forever**, and the failure is indistinguishable from an agent that never started.

The direction was settled by probe, not argument: a real session started from a directory whose name contains a space produced a slug in which the space had become a dash. So Claude Code's own slugger maps it, the reverse walk is right, and **the encoder is wrong**.

## Definition of done

**1. The round-trip law, written first.**
- A property test over generated paths including spaces, dots, underscores and colons: encode, then walk back through the concierge module's reverse form, and get the original back.
- The law asserts the **ROUND TRIP**, never the character class's literal contents. A literal assertion passes at any wrong-but-matching pair — which is exactly how this defect survived.
- The generated set **must** include a segment with a literal space. That is the case the law exists for.

**2. The fix.**
- The encoder's character class gains the space.
- The pinned expectation at line 66 of the encoder's own test moves in the SAME commit — a path and its encoding are one fact written twice, and changing one leaves a test that passes while no longer testing the encoding.
- The two stale doc comments move too (see the widening note).

**3. The evidence.**
- Write the law, run it against the unfixed encoder, and **capture it failing**. Then fix and capture it passing. Both transcripts go in the commit body. That is what the red keystone was for, and a mutation transcript is stronger evidence than a red CI run because it names the reason.
- The PR states explicitly that on-disk slug directory names change, and that no migration is needed because a spaced-path slug never resolved to anything.

## How the reverse half is reached — verified at grooming, 2026-08-24
- The reverse walk takes an optional filesystem adapter defaulting to the real one, and walks the **real filesystem** one hop at a time. So the law builds real directories (mkdtemp + mkdir) and calls it with that default.
- **Do not import the fixture filesystem helper** from the concierge test — it is private to that file, and reaching for it buys nothing since the law wants a real tree anyway.
- **Canonicalise the temp root before encoding.** os.tmpdir() is a symlink on macOS (/var → /private/var) and is not on Linux, so a round trip that encodes the raw path and walks back to the canonical one **passes on ubuntu and fails only on the macOS leg**. Resolve the root once, up front, and generate the cases beneath it. AGENTS.md names this leg as the one that carries signal for path-shape bugs.

## #11 absorbed here at grooming, 2026-08-24 — and why there is no wave 1
#11 held the round-trip law as a zero-claimant wave-1 keystone, specified to land **RED**. Two findings closed that off:

- **It shares a file with this issue.** Both claim the encoder's test. Under the wave contract that is a stack, not a bundle, and prd-39's grooming already settled the one lawful resolution: *"a stack in one file has exactly one lawful bundling — same wave as one issue."* prd-42's own Sequencing says every wave follows prd-39 wave 1, so this is that precedent applied, not a new rule.
- **A red keystone cannot land.** `scripts/gate.sh` runs the suite and exits non-zero on failure, so a PR whose whole content is a failing test has no route to main. The red demonstration was never a landing state — it was evidence, and evidence belongs in the commit body.

Consequence worth naming so it does not read as an error later: **prd42 has no wave 1.** Wave 2 is #12 + #13 + #14 (one PR, three commits); wave 3 is #15 + #45 (one PR, two commits). The wave numbers are not renumbered, because #13 and #14 were already dispatched carrying "w2" in their briefs and the board is not worth desynchronising to close a cosmetic gap.

## Widening recorded at grooming, 2026-08-24 — before the change, not after
The concierge repos module and its test both say, in a doc comment, that the forward transform *"maps only slash and underscore to dash"*. That is **already wrong at HEAD** — it also maps dot, backslash and colon — and this issue makes it wronger. The change is what invalidates the sentence, so it carries the fix.

Checked at grooming: nothing reads that module's source off disk, and no snippet-anchored law matches the wording, so the edit cannot redden a test from another directory.

## PRD open question 1 — CLOSED by enumeration at grooming, 2026-08-24
*"Does any tracked fixture encode a slug by hand? Not yet enumerated."* Now enumerated, by sweeping every tracked ts/tsx/json/md file for dash-encoded path literals and for path literals containing a space:

- The **only** hand-written *forward*-encoding expectation in the tree is line 66 of the encoder's own test — inside this claim.
- The concierge repos test uses a spaced path at lines 97-107, but exercises the **decoder**, which already maps the space. This change does not move it.
- Every other call site computes its slug by calling the real helper, so it moves with the change. **One exception, filed as #45:** the CLI index test hand-rolls a fourth, narrower copy at line 250. That file is claimed by wave 3, never here.
- **No fixture directory names need migrating.**

### LINKED ISSUE #13: prd42 w2: the served page decides containment with the shared primitive
## Fence (may touch ONLY)
- `packages/server/src/server/static.test.ts`

## Blocked by
Nothing.

packages/server/src/paths/containment.ts is consumed, never edited. **Do NOT gate this route:** `GET /*` stays tokenless forever as ADR-0012's bootstrap, and that route class is owned by prd-29 ruling 1. Fixing containment must not drift into changing who may read it. static-csp.test.ts belongs to nobody here — checked at grooming, it asserts no 403 and no containment behaviour, so this change cannot redden it.

## Definition of done
- The `startsWith` prefix comparison at static.ts:119 is replaced by `canonicalize` + `isInside`.
- Test: a symlink placed inside a temp dist pointing outside it is refused.
- The route still answers 200 for an ordinary asset, and its `routeClass` is unchanged.
- The two existing containment tests still pass **for their own reasons**, not incidentally.

## Claim widened at grooming, 2026-08-24 — before the change, not after
The original claim was static.ts alone, but the done-when requires a new symlink test and the test file was outside it. A lane would have stopped on its first test write. Added static.test.ts, which is where the existing containment coverage already lives:

- static.test.ts:152 — *"refuses a sibling directory that merely shares the root as a literal string prefix"*. Its comment names the exact expression this issue replaces, so the prose and the code are one fact: if the implementation stops being a string-prefix check, that comment must move with it.
- static.test.ts:140 — the traversal case, which lands on the SPA fallback with 200 because the URL is normalised before routing. **Keep that outcome.** It is asserted deliberately, and a canonicalising containment check must not turn it into a 403.

## The mutation this must survive
Reverting static.ts to the prefix comparison must make the NEW test fail. A symlink test that passes against the old check is testing nothing — the old check already refuses a path whose resolved string leaves the root; what it misses is a path that stays *textually* inside while resolving outside, which is precisely what a symlink under dist does. Build the case so the resolved-but-uncanonicalised path still starts with the root plus a separator, or the test cannot fail for the reason it claims.

### LINKED ISSUE #14: prd42 w2: a retarget refuses to overlap itself
## Fence (may touch ONLY)
- `packages/server/src/api/retarget.test.ts`
- `packages/server/src/api/retarget-law.test.ts`

## Blocked by
Nothing.

Sibling lanes hold other surfaces: the slug encoder, and the served-page containment lane. packages/server/src/api/index.ts is prd-29's — its route classes are not this issue's, and none needs to change. packages/server/src/api/rotate.test.ts is deliberately out of scope: rotateSession's own signature and behaviour do not change here. If it reddens, stop and ask for a widening rather than editing it.

## Definition of done
- The retarget route holds an in-flight boundary that **shares rotation's map**, so a retarget cannot interleave with a rotation, nor a rotation with a retarget.
- An overlapping retarget is **refused**, never queued and never coalesced.
- Test: two concurrent retarget requests → one proceeds, one is refused, one session directory results.
- retargetSession's doc comment no longer assigns a concurrency obligation to a caller that does not exist.

## Ruling recorded at grooming, 2026-08-24 — read this before writing code
Two findings from reading the tree change what "the same guard" can mean. Both are EXECUTED (read at HEAD), not inferred.

**1. rotateSession's guard COALESCES — it does not refuse.** At rotate.ts:201-215, the in-flight map returns the *first caller's promise* to the second, and its own doc says so deliberately — *"two operators asking at once is ONE boundary, not two — the second caller awaits the first's answer instead of closing a session that is a millisecond old."*

Copying that shape onto retarget would **fail this issue**, and worse than plainly: two retargets to *different* repos would have the second silently receive the first's result — asked to move to B, told "moved to A", status 200. That is not a refusal, and the PRD's rejected alternative (*"a queued retarget writes a session into a repo the operator has already moved away from"*) applies to a coalesced one with equal force.

So: **share the map, not the behaviour.** One WeakMap keyed by recorder, living in recorder/rotate.ts, consulted by both entry points. Rotation keeps coalescing — that is a considered decision with its rationale in its own doc, and this issue must not change it. Retarget refuses. Sharing the *map* is exactly what buys the no-interleaving half of the requirement; sharing the *behaviour* would forfeit the refusal half. The PRD's ruling 3 phrase "shared rather than duplicated" refers to the map.

**2. PRD open question 3 is answered: 409, and NOT 423.** The question was framed as "409 matches the repo's existing vocabulary; nobody has ruled on it." Reading the route settles it — **this route already returns 409 at three sites** (retarget.ts:130, :170, :183), and each carries a discriminating `code` field typed `RetargetRefusalCode`. 409 is this route's refusal *status*; the code is what tells the refusals apart. A 423 would be the only status in the route that means "refused" without joining that vocabulary.

Add a fifth code — `'retarget-in-flight'` or similar — to the union at retarget.ts:107. Checked at grooming: that type is declared there, used only at :171 and :184, and has **no consumer anywhere else in the monorepo** (the web package included), so extending it is wholly inside this claim and breaks no renderer.

## Claim widened at grooming, 2026-08-24 — before the change, not after
The original claim was the two source files, but the done-when requires a concurrency test and every candidate test file sat outside it. Added recorder/rotate.test.ts, api/retarget.test.ts and api/retarget-law.test.ts — the last because a law over this route may well enumerate its refusal codes, and adding a fifth is exactly the edit that would redden it.

## The mutation this must survive
Removing the guard must make the concurrency test fail. Two awaited sequential calls prove nothing — the race only exists across the `await ctx.pollLoop?.stop()` at retarget.ts:144-237, where the from/to pair is snapshotted before the yield. Fire both requests without awaiting the first, and assert on the count of session directories that result, not only on the status codes.

gabriel-canaan:
prd-42 wave 2 — the three code fixes. One commit per issue, reviewed commit by commit.

Closes #12
Closes #13
Closes #14

**Note on shas:** this branch is cherry-picked from three lane branches, so its commits (`4fc97f3`, `e58ce91`, `e02c140`) are not the shas the issue comments cite (e.g. #13's report names `33e7329`). Same work, re-parented onto current main.

## What becomes true

- **`4fc97f3` (#12)** — a repo path containing a space finds its transcripts. The encoder's character class gains the space, its pinned expectation moves in the same commit, and a round-trip law proves it rather than asserting the class's contents.
- **`e58ce91` (#13)** — the served page decides containment with the shared primitive, fails closed when it cannot decide, and reads the path it actually checked.
- **`e02c140` (#14)** — a retarget holds its own boundary, refusing rather than coalescing, and acquires it before anything else in the route.

## Verification

Verified twice, independently, before this PR existed. **Round 1's reviewer passes could not execute anything** — a misconfiguration on my side, since headless `-p` denies any command needing approval — so their findings were reasoning-only and the two models disagreed on severity for the most serious one. Round 2 gave each reviewer its own disposable worktree with execution enabled; both then confirmed everything by running it, and found four defects round 1 had missed.

Four must-fixes came out of that and are fixed in these commits. Each is proven by a mutation that goes red:

| what | mutation | result |
|---|---|---|
| #13's containment throwing 500s and leaking the install path | revert to the raw `isInside` | new tests red |
| #13's TOCTOU (checked one path, read another) | — | closed by construction: one `canonicalize`, all reads use it |
| #14's refusal re-arming the poll loop | re-add `pollLoop.start()` on the refusal path | 3 tests red |
| #14's stale `from` snapshot outrunning the guard | make the boundary never refuse | 6 tests red |
| #12's decorative colon assertion | delete the colon fixture's `mkdir` | 1 test red (passed 11/11 before) |

Gates on this merged branch, run against node 22.22.2: `npm run typecheck` 0, `npm run lint` 0, `VITEST_MAX_WORKERS=6 npm test` 0 (375 files, 6366 passed, 3 expected fail), `npm run build` 0.

## What is the sibling case?

Answered per commit in each body; the three that mattered:

- **#12** — the commit names `cli/index.test.ts:250`'s fourth copy of the transform (filed as #45) but the unnamed sibling was **backslash**, which fails identically to the colon it did name. Both are now stated. The deeper sibling is the premise itself: Claude Code's real transform is `[^a-zA-Z0-9]` with a 200-character cap and a hash suffix, read out of the shipped binary — so the encoder's class is a documented *subset*, and the comments now say so instead of claiming the space was the last gap. #47 carries the rest.
- **#13** — the commit fixed a symlink inside `dist` resolving outside it, and missed that **`isInside` throws where the old string check could not**. The repo's own fail-closed wrapper, `isPathContained`, was one import away — and `static.ts`'s original comment already named the file that defines it. Now imported.
- **#14** — the guard covered `retargetSession`; the sibling was **the other side of the shared map**, and **the route around it**. Both addressed: rotation-onto-retarget is filed as #49, and the route now acquires the boundary before it reads or stops anything.

The table above is the answer, and every entry was executed rather than argued. Two specifics worth naming because they were the weak points:

- #12's round-trip law asserts the **round trip**, never the class's literal contents — a literal assertion passes at any wrong-but-matching pair, which is how the original defect survived. It also canonicalizes the temp root before encoding, so it cannot pass on ubuntu and fail only on the macOS leg.
- #12's colon assertion originally could not fail for the reason it claimed: `toBeNull()` was satisfied equally by "the walk cannot re-encode `:`" and by "that directory does not exist". It now asserts the fixture exists *and* that the refusal reason names the colon segment.

## Fence widenings

Both recorded on their issues **before** the changes, per the runbook:

- **#12** gained `concierge/repos.ts` and `concierge/repos.test.ts`, comment-only — this change is what invalidated their doc comments, so it carries the fix.
- **#13** gained `server/static.test.ts` — the done-when required a new test and the test file sat outside the original claim.
- **#14** gained `recorder/rotate.test.ts`, `api/retarget.test.ts` and `api/retarget-law.test.ts`, for the same reason.

## Deliberate behaviour change, stated not slipped

**`GET /index.html/x` (any URL whose prefix names an existing file) now answers 403 where it fell through to the SPA shell at 200.** A path that cannot be canonicalized cannot be vouched for, so it is refused rather than served. This was **not** in #13's Definition of done and was signed off by the operator explicitly. Controls verified unchanged at 200: `/index.html`, `/assets/app.js`, `/lane/abc`, and `/../../etc/passwd` — that last one deliberately, since it is normalised before routing and its 200 is asserted on purpose.

Also stated, from #12: **on-disk slug directory names change** for any path containing a space. No migration is needed, because a spaced-path slug never resolved to anything, so nothing exists under the old spelling.

## Not closed here — filed, not left hanging

- **#47** (Urgent) — the real transform is every non-alphanumeric, plus a 200-char cap and an override. Read from the binary; the probe it was waiting on is done.
- **#49** (High) — a rotation asked during a retarget is handed the retarget's boundary. Includes a correction: an earlier reasoning-only adjudication claimed no durable corruption, and execution refuted it.
- **#51** — #13 also *widened* the route: a path textually outside root that resolves inside via a symlink into `dist` went 403 → 200, untested either way.
- **#52** — #12's law reaches the concierge through a computed specifier the reachability law cannot see, with a hand-copied return type; also carries the missing `.swarm/coupling.txt` entry.
- **#53** — four tidy-ups, including that `colon:segment` will break a future `windows-latest` leg.

## Reviewer note

I briefed all three fixes, so I should not be the one approving this — `AGENTS.md` is explicit that the reviewer fixes but never reviews their own fix into main. The three lane worktrees are still live if anything comes back.


ciaran-slow:
## Review of #54 — reviewer: @ciaranslow

Read commit by commit against #12, #13 and #14. **One substantive fix pushed as `5fa84f4`**; everything else verified and holds.

Environment note up front: I ran on **node v22.23.2**, not the v22.22.2 in your gate report — v22.22.2 is not installed here and the repo's `engines` floor is `^22.22.2`. Same major line, but stated rather than glossed.

---

### First: what verified in your favour

Every mutation your commit bodies claim, I re-ran myself. All five reproduce **exactly** as written — this is the cleanest evidence trail I have reviewed on this repo.

| # | mutation | your claim | what I measured | verdict |
|---|---|---|---|---|
| #12 | revert the encoder class to `/[/_.\\:]/g` | round-trip law red | `2 failed \| 9 passed (11)` — the law *and* the pinned tilde/space expectation | **stronger than claimed** |
| #12 | delete the colon fixture's `mkdir` | 1 red (was 11/11 before) | `1 failed \| 10 passed (11)`, `AssertionError: expected [ 'a space here', …(2) ] to include 'colon:segment'` | **confirmed, verbatim** |
| #13 | revert to the raw `isInside` + raw `requested` | 2 new tests red | `2 failed \| 17 passed (19)`, both `expected 500 to be 403` | **confirmed, verbatim** |
| #13 | the **original** `startsWith` prefix check (the mutation *the issue* demands) | not claimed | `3 failed \| 16 passed (19)` — the symlink-escape test goes red too | **the symlink test is not vacuous** |
| #14 | re-add `pollLoop.start()` on the refusal path | 3 tests red | `4 failed \| 18 passed (22)` | **confirmed** (the 4th is my new one) |

All **EXECUTED**. Baseline on the branch as submitted, before my commit: **375 files, 6366 passed, 3 expected fail, exit 0** — precisely the numbers in your PR body.

Two more things I checked and want on the record:

- **The gate really ran.** All 7 checks are `SUCCESS`. `build-test-boot (macos-latest, min)` is absent from the check list, which is the shape AGENTS.md warns about — but it is a deliberate `exclude:` in `ci.yml`'s matrix (macOS bills 10×), so 7 *is* the complete expected set. No silently-skipped leg.
- **#13's deliberate 403.** `/index.html/x` → 403 with no path and no `ENOTDIR` in the body, controls unchanged. The pre-fix 500 leaked the absolute install path on the one permanently tokenless route; calling that out yourself, with the measured bodies, is the right call and I would not have found it from the diff alone.

---

### The one substantive finding — fixed in `5fa84f4`

**EXECUTED.** `api/retarget.ts` acquires the boundary before validation — correct, and your defect-2 test proves it. But it **released** the boundary at the end of the close/open, not the end of the route:

```
265: boundary.resolve(retarget)     ← released here
271: ctx.repoPath = to.repoPath
279: await ctx.pollLoop?.reset({…}) ← an await, with the guard DOWN
283: ctx.pollLoop?.start()
```

`await reset()` sits on the far side of the release, so the guard is down for the whole resume. That reinstates **defect 1's exact shape on the winner's own tail**: the poll loop is one object shared by every request, so a second retarget acquiring the boundary in that window calls `stop()`, taking the timer down, and the winner's still-pending `start()` re-arms it while another boundary is mid-flight — and `poll-loop.ts`'s `start()` fires a tick immediately when the timer is down.

Reproduced with a `reset()` gated to park the winner mid-resume, then firing a second retarget into that window:

```
× a second retarget arriving while the winner is still resuming its loop is refused
AssertionError: expected 200 to be 409

PROBE: loopLog phases = ["stop","reset","stop","reset","start"]
```

The loser's `stop` lands **between** the winner's `reset` and its `start` — an interleaved stop/start pair on the shared loop, which is what your own `expect(loopLog.map(phase)).toEqual(['stop','reset','start'])` exists to forbid. The 200 is the plainer statement of it: two retargets both ran to completion, overlapping.

The fix is moving one line — `boundary.resolve(retarget)` below `ctx.pollLoop?.start()` — plus the regression test. Nothing else moves: both refusal paths settle before touching the loop, the `performRetarget` catch still restarts the loop it stopped and releases via your outer safety net, and `rotate.ts`'s `retargetSession` is untouched (it owns no loop to resume).

With `5fa84f4`: **375 files, 6367 passed, 3 expected fail**, `npm run typecheck` 0, `npm run lint` 0.

**One trade-off you should decide on, since it is yours to make.** Releasing later means a throw from `reset()` now reaches `boundary.reject(err)` *effectively* rather than as a no-op, so a rotation that coalesced onto this boundary gets a rejection instead of the retarget's `Rotation`. Given #49 already holds that the coalesced-`Rotation` is a lie to its caller, a rejection seemed the more honest of two bad outcomes — but it is a behaviour choice and I would rather you confirm it than discover it.

---

### For #49, not for this PR — one input class it does not yet name

**REASONED**, not executed. #49 covers a rotation being handed a *successful* retarget's boundary. It does not name the **rejected** case: the reservation is now held from the top of the route, so a rotation can coalesce onto a boundary that is about to be refused (`already-watching`, or a failed validation) and receive `Error('refused: already-watching')`. `POST /api/rotate` then 500s and does not rotate at all.

That window is not small — it spans `validateRetargetTarget`, a real subprocess with a 5 s ceiling. Before this PR a rotation during a retarget simply rotated, so this is a widening, and arguably worse than #49's 200-with-a-lie because the rotation silently does not happen. Same root cause, same fix, and `api/rotate.ts` is out of #14's fence — so it belongs on #49 rather than here. I have not executed it; worth a route-level test when #49 is picked up.

---

- **`#12`'s computed specifier.** Reaching `reverseProjectSlug` through a runtime-computed specifier to dodge `namespace-law.test.ts` clause 1 is legitimate and the comment explains it well, but it is a reachability blind spot bought with a hand-copied return type. #52 already carries this; noting only that I agree it is worth doing rather than leaving.
- **`static.ts`'s SPA fallback** still uses raw `root`: `filePath = path.join(root, 'index.html')`, uncanonicalized and unchecked. Pre-existing, not reachable by a caller (it is the build's own output), and out of #13's fence — but it is the one read in that handler that no longer goes through the resolve-once discipline the rest of it just gained.
- **`retarget-law.test.ts`** was added to #14's fence for a refusal-code enumeration that turns out not to exist in it, so the widening went unused. Fences are "may touch", so this is fine — recording it only because the widening note reads as if the file needed the edit.

---

### What I did not run

- The `windows-latest` leg — it does not exist yet (prd-25). #53's note that `colon:segment` will break it when it lands is correct and I did not verify beyond reading.
- The intra-request TOCTOU race in #13. Your commit is straight about why it is not simulated, and I agree the mitigation is structural; I checked the structure (one `canonicalize`, every read on that value) and did not attempt a race test.
- `npm run build` / the packaging guard locally — CI's 7 green checks cover both, and I confirmed each job actually executed rather than reading the bar.

---

### Disposition

**Rung 2 (substantive) — not merging.** `5fa84f4` changes concurrency behaviour, and per AGENTS.md I do not review my own fix into main. Two things needed:

1. **Your confirmation** on `5fa84f4`, particularly the reject-on-`reset()`-throw trade-off above.
2. **A second pair of eyes** on that commit before it lands. All merges here are manual.

CI will re-run on the pushed commit — worth confirming all 7 legs go green again before anyone merges.

Everything else in this PR I verified and would approve as-is. The mutation-per-defect discipline in these three commit bodies is what made this reviewable in one pass instead of three.

ciaran-slow:
CI on `5fa84f4` — **all 7 legs green**, verified as a real run against that sha rather than a bar carried over from `e02c140`:

```
$ gh run list --branch prd42-w2-one-path-one-spelling --limit 3
5fa84f4  completed   success    2026-08-24T06:47:27Z
e02c140  completed   success    2026-08-24T06:32:01Z
```

```
SUCCESS  build-test-boot (macos-latest, current)
SUCCESS  build-test-boot (ubuntu-latest, current)
SUCCESS  build-test-boot (ubuntu-latest, min)
SUCCESS  pack-smoke (macos-latest, current)
SUCCESS  pack-smoke (macos-latest, min)
SUCCESS  pack-smoke (ubuntu-latest, current)
SUCCESS  pack-smoke (ubuntu-latest, min)
```

Seven is the complete expected set — `build-test-boot (macos-latest, min)` is a deliberate `exclude:` in `ci.yml`'s matrix, not a skipped leg. Since every `build-test-boot` leg reached `Lint` (the step after `Test`), the packaging guard and boot smoke really executed rather than being silently skipped behind a red `Test`.

Still needs the two things from my review above before merge: your confirmation on `5fa84f4` (especially the reject-on-`reset()`-throw trade-off), and a second reviewer on that commit — I wrote it, so I am not approving it.

Issue #47, opened 2026-08-24
prd42 w5: a colon and a backslash round-trip, and the law reaches the concierge without a dodge


gabriel-canaan:
## The defect

The forward transform and the reverse walk disagree on **two more characters than prd-42 knew about**.

- The encoder maps a colon and a backslash to a dash. At HEAD its class is dot, slash, underscore, backslash, colon — and after prd42 w2 (#12) also a space.
- The reverse walk re-encodes each real directory entry with **dot, underscore and space only**. It has never handled a colon or a backslash.

So for any repo path with a colon or a backslash in a segment, the encoder mints a slug the reverse walk cannot resolve. It reports the failure honestly — the walk's "unknown is not absent" contract holds, and the operator gets a reason rather than silence — but the transcript is still not found.

This is prd-42's own thesis (*one path, one spelling*) in the two characters prd-42 did not check.

## How it was found

#12's lane hit it while writing the round-trip law that prd-42 ruling 1 requires. That ruling says:

> for a generated set of paths including spaces, dots, underscores and colons, encoding then walking back must return the original.

**As written that is unsatisfiable**, and not because of the space bug #12 fixes — the colon half cannot pass at all, because the gap is in the reverse walk's logic, which #12's claim only permits it to touch as comment text.

Resolved for #12 by asserting an **honest refusal** for the colon case rather than a round trip (recorded as a deviation on #12, with the law's own doc comment carrying the reason). That keeps the law green and truthful and turns the gap into a tested fact. It does not close the gap.

## Why this cannot be fixed without a probe first

**It is not established which half is wrong.** prd-42 settled the space by running a real session from a directory whose name contained one and reading the slug that appeared — the space had become a dash, so the encoder was wrong and the walk was right.

[...]

PR #56, opened 2026-08-24
prd42 w3: the sweeps — one containment comparison, one forward transform

### LINKED ISSUE #15: prd42 w3: no path prefix comparison survives outside containment.ts
## Fence (may touch ONLY)
- `packages/server/src/paths/prefix-comparison-law.test.ts`

## Blocked by
Blocked by prd42 w2's `server/static.ts` issue — this sweep runs after the known site is fixed, so it lands green rather than red by design. Sweeps come last; one that runs earlier re-lays ground the other waves are digging.

## Definition of done
- Greps `packages/server/src` for path-prefix comparison idioms and fails on any outside `paths/containment.ts`.
- `paths/containment.ts` itself is the only allowlisted home, and the allowlist is asserted non-empty.
- Proves its own detector bites against a rigged snippet — a grep law never shown to fire is decoration.

### LINKED ISSUE #45: prd42 w3: the forward transform has one spelling
## Fence (may touch ONLY)
- `packages/server/src/cli/index.test.ts`
- `packages/server/src/collectors/sessionlog/forward-transform-law.test.ts` (new)

## Blocked by
Blocked by the prd42 w2 slug issue — the character class must be final before a law asserts there is only one of it, and before the hand-rolled copy is replaced by a call to the real one.

Sibling wave-3 lane owns `packages/server/src/paths/prefix-comparison-law.test.ts` (the containment sweep). `collectors/sessionlog/worktree-slug.ts` is **consumed, never edited** — its class was settled in wave 2.

## Why this exists — the sibling case, found at grooming 2026-08-24
`cli/index.test.ts` **already imports `worktreePathToProjectSlug` at line 10** and uses it correctly at three sites (`:1301`, `:1413`, `:1476`). At `:250` it hand-rolls a fourth, narrower copy:

```ts
const projectDir = path.join(claudeProjectsRoot, extraDir.replace(/[/_]/g, '-'))
```

`/[/_]/g` maps neither `.` nor `:` nor — after wave 2 — a space, so this computes a *different* directory than the collector under test looks in. Three sites fixed, one missed, in one file: AGENTS.md's sibling-case shape verbatim, and this PRD's whole thesis (*a duplicated path primitive gets hardened in one copy and keeps the hole in the other*).

It is dormant only because `os.tmpdir()` happens to contain no `.`, `:` or space on either CI leg today. That is a coincidence of the platform, not a property of the test — REASONED, not executed: the macOS leg has not been run against a tmpdir carrying one.

## Definition of done
- `cli/index.test.ts:250` calls `worktreePathToProjectSlug(extraDir)`. The test still passes for the reason it claims — the collector and the fixture agree on the directory *because they compute it the same way*, not because the characters happened to match.
- A law asserts the forward transform has exactly one implementation: nothing outside `collectors/sessionlog/worktree-slug.ts` builds a project slug by replacing path characters with `-`.
- The law is **scoped by what a file is, not by what it is called** — `#649`'s lesson: a guard keyed on a naming convention never saw the three captures that predated it. Sweep the package, do not sweep a prefix.
- The law's allowlist is asserted **non-empty**, and the detector is proven to bite against a rigged snippet. A grep law never shown to fire is decoration.

gabriel-canaan:
prd-42 wave 3 — the two sweep laws. One commit per issue, reviewed commit by commit.

Closes #15
Closes #45

**Note on shas:** cherry-picked from two lane branches, so these commits (`2796945`, `03fa365`) are not the shas the issue comments cite. Same work, re-parented onto current main.

## What becomes true

- **`2796945` (#15)** — a second path-containment comparison outside `paths/containment.ts` fails the build. prd-42 ruling 2 becomes enforceable rather than aspirational.
- **`03fa365` (#45)** — the fourth hand-rolled path-slug transform collapses into the one real implementation, and a law holds it there.

## Verification

Four rounds. Rounds 1–3 ran two independent model passes (Claude Opus, Codex `gpt-5.6-terra`) at xhigh, each executing in its own disposable worktree, plus orchestrator probes. Round 4 was orchestrator probes only — stated plainly because it matters: in each of the first three rounds the model passes found defects the probes missed, so round 4's evidence is thinner than the rounds before it.

**Eight findings were raised and closed across those rounds.** The two that mattered most:

- #45's law originally asserted cardinality and enforced *one regex spelling*. Reintroducing the exact defect as `extraDir.split('/').join('-')` left both the law and its host test green. Ten of eleven probed spellings were missed, including #47's `[^a-zA-Z0-9]` shape.
- Fixing that created a **new** hole one level out: scoping the detector to the join's argument text meant hoisting the transform to a `const` on the preceding line dodged it entirely. A repair to a detector is itself a detector.

Both are closed and proven closed with spellings not used to find them.

Final probe set, baselines law15 = 26 passed, law45 = 32 passed:

| probe | result |
|---|---|
| `slice(0, root.length) === root` with path context | 1 failed — caught |
| `substring(...) === root`, `indexOf(root) !== 0`, parenthesised `startsWith` | 1 failed each — caught |
| hoisted `const` transform | 1 failed — caught |
| comment-only file quoting the idiom | 32 passed — prose not convicted |
| `CODE_EXTENSIONS` → `['.NOPE']` (empty scan) | 1 failed |
| `SOURCE_EXTENSIONS` → `['.NOPE']` (empty walk) | 3 failed |
| plant a fourth containment site | 1 failed |
| shrink the allowlist, site remains | 2 failed — two-way ratchet |
| ordinary parsing without path context | not convicted |

Gates on this merged branch, node 22.22.2: `npm run typecheck` 0, `npm run lint` 0, `VITEST_MAX_WORKERS=6 npm test` 0 (**379 files, 6474 passed**, 3 expected fail), `npm run build` 0.

## What is the sibling case?

- **#15** — the founding idiom was `X.startsWith(root)`. Its siblings are `path.relative(root, X)` + `'..'`, template-literal roots, `indexOf`, `slice`/`substring`, and the parenthesised argument. All are caught now with path context nearby. The residual sibling — any of those shapes with **zero** surrounding `path.*` work — is documented as a deliberate gap, along with the evidence that loosening the gate to close it immediately re-convicts `argv.indexOf(flag) === 0`. The law's own founding idiom is missed under that condition too, and the doc says so.
- **#45** — the commit fixed `cli/index.test.ts:250`, the fourth copy. The sibling was every *other* way to spell the same transform, and then the sibling of the fix itself. Both addressed. The remaining axis — the `path.join` spelling rather than the transform spelling — is named in the doc rather than silently left.

None of the above. Every row in the table is a mutation, and each was run in an isolated worktree with the tree restored and the baseline re-confirmed. The two laws' own detector-bites tests use **rigged snippets, never live code** — a law that proves itself against a real file stops meaning anything the moment that file is fixed, which is exactly what wave 2 did to `server/static.ts`.

## Deliberate residuals, stated not slipped

- **#15's path-context gate.** Shapes are only convicted with `path.*` work nearby. This is a deliberate precision-over-recall trade, documented with the mutation evidence behind it, and it is why ordinary string and array parsing is not convicted.
- **#15's test-file ceiling.** Test-file containment sites are pinned to a count (16, currently exact with zero headroom) rather than allowlisted per site, and the `describe` title says so. One-directional: migrating a site drops the count and gifts a slot of slack. Worth a follow-up, not a blocker.
- **#45's scan boundary.** Uses `git ls-files`, so an untracked file is invisible until staged; #15 walks the filesystem and sees it immediately. Neither is wrong; the difference is now documented.

## Reviewer note

I verified this, so I should not approve it. `AGENTS.md` is explicit that the reviewer fixes but never reviews their own fix into main — and that is not theoretical here: wave 2 was reviewed after two rounds of my verification and a real defect was still found and fixed on top (`5fa84f4`). Round 4 skipped the model passes, so a second pair of eyes is worth more than usual on this one.

Landing unblocks nothing else in prd-42 directly; #47 (Urgent) is the next real gap.


ciaran-slow (approved):
Reviewed commit by commit, in a clean clone at node 22.23.2. Everything below is labelled **EXECUTED** (I ran it and read the exit status) or **REASONED** (argued from the code, nothing run). One fix pushed as `ba0022e`.

**The #45 fix is load-bearing, and the defect really was dormant.** This was the claim most worth checking, and it holds in both directions — EXECUTED:

| mutation at `cli/index.test.ts:250` | result |
|---|---|
| revert transform to `extraDir.replace(/[/_]/g, '-')`, keep new fixture | **FAILS** (1 failed, 8.05s) |
| revert transform **and** fixture literal to `rhizomorph-conductor-workdir` | **passes** (1 passed, 81ms) |

That second row is the important one: it independently reproduces your diagnosis. The old test passed by coincidence of `os.tmpdir()`, and the dot-and-space in `:249` is precisely what converts it into a test that passes for the reason it claims. The comment you left there is accurate, not aspirational.

**Both laws bite against real planted files, not just rigged snippets.** Your probe table used rigged snippets for most shapes; I planted real files under `packages/server/src` — EXECUTED, each caught, each reported with the right snippet:

- `path.posix.relative(root, cand)` + `.split(path.sep)[0] !== '..'` → caught
- `` r.startsWith(`${root}/`) `` template spelling → caught
- `startsWith(root)` with `path.resolve` 4 lines above → caught
- `path.join(claudeProjectsRoot, worktree.replace(/[/_. ]/g, '-'))` → caught by **both** law45 check 1 and check 3

**The context window is honest.** EXECUTED: context at exactly 4 lines above is caught, at 5 lines it is missed. `PATH_CONTEXT_LINES_BEFORE = 4` means exactly what the doc comment says — no off-by-one.

**The `git ls-files` boundary is real and correctly documented.** EXECUTED: the planted hand-rolled transform is invisible to law45 while untracked and fails immediately once `git add`-ed. Exactly as your doc states.

**Post-merge, not just on the branch.** `origin/main` moved to `1b3a70e` after you branched (#57, docs only — no `packages/` changes). EXECUTED: merged `origin/main` in cleanly, no conflicts, and re-ran everything against the actual post-merge tree — typecheck 0, **379 files, 6474 passed, 3 expected fail**. Your stated numbers reproduce exactly. Worth noting because both laws are exact-set/exact-count ratchets, so a `packages/` change on main would have broken them; it didn't.

**CI**: EXECUTED `gh pr checks 56` — all 7 jobs actually executed and passed, both legs, so nothing was silently skipped behind a failing `Test` step.

## One fix pushed — `ba0022e` (mechanical)

`codeOf` strips comments by text rather than tokenizing, so the same characters inside a **string literal** blank real code. EXECUTED — both planted as real files under `packages/server/src`, suite reported **26/26 green** against them:

```ts
const docs = 'https://example.invalid/docs'; if (!requested.startsWith(root)) return docs
// the URL's `//` blanks the rest of the line, containment check included
```
```ts
const openGlob = '/*'
const requested = path.resolve(root, rest)
if (!requested.startsWith(root)) return null
const closeGlob = '*/'
// lines 2-4 blanked wholesale — both the path.* context AND the check
```

I confirmed the mechanism directly by printing `codeOf` output rather than inferring it.

**Why this belongs on `prefix-comparison-law.test.ts` specifically:** `forward-transform-law.test.ts`, in this same PR, found this exact root cause from the other side — the bare `/\//g` regex whose escape abuts the closing delimiter — and documented it with a dedicated test at `:583`. But that law *borrowed* `codeOf` from `prefix-comparison-law.test.ts`, and the file it was borrowed from never got the same treatment. That is the sibling shape your own PR body names.

**Severity, stated plainly: nothing is hidden today.** EXECUTED — swept every `.ts`/`.tsx` under `packages/server/src` comparing raw text against `codeOf` output: **zero** non-comment lines lose an idiom to either blanking. So this is a gap in the "Known, deliberate gaps" list, not a live hole. The fix adds the entry and a test; it changes no detector behaviour.

The test asserts the gap *exists*, so it is two-directional like your zero-context test — EXECUTED: making `codeOf` string-literal aware turns it red, forcing whoever closes the gap to say so.

Gates on `ba0022e`: typecheck 0, lint 0, full suite 0 (379 files, **6475 passed**, 3 expected fail).

- The PR body says test-file sites are pinned at **16**; the code pins `TEST_FILE_KNOWN_DEBT_COUNT = 14` and the doc comment says 14 consistently. REASONED — the code is self-consistent and green, so I read this as a stale number in the PR description only. Worth correcting so a later reader doesn't chase it.
- Both `describe` titles are long enough that vitest wraps them in output. Cosmetic, entirely yours.

## What I did not run

- No Windows leg (prd-25 gap, unchanged by this PR).
- I did not re-run your rounds 1–3 model passes. You flagged that round 4 was probes-only and thinner — I treated that as the place to spend effort, which is where `ba0022e` came from.
- I did not run `scripts/gate.sh`; that is the operator's landing step, not a reviewer's.

Approving. `ba0022e` is mine and it is documentation plus a gap-asserting test with no behaviour change, so I'm comfortable approving alongside it — but flagging it explicitly so you can judge that for yourself rather than take my word for the rung.

PR #65, opened 2026-08-24
Light mode reads by weight, not brightness — bordered cards, square corners, a growing dock

hchristinasmith:
## Summary

A live, interactively-directed design pass on the light theme, done in-session with the operator watching each change in the running dashboard. Not tied to a tracked issue — ad hoc UI tuning per the "anything I am watching happen live" carve-out.

- **Cartouche, not glow.** Light mode's alarm chips (`glow-notice`/`glow-needs-you`/`glow-broken`/`glow-calm`) used the same soft-blur recipe as dark mode, which reads as brightness and all but disappears against warm paper — a direct contradiction of `prd-32` ruling 7 ("severity is re-carried, not re-lit"). Light mode now gets a filled tint + solid ring instead, same class names, zero component changes.
- **Bordered cards.** The header, BURN grid, GAPS panel, and FLEET surface now render as discrete square-cornered cards floating over the page floor, instead of edge-to-edge hairline-divided strips. The SPEND table gained a colored left-edge accent bar + row tint per status. Added a "Lanes" rail beside the fleet canvas.
- **The dock grows, the page scrolls.** Reverses the documented "page never scrolls, panels do" invariant on the vertical axis only — the horizontal-scroll law (`#655`) is untouched. `spend`/`collisions`/`activity`/`trace` no longer clip inside a fixed share; the whole page scrolls instead, and the shell's top dock is sticky so it stays visible. fleet:dock grid share retuned 7:3 → 3:2 at the operator's request; the hero keeps a `minmax(420px, 55vh)` floor so the canvas is never squashed below its own zero-size fallback.
- **Square corners, sitewide.** Every button/chip/field's border-radius zeroed by explicit request — 49 files swept. The three now-fully-unused radius tokens (`--radius-ctl`, `--radius-plate`, `--radius-bead`) were deleted from `theme.css` rather than left declared with no consumer, per this file's own "every declared material is consumed" law.

## Test plan

- [x] `npm run typecheck` — clean across all 5 workspaces
- [x] `npm run lint` — clean
- [x] `VITEST_MAX_WORKERS=6 npm test` — 375 files, 6355 passed, 3 expected fail (unchanged baseline)
- [x] Verified live in-browser at each step: cartouche rendering, card layout, dock growth + page scroll (no horizontal overflow, canvas not squashed at 532px), sticky header stacking, square corners on every button/chip
- [x] Updated the two tests that pinned the old fixed-share/no-scroll behavior (`PanelGrid.test.tsx`, `shell-bounds-law.test.ts`) to assert the new, deliberate design instead of silently deleting coverage


ciaran-slow:
## Reviewed commit-by-commit, verified by running. One fix pushed as `d9e0ec7`.

Everything the PR body claims, I checked rather than took on trust:

- **`npm run typecheck`** clean, **`npm run lint`** clean (846 files), **`VITEST_MAX_WORKERS=6 npm test`** → **380 files, 6476 passed | 3 expected fail**, exit 0 — against the tree *with `origin/main` merged in*, and the 3 expected fails match your stated baseline exactly. `EXECUTED`
- **All 7 CI jobs pass, and the gate steps really ran.** I didn't take the green bar for it — `gh api .../jobs/97605380133` shows `Test`, `Typecheck`, `Lint`, packaging guard and boot smoke each `success`, not skipped. I also checked the missing `build-test-boot (macos-latest, min)` leg before flagging it: `ci.yml:34-36` excludes it deliberately. Not a gap. `EXECUTED`
- **The three deleted radius tokens genuinely had no consumers.** `grep -rn 'radius-ctl|radius-plate|radius-bead' packages/` returns only your own historical prose in `theme.css`. Deleting them rather than leaving them declared-and-unread is the right call under that file's own law. `EXECUTED`
- **The `rounded-full` survivors are correct, not misses.** The 11 remaining are all `h-1.5/h-2` status dots — genuine circles, not buttons/chips/fields. The sweep's scope holds. `EXECUTED`
- **The two rewritten tests are rewritten honestly.** `PanelGrid.test.tsx` and `shell-bounds-law.test.ts` both state the reversal, date it, name who asked for it, and say what property still holds and by what mechanism. `shell-bounds-law` keeps its load-bearing positive assertion (`grid-cols-[minmax(0,1fr)]`), so the horizontal law is still guarded. This is the good version of changing a test.

### One finding — fixed and pushed, needs your confirmation

**The collisions panel's column headings became invisible.** `packages/web/src/panels/collisions/index.tsx:100`

You deleted that panel's `flex-1 overflow-auto` wrapper, correctly, so the table grows with the page. Its two `<th>`s kept `sticky top-0 z-(--z-sticky)`. `position: sticky` resolves against the nearest scrolling ancestor — with the wrapper gone that is now the **document**, so the heading row detaches from its own table, pins to the viewport's top edge, and lands underneath the shell dock, which this same PR raised to `--z-header` (20) over `--z-sticky` (10). A scrolled collision matrix shows no column headings at all.

`--z-header`'s own comment in `theme.css` shows you saw this interaction from the dock's side — *"so it does not disappear behind one"*. The half that got missed is that winning that z-fight is precisely what hides the `<th>`.

It is the only site with that shape. I audited every file in the PR that lost an overflow: `feed`, `ledger` and `trace` also lost theirs but carry no `sticky top-`, and `fleet/index.tsx` keeps its `overflow-auto` so its sticky `<thead>` is fine. `EXECUTED`

`d9e0ec7` drops the sticky (the heading scrolls with its table, matching the other three dock panels) and adds `panels/sticky-scrollport-law.test.ts` pairing the two declarations so it cannot recur silently. Comments are blanked before the grep — `tokens.test.ts`-style — so a file's prose about `sticky top-` can neither satisfy nor trip a law meant to read its code.

**Mutation-verified**, since jsdom never computes `position: sticky` and no render test here could have caught this:

```
git stash push packages/web/src/panels/collisions/index.tsx   # restore the PR's own state
npx vitest run packages/web/src/panels/sticky-scrollport-law.test.ts
→ AssertionError: expected [ 'collisions/index.tsx' ] to deeply equal []
```

Green with the fix, red without it, naming the right file. `EXECUTED`

**This is why I'm not landing it myself:** dropping the sticky is the smaller half. If you want the headings to *stay* visible on a wide matrix — which is a reasonable thing to want — that needs a `top` offset equal to the dock's height, and the dock's height is dynamic (attention strip and replay banner swap in and out). That is a design call and new machinery, and you had the operator in the room. Say which way you want it and I'll land it.

- **Three `row-tint-*` utilities ship with no consumer.** `row-tint-notice`, `row-tint-needs-you`, `row-tint-broken` (and their `cell-rule-*` counterparts) are declared in `theme.css`; only `-working` and `-done` are used, in `ledger/index.tsx:123-126`. Your comment says this is deliberate — five to match the ladder's vocabulary. No test fails, because `tokens.test.ts`'s "every declared material is consumed" law runs off a hardcoded 5-token allowlist and never sees `@utility` blocks. But it does sit oddly beside deleting the radius tokens *for exactly that reason* one screen away. Either is defensible; they just aren't both defensible at once. `EXECUTED`
- **The `min-h-[Nvh]` law no longer sees the floor it was written about.** It bans `min-h-[55vh]`; this PR reintroduces a viewport floor as `grid-rows-[minmax(420px,55vh)]`, which the regex can't match — the guard is scoped by syntax rather than by what the thing *is*, the same shape `AGENTS.md` warns about. Coverage isn't actually lost (the sibling test pins the exact string), so this is stale prose, not a hole. Worth a comment refresh next time you're in there. `EXECUTED`

### What I did not run

I did not open this in a browser. Every claim above about *rendering* — the cartouche reading correctly on warm paper, the card layout, the sticky behaviour I'm describing — is derived from the CSS and the class strings, not seen. You verified those live with the operator watching, which is better evidence than I can produce here; I'm relying on that for the visual half. The one exception is the collisions finding, where the mechanism is decidable from the code and I proved the guard fires by mutation rather than by eye.

Not approving formally yet — the sticky question above is yours to settle first.

ciaran-slow:
@hchristinasmith — one open question for you before this lands, plus a suggested fix. Writing it out in plain terms so it's on the record.

## What went wrong

Sticky positioning always attaches to the nearest parent that scrolls.

The collisions table's two heading cells were marked `sticky top-0`. That worked before, because the panel had its own scrolling box (`flex-1 overflow-auto`) — the headings stuck to the top of *the panel*.

This PR removes that scrolling box on purpose, so the whole page scrolls instead. Good change. But the headings kept their `sticky top-0`. With the panel's scrolling box gone, the nearest thing that scrolls is now **the page itself** — so the headings jump out of the table and stick to the very top of the window.

That's the same place the top dock sits. The dock is solid and sits on a higher layer (`--z-header` is 20, the headings are `--z-sticky` at 10), so the headings end up **behind the dock, completely invisible**. Scroll through a wide collision matrix and there are no column headings at all.

Worth saying: you *did* spot this interaction — your `--z-header` comment says it's there "so it does not disappear behind one". You solved it from the dock's side. The part that got missed is that the dock winning that fight is exactly what hides the headings.

## What I've already done

`d9e0ec7` removes the `sticky` from those two heading cells. The headings now scroll along with their own table, which is what the other three dock panels do. All 7 CI legs pass on it.

I also added `panels/sticky-scrollport-law.test.ts`, which fails if anyone writes `sticky top-*` in a panel that has no scrolling box of its own. Normal tests can't catch this — jsdom has no layout engine and never works out sticky positioning at all — so it's a grep-style law like `shell-bounds-law.test.ts`.

## The question for you

My fix is the *simple* half. It stops the headings being invisible, but it also means they scroll away — on a wide matrix you lose track of which column is which.

**Do you want the headings to stay visible while scrolling?** You had the operator watching, so you know whether that matters in practice. I didn't want to decide it by review edit.

### Option A — leave it as I've pushed it

Headings scroll away with the table. Nothing more to do; consistent with the other dock panels. Fine if the matrix is usually short.

### Option B — make them stick properly, below the dock

The reason they can't just use `top-0` is that they'd need to stop *below* the dock, and the dock's height changes (the attention strip and replay banner swap in and out).

So measure the dock and publish its height as a variable. There's already a hook shaped exactly like this in `tide/TideDock.tsx` (`useElementWidth`) — a height version would be its twin, including the `typeof ResizeObserver === 'function'` guard that keeps jsdom happy:

```tsx
// Shell.tsx
const [dockRef, dockHeight] = useElementHeight()

<div
  style={{ '--dock-h': `${dockHeight}px` } as CSSProperties}
  className="grid min-h-screen grid-cols-[minmax(0,1fr)] ..."
>
  <header ref={dockRef} className="sticky top-0 z-(--z-header) ...">
```

Then the heading cells become:

```tsx
className="sticky top-(--dock-h) z-(--z-sticky) min-w-[14rem] bg-(--surface-panel) ..."
```

No z-index change needed — at 10 the headings correctly slot *underneath* the dock rather than fighting it, which is right once they're offset out from behind it.

**Two things to watch if you go this way:**

1. That table is `border-collapse`. Sticky `<th>` cells are known to render oddly under `border-collapse` in some browsers — borders can drop out while stuck. It's the same class of problem your own theme comment already hit with `box-shadow` on `<tr>`. The cells do already carry `bg-(--surface-panel)`, which is the usual thing that saves you, but it needs a real browser check.
2. The law I added would need to accept this shape — right now it only accepts "has its own scrolling box". If you take Option B, tell me and I'll widen it to also accept a `top-(--dock-h)` offset, so it keeps catching the real mistake without blocking the correct fix.

Happy to implement Option B myself if that's the one you want — just say which.

hchristinasmith:
@ciaran-slow Option A — leave it as you've pushed it. Headings scroll away with the table, consistent with the other three dock tabs. We're not routinely scrolling a wide collision matrix, so this is fine; happy to revisit with the dock-height offset machinery if that changes.

Thanks for the thorough pass — the sticky/scrollport interaction was a real miss on my end, and the mutation-verified law is a good permanent guard for it. Pulled `d9e0ec7` in, gate is green locally on the synced branch (380 files, 6476 passed, 3 expected fail, matching your numbers).

Leaving the `row-tint-*`/`cell-rule-*` unused-utility note and the stale `min-h-[Nvh]` law comment as-is for now — both flagged as offered-not-asserted, and neither blocks anything.

hchristinasmith:
@ciaran-slow Correction — the operator changed their mind after seeing the tradeoff spelled out, and wants option B after all: headings stay visible on a wide matrix.

Pushed as \`13308c5\`:

- \`Shell.tsx\` gets a \`useElementHeight\` hook (twin of \`tide/TideDock.tsx\`'s \`useElementWidth\`, same \`ResizeObserver\` guard), measures the dock's live height, publishes it as \`--dock-h\` on the shell's root.
- The two \`<th>\`s restore \`sticky top-(--dock-h) z-(--z-sticky)\` — no z-index change needed, exactly as you called it: at 10 they now correctly slot underneath the dock's 20 instead of fighting it, because the offset stops them at the dock's own bottom edge rather than the viewport's top edge.
- Widened \`sticky-scrollport-law.test.ts\` to accept \`top-(--dock-h)\` as the second lawful shape, plus a test pinning that collisions specifically took this path (not the scrollport one) so a future edit that silently reverts to \`top-0\` still trips the original law.

On your two watch-items: the table's \`bg-(--surface-panel)\` on both \`<th>\`s is intact, and I didn't hit visible border dropout under \`border-collapse\` in-browser — though I'll be honest, the local fixture only has 3 collision rows, not enough to scroll the header into its actual stuck position, so that's from computed-style verification (\`position: sticky\`, \`top: 146px\` matching the live-measured \`--dock-h\`, \`z-index: 10\`) rather than eyes-on the stuck frame. Full gate green (typecheck/lint/380 files, 6477 tests — one more than your count, matching the new pinning test).

ciaran-slow:
@hchristinasmith Checked `13308c5`. The implementation is right, and the law widening in particular holds up — I tried to break it and couldn't. Two follow-ups pushed as `57e5f13` and `e2e4832`, both the same shape as the original defect.

## Your commit verified

- **The widened law still catches the original bug.** This was my main worry — "widening" a guard is the usual way one quietly stops guarding. Reverted collisions to a bare `sticky top-0` and **both** tests fire, naming the file:
  ```
  × no panel declares `sticky top-*` without owning a scrollport or offsetting below the dock
  × the collisions table took the dock-offset path, not the scrollport one
  ```
  The second one earns its place — it's what stops a future edit swapping the offset out for a scrollport and back without anyone noticing which answer was chosen. `EXECUTED`
- `useElementHeight` is a faithful twin of `useElementWidth`, `ResizeObserver` guard included; `ref` correctly on the `<header>`; `--dock-h` on the shell root where every descendant can read it. `Math.ceil` over my `floor` is the better call — floor would leave a 1px seam. `EXECUTED`
- Documenting the file-level coarseness blind spot yourself, unprompted, is the right instinct. Agreed it's the same trade every grep law here makes.

## Two follow-ups

### 1. Nothing was guarding the *other* end of `--dock-h` — `57e5f13`

The law pins that a panel offsets by `--dock-h`. Nothing pinned that anyone **publishes** it.

Delete the `style={{ '--dock-h': ... }}` line from `Shell.tsx` and all **166 web test files stay green**. `EXECUTED`

That failure is silent by construction, which is what makes it worth a test rather than a comment. An undefined custom property doesn't throw — `top: var(--dock-h)` becomes invalid at computed-value time, `top` falls back to its initial `auto`, and a `position: sticky` element with `top: auto` never sticks. The headings would go back to scrolling away — the exact behaviour option B was picked over — with no failing test, no console error, and nothing on screen to say the offset had stopped existing.

Same shape as the bug this file was written for: a declaration whose meaning depends on something that can move independently of it. So the law now reads both ends. It's conditional (no consumers → the shell owes nobody a measurement) and it asserts the value is *measured*, because a hard-coded `--dock-h: 146px` would satisfy a presence check while being wrong in precisely the mode that motivated measuring it.

Mutation-verified three ways, each failing this test alone: delete the publisher line, replace the `ResizeObserver` with a constant, drop the `ref` off the header.

### 2. A focused panel offsets for a dock that isn't on screen — `e2e4832`

`--dock-h` is chrome-relative. A focused panel has no chrome above it: `PanelFrame` renders it `fixed inset-0` at `--z-focus` (30), opaque, over the dock's `--z-header` (20). And the dock **is** focusable — `PanelGrid.tsx:218` wraps it in a `PanelFrame` like any other surface.

So a focused collisions table holds ~146px of empty space open above its headings, reserving room for a bar that is covered. Not invisible headings this time — headings hanging below a gap, and the gap tracks the height of a dock the operator can't see.

Fixed by zeroing `--dock-h` for the focused subtree only. The `<th>` declaration keeps meaning exactly what it meant — *stop below the chrome* — and in a focused panel the chrome is nothing. It's on the frame rather than the table so any future sticky child adopting the offset gets it too.

Real render test this time rather than a grep law, since jsdom can render the focus toggle. Mutation-verified in both directions: removing the override gives `expected '' to be '0px'`, and the test checks restore too, so a version that zeroed the offset permanently wouldn't pass either.

## On the verification gap you flagged

You were right to call out that 3 rows isn't enough to scroll the header into its stuck position, and I want to name why that matters rather than let it pass as a nicety: **your computed-style check would have read `top: 146px` in both of the cases above.** It confirms the declaration resolved, not that the stuck frame is correct — and in focus mode a correct-looking `146px` is precisely the wrong answer. Not a criticism of the check; it's just load-bearing in a narrower way than it looks.

Neither follow-up is browser-verified either, for the same reason. Both are decidable from the code, and I've pinned both with tests that fail when the mechanism breaks — but the rendered result of option B, stuck frame and all, still hasn't been seen by anyone. **A wide-matrix fixture is the thing that would actually close this**, if you think it's worth one.

## Gate

`typecheck` 0, `lint` 0 (847 files), `VITEST_MAX_WORKERS=6 npm test` **380 files, 6479 passed | 3 expected fail**, exit 0 — your 6477, +1 for the publisher guard, +1 for the focus test. `EXECUTED`

Approving. Both my commits are review fixes on your branch, so I'm not landing it myself — merge is yours whenever you're happy with the two above.

ciaran-slow (approved):
Approving. Hannah's `13308c5` implements option B correctly and its law widening survives a revert-to-`top-0` mutation intact. Two follow-ups pushed as review fixes (`57e5f13`, `e2e4832`) — a guard for the unpublished-`--dock-h` case, and the focused-panel offset. Full gate green: 380 files, 6479 passed, 3 expected fail. See my comment above for the detail, including what is still not browser-verified.

Issue #37, opened 2026-08-24
prd44 w2: the server's event buffer has a stated ceiling


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/recorder/session-recorder.ts`
- `packages/server/src/recorder/session-recorder.test.ts`

## Blocked by
- **#3** (prd40 w1) — it is the live claimant on this lane's own file and adds foldSoFar(). This lane cannot open that file until #3 has landed.
- **prd-40's own open question**, which is the prerequisite ruling 4 consumes: *should `foldSoFar()` be the only reader, with `eventsSoFar()` narrowed to the exporter?* If it is, this lane's bound is far cheaper to state. Do not pre-empt that answer.
- **Wave 0, an operator act** — the ceiling's number is the operator's, not this lane's. It is not in this issue to choose.

Sibling lanes hold the three collectors. **Do NOT touch `recorder/session-log-writer.ts`** — prd44 #31 changed it in wave 1, and it is a different concern.

**Model:** sonnet. **Wave:** 2.

## Definition of done
- `eventsSoFar()`'s growth is bounded by a named constant, not by session length. A law asserts the buffer's length stops rising past the bound while `session.eventCount` keeps climbing.
- Eviction is oldest-first, matching the client's own rule (`web/src/app/streamState.ts:60`, `MAX_EVENTS = 75_000`).
- **Any surface reading a bounded window says so.** Reuse `eventsWindowLabel`'s existing vocabulary rather than inventing a server-side twin — the client already solved saying "this window is partial".
- The two consumers that assume the whole session is in memory are named and handled explicitly, not discovered later: `/api/stream`'s replay promise to a new subscriber, and the exporter. If either keeps needing everything, say which and why in the PR.
- `openSession()` still clears the buffer outright; a rotation is not an eviction.
- A law asserts the bound holds across a rotation, not just within one session.
- **Assert a count, never a wall clock.** And do NOT name any new law `*.bench.test.ts`, nor mark it `// @gate-timing` — `scripts/gate.sh` derives its 4x-load timing set from exactly those and ratchets `.swarm/timing-count`.

**Why it is a ruling and not a patch:** the client capped itself at 75,000 events *and* built a label to admit it. The server kept everything and said nothing. Bounding it changes a promise, so the number is the operator's and the affected consumers are named up front.

[...]

Issue #26, opened 2026-08-23
prd40 w4: a dropped append speaks in the degrade voice


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/server/poll-loop.ts`
- `packages/server/src/server/poll-loop.test.ts`
- `packages/server/src/recorder/session-recorder.ts`
- `packages/server/src/recorder/session-recorder.test.ts`

**Fence widened 2026-08-25, recorded before the change.** This issue is unbuildable as originally fenced. The DoD requires the degrade `collector.error` to be emitted **whether or not its own append succeeds**, and after #4 there is no way to do that from `poll-loop.ts`: `record()` appends first and publishes nothing when the append fails — exactly the disk-full case — and `SessionRecorder`'s emitter is private. Its whole public surface is `sessionId`, `filePath`, `isSealed`, `record`, `closeWith`, `openSession`, `sync`, `foldSoFar`, `eventsSoFar`, `subscribe`. There is no publish-without-append, so one has to be added on the durability boundary itself.

**This widening is what moved the issue to wave 4.** `session-recorder.ts` is also claimed by #69 and #80; three issues cannot share one file in one wave, so they run in priority order — this one first (Now/High), then #69, then #80.

## Blocked by
**Ruling blocker CLEARED 2026-08-25:** prd-40 success 1's named exemption landed on `main` in #86 (it had been stranded on `prd39-paper` since 2026-08-24, cited by this issue but never merged). Success 1 and ruling 1 both now carry it.

Blocked by prd40 w2 (#4) — landed in #79. The degrade path cannot distinguish "the append failed" as its own condition until `record()` appends before it publishes — before that fix a rejected append and a rejected poll arrive at the same `catch` looking alike.

Sibling lane in this wave owns `packages/server/src/api/meta.ts` (#5) — no shared file, which is what prd-40's Sequencing promises for wave 3. `packages/server/src/recorder/session-recorder.ts` was wave 2's and is landed by the time this runs; do NOT reopen it here. The `collector.error` event type itself is prd-11/prd-17 schema and has ten consumers — emit it, never redefine it.

**Model:** sonnet. **Wave:** 4.

## Definition of done
- An append failure produces a `collector.error` that reaches subscribers, so the instrument shows the record is broken rather than continuing to look healthy.
[...]

PR #67, opened 2026-08-24
prd40 w1: the recorder answers from a fold it maintains

### LINKED ISSUE #3: prd40 w1: the recorder answers from a fold it maintains
## Fence (may touch ONLY)

## Blocked by
Nothing — this is the wave-1 keystone.

Sibling lanes own `packages/server/src/api/meta.ts` and `packages/server/src/server/poll-loop.ts`. `packages/core/src/reduce.ts` is owned by prd-11/prd-17 — consume it, never edit it.

## Definition of done
- A running `SessionState` is kept beside the buffer: updated in `record()` through the existing `reduce`, reset in `openSession()`, exposed as `foldSoFar()`.
- No existing caller changes behaviour — this wave is purely additive.
- A spy law asserts the number of `reduce` calls per `foldSoFar()` read is constant as the session grows. Not a wall-clock assertion: measuring without asserting is the defect prd-24 is cleaning up after.
- `foldSoFar()` and `reduceAll(eventsSoFar())` return deep-equal state for the same event stream, proven over the golden-era corpus.

ciaran-slow:
Closes #3

prd-40 wave 1 — the keystone. `SessionRecorder` now maintains a running `SessionState` beside its buffer and exposes it as `foldSoFar()`, so the fold a reader needs is already computed rather than rebuilt from scratch per request.

**Purely additive. No caller is wired yet** — `foldSoFar()` is new surface, the sixteen existing `eventsSoFar()` call sites are untouched, and `eventsSoFar()`'s own body is byte-for-byte unchanged. Wave 3 (`/api/meta`, #5) and prd-44 #37 are what consume it.

## The design

One private `advanceFold(event)`, called from **both** `record()` and `closeWith()` immediately after their existing `buffer.push`, in the same synchronous step. That single call site is the point: a fold that advances on a path the buffer does not is the sibling-publisher defect this repo keeps rediscovering, and two copies of the advance is how it comes back.

- `openSession()` resets the fold alongside its buffer reset.
- The constructor seeds from `reduceAll(this.buffer)`, so a `resumeFrom` preload starts consistent — exercised for real by the boot smoke, which resumed a 15-event session.
- `reduce`'s own boundary detection cannot fight `openSession()`'s reset: by the time a new session's `session.started` folds, the state is already empty. Verified — `eventCount` is 1, not 0 or 2.
- A throwing `reduce` is caught inside `advanceFold` and never becomes a new rejection from `record()`/`closeWith()`, whose contracts belong to the writer. It sets a desync flag; `foldSoFar()` repairs once from the buffer, which is always ground truth.

prd-40's two open questions are left open: whether `foldSoFar()` should become the only reader with `eventsSoFar()` narrowed to the exporter, and #205's fold-order divergence.

## What is the sibling case?

`record()` and `closeWith()` are the two publishers, and a fold advanced by one but not the other is the whole defect class. Removing `advanceFold` from `closeWith` fails a law by name. All four sites that mutate `buffer` — including the constructor's `resumeFrom` — were checked and are paired.

The second-order sibling is the *observation* point: a subscriber reading `foldSoFar()` from inside `emitter.emit`. That invariant was claimed in a doc comment and unasserted; it is now a law, which matters because **wave 2 (ruling 1) reorders these exact lines under this same fence.**

It survived one, and that was the review's finding. The original spy law used `vi.spyOn(core, 'reduce')`, and `reduceAll` is `events.reduce(reduce, state)` binding core's **own module-local** declaration — so the spy sees `advanceFold`'s calls but is blind to every fold inside `reduceAll`. Forcing `foldSoFar()` to re-fold on every read left **all 11 tests green**, so DoD item 3 — "constant as the session grows" — was unproven. That is the shape prd-40 itself names: *measuring without asserting is the defect prd-24 is cleaning up after.*

Fixed, then hardened after cross-model verify found the boundary of that fix:

| mutation | result |
|---|---|
| re-fold on every read | **3 laws fail** |
| re-fold via a deep import — invisible to both barrel-scoped spies | **2 laws fail** |
| never clear `foldDesynced` (every later read O(n) forever) | **1 law fails** |
| drop `advanceFold` from `closeWith` | **1 law fails** |
| swap `push`/`advanceFold` in `record()` | not caught — **equivalent mutation**, `emit` runs after both, so it is unobservable where a subscriber looks |

Reference identity (`toBe`) is what closes the deep-import route, and it is deliberately not "fixed" by copying the state — F6 below depends on that.

## Fence

`packages/server/src/recorder/session-recorder.ts` plus **one recorded widening**: `packages/server/src/recorder/session-recorder.test.ts`. The base fence omitted the test file while the DoD demanded a spy law and a corpus proof — widening recorded before the change, per AGENTS.md. Untouched: `api/meta.ts` (#5), `server/poll-loop.ts` (#26/#4), `core/src/reduce.ts` and `state.ts` (prd-11/prd-17, consumed only), `session-log-writer.ts` (#46), `rotate.ts` (prd-16).

One flagged deviation: the corpus specifier is routed through a `const` before a dynamic `import()`, because `?raw` has no ambient declaration in the server tsconfig (TS2307) and a local `declare module` fails as augmentation (TS2664). Judged acceptable — every realistic break throws at test time and CI runs `test` before `typecheck`, so the failure surfaces earlier, not later. No tsconfig change, no widening.

## Verification

Planned, built and verified by three separate agents; verify ran on **Opus** against a **Sonnet** build, so the independence is real rather than nominal — and it produced the finding above rather than agreeing.

| gate | result |
|---|---|
| `npm run typecheck` | clean, 5 workspaces |
| `npm run lint` | clean, 846 files |
| `VITEST_MAX_WORKERS=6 npm test` | **379 files, 6488 passed**, 3 expected-fail |
| `npm run build` | clean |
| boot smoke | `/api/meta` 200, `/` 200, resumed a 15-event session |
| gate timing set | **6 files, unchanged** — no ratchet movement |

Not merged. `scripts/gate.sh` was not run: in this repo running it *is* the landing — past `:167` it merges to local `main`, then `npm install`, `npm run build`, `git push origin main`.

## Non-blocking, recorded rather than silently dropped

- **F4** the bare `catch {}` in `advanceFold` is unobservable — a follow-up issue, outside this fence.
- **F5** the corpus law would pass on an empty event list; real risk near zero since `eras.test.ts` already asserts non-empty.
- **F6** `foldSoFar()` returns the live object, so a caller could corrupt the recorder's fold. Correct for wave 1 (no caller exists) but **wave 3's `/api/meta` wiring must handle it** — and it must not be fixed by copying, since the identity law above depends on the reference.


gabriel-canaan (approved):
## Approved — with two review commits pushed to the branch

Reviewed the single commit `fcb341d` against #3's DoD. Fixes are `652e9c7` and `5ff9bd6`, both test-only.

### What verified in the author's favour — all `EXECUTED`

Everything the PR body claims about itself, I re-ran rather than took:

| claim | result |
|---|---|
| suite | `379 files, 6488 passed, 3 expected-fail` — the reported numbers, exactly |
| `npm run typecheck` / `npm run lint` | clean, clean (846 files) |
| CI | all **7** jobs executed and passed — `build-test-boot` ×3 and `pack-smoke` ×4 (`gh pr checks 67`) |
| purely additive | confirmed: `grep` finds **zero** non-test callers of `foldSoFar()`, and the diff contains **no `-` line touching `eventsSoFar`** |
| the `?raw` deviation is forced | **confirmed by running it.** Swapping the `const`-routed specifier for a literal one and running `npm run typecheck` gives exactly `error TS2307: Cannot find module './era-1/recording.jsonl?raw'` at `packages/core/src/eras/corpus.ts(1,27)`. The workaround is justified, and its comment is accurate. |

I re-ran the PR body's own mutation table plus five more. Every entry it lists reproduced:

| mutation | result |
|---|---|
| drop `advanceFold` from `record()` | 9 of 15 fail |
| drop `advanceFold` from `closeWith()` | 1 fails |
| `foldSoFar()` re-folds every read | 3 fail |
| never clear `foldDesynced` in `foldSoFar()` | 1 fails |
| drop `openSession()`'s fold reset | 1 fails |
| seed the ctor with `initialSessionState()` instead of `reduceAll(this.buffer)` | 1 fails |
| `advanceFold` swallows the throw without flagging | 2 fail |
| `advanceFold` no-ops after the first event | 7 fail |
| swap `push` / `advanceFold` in `record()` | **survives** — equivalent, as the body says: `emit` runs after both, so no subscriber can observe the order |

The self-refuting spy law is the strongest thing in this PR. Finding that `vi.spyOn(core, 'reduce')` is structurally blind to `reduceAll`'s module-local binding, and then closing the route with reference identity rather than a third spy, is the right resolution — a spy proves *how many times*, identity proves *whether at all*, and only the second is route-independent.

### The two fixes — each proven by mutation, not argued

**`652e9c7` — `openSession()`'s desync clear was the one unlaw'd line.**

Of the fifteen mutations above, exactly one survived the whole suite:

```
delete `this.foldDesynced = false` from openSession()   →  16/16 green
```

This is the sibling of the law the file *does* pin. The body's own rationale for laws the `foldSoFar()` clear — "after a single caught throw EVERY later read is O(n) forever" — applies one boundary later: a flag raised by a throwing `reduce` in the session that **ended** survives the rotation, so the first `foldSoFar()` read of the **new** session pays a full `reduceAll` rebuild for a desync that is no longer real. Same shape, same cost, no law.

The new test forces a desync, deliberately does **not** read `foldSoFar()` (that would repair it), rotates, records into the new session, and asserts `reduceAll` is never reached. With it present that deletion fails 1 of 16; dropping the whole reset fails 2 of 16.

**`5ff9bd6` — F5 promoted from "risk near zero" to a red test.**

F5 was recorded as non-blocking and reasoned. It is real, and I confirmed it by mutation rather than by argument:

```
replace the corpus read with `{ events: [] }`   →  16/16 green   (before this commit)
                                                →   1 of 16 fails (after)
```

`initialSessionState()` deep-equals `initialSessionState()`, so the whole corpus law holds vacuously on an empty event list. Measured, era-1 carries **100 events across 15 event types** — `pane.activity`, `llm.usage`, `tool.activity`, `worktree.*`, `branch.*`, `commit.landed`, `agent.*`, `trace.span`, `llm.cost` — so the law is substantive today. Two lines now keep it that way: the corpus is non-empty, and the resulting fold is not the initial state. Costs nothing, and the argument for skipping it was longer than the fix.

Both are rung-1 mechanical (missing test cases). Neither touches production code, and the fence is unchanged — same one recorded widening.

- **F6 is the one to carry forward, and the identity law makes it sharper than the body says.** `foldSoFar()` returning the live reference is fine now, and the doc's comparison to `api/meta.ts` / `cli/run.ts` is *not* quite symmetric: `reduceAll()` hands every caller a private fold, so a mutating reader corrupts only itself. `foldSoFar()` hands out the recorder's only fold, and nothing sets `foldDesynced` when a caller mutates it — so there is no repair path. Wave 3's `/api/meta` wiring is where that bites, and the identity law (correctly) rules out fixing it by copying. `Readonly<SessionState>` on the return type, or a `freeze` behind a dev flag, are the two shapes that survive both constraints. Worth an issue against wave 3 rather than a change here.
- **Dead branch, cosmetic.** In the emit-subscriber law, `recorder.foldSoFar() === reduceAll(recorder.eventsSoFar()) ? true : deepEqualFold(recorder)` — the left side is *always* false, because `reduceAll` always allocates. It reads as though identity might hold there. `deepEqualFold(recorder)` alone says the same thing. Left alone; not worth a commit of its own.
- `SessionState` is entirely plain objects, arrays and `Record`s (checked `state.ts:928`), so `deepEqualFold`'s `JSON.stringify` comparison is sound here — no `Map`/`Set`/`Date` to flatten. Fine as written; just noting it was checked rather than assumed.

### What I did NOT run

`scripts/gate.sh` — running it *is* the landing in this repo, and that is the operator's step, not the reviewer's. Landing is handed back.

ciaran-slow:
## Both review commits verified independently, and the cosmetic note taken — `932251d`

Re-ran both mutation claims rather than accepting them. **Both reproduce exactly.**

`652e9c7` — deleting `openSession()`'s `foldDesynced = false`:
```
with the new law present   ->  1 of 16 fails
with it removed (pre-652e9c7) ->  15/15 green
```
Confirmed: that line was genuinely unlaw'd, and the reasoning is right — a flag raised by a throwing `reduce` in the session that *ended* survived the rotation, so the first read of the *new* session paid a full rebuild for a desync that no longer existed. Same shape as the `foldSoFar()` clear, one boundary later. I had law'd one and missed its sibling.

`5ff9bd6` — the empty-corpus case:
```
with the guard present  ->  1 fails: "expected 0 to be greater than 0"
with it removed         ->  16/16 green
```
**This one lands on my reasoning, not just my code.** I recorded F5 as "real risk near zero" and left it. The reviewer's line — *the argument for skipping it was longer than the fix* — is correct, and it is the second time in this PR that a law of mine held for a reason other than the one it claimed. Two lines close it.

### The cosmetic note, taken

Fixed in `932251d`. The mid-emit law read `foldSoFar() === reduceAll(...) ? true : deepEqualFold(...)`, and the left arm is dead by construction — `reduceAll` allocates every call. Worse than dead: it implies identity *might* hold mid-emit, sitting right beside a sibling law whose entire point is that identity holds across reads and nowhere else. `deepEqualFold` alone says what is meant, and the law still bites — emitting before the fold advances gives `equal: false` and fails it.

### One correction to my own verification

My first two attempts at reproducing `5ff9bd6` reported green and proved nothing. Cause: I had run `git checkout 652e9c7~1 -- <file>`, which **stages**, and my subsequent `git checkout -- .` could not undo a staged change — so the pre-review test file sat in the index and the baseline was silently 15 tests, not 16. Everything I concluded before `git reset --hard` was void. Flagging it because a green mutation result from the wrong tree is exactly the failure mode this PR already found twice.

### Gates at `932251d`

typecheck clean · lint clean (846 files) · full suite **379 files, 6489 passed** + 3 expected-fail · fence unchanged, still the one recorded widening.

### F6 carried forward as #69

Filed against wave 3 with the sharpened form, because the asymmetry is the part worth writing down: `reduceAll()` hands each caller a **private** fold, so a mutating reader corrupts only itself, while `foldSoFar()` hands out the recorder's **only** fold and nothing sets `foldDesynced` when a caller mutates it — so there is no repair path. Silent and permanent, and invisible until a caller exists. Both shapes you named (`Readonly<SessionState>`, or a dev-flag freeze) are recorded there, along with why copying is ruled out by the identity law.

Landing is yours.

Issue #4, opened 2026-08-21
prd40 w2: an event reaches the log before it reaches a subscriber


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/server/poll-loop.ts`
- `packages/server/src/recorder/session-recorder.ts`
- `packages/server/src/recorder/session-recorder.test.ts`
- `packages/server/src/server/poll-loop.test.ts`
- `docs/prds/prd-40-the-record-survives-the-write.md`
- `CHANGELOG.md`

**Fence widened again 2026-08-25 (second)**, adding `prd-40` and `CHANGELOG.md`, recorded before the change. The plan below amends prd-40 ruling 1 for the third time — the ruling's "the existing `catch` is reused as-is" sentence is a defect once the append moves first — and the change is user-visible, so it carries a `[Unreleased]` entry. Both belong in the same commit as the code they describe.

**Fence widened 2026-08-25**, adding the two test files. Recorded before any change, per AGENTS.md. The DoD below requires three tests and the fence named no file they could live in — the same unbuildable-as-fenced defect that forced the 2026-08-24 widening. Source scope is unchanged. #3 carried `session-recorder.test.ts` as pre-authorised beside its source file; this makes that explicit for both halves.

**Fence widened 2026-08-24**, from `poll-loop.ts` alone. Review of #25 found this issue unbuildable as originally fenced: the DoD below requires the append to precede the buffer push and the subscriber emit, and both of those are in `session-recorder.ts` — `record()` at `:117-125` and `closeWith()` at `:136-154` — not in the poll loop. prd-40 ruling 1 was corrected to name all three publishing sites (`84b895c`), and its Sequencing declares this wave as ONE issue fencing both files: splitting them would let one half land without the other, which is the half-fix the ruling exists to forbid.

## Blocked by
**UNBLOCKED 2026-08-25.** Wave 0's operator act has landed: **ADR-0029, "A recording may repeat a fact"** (`docs/adr/0029-a-recording-may-repeat-a-fact.md`), accepted, amending ADR-0011. It rules that a recording MAY contain the same fact twice and a reader must not treat that as corruption — at-least-once on the poll path, no read-side dedupe. **The lane cites that ADR; it does not re-derive whether a duplicate is allowed.**

Two findings from the ADR's evidence bear directly on how this is built, and one of them corrects prd-40's own wording:

[...]

PR #68, opened 2026-08-24
prd45 w1: the checks that could fail and say nothing now say so

### LINKED ISSUE #42: prd45 w1: a gate check that cannot run says so

### LINKED ISSUE #43: prd45 w1: a red leg reports its remaining gates
## Fence (may touch ONLY)
- `.github/workflows/ci.yml`

## Blocked by
Nothing. Sibling lanes hold `scripts/gate.sh` + `packages/server/src/gate-honesty-law.test.ts`
and `packages/web/src/recordings/` — do not enter either.

## Definition of done

Ruling 4. This discharges prd-24's residual 2, which retired to `done/` with no
owner named; prd-25's leg design already prices the defect in and benefits.

- Every step after `Test` on every leg carries `if: always()`, so a red `Test` no longer skips the gates behind it. Today `:66` Typecheck, `:69` Lint, `:72` the packaging guard and `:97` the boot smoke carry **no `if:` at all** — a red leg is four unrun gates, a fact `AGENTS.md` already states and CI does not honour.
- A leg that fails `Test` still reports Typecheck, Lint, packaging and boot smoke results rather than skipping them silently.
- The job's overall conclusion is still failure when any step failed — `if: always()` must not convert a red leg into a green one.
- The matrix-conditional `Setup Node` steps at `:43` and `:49` keep their existing `if:` semantics; `always()` is not bolted onto a step whose condition is a matrix selector.

## Proving it

`EXECUTED`, with exit statuses. Two directions, because a change that makes
everything run must not also make everything pass:

- **A rigged failing `Test` still reports the later steps.** Break a test on a branch, push, and show the run: `Test` red, and Typecheck/Lint/packaging/boot-smoke each reporting a real conclusion rather than `skipped`. `gh run view <id> --json jobs -q '.jobs[].steps[] | "\(.conclusion) \(.name)"'` is the read that shows it.
- **The leg is still red overall.** Same run: the job's conclusion is `failure`, not `success`. This is the mutation that matters — `if: always()` is one keystroke away from `continue-on-error`, which would hide exactly what this issue exists to expose.
- **A clean run is unchanged.** All steps `success`, no new skips, no new failures.

## Two tests read this file — neither should need touching

Checked by hand at grooming, stated here so you do not flinch:
`packages/app/src/host/node-floor-law.test.ts:51,55` asserts `node-version:`
strings and a count of exactly 2; `packages/app/src/host/boot-line.test.ts:72`
asserts the boot-line marker appears in the workflow. Both are text `toContain`
on strings `if: always()` does not touch. If your change does make either go red,
that is a fence widening — record it on the PR **before** making it.

## Out of scope

No new job, leg or step. The `windows-latest` leg is prd-25's. Nothing about
what CI runs, only about what it still reports when something fails.

### LINKED ISSUE #44: prd45 w1: the recordings law walks what it claims
## Fence (may touch ONLY)

## Blocked by
Nothing. Sibling lanes hold `scripts/gate.sh` + `packages/server/src/gate-honesty-law.test.ts`
and `.github/workflows/ci.yml` — do not enter either.

## Definition of done

prd-24 ruling 3, which still governs and is **not** re-ruled by prd-45: a law's
walked scope is derived from what it claims, never hardcoded. This is the last of
the three flat walks that ruling named.

- `no-live-fleet-law.test.ts:31` walks the governed root recursively, so a file added at any depth is checked or turns the law red. The fix pattern is this law's own siblings' `walkSourceFiles` — reuse it, do not write a third walker.
- `:39`'s floor is **derived from the walk**, not the hardcoded `toBeGreaterThan(3)` it is today.
- The `FORBIDDEN_PATTERNS` set at `:19` is unchanged — what the law forbids is not this issue; only the scope it forbids it over.

## This defect is LATENT — say so, and prove it stays fixed anyway

Measured at grooming: `packages/web/src/recordings/` has **zero
subdirectories**, so the flat `readdirSync` currently sees all 13 source files
and the floor of 3 has enormous headroom. **It hides nothing today.** Do not
write a commit message claiming it does.

What makes it real is prd-24's own evidence for the sibling law: the `lab/` walk
was flat too, and it missed `lab/branching/geometry.ts:1` importing
`../../scene/palette.js` — the exact import the law forbids, invisible because
the walk did not descend. This one fires the day someone adds a subdirectory.

So the proof is a fixture, not the current tree. `EXECUTED`, both directions:

- **The deep walk sees a nested file the flat one missed.** Create a subdirectory under the governed root containing a file that matches a forbidden pattern. Before: the law passes (the file is invisible). After: the law fails, naming that file. Paste both. Remove the fixture, or keep it as a committed negative if it reads as one.
- **The derived floor cannot be satisfied vacuously.** Show that a walk returning fewer files than the root actually contains fails, where the hardcoded `> 3` passed.
- **The current tree still passes.** 13 source files, no forbidden patterns, law green.

## Out of scope

The two walks prd-24 named as only vacuously correct —
`interaction/no-model-call-law.test.ts` and `connect/index.test.tsx` — are
**deliberately not yours**: taking them would widen this fence to the whole web
package. prd-24's `geometry.ts` carve-out is a decision owed to whoever takes
them, not to you.

gabriel-canaan:
prd-45 wave 1 — the three places a check could fail and say nothing. One commit per issue, reviewed commit by commit.

Closes #42
Closes #43
Closes #44

**Note on shas:** cherry-picked from three lane branches onto current `origin/main`, so these commits are not the shas the issue comments cite. Same work, re-parented.

## What becomes true

- **`4c1da73` (#42)** — every check in `scripts/gate.sh` distinguishes "could not run" from "ran and found nothing". Eight guards, each fixed and each proven in both directions. `packages/server/src/gate-honesty-law.test.ts` is the new law holding the line.
- **`b779435` (#43)** — a red `Test` step no longer skips the four gates behind it. Four `if: always()` lines, and nothing else.
- **`4f1de19` (#44)** — the recordings law walks the governed root recursively, and its floor is derived from the tree rather than typed.

## Verification

Gates on this merged branch, node 22.22.2: `npm run typecheck` exit 0 · `npm run lint` exit 0 (847 files) · `VITEST_MAX_WORKERS=6 npm test` exit 0 (**380 files, 6520 passed**, 3 expected fail).

Two rounds. Round 1 found a real defect in two of the three commits; both were sent back and both are fixed. The findings, because the fixes are the interesting part:

- **#42's law swept two idioms and missed the third.** It swept `|| true` and `|| echo` but never `2>/dev/null` — while `gate.sh` carries seven occurrences of it, five undeclared, **including the lines the commit body itself named as "a real ninth instance"** of the very defect class the law exists to close. So the law was green while that class was live in the file it governs. The DoD's own words: *"a tolerance nobody can enumerate is how prd-39's class survived."* Now fixed as an idiom table with nine declared tolerances, and the ninth instance is declared as a dated `KNOWN GAP` with its reason and its deferral — enumerated rather than fixed, since fixing it is out of this issue's fence.
- **#44's "derived" floor was a literal.** `toBeGreaterThanOrEqual(13)` under a comment claiming derivation from the walk. Proven vacuous: with 14 real files, a walk that drops a one-file subdirectory returns 13 and passes. Now `expect(walk).toEqual(realSourceFileNames())`, where the right side is an independent traversal (node's own `readdirSync{recursive:true}`) — a second mechanism, so a bug in the hand-rolled recursion cannot hide in both sides.

Mutation evidence for the two fixes, run on this merged branch:

| mutation | result |
|---|---|
| #42 law, untouched tree | 45 passed |
| plant an undeclared `2>/dev/null` swallow beside a printed verdict | 1 failed — names the planted line |
| #44 law, untouched tree | 3 passed |
| add a nested source file, walk intact (control) | 3 passed — no false positive |
| add a nested source file, remove the recursion | 1 failed — names `nested/leak.ts` |

## What is the sibling case?

- **#42** — the eight guards' ninth sibling is the timing-file discovery at `gate.sh:149-152`: a `grep -rlF` and a `find`, both `2>/dev/null`, feeding a process substitution whose producers' exit status is never checked. That is the identical `:83` shape one screen away. It is **deliberately not fixed** — the issue's DoD says "work that list; do not invent additions to it", and prd-45 ruling 1 already names it as the successor's job. It is declared in the law as a dated known gap so it cannot be forgotten.
- **#43** — the sibling is any step added after `Test` in future without its own `if: always()`. Nothing in the tree enforces that; the commit records it as a gap rather than pretending otherwise.
- **#44** — the sibling is every other scope-assertion in `packages/web`. Two remain flat-walked and are named: `interaction/no-model-call-law.test.ts` and `connect/index.test.tsx`, both out of fence (prd-24 named both).

The table above is the answer for #42 and #44 — each row was executed on this branch with the tree restored afterwards. For **#43 the honest answer is: none was run**, because none can be. See below.

## #43's evidence is incomplete by construction — and it is the reviewer's to finish

`ci.yml` triggers on `push: branches: [main]` and `pull_request` only. A lane branch push fires no run, and lanes do not open PRs, so **the live-CI half of #43's definition of done could not be collected in the lane**. `act` and `actionlint` are not installed. The lane executed what it could — a pyyaml shape assertion per step, a `continue-on-error` regression grep, and the two neighbour tests — and explicitly refused to claim a run it never observed.

**This PR is the first place that evidence can exist.** The recipe, for whoever reviews:

1. On this branch, break one test deliberately and push.
2. `gh run view <id> --json jobs -q '.jobs[].steps[] | "\(.conclusion) \(.name)"'`
3. Assert both: Typecheck / Lint / packaging / boot-smoke each report a **real conclusion rather than `skipped`**, and the job's overall conclusion is still **`failure`** — `if: always()` is one keystroke from `continue-on-error`, which would turn a red leg green.
4. Revert the break.

## Landing — read this before running anything

**`scripts/gate.sh` must not land this PR.** #42's fence *is* `gate.sh`: the gate would be merging, installing and pushing using the very script this PR rewrites. #42 books the landing as a hand-landing for exactly that reason, and prd-39 booked its wave 0 the same way. Because one PR is one (milestone, wave), #43 and #44 ride in that same manual landing.

## Fence widenings

None. Each commit touched only its issue's fence: `scripts/gate.sh` + `packages/server/src/gate-honesty-law.test.ts` · `.github/workflows/ci.yml` · `packages/web/src/recordings/no-live-fleet-law.test.ts`.

## Reviewer note

I verified this and sent two of the three commits back, so I should not approve it. `AGENTS.md` is explicit that the reviewer fixes but never reviews their own fix into main. Both round-1 findings were mine; a second pair of eyes on the fixes is worth more than usual, particularly on #42's tolerance table — a declared tolerance is only as good as its reason, and there are now nine of them.


gabriel-canaan:
## Verify: not merge-ready — 6 must-fix, 6 worth-later

Two independent passes (Claude Opus at `--effort xhigh`, Codex `gpt-5.6-terra` at `model_reasoning_effort=xhigh`), each in its own disposable worktree, neither seeing the other's output. Builder was Sonnet 5, so no seat shared the builder's family.

**I re-executed the decisive probe for six of these myself** rather than relaying a seat's claim — `AGENTS.md` is explicit that nothing from a seat may be posted unverified, and that a unanimous verdict is the tell. The two passes agreed on exactly **one** finding; everything else was found by one and missed by the other, which is the opposite of the manufactured-consensus shape. **I did not run the cross-review or Fable reconcile seats** — independent execution against the code is what those stages are a proxy for, and I had already done it. Stating that rather than implying the full pipeline ran.

Worth recording: **my own two orchestrator rounds found none of the six below**, and finding 4 is a hole my round-1 finding opened.

---

## Must fix before merge

### 1. `gate.sh:215` — the containment postcondition proves containment in `HEAD`, not `main`
*Both passes. EXECUTED by me.* #42's ruling 2 requires proving the lane commit is contained **in main**; the line checks `HEAD`, while its own failure text says "not contained in main". Nothing in the script asserts the root checkout is on `main` — and this repo's dispatch preflight exists precisely because the main checkout "is routinely parked on a feature or audit branch".

```
root parked on: audit-branch
shipped   (--is-ancestor LANE_SHA HEAD) -> exit 0   "merge complete"
required  (--is-ancestor LANE_SHA main) -> exit 1   NOT in main
```

Opus adds the sharper half: the check sits **before** `MERGED=1`, and `workmux merge` has already deleted the branch and worktree by then, so a false fire prints `">>> HOLDING — not merged"` after cleanup — the exact outcome the `MERGED` flag exists to prevent. The predicate it replaced was self-limiting (a branch ref can only be missing after cleanup ran); this one has no such floor. One word: `HEAD` → `main`. Note the law cannot catch it — its own fixture runs with the root on `main`, encoding the same premise.

### 2. `gate-honesty-law.test.ts:193` — the law enrols itself in gate.sh's timing set
*Opus. EXECUTED by me.* The tolerance needle contains the literal `// @gate-timing`, which is the marker `gate.sh:151` greps for.

```
origin/main:  TCOUNT=6
this branch:  TCOUNT=7   <- packages/server/src/gate-honesty-law.test.ts
```

The issue said that contract must stay undisturbed. Two consequences: a subprocess-heavy `git init`/`rebase` test is moved out of the 4x load batches into the serial timing pass, and the first landing ratchets `.swarm/timing-count` to 7 — so the successor that fixes the declared KNOWN GAP edits that same tolerance line, drops it back to 6, and **hard-fails the operator's landing** naming a fault that does not exist. Fix is re-anchoring on a marker-free substring of the same line.

### 3. `.github/workflows/ci.yml:75,101` — `if: always()` turns two honest `skipped`s into unearned greens
*Opus, EXECUTED with controls; mechanism confirmed by me.* The stated goal (a red **Test** no longer skips the four gates) works and is a real win. The collateral is the **Build**-red and **Install**-red paths, which those steps now also run on.

- The packaging guard only ever filters for *unexpected* files (`unexpected = files.filter(...)`; passes when none). With no `dist`, it reports `Packaging guard passed — 4 files, all allowlisted` at exit 0. Zero build output is indistinguishable from a correct pack.
- The boot smoke is worse: `packages/server/bin/rhizomorph.mjs:36` falls back to TypeScript source (`'dist/cli/index.js not built'`), so the step that exists to prove *the built artefact* boots reports green without touching a build product.

A Build-red leg now reads: Build ❌ · Test skipped · Typecheck ✅ · Lint ✅ · Packaging ✅ · Boot smoke ✅. Before this commit all four read `skipped` — the honest answer. In-fence repair: `id: build` on Build, then `if: "!cancelled() && steps.build.outcome == 'success'"` for the packaging guard and boot smoke, `if: "!cancelled()"` for typecheck and lint.

### 4. `no-live-fleet-law.test.ts:104` — the repair deleted the only non-vacuity guard
*Opus. EXECUTED by me.* The replaced assertion was `toBeGreaterThan(3)` under the name *"has source files to check at all — an empty walk proves nothing."* `toEqual(realSourceFileNames())` fixed the rotting-literal problem but `[]` equals `[]`.

```
CONTROL, untouched tree:                       3 passed
MUTANT, RECORDINGS_DIR -> an empty directory:  3 passed   <- whole law vacuous
```

All three tests pass, including a forbidden-pattern sweep iterating zero files. The sibling `lab/no-live-fleet-law.test.ts:158` still carries this guard. Restore a `> 0` floor above the equality — a non-vacuity floor, not a completeness floor, so it does not reintroduce the literal that rots.

**This one is on the review process, not the lane.** My round-1 finding said the floor was a literal masquerading as derived; the repair removed the floor. A fix to a detector is itself a detector.

### 5. `gate.sh:159` — the ratchet dies silently on an all-digit but oversized count
*Codex. EXECUTED by me, under bash (the real interpreter).* The new `case` guard rejects non-digits, but a 20-digit value passes it and then the comparison itself fails.

```
CONTROL prev=9  (should hold): HOLD
CONTROL prev=3  (should pass): PROCEEDS
CORRUPT 20-digit:  bash: [: 99999999999999999999: integer expected
                   PROCEEDS        <- ratchet silently dead
```

A check that could not run, printing no fault — prd-45's thesis, in the commit closing prd-45's issue.

### 6. `gate.sh:103` — a predictable `/tmp` path, written through a symlink
*Codex EXECUTED; Opus rates it minor (#9). EXECUTED by me.* `NUL_LIST="/tmp/gate-nul-list-$H"` with `>` follows a pre-existing symlink at that predictable path.

```
victim before: 480 bytes
victim after :   4 bytes        (symlink still a symlink)
CONTROL, no symlink: writes a real file, victim untouched
```

Introduced by this commit's `-z` rewrite. `mktemp` closes it. Two sibling paths share the shape (`/tmp/gate-nul-$H.log`, `/tmp/gate-timingcount-$H.log`), and the list leaks when `fail` fires inside the loop. I place this in must-fix rather than minor because it is a data-loss path this PR introduces and the fix is one line — but Opus's lower rating is defensible, since exploiting it needs local access.

---

## Worth doing later

7. **The idiom sweep is a three-string allowlist, and the dominant form of the defect is not a string.** *Opus, EXECUTED.* Planting a plausible guard after `gate.sh:116`: `2>/dev/null` is caught (control), while `|| :`, `2> /dev/null`, `&>/dev/null`, `||true`, `RES=$(cmd | tail -1)` and `cmd || { echo "(could not check)"; }` all pass 45/45. The last shape is **already live in the governed file**: `gate.sh:82` captures `n=$(git log --oneline main..HEAD | wc -l)` and never reads the status — with `.git/HEAD` removed it holds with *"no commits on the branch"*, a fault it did not earn, and the law reads green. A structural rule (`$(...)` assignment whose next line is neither `_RC=$?` nor `|| fail`) would catch line 82 and five of the six evasions at once. Relatedly, `:226`'s "sweep is not vacuous" control never runs the sweep.
8. **The coupling the law's own doc names is not in the registry.** *Opus, EXECUTED.* Rewording any anchored line in `gate.sh` gives `Tests: no tests` — a collection failure taking all 45 assertions out of service from a directory the editor never entered. Out of #42's fence; the one-line `.swarm/coupling.txt` entry lands separately. The existing gate.sh entry's line references were also moved ~50 lines by this PR.
9. **`gate.sh:69`'s fence audit still reads `--name-only` line-delimited** — the quoting defect `-z` fixed one screen away. With an in-fence `café.ts` it reports a fence violation for a file entirely inside its fence. Fails safe (false hold, never false pass); `-c core.quotePath=false` closes it.
10. **Declared tolerance #7's reason is true but incomplete.** *REASONED.* `merge-base --is-ancestor` returns 1 for *not an ancestor* and 128 for *bad object*; the discarded stderr was the only thing distinguishing them, and both land on "the rebase did not take". The other eight reasons check out, including both dated KNOWN GAP entries.
11. **#44's nested fixture is not committed**, so the recursive branch is never exercised by the committed suite. Codex rates this must-fix; Opus and I rate it later, because #44's own DoD permits *"remove the fixture, or keep it as a committed negative"*, and the equality assertion re-arms the instant a subdirectory exists (verified both directions).
12. **`gate-honesty-law.test.ts:399,428` assume non-root** — they assert a write to a `0444` file fails, which is false as root, so they go red in a root container.

---

## What is genuinely good, said plainly

Every one of #42's eight guard fixes is proven in both directions, and the extract-the-real-lines-from-the-tracked-script technique is the right call rather than reimplementing the guards in the test. #43's commit body refuses to claim a CI run it could not observe, which is the behaviour this PRD is about. #44's equality-against-an-independent-traversal is a better property than the sibling law's hardcoded floor, and no file can ever be silently ungoverned once a subdirectory exists. Fences are exactly as declared, with no scope creep in any of the three commits. Gates re-run and confirmed independently: typecheck 0 · lint 0 (847 files) · full suite 0 (380 files, 6520 passed, 3 expected fail).

The findings above are concentrated in the two places this PR could not test itself: the landing script it must never run, and a CI workflow that only fires on `main` and `pull_request`.

gabriel-canaan:
## Round 2 — all six must-fix items closed, each re-checked with a spelling it was not built against

Branch re-bundled from the three amended lane commits: `3102b19` (#42) · `c161cd1` (#43) · `d57ae6f` (#44).

The re-check rule matters here more than usual, because must-fix 4 *was* a hole opened by round 1's repair. Re-running the probe that found a defect only proves that input is caught. So every row below uses a different input.

| # | fix | re-check (new spelling) | result |
|---|---|---|---|
| 1 | `gate.sh` containment `HEAD` → `main` | root on a **detached HEAD** at a commit containing the lane sha (round 1 used a *named side branch*) | old form exit 0 "merge complete"; amended exit 1 holds; control (genuinely merged) passes |
| 2 | tolerance needle re-anchored off the `// @gate-timing` marker | re-ran gate.sh's own discovery verbatim | **TCOUNT 7 → 6**; the law is no longer in the timing set |
| 3 | `ratchet` reads the comparison's own exit code | `2^63` (one past int64 max), `int64 max` exactly, 22-digit with leading zeros (round 1 used a plain 20-digit) | `2^63` → **HOLD "cannot compare"** (was PROCEEDS); int64-max → holds as a ratchet; controls 5<6 hold, 5<4 pass |
| 4 | non-vacuity floor restored above the equality | governed dir **non-empty but every file filtered out** (round 1 used an *empty* dir) | `expected 0 to be greater than 0` — 1 failed / 3 passed; control 4/4 |
| 5 | `mktemp` replaces the predictable `/tmp` path | — | 7 `mktemp` uses; the predictable-path write is gone |
| 6 | CI steps gated on `steps.build.outcome`, not `always()` | — | `id: build`; `!cancelled()` on typecheck/lint, `!cancelled() && steps.build.outcome == 'success'` on packaging guard and boot smoke |

Gates on the re-bundled branch, node 22.22.2: `typecheck` 0 · `lint` 0 · suite 0 — **380 files, 6530 passed**, 3 expected fail (up from 6520; the new guards).

### Still outstanding, and it is the reviewer's to collect

**#43's live-CI evidence has not been gathered.** Its fix now *reduces* what runs on a Build-red leg, which is correct — but that makes the live proof more valuable, not less. What must be shown on this PR, per the recipe in the description: a **Test**-red leg still reports all four gates (the original win must survive the repair), a **Build**-red leg reports packaging and boot smoke as `skipped` rather than `success`, and both legs still conclude `failure`. Until that runs, item 6 above is verified by shape only.

### Deferred — the six worth-later findings are not addressed here

They remain as posted: the three-string idiom allowlist and the live unchecked pipeline at `gate.sh:82`; the missing `.swarm/coupling.txt` entry; `gate.sh:69`'s line-delimited fence read; tolerance #7's incomplete reason; #44's uncommitted nested fixture; and the two root-container assertions. Each needs a home before this lands, so they are not lost — an issue apiece rather than a silent deferral.

gabriel-canaan:
## #43's live-CI evidence — collected, and it took three probes

The half of #43's definition of done that no lane could reach (`ci.yml` fires only on `push: branches:[main]` and `pull_request`) is now gathered on this PR. Branch is back to `d57ae6f`; no probe commit survives.

### The evidence

**A red `Test` leg still reports all four gates** — the original win, and it survives the round-2 repair. Probe `bc2fc05`, run 32789109097, identical on macos-current, ubuntu-current and ubuntu-min:

```
Build      success
Test       FAILURE      <- rigged
Typecheck  success      <- ran, not skipped
Lint       success      <- ran
Packaging  success      <- ran
Boot smoke success      <- ran
job conclusion: failure
```

**A red `Build` leg now skips the two steps that cannot earn a verdict.** Probe `b678108`, run 32790386326, identical on all three legs:

```
Build      FAILURE
Test       skipped
Typecheck  failure      <- ran (!cancelled()), failed honestly
Lint       failure      <- ran, failed honestly
Packaging  SKIPPED      <- under always() this reported "passed — 4 files, all allowlisted"
Boot smoke SKIPPED      <- under always() this reported green off TypeScript source
job conclusion: failure
```

Both legs conclude `failure`. `if: always()` did not become `continue-on-error`, and the outcome-gated form did not become one either.

**A clean run is unchanged:** run 32788556721 on `d57ae6f`, all seven checks pass.

### The middle probe was invalid, and that is worth recording

Probe `c81beb2` was meant to be the Build-red case and was not. It put a **type** error in a file **nothing imports** — esbuild strips types without checking them, and an unimported file never enters the bundle graph — so `Build` succeeded and `Typecheck` took the failure instead. The tell was `pack-smoke` passing, which it could not have done against a genuinely broken build.

That is this PR's own defect class, committed while verifying this PR: a check that appears to have run and has actually tested something else. Probe C corrected both faults — a real syntax error in `packages/server/src/cli/index.ts`, the actual esbuild entry point, with `npm run build` confirmed exiting 1 **locally before pushing**, which is the control probe B never had.

It bought a third data point rather than nothing: a **Typecheck**-red leg reports Lint, packaging and boot smoke and still concludes `failure`.

### Status

All six must-fix findings are closed and re-verified with fresh spellings; the live-CI half of #43 is now evidence rather than a claim. The three deferred `gate.sh` findings have a home in **prd-46** (blessed 2026-08-25, milestone `prd46`), groomed as **#70** (wave 1, the structural predicate plus the live `gate.sh:82` instance, folded as a stack) and **#71** (wave 2, `:69`'s quoting). **prd-45 ruling 4 carries a dated amendment note** recording that `always()` was falsified in practice and naming the mechanism that replaced it — its verdict is unchanged, so existing citations still resolve.

Taking this out of draft. **It must not be landed with `scripts/gate.sh`** — #42's fence is that script, so this is a hand-landing, as #42 books and prd-45 wave 0 requires.

ciaran-slow (approved):
## Second pair of eyes — the code verifies; four corpus/tracker items to close before it lands

Reviewed commit-by-commit in a clean detached worktree with `origin/main` **merged in** (the branch was 12 commits behind), node 22.23.2. The author asked for a second reviewer because both round-1 findings were their own — this is that pass. Everything below labelled `EXECUTED` I ran myself and read the exit status; where I only reasoned from the code I say so.

---

**The post-merge tree is green, and it is a bigger tree than the one the author measured.** Main's 12 commits touched `packages/web/src/recordings/` — #44's own fence — so the derived-floor law was exercised against a tree it was not written on:

```
npm run typecheck   exit 0
npm run lint        exit 0   (848 files)
VITEST_MAX_WORKERS=6 npm test
                    exit 0   381 files, 6548 passed, 3 expected fail
```

**#43's live-CI evidence is exactly as described.** I pulled all three runs and read every step conclusion on all three legs, not just the summary. Test-red probe `bc2fc05` (run 32789109097):

```
Build success · Test FAILURE · Typecheck success · Lint success
Packaging success · Boot smoke success        job conclusion: failure
```

Build-red probe `b678108` (run 32790386326):

```
Build FAILURE · Test skipped · Typecheck failure · Lint failure
Packaging SKIPPED · Boot smoke SKIPPED        job conclusion: failure
```

Identical on macos-current, ubuntu-current and ubuntu-min. Both legs conclude `failure`; neither `if:` degraded into `continue-on-error`. This is the half of #43 no lane could reach, and it is now real evidence rather than a claim.

**Must-fix #2 is genuinely closed, re-derived independently.** I ran `gate.sh`'s discovery pipeline by hand against the merged tree rather than through the law's harness:

```
branch:      6 files   (3 bench + 3 @gate-timing)
origin/main: 6 files   (same six)
```

The law file is absent from both. The ratchet will not move.

**The laws bite. Five mutations, tree restored and verified clean after each:**

| mutation | result |
|---|---|
| `gate.sh` containment `main` → `HEAD` (revert must-fix #1) | law RED at collection |
| plant an undeclared `2>/dev/null` guard | RED — **names the planted line**, 1 failed / 53 passed |
| revert the `CMP_RC` three-way to round-1's `&&` form | law RED at collection |
| #44: add `recordings/nested/leak.ts`, walk intact (control) | 4 passed — no false positive |
| #44: same file, recursion removed | RED — names `nested/leak.ts`, 1 failed / 3 passed |
| #44: governed root → populated dir where every file filters out | RED — `expected 0 to be greater than 0` |

The last row is must-fix #4's non-vacuity floor doing its job on a *populated* directory, not merely an empty one — the case that distinguishes it from the hole round 1 opened.

**A sibling I went looking for and did not find.** The `dirty` count changed from a direct pipe to `printf '%s' "$STATUS_OUT" | grep -v … | wc -l`. `$( )` strips the trailing newline, and `wc -l` counts newlines — so a single stranded file looked like it could count as 0 and report a clean worktree. EXECUTED: grep re-adds the terminator, both forms return `1`. No defect. The display sibling one line down correctly uses `printf '%s\n'` while the counting one uses `printf '%s'` — that asymmetry is right, not a slip.

Also confirmed: `bash -n` clean · no `set -e` added · exactly one `EXIT` trap in the file, so `trap - EXIT` at :129 clobbers nothing · all four `mktemp` scratch files removed on both branches · fences exactly as declared, three commits, no probe commit survives in the branch.

---

## Findings — none are code defects; all four are corpus/tracker

### 1. prd-45 ruling 4 carries no amendment note. `EXECUTED`

The third comment states: *"prd-45 ruling 4 carries a dated amendment note recording that `always()` was falsified in practice and naming the mechanism that replaced it."* It does not — on this branch or on `origin/main`:

```
$ git show origin/main:docs/prds/prd-45-the-earned-verdict.md | sed -n '131,134p'
## Ruling 4 — a red leg still produces the evidence of its remaining gates

Every step after `Test` in `ci.yml` carries `if: always()`.
```

This matters more than a stale sentence usually would, because ruling 4 prescribes precisely the thing this PR **falsified with live CI evidence**. A future reader citing it reintroduces the unearned-green defect on the Build-red path. The deviation *is* properly recorded on issue #43 (comment of 23:02:32Z, before the change) — it is only the ruling itself that was never amended.

`docs/prds/` is in none of the three fences, so this is a fence widening to record on the PR **before** making it, or a separate docs commit. Suggested text, so it is a paste rather than a drafting job:

> **Amendment, 2026-08-25 (PR #68).** `if: always()` was falsified in practice and is not what landed. On a Build-red leg it also runs the packaging guard and the boot smoke, both of which report green having verified nothing: the guard only flags *unexpected* files and passes over an absent `dist/`, and `packages/server/bin/rhizomorph.mjs` falls back to TypeScript source when `dist/cli/index.js` is missing. What landed instead: `if: "!cancelled()"` on Typecheck and Lint, and `if: "!cancelled() && steps.build.outcome == 'success'"` on the packaging guard and boot smoke. The ruling's verdict — a red leg still produces the evidence of its remaining gates — is unchanged; only the mechanism is.

### 2. Four of the six worth-later findings have no home. `EXECUTED`

Round 2 committed to *"an issue apiece rather than a silent deferral."* Round 3 named homes for two. Checking every open issue:

- item 7 (structural predicate + the live `gate.sh:82`) → **#70** ✓
- item 9 (`:69`'s line-delimited fence read) → **#71** ✓
- item 8 (coupling registry entry) → no issue
- item 10 (tolerance #7's reason incomplete re: exit 128) → no issue
- item 11 (#44's uncommitted nested fixture) → no issue
- item 12 (the two root-container assertions) → no issue

I have not filed these — boarding an issue here means milestone, timeline, priority and type, which is the groomer's act, not the reviewer's.

### 3. `.swarm/coupling.txt:45`'s line references are stale, and this PR widens the gap. `EXECUTED`

The tracked entry cites `gate.sh:100-101`, `:104`, `:110`, `:138`. Measured by anchor text:

```
                              coupling.txt says   origin/main   this branch
@gate-timing discovery                 100-101           119           166
"timing pass matches ZERO"                 104           123           170
the ratchet's shrink message               110           129           195
.swarm/timing-count                        138           125           172
```

It was already wrong on `main` before this PR — worth saying plainly, since it is not a defect this branch introduced. It is item 8's territory and belongs in whatever issue gets filed for it.

### 4. No prd-46 PRD document exists. `EXECUTED`

The comment describes prd-46 as *"blessed 2026-08-25"*, and the `prd46` milestone plus #70/#71 are real. But `docs/prds/` contains no prd-46 file, on `main` or here. AGENTS.md has the PRD blessed before a single issue exists, so either the doc is unpushed or the two issues are ahead of it.

---

## Verdict

**The code is sound and I found no defect in it.** All six round-2 must-fix items are closed and I re-proved four of them by a route the fix was not built against. The extract-and-run-the-real-lines technique in `gate-honesty-law.test.ts` is the right call, and the mutation table above is what convinced me — the law fails for the right reasons and stays green under a legitimate control.

Finding 1 should land with this PR or immediately after it; the other three are tracker hygiene that should not hold the queue.

**Landing:** per #42, `scripts/gate.sh` must not land this — the gate would be merging using the very script this PR rewrites. Hand-landing, operator's step. I have not merged it.

### Not run

`scripts/gate.sh` (never, by design) · `workmux merge` · any push to `origin`. The `windows-latest` leg does not exist yet, so nothing here says how these `if:` conditions behave there.

gabriel-canaan:
## Landing gate: one genuine blocker, and it was created by round 2's own repair

An independent Codex seat reviewed **the fixes** (not the original commits) at `d57ae6f`, since nobody but me had checked them. It returned NOT SAFE TO LAND with three items. **I re-executed all three myself.** One blocks, two are cheap corrections. Flagging to whoever is mid-review: this is input, not a fix pushed under you — nothing has been changed on the branch.

### BLOCKER — the law now races itself under the gate's own load batches

**`packages/server/src/gate-honesty-law.test.ts:457`. EXECUTED by me, with control.**

Round-2 repair #2 re-anchored a tolerance needle so the law would stop enrolling itself in the timing set (TCOUNT 7 → 6). Correct — and it has a consequence nobody traced: the timing set is exactly what `gate.sh:207-209` **excludes** from its four concurrent load runs. Leaving the timing set means the law now runs in **all four simultaneously**.

Two of its fixtures manipulate one fixed global path, `/tmp/gate-nul-list-t42` — they `unlink` it, `symlink` it at a victim file, redirect into it, and `unlink` it again. Four concurrent copies of that is a race.

```
CONTROL, one run alone:              54 passed
FOUR concurrent, as gate.sh:209:     run 1 exit 0
                                     run 2 exit 1   ← 1 failed
                                     run 3 exit 1   ← 1 failed
                                     run 4 exit 1   ← 1 failed
```

The failing test is, precisely, `EXECUTED — the OLD static /tmp path IS followed by > through a pre-existing symlink, corrupting its target`.

So **the load gate would fail on this PR's own law** — a false hold, on the script whose subject is not printing verdicts it did not earn. Fix is contained: give those fixtures unique paths (`mktemp`), the same remedy repair #5 applied to `gate.sh` itself one file over.

This is the **third** time in this PR's review that a repair opened an adjacent hole: round 1's floor fix made the recordings law vacuous, round 1's sweep fix put the timing marker into the law, and now round 2's marker fix moved the law into the concurrent batches. Worth stating plainly rather than treating each as a one-off.

### Cheap correction — the sweep claims a class it matches by literal

**`gate-honesty-law.test.ts:211-228`. EXECUTED (Codex), mechanism confirmed by me.** The sweep's doc says a bare stderr redirect "is swept below", but it matches the literal string `2>/dev/null`. A guard spelled `command 2> /dev/null` — one space — passes. Codex measured the count moving 7 → 8 for the exact spelling and staying at 7 for the spaced one.

The *class* fix belongs to prd-46 **#70**, which is filed for exactly this and should not be pulled forward. But a law about honesty should not contain a sentence overstating its own reach: either make the match whitespace-tolerant or narrow the claim to the spelling it actually covers. One line either way.

### Cheap correction — a declared tolerance's reason overstates the guarantee

**`gate-honesty-law.test.ts:187`. EXECUTED (Codex), provenance checked by me.** Tolerance #4 justifies `workmux path "$H" 2>/dev/null` on the grounds that *"the very next line ([ -d "${W:-}" ] on :24) checks the result and falls back"*. The existence check accepts **any** existing directory, not the right one — so a `workmux path` that fails while emitting some other real directory is used rather than falling back, and the gate could audit one worktree while merging a different handle.

`gate.sh:23-24` is **pre-existing and untouched by this PR** — I verified it appears in no hunk of the diff. Only the *reason* is new. So this is a wording fix on a tolerance, not a code defect this PR introduced. Two of the nine reasons have now been found to overstate (this and #7's `--is-ancestor` exit-code conflation); the other seven check out.

### What the seat confirmed as correct

Containment holds for an unmerged lane with the root detached **and** with `main` checked out in another worktree, passes after a real merge, and holds cleanly mid-rebase. `CMP_RC` permits ordinary non-shrink values and holds at the first signed-integer overflow. Timing discovery finds the six intended files and excludes this law. The recursive recordings law passes with a safe nested file and fails on a nested `useFleet` mutation. Typecheck, lint, `bash -n` and CI YAML parsing all pass.

### Caveat on the seat's environment, stated because it matters

Its host could not run the full suite — Node's `child_process` returned `spawnSync git EPERM` even where git itself succeeded — so it used direct bash probes. That is why I re-ran every finding here rather than relaying: the concurrency evidence above is mine, on this machine, with a passing single-run control.

A second seat (Opus) is still running; I will post anything it adds.

gabriel-canaan:
## Second landing-gate seat (Opus): same single blocker, independently — plus three follow-ups

Both seats ran blind to each other and **converged on one blocker**, which is the opposite of the manufactured-consensus shape `AGENTS.md` warns about — they differ on everything else. Opus re-verified the gates itself rather than trusting the posted numbers (380 files, 6530 passed, 3 expected fail — matched exactly) and reports the diff scope clean: 4 files, 3 commits, one per issue, nothing unrequested. It could not open an adjacent hole in five of the six repairs.

### The blocker, with harder evidence than mine

```
4 concurrent runs x 4 rounds:                    2/4, 3/4, 4/4, 4/4 red  -> 13 of 16
CONTROL, same 4x on a sibling law (no shared path):  0/4 red
CONTROL, single run:                                 54/54 pass
```

Two failure modes, not one: `ENOENT: unlink '/tmp/gate-nul-list-t42'`, and an assertion failure where one run's `>` truncated the victim *after* another had re-created the symlink.

**Why it blocks rather than annoys** — and this is worse than I stated earlier. `gate.sh:210` runs the suite 4x concurrently per batch, and the load argument is mandatory for anything touching tests (`gate.sh:138`). So the documented landing path runs into `gate.sh:215` → `fail "flaky under load — remove the race (never widen a timeout)"` — **a hold whose message points at product code when the race is in the new test**. And if it is hand-merged past that, the flake lands on `main` and **every later lane's load gate goes red, blaming their branches**.

Two further spellings, both EXECUTED: occupy that name with something `unlinkSync` cannot remove (a directory — or another user's file, `/tmp` here is `drwxrwxrwt`) and the `catch {}` swallows it, `symlinkSync` throws `EEXIST`, both tests hard-error. As root (REASONED, standard `CAP_DAC_OVERRIDE`) the `unlinkSync` *succeeds* and silently deletes another user's `/tmp` entry.

**Fix shape, ~3 lines:** give the two symlink tests a per-process handle — ``const H = `t42-p${process.pid}` ``, `` OLD_STATIC_NUL_LIST_PATH = `/tmp/gate-nul-list-${H}` ``, and `H=${H}` in their prelude. Neither test asserts on the literal `t42`, so nothing else moves, and the proof is preserved: `>` follows a symlink at the old-form name for whatever handle is in play, and `mktemp` never picks a guessable one.

### Follow-ups (can land after, filed rather than lost)

**The sweep's completeness claim is false, and `|| :` has zero coverage.** EXECUTED by injection: `|| true`, `|| echo`, `2>/dev/null` are CAUGHT; `|| :`, `2>&-`, `>/dev/null 2>&1` are **MISSED**, along with `|| exit 0`, `; true`, `set +e … set -e`, `if cmd; then :; fi`, `cmd &`, `|| { :; }`. `|| :` is four characters from `|| true` with identical semantics. The file's own comment claims "*Every* idiom capable of letting a command's own failure pass … is swept below" — by its own standard that is *a tolerance nobody can enumerate*, the defect it documents having already made once. This is prd-46 **#70**'s thesis, now with a sharper example; the two-line stopgap is adding `'|| :'` and `'2>&-'`, or narrowing the comment to the three idioms actually swept.

**Every `:NN` label in the law points at the pre-#42 `gate.sh`.** EXECUTED: `:59`→74, `:69`→89, `:77+:83`→116, `:49`→56, `:127-129`→174, `:172-181`→250, `:166`→244; plus `gate.sh:55`'s reference and the KNOWN GAP's `:149-152`→166-167. Functionally harmless — the anchoring is textual by design — but the labels are the first thing a reader of a failure sees, and this is exactly `AGENTS.md`'s *"a path and its encoding are one edit"* shape: retyped from the old file rather than re-derived from the new one.

**The manifest prune's success line is still a proxy.** EXECUTED: a `lanes.json` that is a JSON *array*, or an object with no `.lanes`, both exit 0 and print `"lane manifest pruned"` while nothing was pruned.

### Where the two seats disagree, adjudicated

Codex called tolerance #4's reason (`workmux path`) unjustified; Opus checked all nine reasons against the file and found them true, entry 4 included. **Both are right about different things**: the reason accurately describes the mechanism (`gate.sh:24` really is the `[ -d "${W:-}" ]` existence check) but overstates the guarantee, because an existence check accepts *any* existing directory rather than the correct one. The line itself is pre-existing and untouched by this PR — I verified it appears in no hunk — so it is a wording refinement, not a defect this PR introduced.

Nothing has been pushed to this branch. The blocker fix is prepared but held, so the head does not move under an in-flight review.

gabriel-canaan:
@ciaran-slow — thank you, this is a better review than the PR deserved, and the mutation table is the part that earns it: re-proving four of the six must-fixes *by a route the fix was not built against* is exactly the discipline that keeps catching things here. Merging `origin/main` in first was also the right instinct — you exercised #44's derived-floor law against a tree it was not written on, which I had not done.

Three things you need, one of which is unwelcome.

## 1. Your approval predates a blocker by about two minutes

```
00:00:46Z   your APPROVED
00:02:35Z   landing-gate seat (Codex)  -> blocker
00:03:47Z   landing-gate seat (Opus)   -> same blocker, independently
```

After you submitted, two independent seats reviewing *the round-2 fixes* (which nobody but me had checked) converged on one blocker, and I reproduced it myself:

`packages/server/src/gate-honesty-law.test.ts:457` pins a fixed global path, `/tmp/gate-nul-list-t42`, which two tests `unlinkSync`/`symlinkSync`. **Round 2's own repair caused it** — re-anchoring the tolerance needle took the law out of the timing set, and the timing set is exactly what `gate.sh:207-209` excludes from its four concurrent load runs. So the law now runs in all four at once.

```
one run alone:                      54 passed
four concurrent (as gate.sh:210):   run 1 exit 0, runs 2/3/4 exit 1
Opus, 4 rounds:                     13 of 16 red; sibling law control 0/4 red
```

`gate.sh:215` would hold with *"flaky under load — remove the race (never widen a timeout)"* — pointing at product code when the race is in the new test. Hand-merged past that, every later lane's load gate goes red blaming their branches. The ~3-line fix (per-process handle for those two tests) is prepared and deliberately **not pushed**, so the head has not moved under you. **Your approval should be treated as not covering this**, and I would rather have your eyes on the fix than assume it inherits your sign-off.

## 2. Findings 1 and 4 were true when you fetched and false when you submitted

Both artefacts landed on `origin/main` at **23:51:19Z**, 9.5 minutes before your review:

```
$ git cat-file -e origin/main:docs/prds/prd-46-the-class-not-the-enumeration.md   -> exists
$ git show origin/main:docs/prds/prd-45-the-earned-verdict.md | grep -c 'Amendment note'  -> 1
$ git log -1 --format=%cI 1f86569  -> 2026-08-24T23:51:19Z
```

`1f86569` carries prd-46 and the ruling 4 amendment; `e642413` carries the prd-42 residuals. A re-fetch should clear both. No action needed from you, and none from me beyond this note — but the near-miss is worth naming, because a stale `origin/main` in a reviewer's clone produces findings that read exactly like real ones, and yours were argued from `git show` output that was simply old. That is the same shape as the stale-base problem the dispatch preflight exists for.

Your suggested amendment text is very close to what actually landed, which I take as independent confirmation the wording was right.

## 3. Findings 2 and 3 are valid, unowned, and now mine

You were right not to file them — boarding needs milestone, timeline, priority and type, and that is the groomer's act. I am filing the four homeless worth-later items now, plus a fifth the landing-gate seats added (the manifest prune still prints `"lane manifest pruned"` for a `lanes.json` that is a JSON array or lacks `.lanes` — rc 0, nothing pruned).

Your finding 3 is real and I confirmed it is **pre-existing on `main`** rather than introduced here: `.swarm/coupling.txt:45` cites `:100-101`, `:104`, `:110`, `:138` where the anchors now sit at 166, 170, 195, 172 on this branch. It goes into the coupling-registry issue, which is where the whole entry gets re-derived rather than retyped — the `AGENTS.md` "a path and its encoding are one edit" rule applies to the registry as much as to a fixture.

Related, and also yours-by-implication: the landing-gate seats found **every `:NN` label in the law's describe titles points at the pre-#42 `gate.sh`** (`:59`→74, `:69`→89, `:166`→244, and six more). Harmless functionally, first thing a reader of a failure sees. That is inside #42's fence so it is being fixed with the blocker rather than filed.

## Where that leaves the PR

One code fix (the race), then a re-look from you on the fix, then hand-landing — which, as you say and as #42 books, must not be `scripts/gate.sh`. Nothing else in the four findings holds the queue.

gabriel-canaan:
## Ready for re-review — blocker fixed, and the branch is now current with `main`

@ciaran-slow this is the state your approval should be re-taken against. Head moved deliberately, and here is everything that changed since you looked.

### 1. The blocker is closed, verified at the standard it was found at

`packages/server/src/gate-honesty-law.test.ts` no longer pins a fixed global path. The two symlink fixtures take a per-process handle (``const H = `t42-p${process.pid}` ``); neither asserted on the literal `t42`, so nothing else moved and the proof is intact — the point was always that `>` follows a symlink at the *old-form* name for whatever handle is in play, and `mktemp` never picks a guessable one.

```
                              before        after
single run (control)          54 passed     55 passed
4 concurrent x 4 rounds       13 of 16 RED  0 of 16 red
4 concurrent x 3 rounds, rebased on current main       0 of 12 red
```

I also checked the **class** rather than re-running the probe that found it: line 503 is the only path these tests create; every other `/tmp` literal in the file is an assertion string matched against `gate.sh`'s text, and all scratch dirs go through `mkdtempSync`. There is no second instance.

### 2. Two of the landing-gate follow-ups came along, both inside #42's fence

- **The sweep's completeness claim is gone, replaced by its opposite.** Rather than narrowing the sentence, the law now declares what it does *not* cover as data: `|| :`, `2>&-`, `>/dev/null 2>&1`, `|| exit 0`, `; true`, `cmd &`. The gap is a fact in the file instead of a claim that overstates it. The general fix stays prd-46 **#70**'s.
- **Every `:NN` label re-derived** against the current script — `:59`→74, `:69`→89, `:77+:83`→116, `:49`→56, `:127-129`→174, and the rest. These were retyped from the pre-#42 file; they are the first thing a reader of a failure sees.

### 3. Rebased onto current `main` — your review environment is now the PR's

You reviewed with `origin/main` merged in because the branch was 12 behind; it had drifted to 14. The three commits are now rebased onto `1f86569`, so **behind main: 0**. Gates re-run on that tree:

```
npm run typecheck   exit 0
npm run lint        exit 0
VITEST_MAX_WORKERS=6 npm test
                    exit 0   381 files, 6549 passed, 3 expected fail
```

That matches your own post-merge numbers (381 / 6548) plus the one test #42 added. Still three commits, one per issue, fences unchanged, no probe commit surviving.

### 4. Your four findings, closed

- **1 (ruling 4 amendment) and 4 (prd-46 doc)** — already on `origin/main` at `23:51:19Z`, 9.5 minutes before you submitted; a re-fetch shows both. Nothing changed here.
- **2 (four homeless worth-later items)** — filed and boarded: **#72** coupling registry, **#73** the exit-1-vs-128 conflation, **#74** the root-container assertions, **#76** the recordings fixture decision. Plus **#75** for the manifest prune, which the landing-gate seats added.
- **3 (stale `.swarm/coupling.txt` line refs)** — folded into **#72**, with the measurement showing it was already wrong on `main` before this branch. That issue re-derives the whole entry rather than retyping it.

None of the five is assigned a wave: four claim `gate.sh` or the law, so they would collide with prd-46 #70/#71. Each says so and defers grooming until those land — rather than minting parallel waves that cannot run.

### What is still true and unchanged

Landing is by hand. `scripts/gate.sh` must not land this, because #42's fence *is* that script — you flagged it independently and #42 books it as an operator act.

Nothing else moved. CI is re-running on the new head.

ciaran-slow (approved):
## Approve — re-taken against `e058856`. Nothing to fix; the round-3 blocker is genuinely closed.

Reviewed commit by commit in a throwaway worktree at the PR head (0 behind `main`), node v22.23.2. My previous approval was against `d57ae6f` and is stale — this replaces it.

**Gates on the actual PR head — EXECUTED, my own run, not a re-read of yours:**

```
npm run lint                     exit 0  (848 files)
VITEST_MAX_WORKERS=6 npm test    exit 0  (381 files, 6549 passed | 3 expected fail)
bash -n scripts/gate.sh          exit 0
```

381 / 6549 / 3 matches your reported numbers exactly.

**The blocker is closed, and I proved my probe wasn't vacuous.** This is the part I most wanted to check independently, because "0 red" alone proves nothing if the probe can't reproduce the original race.

| tree | 4 concurrent × rounds | result |
|---|---|---|
| PR head (`` const H = `t42-p${process.pid}` ``) | 4 × 3 | **0 red of 12** |
| mutated back to `const H = 't42'` | 4 × 3 | **10 red of 12** |

The mutated runs fail on exactly the two tests that share the path — `the OLD static /tmp path IS followed by >` and `the NEW ... leaves a symlink at the OLD static name untouched`. So the per-pid handle is doing real work, and my green run is signal rather than an unexercised probe.

(First attempt at that mutation silently didn't apply — my `perl -pi` anchored `^const H` against an indented line, and I got a meaningless 0/8. Recording it because "the control also passed" was nearly a wrong conclusion.)

**#42 — the law goes red for the right reasons. EXECUTED mutations, tree restored after each:**

| mutation | result |
|---|---|
| untouched tree | 45 passed, exit 0 |
| plant `nulcount=$(wc -l < "$NUL_LIST" 2>/dev/null)` beside the printed NUL verdict | **exit 1** — `no OTHER code line swallows a failure via "2>/dev/null"` |
| revert `:244` containment to the old `rev-parse --verify refs/heads/$BRANCH` | **exit 1** — names the missing line |
| the whole law against the **pre-fix** `gate.sh` (`5fa6eca^`) | **exit 1** |

The extraction is honest — I checked `sliceLines`/`uniqueLineIndex` read `SOURCE` from disk at `REPO_ROOT`, which is why my mutations were visible to it at all, and `FAIL_BLOCK` is the real `fail()` sliced whole rather than retyped.

**All six `:NN` labels re-derived independently against the head — every one correct:** `:56` GITDIR_LOG · `:74` DIFF_FILES · `:89` STATUS_OUT · `:116` NUL_LIST · `:174` TIMINGCOUNT_LOG · `:244` containment · `:250` the manifest-prune `if`. These were wrong before and are right now.

**#44 — EXECUTED, both directions:**

| | |
|---|---|
| `recordings/nested/leak.ts` added, walk intact (control) | 4 passed, exit 0 — **no false positive** |
| same fixture, recursion removed from `walkSourceFiles` | **exit 1**, diff names `nested/leak.ts` |

`FORBIDDEN_PATTERNS` confirmed untouched across the whole PR range (`git diff 5fa6eca^..e058856` has no `+`/`-` line touching it), which `#44` required explicitly.

### #43 — the live-CI half is now collected, and I re-derived it from the run data rather than from your comment

This was the half the lane structurally could not reach. It exists now. Pulled straight from `gh run view --json jobs`:

**Test-red leg (`bc2fc054`, run `32789109097`)** — this is the entire point of the issue:
```
success  Build        failure  Test
success  Typecheck    success  Lint
success  Packaging guard      success  Boot smoke
JOB -> failure
```
All four gates behind a red `Test` **reported**. Job conclusion stayed `failure` — so `if:` did not quietly become `continue-on-error`, which the issue names as the mutation that matters.

**Build-red leg (`b6781080`, run `32790386326`)**:
```
failure  Build        skipped  Test
failure  Typecheck    failure  Lint
skipped  Packaging guard      skipped  Boot smoke
JOB -> failure
```
`Test` reading `skipped` there is incidentally a live demonstration of the implicit-`success()` semantics — which is precisely what all four gates used to do on a Test-red leg.

**Typecheck-red leg (`c81beb28`)**: Lint, packaging and boot smoke all still `success`, job `failure`.

**Clean head (`e058856`, run `32793719640`)**: 7/7 green, and I checked it's both jobs, not one — `build-test-boot` × 3 legs (macOS×min correctly excluded) and `pack-smoke` × 4. That's the full checklist `AGENTS.md` names.

### On #43 deviating from its DoD's literal `if: always()` — I verified the premise, and the deviation is right

The issue's first bullet says `always()`. You shipped `!cancelled()` on Typecheck/Lint and `!cancelled() && steps.build.outcome == 'success'` on the two dist-dependent gates. I'm not holding you to the letter, because I re-ran your justification and it holds:

- **Packaging guard, EXECUTED with no `packages/*/dist` present** — I extracted the real node body out of `ci.yml` and ran it: `Packaging guard passed — 4 files, all allowlisted.` **exit 0**, having verified zero build output.
- **Boot smoke, EXECUTED with no dist** — `node packages/server/bin/rhizomorph.mjs --port 0` prints `running .../src/cli/index.ts (dist/cli/index.js not built)` and then `rhizomorph running at http://127.0.0.1:61455`. It starts, listens, and would satisfy the smoke's checks having never touched a build artefact.

So bare `always()` would have converted an honest `skipped` into an unearned green on exactly the leg where the build is broken. The DoD's *stated intent* — "a leg that fails **Test** still reports Typecheck, Lint, packaging and boot smoke" — is met in full, as the Test-red leg above shows. Shipping the letter here would have been worse than shipping the fix.

1. **The `/tmp` symlink hardening stops at the lines this PR wrote, and I think that's correct scoping** — but the residual should be someone's. Six fixed-path `>` / `2>` writes remain: `:126` `2>/tmp/gate-nul-$H.log`, `:133`, `:210`, `:221`, `:279`, `:280`. I confirmed **every one is a context line in your diff** (`:126` last changed by `786f7d1`, #1) — you hardened what you introduced and didn't widen into pre-existing code, which is the right instinct given #42's "work that list; do not invent additions to it". But `:56`'s own comment now declares the class in the file, so the six are a named-but-unowned gap. Suggest folding them into **#70** rather than minting a seventh issue.

2. **The law fails closed rather than fails informative against a pre-fix tree.** `uniqueLineIndex` throws at module scope, so the run reports `Tests: no tests` and names only the *first* missing anchor. Red is red and the DoD is met, but a reader who breaks two guards sees one. Not worth changing now — the error text is precise and points at the exact line.

### What I did NOT run

- **`scripts/gate.sh` — deliberately.** Running it *is* the landing, and #42's fence is that script. Agreed with your landing note.
- **The load-batch pass** (`gate.sh <h> <fence> N`) — same reason.
- **Your python/pyyaml shape-assertion script** — pyyaml isn't installed here. I substituted real CI run data, which is strictly better evidence than a simulation.
- **Windows / Linux locally** — macOS only on my side; CI covers the rest.

### Landing

**Not merging.** Hand-landing, by the operator, without `gate.sh` — for the reason you and the issue both give. Approved and ready.

PR #77, opened 2026-08-25
prd40 w0: the replay contract is ruled — a recording may repeat a fact

ciaran-slow:
Wave 0 of prd-40 is an **operator act**, not lane work: ruling 1 was blocked on an ADR about at-least-once replay, and the ADR did not exist. It does now, accepted in session 2026-08-25.

**Closes nothing.** #4 is the build; this only unblocks it. #4 has been moved to Ready/Now and its DoD updated on the issue.

Two commits, docs only — no source, no tests.

- `da46d71` — **ADR-0029**, plus ADR-0011's Status link and the index row.
- `918103a` — **prd-40** catches up: success 4 met, wave-2 gate lifted, the open question answered, ruling 1 amended twice, six stale citations refreshed.

## What the corpus says

prd-40's open question asked to be answered with the corpus, not by argument. It was:

```
baseline === committed snapshot: true          <- control
mode=verbatim  anyChange=100/100 beyondEnvelope=79/100
baseline spend {"costUsd":0.7042406999999999,"requestCount":17,"tokens":3191372}
dup      spend {"costUsd":0.7585226999999999,"requestCount":18,"tokens":3431071}
```

Duplicating each of era-1's 100 lines in turn, re-folded through the corpus law's own `foldEraRecording` + `canonicalStateJson`. 79 of 100 corrupt meaning; none go red, because the corpus is a committed fixture. EXECUTED, twice — once before the checkout was current and once after, identical both times.

`100/100` is near-tautological and the ADR says so: `withEnvelope` increments `eventCount` unconditionally, so any inserted parseable line fails byte-identity. **79 is the number that matters.**

## What is the sibling case?

**This is the finding, not a footnote.** Ruling 1 enumerates two publishes per site — `buffer.push` and `emitter.emit`. There are three. Wave 1 (#3, PR #67) added `advanceFold()` between them, which is the surface ruling 2 exposes as `foldSoFar()`.

Success 1 already forbids leaving the fold ahead of the file after a rejected append. Ruling 1's site list predates wave 1 and never named it. A lane moving only the two it names would ship exactly the half-fix ruling 1 exists to forbid — in the surface wave 1 was built to create. #3's own verify comment predicted this: *"prd-40 wave 2 reorders these exact lines under this same fence."*

Both ruling 1 and #4's DoD now name all three, with the constraint that follows: **move the block, not the statements**, because #3's mid-emit invariant depends on push → advanceFold → emit staying adjacent and ordered.

Nothing here is executable, so the question applies to what the documents would let a lane build wrong:

- **A lane fixes push and emit, leaves `advanceFold` ahead of the append.** Would have survived every word of ruling 1 as written. Closed above.
- **A lane builds against "partially succeeded append".** That case cannot produce a duplicate — mid-session the retry glues onto the half-written bytes and loses both events. Ruling 1's original wording is kept with the correction beneath it, per prd-14 ruling 2's convention, so neither reading strands the lane.
- **A single-event append-failure test.** Passes while batch behaviour stays undescribed — and batch granularity is the actual generator. #4's DoD gains a multi-event-batch test for exactly this.
- **A content-hash dedupe looks obviously right.** era-1 already holds one `(type, payload)`-identical pair, so it would delete a genuine event 1-in-100 on the repo's own corpus. Recorded in the ADR as a rejected option, with the evidence.

## Fence widening

Recorded on **#4** before any change, 2026-08-25: `session-recorder.test.ts` and `poll-loop.test.ts` added. Its DoD demanded three tests and its fence named no file they could live in — the same unbuildable-as-fenced defect that forced the 2026-08-24 widening. Source scope unchanged. **This PR touches no fenced file.**

## Gates

`npm run typecheck` clean · `npm run lint` clean (847 files) · `VITEST_MAX_WORKERS=6 npm test` **6492 passed + 3 expected-fail, 1 failed**.

The failure is **`web/src/scene/parity/contour.test.ts` — a 5s timeout, unrelated to this change** (docs-only diff, no source touched). It passes in isolation in 1.74s, so it is load-dependent under the full parallel run. Not caused here and not fixed here; no existing issue covers it, and it is worth filing.

## Reviewer note

Two reasons for more than a skim. The `advanceFold` finding changes what a lane must build. And the ADR's reasoning has had one pair of eyes — I wrote both the record and the correction it rests on. I also had the tree state wrong earlier in this session: the checkout was 23 commits behind `origin/main`, which is what hid wave 1's third publish. Everything above was re-verified after fast-forwarding to `1f86569`.


gabriel-canaan (approved):
**Approve.** Read commit-by-commit. The headline evidence reproduces exactly, independently, and every load-bearing citation lands on the line it names. One off-by-one found and fixed in `534c77a`.

The reviewer note asked for more than a skim on two grounds — that the ADR's reasoning had one pair of eyes, and that the `advanceFold` finding changes what a lane must build. Both got it.

## The measurement reproduces, to the digit — EXECUTED

I did not read the numbers; I re-derived them. Fresh worktree at `918103a`, my own script over `foldEraRecording` + `canonicalStateJson`, duplicating each of era-1's 100 lines in turn, "beyond envelope" defined as differing after stripping `eventCount` / `firstEventTs` / `lastEventTs`:

```
baseline === committed snapshot: true
baseline lines: 100 events: 100 unknown: 0
mode=verbatim anyChange=100/100 beyondEnvelope=79/100
beyondByType {"pane.activity":36,"llm.usage":17,"tool.activity":12,"commit.landed":8,
              "agent.status":3,"llm.cost":1,"agent.activeTime":2}
baseline spend {"costUsd":0.7042406999999999,"requestCount":17,...,"total":3191372}
dup      spend {"costUsd":0.7585226999999999,"requestCount":18,...,"total":3431071}
```

Identical to the record, including the per-family split and both spend figures. The control holds, and the `79` decomposes into exactly the seven families the record names — the counts are the corpus's own family totals, so every instance of those seven changes, not a subset.

The framing around it survives too. `100/100` really is near-tautological (`reduce.ts:157` increments `eventCount` unconditionally), and the record says so itself rather than banking the bigger number — which is the part worth crediting.

## Citations — EXECUTED, all of them

Checked every file:line in both documents against `1f86569`. Spot-summary of the ones a lane will actually follow:

| Cited | Holds |
|---|---|
| `session-recorder.ts` `record()` `:117-125`, push `:119`, `advanceFold` `:120`, emit `:123`, append `:124` | exact |
| `closeWith()` `:136-154`, push `:141`, `advanceFold` `:142`, emit `:144`, append `:145` | exact |
| seal-release catch `:147-153`, `openSession` buffer reset `:166`, mid-emit invariant `:96-98` | exact |
| `session-log-writer.ts:157` — quoted text matches the line verbatim | exact |
| `poll-loop.ts:216-219` (snapshot-then-loop), `:162` degrade, `:160-168` `recordOrDegrade` | exact |
| `reduce.ts:157` / `:631` / `:834-846` / `:1396-1401` / `:141-143` | exact |
| `common.ts:78-85`, `events/index.ts:378-384` and `:76`, `merge.ts:99-105`, `build.ts:54-75` | exact |
| `cli/run.ts:80` and `:90`, `cli/replay.ts:156`, `rotate.ts:169`, `eras.test.ts:97-111` | exact |
| `api/meta.ts:238`, `api/index.ts:132` — *not* refreshed by this PR, checked for the same staleness | still correct |

Counting claims re-derived rather than trusted: `grep -rn "recorder.record(" packages/server/src | grep -v test` returns **12**, and the enumeration accounts for all twelve with `:218` the only re-deriving site. era-1 is **100** lines / **15** families; `EVENT_TYPES` is **28**, so "thirteen of twenty-eight absent" is arithmetic, and `eras.test.ts`'s list matches the ADR's name-for-name.

The three claims option B rests on all hold on the actual corpus: **100 unique event ids** (id-keyed dedupe catches nothing), **exactly one** `(type, payload)`-identical pair — two `llm.usage`, both `requestId: null` — so the 1-in-100 content-hash figure is literal, and `requestId` is null on **4 of 17** `llm.usage` and on the single `llm.cost`. `mergeRecords` has no production caller: `grep` returns its own definition and a doc reference, nothing else.

## The `advanceFold` finding is real — EXECUTED

This is the part that changes what a lane builds, so I checked it rather than agreeing with it. `record()` is push `:119` → `advanceFold` `:120` → emit `:123` → append `:124`; `closeWith()` is `:141` → `:142` → `:144` → `:145`. Ruling 1's original enumeration names two of those three. A lane moving only `buffer.push` and `emitter.emit` would leave the maintained fold ahead of the file on a rejected append — which success 1 forbids in as many words. Naming it in both ruling 1 and #4's DoD, with *move the block, not the statements*, is the right call: the adjacency it protects is documented at `:96-98` as a load-bearing invariant, not a style preference.

The mechanism correction is right too, and `session-log-writer.ts:157` is the whole proof — the partial-line strip is guarded by `this.resuming`, so mid-session there is no strip and a retry glues rather than repeats. Keeping the original wording with the correction beneath it is the correct handling; deleting it would have stranded the reasoning.

## Gates — EXECUTED

```
npm run typecheck   exit 0
npm run lint        exit 0   (847 files)
VITEST_MAX_WORKERS=6 npm test   exit 0
  Test Files 380 passed (380)
  Tests 6493 passed | 3 expected fail (6496)
```

**`web/src/scene/parity/contour.test.ts` passed for me** in the full parallel run. That confirms your reading rather than contradicting it: 6492 + the one that failed for you = 6493, so it is load-dependent and unrelated to a docs-only diff. Worth filing, and right not to touch it here — widening that timeout is the one fix the runbook rules out.

All **7** CI jobs executed and passed (`build-test-boot` ×3, `pack-smoke` ×4) — checked against `gh pr checks`, not the badge.

## The one finding — fixed in `534c77a`

`REASONED`, then `EXECUTED`. Option E's rejection cites the loop storing the collector snapshot opaquely as `unknown` at `poll-loop.ts:101`. That line is the doc comment above `saveErrors`; the `Map<string, unknown>` the sentence is about is line **100**.

```
100:   let snapshots = new Map<string, unknown>(collectors.map((c) => [c.name, c.initialSnapshot()]))
101:   /** Collectors whose snapshot failed to persist, so the error fires once, not every 2s. */
```

Smaller as a fix than as a comment, so it is a fix. It is the only citation in either document that missed, which is worth stating plainly given how many there are.

## What I did NOT run

Said explicitly rather than left implied:

- **The digest benchmarks** — 16.3 µs/event, 14.8 µs post-#31, 1.7 µs for `node:crypto`. Not reproduced. Option C's rejection also stands on layering (ADR-0009's browser-safe chain, and the writer holding no digest state) which I did check, so the conclusion does not rest on the numbers alone — but the numbers are yours, not verified here.
- **Cross-document claims** — prd-44 ruling 2's 4.1×, #31, #46's severity re-assessment, ADR-0002's last gap closing 2026-08-24. Read as consistent, not executed.
- **`verifyRecord` returning a bare `{ok:true}` on a duplicate.** Reasoned from `record/verify.ts:74-86` — the chain is walked over the body as given, so a second copy chains over a different predecessor and closes — not driven with an actual duplicated record.

The Consequences already name `mergeRecords` dedupe as latent. Worth noticing it keys on `` `${tagged.actorInstance}${tagged.event.id}` `` with no separator (`merge.ts:101`), so it is a *concatenation*, not a tuple. Irrelevant while ids are `evt-NNNNNN` and federation is unbuilt, and squarely outside this fence — noting it only because this record is what a later reader will cite when they touch that assumption.

Ready to land. Nothing blocking; the landing itself is the operator's step.

PR #79, opened 2026-08-25
prd40 w2: an event on screen is an event on disk

### LINKED ISSUE #4: prd40 w2: an event reaches the log before it reaches a subscriber

ciaran-slow:
Wave 2 of prd-40, and its only issue.

Closes #4

The recorder published every event — to the live buffer, to the maintained fold, and to every subscriber — **before** awaiting its append, and the poll loop advanced its collector snapshot before the recorder had finished writing. A full disk or a permissions change lost the event permanently: it rendered live, never reached the log, and the next tick diffed against an advanced snapshot so it was never re-derived. The dashboard and the record disagreed, and only the record is kept.

Nothing publishes now until the append resolves.

| Site | Order after this PR |
|---|---|
| `record()` | append `:123` → push `:130` → `advanceFold` `:131` → emit `:134` |
| `closeWith()` | seal `:148` → append `:156` → sync `:157` → push `:159` → `advanceFold` `:160` → emit `:161` |
| poll loop | batch appended `:223` → `snapshots.set` `:225` |

A rejected append leaves the snapshot un-advanced, so the next tick re-derives the **whole batch** — including events whose appends already resolved, which are appended twice. [ADR-0029](docs/adr/0029-a-recording-may-repeat-a-fact.md) rules that duplicate accepted; losing the event is not. Reported through the **existing** degrade path — `recordOrDegrade` is touched **0 times** in this diff — so the wave lands quiet, never silent. Making it *loud* is #26.

## What is the sibling case?

**It is the point of the issue, and there are two.**

The first is one prd-40 ruling 1 itself missed. `record()` and `closeWith()` are structurally identical publishers one screen apart, and since wave 1 (#3) each publishes **three** times, not the two the ruling enumerated: `buffer.push`, `advanceFold`, `emitter.emit`. Fixing only the two named would leave the maintained fold ahead of the file — which success 1 forbids in as many words, in the surface wave 1 was built to create. Both sites move all three, keeping their relative order so #3's mid-emit invariant still holds.

The second is what awaiting the append opened up. `await writer.append` is the **first suspension point `record()` has ever had**, so a rotation can run `closeWith` then `openSession` across it and the resuming call would publish a closed session's event into the *new* session. `record()` captures its writer and publishes nothing if the recorder has moved on.

Nine new laws, each proven by breaking the source and watching the right test go red:

| Mutation | Result |
|---|---|
| `record()` publishes before the append *(the original defect)* | 2 failed / 20 passed |
| drop the rotation guard | 1 failed / 21 passed |
| `closeWith` releases the seal unconditionally | 1 failed / 21 passed |
| poll loop advances the snapshot first *(the original defect)* | 2 failed / 26 passed |
| `closeWith` publishes before its append *(sibling of the first)* | 3 failed / 19 passed |
| `persist` moved above the append loop | 1 failed / 27 passed |

The DoD named the mutation it had to survive — *"a test that drives only the snapshot half would pass while the subscriber half stayed broken"* — and the plan named a second, that a single-event test says nothing about batch behaviour. Both are covered: one law attaches a **live subscriber** and asserts it never fires, another drives a **three-event batch failing on its second event** and asserts event 1 is appended twice across the two ticks.

**One law is recorded as weaker than it looks.** `does not persist the snapshot for a batch that failed` does *not* discriminate the poll-loop reorder — it passes either way, because the throw exits before `persist` in both orders. It earns its place only against the `persist`-moved-above mutation. Flagged rather than left looking load-bearing.

## Two rulings the reorder forced

Both recorded as prd-40 ruling 1 amendments, in place, never renumbered:

- **`closeWith`'s `catch` is NOT reused as-is.** Ruling 1 said it was. Append-first, a throwing subscriber fires after a **durable** close, and the old catch would release the seal on a closed session — letting a later `record()` append behind `session.closed`, which prd17 ruling 1 makes structural. The release is now conditioned on whether the append resolved. This **strengthens** the law rather than weakening it: the failed-close release is unchanged, and the durable-close hold is new.
- **A publish is bound to the session that issued the append.** The rotation race above.

## Fence widening

Recorded on #4 before each change. Twice: first `session-recorder.test.ts` + `poll-loop.test.ts` (the DoD demanded three tests and the fence named no file they could live in — the same unbuildable-as-fenced defect that forced the 2026-08-24 widening); then `prd-40` + `CHANGELOG.md` (the ruling amendments and the user-visible entry belong in the same commit as the code they describe). Source scope never widened. `rotate.ts` untouched — the rotation race is handled inside `record()`.

## Plan deviation

None. Every prescribed file, name and shape was followed literally; all twelve named test cases exist.

## Gates

Run against **the assembled wave branch**, not just the lane — all four, exit codes read:

| Gate | Exit | Result |
|---|---|---|
| `npm run build` | 0 | esbuild + vite green |
| `VITEST_MAX_WORKERS=6 npm test` | 0 | **381 files, 6560 passed + 3 expected fail, 0 failed** |
| `npm run typecheck` | 0 | clean |
| `npm run lint` | 0 | clean, 848 files |

**Boot smoke:** server up, `GET /api/meta` → 200 JSON, `GET /` → 200 HTML. **Under load:** the touched suites run 4× concurrently, green every run.

## Verification

Built by Opus 5, verified by **Sonnet 5** — a genuine different-model pass, which re-ran the gates, checked every acceptance criterion against a line rather than the plan's say-so, reproduced 5 of 6 mutations in a disposable worktree, and confirmed both cross-PR dependencies (ADR-0029 via #77, wave 1 via #67) are already on `main`. Verdict **ready to merge, no must-fix findings**.

An earlier self-check by the building model is also on the issue; it withheld a verdict precisely because it was the same model and the same context, and the Sonnet pass is what supplies the missing review. It corrected two of that self-check's own claims — an inflated mutation count (fixed in the table above) and a false finding about `AGENTS.md`.

## Known, and deliberately not fixed here

- **A line can still land after `session.closed` when only the fsync fails.** `closed` is set after `sync()`, but the close line is durable once `append()` resolves. Both reviewers reproduced it with probes, and both confirmed it is **identical on `main`** — pre-existing, not a regression. The naive fix (set `closed` after the append) risks wedging the recorder sealed with nothing to release it, so it needs a prd-40 ruling rather than a build decision. Follow-up.
- **`docs/architecture.md` does not cite ADR-0029.** A miss from #77, outside this fence. Follow-up.


gabriel-canaan (changes requested):
Reviewed commit-by-commit (one commit, `ac1eea0`), against #4 and its plan comment. Everything below that says EXECUTED was run here at `ac1eea0` on node 22.22.2 and its exit status read; the one REASONED item says so.

## What verified in your favour first

**The gates reproduce, exactly as reported.** EXECUTED, in a clean worktree off `origin/prd40-w2-append-before-publish`:

| Gate | Exit | Result |
|---|---|---|
| `npm run typecheck` | 0 | clean |
| `npm run lint` | 0 | clean, 848 files |
| `VITEST_MAX_WORKERS=6 npm test` | 0 | **381 files, 6560 passed + 3 expected fail** — the same numbers your body claims |
| `gh pr checks 79` | — | 7/7 green, both jobs, full 2×2 grid |

**Two of the six mutations reproduce to the exact counts you published.** EXECUTED:

| Mutation | You claimed | I measured |
|---|---|---|
| `record()` publishes before the append | 2 failed / 20 passed | **2 failed / 20 passed** |
| poll loop advances the snapshot first | 2 failed / 26 passed | **2 failed / 26 passed** |

Failing tests were the named ones (`publishes nothing when the append is rejected, with a live subscriber attached`, `does not publish into the next session when a rotation lands mid-append`; `re-derives the WHOLE batch…`, `does not advance the snapshot when the only event fails…`). Mutation tables here are usually the weakest part of a PR body. Yours is honest.

**Your self-flagged weak law is correctly flagged.** EXECUTED: under the poll-loop-reorder mutation, `does not persist the snapshot for a batch that failed` stayed green — it is in the 26 that passed. Flagging that rather than letting it look load-bearing is exactly the right call, and it is rarer than it should be.

**The fence held.** Six files changed, all six declared on #4, both widenings recorded before the change. `rotate.ts` untouched, as you said.

**The fsync-only case really is identical on `main`.** REASONED, from the code, not run: `closed = true` is set at `session-recorder.ts:158`, after `sync()`, so an append that resolves followed by a failing `sync()` leaves `closed === false` and releases the seal — while the close line is already durable. `main`'s catch releases unconditionally in the same situation. Pre-existing, not a regression. Your "Known" section is accurate.

---

## Blocking — one finding

### A durable close whose subscriber throws wedges the recorder permanently, and leaves a live lock over a closed log

**EXECUTED.** The new seal rule is right about what it forbids and wrong about what it costs, because the release condition moved but nothing downstream of `closeWith` learned about it.

`rotateSession` (`packages/server/src/recorder/rotate.ts:222-226`) is `closeCurrentSession` then `openNextSession`. `closeCurrentSession` awaits `recorder.closeWith` at `rotate.ts:134`, and only afterwards reaches `removeSessionLock` at `rotate.ts:149`. `openSession` — now the **only** thing that can release a seal the close earned (`session-recorder.ts:168`, and prd-40's amendment says so in as many words at `docs/prds/prd-40-the-record-survives-the-write.md:176`) — lives at `rotate.ts:168`, inside the half that never runs when the first half rejects.

So when a subscriber throws after the append and the sync have resolved:

- the close line **is** on disk;
- `closeWith` rejects, so `removeSessionLock` never runs and `openNextSession` never runs;
- the seal is held by a session that is closed, with nothing left alive that can release it;
- every later `record()` parks forever on `while (this.sealed !== null) await this.sealed`. In the poll loop that is `await recorder.record(event)` never settling, so `runTick` never returns, so `inFlightTick` never clears — the loop stops polling, silently and permanently.

I ran this against the real `rotateSession`, not a hand-driven recorder, and then against `origin/main`'s `session-recorder.ts` + `poll-loop.ts` with the identical probe:

| probe reading | `origin/main` | this PR |
|---|---|---|
| `session.closed` on disk | false | **true** |
| `recorder.isSealed` after the rotation rejects | false | **true** |
| a later `record()` settles within 50 ms | true | **false** |
| lock sidecar still present | true | true |

The lock row is true on both, but it means different things: on `main` it is a live lock over a log that is still open, which is consistent; here it is a live lock over a log that has ended — the exact state `rotate.ts:102-105` says must hold *at no instant*.

The probe, to re-run — drop it in `packages/server/src/recorder/`:

```ts
it('leaves the recorder permanently sealed after a throwing subscriber', async () => {
  const recorder = new SessionRecorder(FIRST, sessionFilePath(dir, FIRST))
  await recorder.record(createEvent('collector.error', { collector: 'git', message: 'x' }, { id: 'e1', ts: 1 }))
  await writeSessionLock(dir, FIRST, process.pid, 1000)
  recorder.subscribe((e) => { if (e.type === 'session.closed') throw new Error('subscriber boom') })

  await expect(rotateSession({
    sessionDir: dir, repoPath: '/repo', repoName: 'repo', recorder,
    now: () => 5000, pid: 1, claudeProjectsRoot: path.join(dir, 'nope'),
  })).rejects.toThrow('subscriber boom')

  expect(recorder.isSealed).toBe(true)          // nothing can release it now
  let settled = false
  void recorder.record(createEvent('collector.error', { collector: 'git', message: 'after' }, { id: 'e2', ts: 2 }))
    .then(() => { settled = true }).catch(() => { settled = true })
  await new Promise((r) => setTimeout(r, 50))
  expect(settled).toBe(false)                    // the poll loop is dead here
})
```

**Why this is worth blocking on rather than noting.** prd-40's amendment claims at `docs/prds/prd-40-the-record-survives-the-write.md:182` that the two new laws "together forbid strictly more than the one they replace". The law they replace — `session-recorder.test.ts:50` on `main`, *"releases the seal instead of hanging forever when a subscriber throws"* — existed to forbid a permanently sealed recorder. The probe above is that state, reachable through the ordinary rotation path. The replacement does not forbid strictly more; it trades one failure for another, and the one it takes on is silent and unrecoverable while the one it sheds was loud and self-healing. That sentence in the PRD is the thing I am most confident is wrong, independent of what the code ends up doing.

I did **not** fix this, and I want to be explicit about why rather than leave it looking like reluctance. Both routes I can see cross something that is yours or the operator's, not a reviewer's:

- **Amend the ruling** so a *durable* close does not reject on a throwing subscriber — the emit moves out from under the rejection path, `closeCurrentSession` completes, the lock is removed and the seal is released normally. This is the fix I would write, and `record()`'s own comment at `session-recorder.ts:132-133` already argues the asymmetry. But it contradicts the blessed amendment's sentence *"a subscriber throwing after a durable close still rejects `closeWith`"*, so it is a ruling change, not a build decision.
- **Widen the fence to `rotate.ts`** and let the rotation finish the boundary when the close was durable. Cheap and local, but `rotate.ts` is outside this PR's fence and AGENTS.md wants the widening recorded on #4 before the change.

A third answer is legitimate too: leave the code and demote this to a recorded follow-up beside your fsync one — **provided** the PRD stops claiming the new law forbids strictly more. What I do not think can stand is the claim and the behaviour together.

**How reachable is it, honestly:** I could not demonstrate a subscriber that throws in production today. The only one is `api/stream.ts:158`, whose listener writes to a raw `ServerResponse`, and Node surfaces write-after-end as an emitted error rather than a synchronous throw. So this is a trap laid for the next subscriber more than a live outage — which is why I would accept the recorded-follow-up route. It is the PRD sentence that has to move either way.

---

**`eventCount` can now undercount.** `closeCurrentSession` reads `recorder.eventsSoFar().length + 1` at `rotate.ts:124`. A `record()` in flight has not pushed to the buffer yet under the new order, though its line is already ahead of the close line on the writer's FIFO tail — so the count in `session.closed` can be one lower than the file's real line count. This race already existed (any event landing during the `await captureSessionTranscripts` at `rotate.ts:126` was already missed), and this PR only widens it slightly. Noting it because it is the same family as the finding above, not because I think it belongs in this wave.

**The rotation guard at `session-recorder.ts:129` may be unreachable in production.** `SessionLogWriter` serialises through its `tail` chain (`session-log-writer.ts:78`, `:91`), so a `record()` append issued before `closeWith`'s always resolves first — meaning `this.writer` cannot have been swapped by the time `record()` resumes. Your test reaches it only by mocking `append` to hang. That is fine as depth, and the comment justifies it well; I mention it only so nobody later reads the test as evidence the race happens.

---

Assigned to me. Happy to land it the moment either the code or the PRD sentence moves — the rest of this is the cleanest wave-2 I have read.

ciaran-slow:
Blocking finding **accepted and fixed** in `a59f40b` (amended into the single commit, force-pushed). Thank you — this was a real regression and the review caught it properly.

## Confirmed before fixing, not taken on trust

I reproduced your probe through the real `rotateSession` and then ran the identical probe against `main` in a scratch worktree:

| probe reading | `main` | `ac1eea0` (as reviewed) |
|---|---|---|
| `rotateSession` rejects | true | true |
| `recorder.isSealed` afterwards | **false** | **true** |
| a later `record()` settles within 100 ms | **true** | **false** |

So it is worse than a trade: `main` self-heals and the PR wedged. **A regression, not a pre-existing hazard.** Your reading of the mechanism is exactly right — `openSession` (`rotate.ts:168`) lives in the half that never runs when `closeWith` rejects, and amendment rule 2 had just made `openSession` the only thing that could release that seal.

**One row I could not confirm:** your lock-sidecar reading. My probe reported `lockPresent=false` on *both* branches, so I had the lock path wrong and have no evidence either way. Not disputing it — flagging that I did not reproduce it.

**You were right that the PRD sentence was the load-bearing error.** I had even written the wedge hazard into #80's body for the fsync case and failed to apply it to the subscriber case, which is the one the code actually implemented. That is the sibling-case shape, in my own reasoning about sibling cases.

## The fix — your first route

A durable close no longer rejects. `session-recorder.ts`: the `try/catch` now covers **only** append + sync; past that point the publish runs outside the rejection path and a throwing subscriber is reported via `console.error`, never propagated. The `closed` flag is gone — it was only ever standing in for "which side of the durable boundary are we on", and the boundary is now structural.

Rotation therefore always reaches `removeSessionLock` and `openSession`, the seal is released normally, and the recorder is reusable. Inside the fence — **`rotate.ts` untouched**, so no third widening.

prd-40 ruling 1's third amendment is rewritten as three rules, with rule 3 recorded as *found in review of this PR*, and the "forbids strictly more" claim is now stated honestly: it is true **with** rule 3, and it was not true without it. CHANGELOG updated to match.

## New laws, both mutation-proven

- `a subscriber throwing after a durable close neither fails the close nor reopens the log` — asserts `resolves`, the error reported, and `isSealed` still `true`.
- `leaves the whole rotation able to finish when a subscriber throws` — drives the **real `rotateSession`**, per your probe, and asserts the rotation resolves, the seal is released, and a later `record()` settles.

| Mutation | Result |
|---|---|
| M7 · a throwing subscriber rejects the durable close again *(this regression)* | **2 failed** / 21 passed — both new laws |
| M8 · release the seal after a durable close *(breaks prd17 ruling 1)* | **1 failed** / 22 passed |

Two mutations, because the two properties are independent: M8 confirms the fix did not buy the no-wedge guarantee by giving back the durable-close hold.

## Gates, re-run in full after the change

`npm run build` 0 · `VITEST_MAX_WORKERS=6 npm test` 0 — **381 files, 6561 passed + 3 expected fail** · `npm run typecheck` 0 · `npm run lint` 0, 848 files. Boot smoke green. `src/recorder` + `src/server` 4× concurrent: 258/258 every run.

**Worth knowing, and I am not hiding it:** my first full-suite run after the fix came back **red — 4 failed**, all `packages/web/src/scene/**` 5 s timeouts (`SceneView.test.tsx`, `marks.test.ts` ×2, `parity/contour.test.ts`). All four passed in isolation (237/237) and the clean full re-run was green. My machine was running scratch worktrees at the time. So the known `contour.test.ts` flake is **not a single test** — it is at least four in that scene family, and it reproduces under real load. That deserves its own issue; it is not this wave's, and I have not filed it yet.

## Your two non-blocking notes

- **`eventCount` can undercount** — agreed, and agreed it does not belong in this wave. It is the same family as #80; I would rather it join that issue's ruling than get patched here. Say the word and I will add it.
- **The rotation guard may be unreachable in production** — agreed, and the point about nobody misreading the test as evidence of a live race is well taken. The law is depth against the writer's FIFO guarantee changing, not a claim that it fires today.

Ready for another look.

Issue #82, opened 2026-08-25
prd46 w6: the rise fixtures derive their delta from the tolerance they test


gabriel-canaan:
## The defect

Three fixtures in `packages/server/src/gate-honesty-law.test.ts` hardcode a ±1 delta against the rise tolerance, so changing `RISE_TOLERANCE` breaks them with messages that name neither the tolerance nor the change.

Found during the fix re-review of prd45 w2 (#48). EXECUTED: mutating `RISE_TOLERANCE=0` to `=1` in `scripts/gate.sh` —

```
3 tests red at RUNTIME:
  expected 'RISE_NOTE=[]' to contain 'ROSE from 8 to 9'
  expected 'RISE_NOTE=[]' to contain 'ROSE from 2 to 3'   (x2)
```

The other ~70 tests still report, which is the improvement #48 already landed: before its anchor fix, the same mutation killed the whole file at *collection* with `expected exactly one line … containing "RISE_TOLERANCE=0", found 0`. So the failure moved from collection to runtime — better, and still not good, because the operator reading those three messages cannot tell that a tolerance change caused them.

## Why it matters

`scripts/gate.sh:181-183` promises that a human changes the tolerance *"by editing this one line"*. That promise is not currently true: editing that one line also requires updating three fixtures, and nothing says so at the point of failure.

This is the same class the wave was about — a test that cannot fail for the reason it claims, inverted: a test that fails for a reason it does not name.

## Why it was deliberately NOT fixed in #48

The fix is a **behaviour change** to test data. #48's final round was scoped to text-only edits precisely so it would be inert under the fix-review rules and not re-trigger a full independent-seat pass. Widening the deltas would have re-opened that gate for a non-blocking cosmetic improvement. Recorded here instead, with both seats' agreement that it does not block.

The comments in `gate.sh:181-183` and `gate-honesty-law.test.ts:754` were softened in #48 to state the truth in the meantime — that the edit no longer breaks collection, that three fixtures still assume a tolerance of 0, and that changing it means updating them. **This issue removes the caveat rather than restating it.**

## Definition of done

- The three rise fixtures derive their delta from `RISE_TOLERANCE` rather than hardcoding ±1, or assert in a way that names the tolerance when they fail.
- EXECUTED both directions: with the tolerance at 0 the suite is unchanged, and with it at 1 the suite still passes (or fails with a message that names the tolerance).
[...]

Issue #75, opened 2026-08-25
prd46: the manifest prune says it pruned only when it did


gabriel-canaan:
## The defect

`scripts/gate.sh`'s lane-manifest prune prints `"lane manifest pruned: <handle>"` in cases where **nothing was pruned**. prd-45 wave 1 fixed the swallowed-exception facet of this block (the `try/catch{}` that hid both a `JSON.parse` failure and a `writeFileSync` failure); the success line is still a proxy for "node exited 0" rather than for "a lane was removed".

EXECUTED during the landing-gate review of PR #68, against the post-wave-1 script:

```
merged, lanes.json is a JSON ARRAY       rc=0  ->  "lane manifest pruned: t42"   (file written back UNCHANGED)
merged, object with no .lanes key        rc=0  ->  "lane manifest pruned: t42"   (nothing pruned)
```

Both are shapes a hand-edited or half-migrated manifest really takes.

## Why it matters

The manifest exists so observers do not fence a ghost — `AGENTS.md` records that three landed lanes read OFF-FENCE in a live UI before the prune existed. A prune that reports success while leaving the entry in place recreates exactly that state, and now does so *with a line saying it did not happen*.

This is prd-45 **ruling 2** — *a postcondition asserts the fact it claims* — and prd-46 **ruling 1**'s class: the verdict is derived from the exit status of the thing that ran, not from the fact the operator cares about.

## Definition of done

- The success line is printed only when the handle was actually removed from the manifest. "The file parsed and was rewritten" is not the fact being claimed.
- A manifest whose shape the prune does not understand (array, missing `.lanes`, wrong types) is a **hold or an explicit report**, not a silent success — consistent with everything wave 1 did to its neighbours.
- A manifest that legitimately has no entry for this handle still succeeds quietly; that is not an error and must not become one.
- A missing manifest file remains a no-op, unchanged.
- EXECUTED for all four shapes above, with the controls: a normal prune still prints success and really removes the entry.
- `fail()` and the `MERGED` flag are reused, never reimplemented.

## Proving it — you cannot run the gate

`scripts/gate.sh` **is** the landing: it merges into local `main`, installs, builds, and ends with `git push origin main`. **Never run it.** Extract the prune block and exercise it against fixture manifests in a scratch directory, with exit statuses.

## Fence (may touch ONLY)
- `scripts/gate.sh`
- `packages/server/src/gate-honesty-law.test.ts`

## Wave — deferred deliberately

[...]

Issue #74, opened 2026-08-25
prd46 w4: the law's unwritable-file proofs hold as root, or say they do not


gabriel-canaan:
## The defect

Two assertions in `packages/server/src/gate-honesty-law.test.ts` (around `:399` and `:428`, and again in the manifest-prune block) prove a guard holds by making a file unwritable with mode `0444` and asserting the write fails.

**As root, mode `0444` does not prevent writing** — `CAP_DAC_OVERRIDE` ignores it. So those tests assert something that is false for the root user, and they go red in any root container.

REASONED on the mechanism (standard POSIX/Linux capability behaviour), and flagged independently by both landing-gate seats during the verification of PR #68. Nobody has executed it, because nobody ran the suite as root — which is precisely the gap.

## Why it is worth fixing rather than ignoring

CI runs as a normal user today, so this is latent. It stops being latent the moment anyone runs the suite in a plain `docker run` (root by default), and the failure will look like a real defect in the gate rather than a wrong assumption in the test — costing whoever hits it the same investigation twice.

It is also the shape this repo keeps finding: **a test that cannot fail for the reason it claims**, inverted — a test that fails for a reason it did not claim.

## Definition of done

- Either the tests skip themselves when running as root, with the reason stated in the skip, or they achieve unwritability by a means root also respects (an unwritable *directory* the file must be created in, an immutable attribute, or a path that cannot exist).
- EXECUTED both ways: as a normal user (still passes, unchanged) and as root (passes or skips, never a false red).
- Whichever route is taken is stated in the test's own text, so the next reader knows the constraint was considered.

## Fence (may touch ONLY)
- `packages/server/src/gate-honesty-law.test.ts`

## Blocked by
Wave 3 (#73) — it claims the gate-honesty law, and two concurrent lanes in one file are a rebase conflict at landing. Not a dependency on its content. Wave 1 (#70) landed 2026-08-25.
Also gated on prd-46 wave 0, the operator's own act.
**Model:** sonnet. **Wave:** 4.

---

## Correction, 2026-08-26 — "wave 1 LANDED" was true of a branch, not of `main`

The Blocked by section above says **wave 1 (#70) LANDED 2026-08-25**. That was true of the
integration branch `prd46` and false of `main`. PR #85's base was `prd46`, not `main`, so
`main` never carried the `N_RC=$?` fix or the structural predicate — a lane cut from `main`
on the strength of that sentence would have got a `gate.sh` without wave 1.

[...]

Issue #73, opened 2026-08-25
prd46 w3: nothing the gate or its law prints outruns what it established


gabriel-canaan:
## The defect — one fact, three live instances

`scripts/gate.sh` still prints two verdicts it has not earned. prd-46 ruling 1 replaced the *enumeration* of swallowed-failure spellings with a structural predicate; these are the two places where the **verdict itself** outruns what was actually established. Both verified on main after #70 landed.

### 1. A corrupt ref and a genuine non-ancestor are told the same story

Declared tolerance #7 justifies a discarded stderr on the rebase-ancestry check with *"exit code still routed through fail()"*. True, and incomplete: `git merge-base --is-ancestor` returns **1** for *not an ancestor* and **128** for *bad object*. The discarded stderr was the only thing distinguishing them, so both land on `"the rebase did not take"`. An operator whose ref is corrupt is told their rebase failed.

Verified present on main: the tolerance entry at `gate-honesty-law.test.ts:203`, needle `merge-base --is-ancestor main HEAD 2>/dev/null || fail`.

### 2. The manifest prune says it pruned when it pruned nothing

#42 made node's own exit code decide which line prints, which closed the swallowed-`catch` half. It does not close this half: a `lanes.json` that is a **JSON array**, or an object with **no `.lanes` key**, parses and writes back fine — node exits 0, the gate prints `lane manifest pruned: <handle>`, and nothing was pruned.

Verified present on main: the `if/else` still branches on node's exit status alone.

Why that matters beyond tidiness: the manifest exists so observers do not fence a ghost — AGENTS.md records three landed lanes reading OFF-FENCE in a live UI before the prune existed. A prune that reports success while leaving the entry recreates exactly that state, now with a line claiming it did not happen.

### 3. The law's own failures do not name the tolerance that caused them

Three rise fixtures in `packages/server/src/gate-honesty-law.test.ts` hardcode a
±1 delta against `RISE_TOLERANCE`, so changing that tolerance breaks them with
messages naming neither the tolerance nor the change. EXECUTED (from #82, merged
here): mutating `RISE_TOLERANCE=0` to `=1` in `scripts/gate.sh` turns 3 tests red
at runtime with `expected 'RISE_NOTE=[]' to contain 'ROSE from 8 to 9'` and two
siblings — while ~70 other tests still report, which is the improvement #48
already landed (before its anchor fix the same mutation killed the file at
collection).

[...]

PR #83, opened 2026-08-25
prd45 w2: a risen ratchet is reported, and the recursive walk is exercised

### LINKED ISSUE #48: prd45 w2: an accidental timing enrolment cannot raise the ratchet unseen

### LINKED ISSUE #76: prd45 w2: the recordings law's recursive branch is exercised, or its silence is stated
## The defect

`packages/web/src/recordings/no-live-fleet-law.test.ts` walks its governed root recursively (prd45 w1, #44) and compares the walk against an independent traversal — a good property, and better than the sibling law's hardcoded floor. But `packages/web/src/recordings/` has **no subdirectories**, so the recursive branch is never exercised by the committed suite.

EXECUTED during the verification of PR #68, both directions:

```
nested leak.ts present, recursion intact:   the forbidden-pattern test bites the leak
nested leak.ts present, recursion removed:  the equality test names the missing file
no nested file, recursion removed:          3 passed   <- the recursion is unexercised today
```

So the law re-arms correctly the instant a subdirectory exists. What it cannot do today is *prove* it re-arms, from the committed tree alone.

## The honest counter-argument, because this issue is a judgement call

#44's own definition of done explicitly permits either outcome: *"Remove the fixture, or keep it as a committed negative if it reads as one."* The lane removed it, which was within its instructions, and one reviewer rated committing it a must-fix while two others rated it worth-later. This issue exists so the disagreement has a home, not because the lane did something wrong.

The case **for** committing a fixture: the recursion is the entire subject of #44, and a branch that is never taken is a branch nobody knows is broken. prd-24's evidence for the sibling law is exactly this — `lab/branching/geometry.ts` imported a forbidden module unseen for as long as the walk stayed flat.

The case **against**: a fixture directory in a governed source tree is itself something future readers must understand and not delete, and the equality assertion already fails the moment a real subdirectory appears with anything in it.

## Definition of done

- A decision, taken and stated: either a committed negative fixture that makes the recursive branch exercised today, with a comment saying why it exists and that it must not be deleted, or a note in the law's own text recording that the recursion is deliberately unexercised until the directory grows, and what re-arms it.
- If a fixture lands, EXECUTED both directions on the committed tree: the law green with it, and red if the recursion is removed.
- `FORBIDDEN_PATTERNS` unchanged either way — what the law forbids is not this issue.

## Fence (may touch ONLY)

## Wave — deferred deliberately

Not assigned a wave. It claims the same directory as prd45 w1's #44, which is in PR #68 and unlanded. Groom into a wave once that lands; it collides with nothing else and could go at any time after.

gabriel-canaan:
prd-45 wave 2 — the ratchet that only looked one way, and a walk that was never taken. One commit per issue, fences disjoint.

Closes #48
Closes #76

**Note on shas:** cherry-picked from two lane branches onto current `main`, so these are not the shas the issue comments cite.

## What becomes true

- **`a47cba7` (#48)** — a risen timing-count is reported rather than silently absorbed into the floor. The ratchet failed only on a **drop**; a rise was written into the floor unnoticed, so a file that enrolled by accident raised the floor and a later *correct* removal then failed the landing **naming the wrong change**.
- **`b96579a` (#76)** — the recordings law's recursive branch is exercised by the committed tree, not merely provably correct.

## Why #48 is not hypothetical

prd-45 wave 1's own law enrolled itself in the timing set by quoting the `@gate-timing` marker in a tolerance table — TCOUNT 6 → 7 — and **nothing in the tool noticed**. An external review seat caught it. That is this issue's thesis, demonstrated on the issue's own PRD.

## The design decision, stated because it is arguable

**A rise is REPORTED, not failed.** Hard-failing every rise punishes the legitimate act of adding a timing test, and a check that annoying is suppressed — or has its ratchet file deleted — within a week, which is worse than the gap it closes. `RISE_TOLERANCE` is declared data (prd-45 ruling 3), not a condition folded into a regex.

An independent seat judged this premise on request and concluded it *"does break the compounding chain, but only just"*: it gives the operator a signal at the moment the floor moves, when an accidental enrolment is still fresh and attributable. Two residuals are named rather than hidden — the note says *that* a rise happened, not *what* rose (the ratchet stores a count, not a list), and it prints after the write rather than before. Both are filed, not fixed here.

## Verification

Gates on this merged branch, node 22.22.2: `typecheck` 0 · `lint` 0 · `VITEST_MAX_WORKERS=6 npm test` 0 — **381 files, 6567 passed**, 3 expected fail. `gate-honesty-law.test.ts` alone 73/73; **4× concurrent: 0/4 red** (that file raced itself once, on a fixed `/tmp` path).

**Four rounds, two of them independent-seat passes** (Claude Opus and Codex `gpt-5.6-terra`, both at xhigh, each in its own disposable worktree, blind to each other). Both returned **safe to merge** on the final pass.

Findings raised and closed, each re-checked with a spelling it was not built against:

| # | finding | closed by |
|---|---|---|
| 1 | the rise report claimed *"this becomes the new floor"* while an unwritable floor left it unchanged and the gate walked on to merge and push | the write's status is checked; `fail()` fires pre-merge, so recovery reads `HOLDING — not merged` |
| 2 | three `:NN` citations staled by #48's own insertions | re-derived — and then re-derived twice more, as each subsequent fix shifted the file again |
| 3 | a leading-zero floor (`08`) errored bash arithmetic, killing the rise check; `010` announced a false rise | non-canonical floors are rejected, matching the neighbouring "a corrupt ratchet must not be silently trusted" |
| 4 | the tolerance was pinned by **value** under a test name claiming it checked the **variable** | anchored on `RISE_TOLERANCE=`; mutating the tolerance now names it |
| 5 | that guard's remedy advice — *"a human clears or fixes it"* — could destroy a real drop hiding behind a corrupt-looking floor | the message now requires comparing the count before clearing, and says correcting the value is the safe act |

Mutation evidence on the final tree, from the seats and reproduced here: `/dev/full` (a **write**-time failure, where the earlier round used `chmod 444`, an **open**-time one) still holds; 19 spellings against the leading-zero guard reject nothing legitimate, bare `0` included; reverting the per-process `/tmp` handle produces **5/8 red**, proving the concurrency harness still detects a race; deleting `|| fail` reddens exactly one test, on point.

## What is the sibling case?

**It kept being the answer.** Four consecutive rounds here ended with the fix opening the next hole: removing a floor made a law vacuous; a sweep repair planted the marker that enrolled the law; re-anchoring that marker moved the law into the concurrent batches it had been excluded from; and fixing the citations then staled two more, twice. Every round's finding was a structural sibling of the previous round's *repair*, not of the original defect. The lane records this in `a47cba7`'s body.

The residual siblings are named and filed: **#82** (three fixtures hardcode a ±1 delta, so changing the tolerance reddens them with messages naming neither), **#75**, **#73**, **#74**, **#72**.

None of the above — and the strongest check is that **reverting #48's `gate.sh` hunk while keeping its test file reddens at collection**, so a silent revert of this fix cannot land. Six mutations were run against the ratchet: neutering the rise condition (4 red), `-gt`→`-ge` (1 red, the HOLD test), inlining the tolerance as a literal (1 red — ruling 3 genuinely enforced), deleting the report line, changing the tolerance value, and the full revert.

For **#76**, the four cases that matter: committed tree green; recursion disabled with the fixture present → red, *naming* `recursion-fixture/walked-marker.ts`; fixture deleted with recursion intact → green, which is exactly what its own doc claims; both removed → green. The fixture is **evidence, not enforcement**, and says so in its file doc.

## Fence widenings

None. `a47cba7`: `scripts/gate.sh` + `packages/server/src/gate-honesty-law.test.ts`. `b96579a`: `packages/web/src/recordings/` only. Proven disjoint by `scripts/fence-lint.sh 48 76` before dispatch.

## Landing

`scripts/gate.sh` **must not land this** — #48's fence is that script, so the gate would be merging using the very code this PR changes. Hand-landing, as prd-45 wave 0 books and as wave 1 was landed.

## Reviewer note

I verified this and requested every fix in it, so I should not approve it. The final round was gated by two independent seats *on the fixes themselves* rather than only on the original commits — which is the check that caught wave 1's landing blocker, and the reason it is now written into the review process rather than left to judgement.


ciaran-slow (approved):
**Approved.** Reviewed commit-by-commit in a clean clone at `b96579a`, node v22.23.2. Everything below is labelled `EXECUTED` (I ran it, read the exit status) or `REASONED` (argued from the code, nothing run).

**The PR does not commit the defect it exists to catch.** This was the first thing I checked and the most load-bearing: `a47cba7` adds 314 lines to `gate-honesty-law.test.ts` that discuss the timing marker at length, in a file under `packages/server/src` — exactly how #42 enrolled itself. Running gate.sh's own discovery (`gate.sh:165-167`) verbatim:

```
EXECUTED
origin/main    TCOUNT=6
PR head        TCOUNT=6      identical file lists, byte for byte
```

The `'// @gate-' + 'timing'` split holds. `REASONED`: the "does NOT enroll this law file" test is what keeps it holding.

**The octal hazard is real, not theoretical.** I did not take the commit message's word for this — `EXECUTED` under `bash` (not the zsh I first ran it in, which does not reproduce it):

```
$(( 10 - 08 ))   ->  bash: 08: value too great for base
$(( 10 - 010 ))  ->  2        # 010 read as octal 8: an honest hold at 10 reads as a rise from 8
[ 9 -lt 08 ]     ->  rc=1     # the test builtin reads decimal
[ 9 -lt 010 ]    ->  rc=0     # the two consumers genuinely disagree
```

**The guard's boundaries are right.** `EXECUTED`, 15 spellings against `case (0[0-9]*)`: accepts `0 1 6 10 20 100 1000000`, rejects `00 000 007 08 09 010 019`. Bare `0` is accepted — it does not overreach.

**The sibling case is closed, and I checked it rather than trusting the claim.** `EXECUTED` (grep): `$PREV` has exactly two consumers — `:224` (`-lt`, decimal-safe) and `:229` (`$(( ))`, now guarded). There is no third. Writes of gate state are exactly two — `:264` (guarded by this PR) and the `lanes.json` prune at `:306` (already guarded). No unguarded sibling remains.

**`fail()` at `:265` really is pre-merge.** The claim the recovery line depends on. `REASONED` from control flow: the `LOAD` block closes at `:271`, `MERGED=1` is at `:288`. So a floor-write failure prints `HOLDING — not merged`, which is true.

**Every `:NN` citation resolves.** I expected this to be where the PR frayed — it is the thing the commit message claims most confidently and the easiest to get wrong. `EXECUTED`: all 19 distinct citations (`:24 :56 :74 :89 :116 :154 :166 :187 :190 :216 :227 :232 :240 :243 :245 :264 :266 :287 :293`) land on exactly the line they describe. `:187` → `RISE_TOLERANCE=0`, `:216` → the leading-zero guard, `:264` → the write, `:266` → the report, `:293` → the prune. None stale.

## Mutations

`EXECUTED`, each reverted after:

| mutation | result |
|---|---|
| `scripts/gate.sh` reverted wholesale to `origin/main`, test file kept | **red at collection** — `Test Files 1 failed`, `no tests`, `sliceLines` throws. A silent revert cannot land. |
| recursion disabled in `walkSourceFiles` | **1 of 4 red**, naming `recursion-fixture/walked-marker.ts` in the diff |
| `\|\| fail` deleted from the floor write | **exactly 1 red**, and it is the on-point one (READ-ONLY floor HOLDS) |
| rise condition neutered | **5 red** (one more than your 4 — my mutation also dropped the `$RISE_TOLERANCE` reference, so it is a broader cut, not a discrepancy) |
| `RISE_TOLERANCE` 0 → 5 | **4 red at runtime with real assertion diffs**, collection intact — your item-3 caveat is exactly right |

## Gates

`EXECUTED` in a clean clone, node v22.23.2 — reproduces your reported numbers exactly:

```
VITEST_MAX_WORKERS=6 npm test   381 files, 6567 passed | 3 expected fail   rc=0
npm run typecheck               rc=0
npm run lint                    849 files, rc=0
gh pr checks 83                 7/7 pass — the full two-job grid
```

`gh pr checks` was 5/7 with two ubuntu legs pending when I started; both have since passed, so all of `build-test-boot` ×3 and `pack-smoke` ×4 genuinely executed.

**What I did not run:** `scripts/gate.sh` itself, deliberately — #48's fence *is* that script, so the gate would be landing using the code under review. Your "hand-land this" note is right and I did not work around it. I also did not run the 4× concurrency probe; your 0/4 stands unreproduced by me.

## One fix pushed — `174367d`

Mechanical, text only. `b96579a` adds `recordings/recursion-fixture/`, which makes the file doc's own earlier sentence false as of that commit:

> `recordings/` has zero subdirectories today, so that gap is latent here, not live

It sat 16 lines above the new #76 paragraph asserting the opposite, so a top-down reader was told the branch is unexercised, then that a fixture exercises it. `EXECUTED`: `find packages/web/src/recordings -type d` returns `recursion-fixture`. Reworded to the past tense it was always describing, pointed at the #76 paragraph. Your `:NN` sweep would not have caught it — it is a prose fact, not a citation.

`EXECUTED` after: lint 0, typecheck 0, `no-live-fleet-law.test.ts` 4/4. No assertion, fixture or walker touched. The full suite ran before this comment-only change, not after.

**#82 undercounts by one, and its DoD would ship the miss.** `EXECUTED`:

```
RISE_TOLERANCE=1  ->  3 failed | 70 passed
RISE_TOLERANCE=2  ->  4 failed | 69 passed    <-- adds 'ROSE from 5 to 7'
```

The fourth is the plain-rise fixture — floor `5`, `TCOUNT=7`, delta **2**, not 1. "Three" is correct at the tolerance #82 measured and wrong at ≥ 2. #82's DoD names *"the three rise fixtures"* and proves itself *"with it at 1"* — the one value the fourth fixture already survives, so a three-of-four fix passes that criterion with the defect intact. Same for the softened comments at `gate.sh:186` and the law's declared-variable test: literally true (three *do* have delta 1), operationally short by one.

Detail and a suggested amendment posted on #82 — nothing to change here.

**Not merging.** Per this repo's convention the merge is the operator's manual step, and your hand-landing note applies.

Issue #81, opened 2026-08-25
prd40 w3: architecture.md says what a replay may now contain


ciaran-slow:
## Fence (may touch ONLY)
- `docs/architecture.md`

## Blocked by
Nothing. ADR-0029 is accepted and on `main` (#77). Shares no file with #5 (`api/meta.ts`), #26 (`poll-loop.ts`) or #69 (`session-recorder.ts`), so it bundles with wave 3.

**Model:** sonnet. **Wave:** 3.

## The gap

`grep -n "0029\|repeat a fact\|at-least-once" docs/architecture.md` returns **nothing**. ADR-0029 changed the replay contract — a recording may now contain the same fact twice, at-least-once on the poll path, with no read-side dedupe — and `architecture.md` does not mention it.

The convention is the one the verify stage enforces and `docs/adr/README.md` sets: **`architecture.md` stays the current-state picture; the ADR carries the reasoning**, and an ADR that changed the architecture must be cited from architecture.md's decisions log. ADR-0011 is cited there; its amendment is not.

Found during the verify pass on #4 and confirmed independently by a second reviewer. It is a miss from **#77**, not from #4 — `architecture.md` was outside that PR's fence, and outside #4's.

## Why it matters more than a missing link

`architecture.md` is what a new reader — human or agent — is pointed at for how the system works now. Today it describes a recording as if every line names a distinct fact. Anyone reasoning about replay, the fold, or `eventsSoFar()` from that document will get the delivery guarantee wrong, and the ADR that would correct them is not reachable from it.

## Definition of done
- `architecture.md`'s decisions log cites **ADR-0029** by number, with a one-line statement of what changed: at-least-once on the poll path, at-most-once on the other eleven `recorder.record` sites, no read-side dedupe.
- Wherever `architecture.md` describes the session record or replay, it no longer implies exactly-once delivery. **Check the text, not just the log** — a citation appended to the decisions log while the body still says the opposite is the failure mode here.
- No ADR is edited. The log is append-only; ADR-0029 and ADR-0011 already carry the reasoning and the amendment link.

## Not in scope
The seven non-idempotent reducer arms ADR-0029 names. They are a cost the ADR accepted deliberately, not a defect this issue fixes.

ciaran-slow:
# Plan — #81 · prd40 w3: architecture.md says what a replay may now contain

Planned against `main` at `80c796e`. **Follow this to the letter. If it is wrong, stop and say so.**

[...]

Issue #80, opened 2026-08-25
prd40 w6: a fsync failure cannot leave a line after session.closed


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/recorder/session-recorder.ts`
- `packages/server/src/recorder/session-recorder.test.ts`

## Blocked by
- **#4** (prd40 w2) — the `closed` flag this issue moves does not exist until #4 lands.
- **An operator ruling.** See "The ruling this needs" below. There are two bad outcomes and picking between them is not a lane's call.

**Live fence conflict with #69**, which holds the same two files, so this cannot share a wave with it. Sibling lanes hold `api/meta.ts` (#5) and `poll-loop.ts` (#26).

**Re-waved 2026-08-25, wave 4 → wave 6.** Three issues claim `session-recorder.ts` — #26 (wave 4), #69 (wave 5) and this one — and AGENTS.md forbids bundling across a live fence. This one runs last because it is the only one of the three still gated on an operator ruling, so it would have held the others up.

**Model:** sonnet. **Wave:** 6.

## The defect

`closeWith` sets `closed = true` **after** `await this.writer.sync()`, but the `session.closed` line is already on disk once `append()` resolves. An append that succeeds followed by a **fsync that fails** therefore takes the `!closed` branch, releases the seal, and lets a later `record()` append behind `session.closed`.

That is the outcome **prd17 ruling 1** makes structural, and the outcome prd-40 ruling 1's third amendment says it prevents. The code comment justifies the guard with *"releasing then would let a later record() append behind session.closed"* — a reason that applies once the **append** alone has resolved, one step earlier than where `closed` is actually set.

EXECUTED, by two reviewers independently, with throwaway probes mocking `SessionLogWriter.prototype.sync` to reject once:

```
isSealed after failed sync = false | lines on disk = ["session.closed","collector.error"] | closed is last = false
```

## It is pre-existing, not a regression

The same probe run against `main` in a separate worktree, before #4:

```
MAIN: isSealed=false lines=["session.closed","collector.error"] closedIsLast=false
```

Identical. #4 does not introduce this and does not worsen it. #4 *does* change the surrounding behaviour in the fsync case — the close event is no longer pushed or folded, so the fold sits **behind** the file rather than ahead of it, which is the safer side of the same disagreement.

## The ruling this needs

The naive fix — set `closed = true` immediately after the append — has a real cost that the amendment did not weigh:

[...]

Issue #69, opened 2026-08-24
prd40 w5: foldSoFar() cannot be corrupted by the caller that reads it


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/recorder/session-recorder.ts`
- `packages/server/src/recorder/session-recorder.test.ts`

## Blocked by
- **#3** (prd40 w1) — this is a property of the `foldSoFar()` it introduces; there is nothing to constrain until it has landed.

Sibling lanes hold `packages/server/src/api/meta.ts` (#5) and `packages/server/src/server/poll-loop.ts` (#26). **Do NOT touch either.** This issue makes the recorder's own return safe; it does not wire any caller.

**Re-waved 2026-08-25, wave 3 → wave 5.** Not a change of scope or priority. `packages/server/src/recorder/session-recorder.ts` turned out to be claimed by three issues — this one, #26 and #80 — and AGENTS.md forbids bundling across a live fence, so they run in consecutive waves in priority order: #26 (Now/High) in wave 4, this one in wave 5, #80 in wave 6. #26's claim on the file was invisible when this wave was groomed: its declared fence was `poll-loop.ts` alone, and the widening it needs became visible only when its plan was written. `scripts/fence-lint.sh` compares fences **as declared**, so it passed on the old composition and could not have caught this.

**Model:** sonnet. **Wave:** 5.

## Definition of done
- A caller that mutates what `foldSoFar()` returned cannot corrupt the recorder's fold. A law proves it: mutate the returned state, then assert a later `foldSoFar()` (and `reduceAll(eventsSoFar())`) still answer correctly.
- **The fix must not be a copy.** #3's identity law asserts `foldSoFar()` returns the same object across reads — that is what makes the "no re-fold on any route" guarantee route-independent, including against a deep import that both barrel-scoped spies miss. Copying on every read would defeat it, and would reintroduce a per-read cost that grows with the session, which is the whole thing #3 removed.
- Whatever is chosen is asserted, not merely typed: a `Readonly<SessionState>` return is a compile-time signal only, so if that is the route, the law must show the runtime behaviour it buys (or state plainly that it buys none, and why that is enough).
- The existing laws stay green, unmodified in intent: the identity law, the mid-emit invariant, the one-rebuild law, the two spy laws, and the corpus law.
[...]

PR #89, opened 2026-08-25
docs(prd): prd-40's Sequencing says what the waves actually are

ciaran-slow:
prd-40's Sequencing described three waves and stopped. There are six, and wave 3 is not the pair it names. One commit, one file.

Closes nothing. This is the PRD catching up to grooming, the same shape as #86.

## What changed the plan

**One file.** `recorder/session-recorder.ts` is claimed by #26, #69 and #80 at once. AGENTS.md forbids bundling across a live fence — two issues claiming one path is a rebase conflict already scheduled — so they run in consecutive waves, ordered by the Timeline and Priority already on them rather than by preference:

| Wave | Issue | |
|---|---|---|
| 3 | #5 · #81 | parallel, fenced apart |
| 4 | #26 | Now/High, so it takes the contested file first |
| 5 | #69 | |
| 6 | #80 | last — the only one still gated on an operator ruling |

**#26 moved because it was unbuildable where it was fenced.** Its DoD requires the degrade `collector.error` to be emitted *whether or not its own append succeeds*. After wave 2 the poll loop cannot do that: `record()` appends before it publishes, so it publishes nothing when the append fails — precisely the disk-full case the alarm exists for — and the recorder's emitter is private. Publish-without-append belongs on the durability boundary, which is outside `poll-loop.ts`.

## What is the sibling case?

**The original wave 1–3 text is kept, not rewritten** — and that is the sibling of how ruling 1's amendments were handled in #79. Waves 1 and 2 ran as planned and landed; wave 3's text is the record of what was expected. A reader of only the correction loses why the split happened, and a reader of only the original builds against a fence that no longer exists. Both need to be on the page.

The other sibling is in the findings themselves. Having found that `#26`'s fence was wrong, the question is whether the *other* wave-3 issue had the same problem — and #5 did, in a different form: it shares no file with #69 but does share `foldSoFar()`'s return type. Looking for one instance of a class rather than fixing the one instance found is what turned a single re-wave into the two recorded couplings below.

Nothing executable here, so: what could a reader still get wrong?

- **Dispatching #26 against `poll-loop.ts` alone.** Its fence now carries the widening, recorded before any change, and the Sequencing says why.
- **Assuming `fence-lint` clean means a wave is safe.** It compares fences **as declared**. Both couplings this wave hit are invisible to it, and both are now written down so the next PRD expects them: *a widening that has not happened yet* is not in any fence, and *a shared type is not a shared path*.
- **Re-deriving whether the degrade event may skip its append.** Success 1's named exception is on `main` since #86 — it had been stranded on `prd39-paper`, pushed the day after that branch's PR merged. The Sequencing records that it was found stranded rather than absent, so nobody re-rules it.

## Gates

`npm run typecheck` clean · `npm run lint` clean · doc laws 21/21 · every markdown link resolves on disk · no personal paths.

Suite and build not run: one markdown file, no code. CI runs the full grid regardless.


gabriel-canaan (approved):
Reviewed at `78f0ae1`, one commit, one file. Every claim below was checked by running something; nothing here is reasoned-only unless labelled.

## What verified in the PR's favour — all EXECUTED

**The wave table matches the live tracker exactly.** Issue titles carry their own wave numbers, and all five agree with the section: `#5` w3, `#81` w3, `#26` w4, `#69` w5, `#80` w6.

**The stated ordering reasons are real, not reconstructed.**
- `#26` is `Now / High` on the board — the reason given for it taking the contested file first.
- `#80` is the only one of the three still carrying an operator-ruling blocker (its Blocked-by names one) — the reason given for it going last.
- `#26`'s fence on the issue already carries the widening *and* the "recorded before the change" note, and it does include `session-recorder.ts`, so the PR's account of the widening is the issue's own.

**The wave-3 pair is genuinely fenced apart.** `#5` → `packages/server/src/api/meta.ts`; `#81` → `docs/architecture.md`.

```
$ scripts/fence-lint.sh 5 81
── overlaps ──
  none
fence lint PASSED     (exit 0)
```

**The `#5` ↔ `#69` type coupling is resolved where the PR says it is.** The read-only binding is on `#5` as a comment timestamped `2026-08-25T03:09:22Z` — 29 minutes before this PR opened — and `#69`'s body really does name the two shapes that make it necessary (`Readonly<SessionState>` on the return, or a dev/test deep freeze). The claim rests on a link that exists.

**Both code claims are true as stated.**
- `private readonly emitter = new EventEmitter()` — `packages/server/src/recorder/session-recorder.ts:52`. The emitter is private.
- `record()` awaits `writer.append(event)` and only then reaches `this.emitter.emit('event', event)`. A failed append throws at the await, so nothing publishes — which is precisely the disk-full case the PR argues makes `#26` unbuildable from the poll loop.

**The history claims hold.** PR #67 closed `#3`, PR #79 closed `#4`, both merged. Success 1's exception was stranded, not absent: `8ba34e3` on `prd39-paper`, re-landed as `25346f7` in #86 — and #86's file list is `docs/adr/0029-...md`, `docs/prds/prd-40-...md`, `docs/prds/prd-41-...md`, matching "along with a prd-41 pointer stranded beside it".

**Gates, re-run rather than taken on report:** `typecheck` clean · `lint` clean (849 files) · all 39 `*law*.test.ts` files, 542 tests, passed · no personal paths in the added lines · the file's single markdown link resolves on disk.

**CI actually executed, not merely green:** all 7 expected jobs completed SUCCESS — 3 × `build-test-boot` (ubuntu current/min, macOS current) + 4 × `pack-smoke`. That is the full grid AGENTS.md describes, with the documented macOS × min-node exclusion.

I did **not** run the full suite or `npm run build`. The PR says so itself and the reasoning is sound for a markdown-only diff; CI ran both.

## Offered, not asserted — a fourth claimant on the contested file

Non-blocking, and it does not change the wave order this PR proposes.

The section says `session-recorder.ts` "turned out to be claimed by three issues at once." Within prd-40's own sequencing that is right. Across the open backlog it is four — **#37** (`prd44 w2: the server's event buffer has a stated ceiling`) is fenced to exactly `packages/server/src/recorder/session-recorder.ts` and its test.

It cannot collide today, which is why this is not blocking: #37 is `Later / Medium / Backlog` and still held by two live blockers — prd-40's own open question ("Should `foldSoFar()` be the only reader…?", still *Open, not ruled* in this same file) and an operator wave-0 act.

The part worth a follow-up is on **#37**, not here: its Blocked-by names only **#3** as "the live claimant on this lane's own file", and #3 landed in #67. When the open question is answered, that line reads as cleared — while #26, #69 and #80 are queued on that file across waves 4–6. Fence-lint cannot catch it either, for the reason this section already gives: it compares fences as declared, and nobody runs it across two milestones' issues at once.

That is a third instance of exactly the class this section was written to record, which is an argument for the section, not against it.

**Approving.** The record is accurate, the couplings it adds are the ones tooling misses, and keeping the original wave 1–3 text rather than rewriting it is the right call for the same reason the PR gives.

Issue #60, opened 2026-08-24
prd29 w2b: the stream authenticates by cookie, never by query param


KelliherL:
prd-29 ruling 4, confirmed by the operator 2026-08-24: `EventSource` cannot set a header, so the stream authenticates by cookie — and never by query param.

## Fence (may touch ONLY)
- `packages/server/src/server/static.ts` — the cookie set beside the existing meta-tag stamp
- `packages/server/src/api/security.ts` — the cookie accepted as an alternate credential
- `packages/server/src/api/stream.ts` — the `preHandler`
- `packages/server/src/api/index.ts` — that one `ROUTE_CLASSES` row
- `packages/server/src/api/route-class-law.test.ts`, `gated-reads.test.ts`, `security.test.ts`

## Blocked by
- **#59** — it shares `ROUTE_CLASSES`, `api/security.ts` and the same laws; this is a stack, not a parallel wave.
- **Coordinate with #4** (*an event reaches the log before it reaches a subscriber*), which owns ordering around `api/stream.ts`. Sequence, don't overlap.

**Model:** sonnet. **Wave:** 2b.

## Definition of done
- The HTML serve that already stamps the capability meta tag also sets it as an **HttpOnly, SameSite=Strict** cookie.
- That cookie is accepted as an alternate credential on **`gated-read` routes ONLY**. A mutation never honours it — an ambient credential on a mutation is CSRF re-invented, and a law asserts the refusal rather than trusting the wiring.
- `GET /api/stream` carries the gate and reads `gated-read`.
- **Spec-level `Last-Event-ID` auto-resume survives untouched**, and the client does not change at all — `useEventStream.ts` sends the cookie automatically. If the client needs an edit, the design is wrong; stop and say so.
- The credential never appears in a URL, a query param or a request line. The query-param and fetch-streaming alternatives are rejected on the record; this issue may not reopen them.
- Open questions this issue does **not** answer, and must not silently decide: whether the cookie rotates on `/api/rotate` (prd-29 inherits prd-23's open question), and dev-mode behaviour under `vite dev`, where no serve stamps anything.

hchristinasmith:
## What Was Built

[...]

PR #90, opened 2026-08-25
prd29 w3: the read axis joins the contract coverage law (#61)

### LINKED ISSUE #61: prd29 w3: the read axis joins the contract coverage law
prd-29's wave 3: the read side joins the contract oracle **by declared enumeration, not ambition** — prd-24 ruling 2's phrasing, and the residual prd-24's closing amendment handed to prd-29 when it retired.

## Fence (may touch ONLY)
- `packages/contract/` — the coverage law, its parser, and the read-side contract tests it enumerates

## Blocked by
**#58, #59, #60.** The read axis cannot enumerate a boundary that is still moving. prd-29's own sequencing gated this on #428, which died with the old repo — `packages/contract/` exists now, so the real gate is the three gating issues above.

## Definition of done
- `packages/contract/`'s coverage law grows a read axis: a named enumeration of the read-seam modules, where adding a seam module without a contract test fails the build. The write side's discipline (inline literal headers, names imported from the one capability module) extends rather than loosens.
- The parser question prd-29 left open is decided **in this issue, in its own words**: whether the read axis extends the existing coverage law's parser or gets its own walker. Say which and why in the PR body.
- The `vite dev` blind spot is **named, not silently inherited**: an `inject` oracle cannot see a token delivered at serve time, so it cannot witness dev-mode behaviour. State that limit in the law's own doc comment rather than letting a green suite imply coverage it does not have. (Same open question prd-23 and prd-29 both carry; this issue records it, it does not rule it.)

hchristinasmith:
Closes #61

This is the last issue in prd-29 — no sibling issues remain in this PRD after this one.

## What changed

`packages/contract/src/contract-coverage-law.test.ts` grows a read axis beside the existing write-side one: `EXPECTED_READS` enumerates the 13 `gated-read` routes that get a contract test (of 14 `gated-read` routes total), each proven end-to-end against the real `buildApp` via `packages/contract/src/harness.ts`. Adding a `gated-read` route with no contract test now fails the build, closed-world, both directions — same discipline as the write side's `EXPECTED`.

13 new contract test files under `packages/contract/src/`, one per route, each with a happy-path case and refusal cases (tampered token, stripped token).

## The parser question — decided here

The read axis reuses the coverage law's existing parser **strategy** — a scoped regex over a closed-world array literal, self-tested against a synthetic sample — but points it at a **new source**: `ROUTE_CLASSES` in `packages/server/src/api/index.ts`, not the web mutation law `routesIn` reads.

`READ_MODULES` on the web side (`replay/mutating-calls-law.test.ts`, prd-29 ruling 6, already shipped) has only ONE row — the shared header module (`recordings/capabilityRead.ts`) — because every read seam funnels its header through that one module rather than each seam declaring its own. It cannot serve as a per-route enumeration at all. `ROUTE_CLASSES` is the actual security-relevant source of truth: it's what the gate-presence law (ADR-0024) already keeps honest against the real running app, so keying on it means a new `gated-read` route with zero web callers still fails this law loudly instead of being invisible to a module-discovery walker. No new AST or filesystem walker — same regex technique, new source, and a simpler shape than the write side's `routesIn` needs (no `file:` expression to dodge).

## The `/api/stream` exclusion

Named, not silent: a dedicated test asserts the raw parse of `ROUTE_CLASSES` finds all 14 `gated-read` rows, and that `EXPECTED_READS` deliberately covers 13, excluding exactly `/api/stream`. It's cookie-authenticated (ruling 4 / #60) because `EventSource` can't send a header at all — there's no `capabilityRead` caller and no fetch-shaped client this harness could drive through a response-based refusal assertion. It's proven by its own dedicated non-contract test (`stream.test.ts`), not by this law.

## The boundary with the already-shipped web law

This law proves "every server-declared `gated-read` route has an end-to-end proof of refusal, driven through a real client against the real server" — by route, outside-in. `READ_MODULES`/`assertHeaderBlocksExact` (already shipped) prove "which files may build a read's header block, and is its shape exactly right" — by file, inside-out. Neither duplicates the other; this package never re-implements any part of `assertHeaderBlocksExact`.

## Why some contract tests check `h.app.inject` + 401 directly instead of `rejects.toThrow`

Each `EXPECTED_READS` row carries a `refusalShape`. 6 of 13 routes' clients throw on refusal (`sessions`, `session-events`, `lab/checkpoints`, `lab/experiments`, `lab/estimate`, `lane-index`) and are checked the same way the write side is: `/rejects\.toThrow/`. The other 7 (`lanes`, `transcript`, `lane-index/:handle`, `session-preview`, `concierge/repos`, `meta`, `doctor`) **swallow** a 401 into a sentinel value (`null`, `{kind:'absent'}`, a React `status:'absent'` state) rather than rejecting — `rejects.toThrow` would never fire on these and requiring it would make the files impossible to write correctly. Their per-file check instead requires both `tamperCapabilityToken`/`stripCapabilityToken` driven AND a direct `h.app.inject(...)` call asserting `401` on the row's own route — closing the same "filename is not coverage" vacuity gap the write law's own doc comments record being bitten by once already (a placeholder resolving to the sentinel for the wrong reason would otherwise satisfy the law while proving nothing).

Two of the swallow-shaped tests (`transcript.contract.test.ts`, `lane-index-entry.contract.test.ts`) needed to go further still, since a 401 and a legitimate "not found" both fold to the identical `status: 'absent'` shape client-side: their refusal assertions check the distinct `reason` text each path actually produces (the gate's own sentence vs. the route's own "no such lane" text), verified by tracing the real code paths and running the tests, not guessed.

## What is the sibling case?

Every other `gated-read` route besides the one under test in each file — the coverage law's closed-world checks (declared routes === EXPECTED_READS, both directions; present files === EXPECTED_READS' files, both directions) are what catch a route added to `ROUTE_CLASSES` tomorrow with no sibling contract test, structurally, rather than relying on a reviewer noticing by eye.

None that were checked — two mutation tests were actually run pre-commit (stripping the `preHandler` gate from `/api/sessions`, a throws-shaped route, and from `/api/meta`, a swallows-shaped route): both went red on the refusal assertions and green on the happy path, then reverted. See review notes below.

## Fence widening — recorded before the change

Issue #61's fence was `packages/contract/` only. `packages/web/package.json` gained 8 additive `exports` entries (`./replay/api`, `./fleet/manifest`, `./drawer/useTranscript`, `./lab/api`, `./lab/launch/estimate`, `./recordings/laneIndex`, `./lane-page/laneIndex`, `./connect/meta`) so these tests can import real `@rhizomorph/web/*` clients through the package's public boundary rather than a deep relative path — the law's own existing check (`/from '@rhizomorph\/web\/[^']+'/`) already requires that discipline, so loosening it to route around missing exports would itself have been the weakening the issue explicitly forbids. No existing entry was touched, reordered, or removed.

## The `vite dev` blind spot

Recorded, not ruled, in the law's own doc comment directly above the new `describe` block: this harness boots the real `buildApp` and serves a real stamped shell, so every assertion is true of the packaged app — but `vite dev` serves `index.html` itself with neither the capability meta tag nor the cookie stamped in (ADR-0012's recorded dev-mode gap), so there is no token for an inject-backed oracle to witness there at all. A green suite here is coverage of the shipped app, not of dev mode.

## Review

Pre-commit review (dispatched `reviewer` agent, not `/code-review`) verified: fetch-stub isolation across all 13 files (re-ran the suite under `--no-file-parallelism`/`--no-isolate`, no leak); all 7 swallow-shaped files' direct inject calls target their own row's route, not a copy-paste from another file; the `transcript`/`lane-index-entry` refusal assertions genuinely distinguish gate-refusal from legitimate-absence by reason text, not vacuously; `EXPECTED_READS.module` matches every file's real import; `packages/web/package.json`'s diff is pure addition; the parser regex matches the real `ROUTE_CLASSES` file, not just its self-test's synthetic sample; two mutation tests (above) actually executed and reverted clean. No findings required a fix.

`npm run typecheck`, `npm run lint`, `VITEST_MAX_WORKERS=6 npm test` all clean (394 test files, 6615 tests passed, 3 pre-existing unrelated expected-fails), confirmed independently by the reviewer in a second run.


ciaran-slow (commented):
## Review of #90 — verified by mutation, one real hole found and fixed

Reviewed commit-by-commit against #61. Fix pushed as `0ecc2fb`. **Substantive rung — I am not merging this; it needs your confirmation comment first.**

I did not take the PR body's mutation claims on trust — I re-derived them, and then went further than the two the reviewer ran.

**EXECUTED — the strongest one.** I neutered `requireCapabilityToken` globally (early `return` before any refusal), then ran the whole contract package:

```
Test Files  18 failed | 1 passed (19)
     Tests  31 failed | 38 passed (69)
```

**All 13 new read files went red.** Not 11, not "most" — all 13: `sessions`, `session-events`, `lanes`, `transcript`, `lab-checkpoints`, `lab-experiments`, `lab-estimate`, `lane-index`, `lane-index-entry`, `session-preview`, `concierge-repos`, `meta`, `doctor`. The single file that stayed green was the coverage law itself, which reads source text rather than driving the server — correct. **None of the 13 is vacuous.**

**EXECUTED.** I picked the case the reviewer did *not* test and that the retro flags as hardest — `/api/transcript/:lane`, where a 401 and a genuine "no such lane" both fold to `status: 'absent'`. Stripped its `preHandler` at `api/transcript.ts:679`:

```
× a tampered token is refused by the real gate ...
× a page served without the token still reaches the wire bare ...
  Tests  2 failed | 1 passed (3)
```

Both refusal tests red, **happy path still green** — which is the part that matters: the two paths really are distinguished by reason text, not passing for a coincidental reason. The `reason`-string work in `transcript.contract.test.ts` and `lane-index-entry.contract.test.ts` is the best thing in this PR.

**EXECUTED.** All 7 `swallows` files' direct `h.app.inject` calls target their own row's route — no copy-paste. Checked each.

**EXECUTED.** `gh pr checks 90`: all 7 expected jobs pass (3 × `build-test-boot` + 4 × `pack-smoke` — the documented grid). I confirmed the steps genuinely ran rather than trusting the green bar; on `build-test-boot (ubuntu-latest, current)`, `Build`/`Test`/`Typecheck`/`Lint`/`Packaging guard`/`Boot smoke` all report `success`. The one `skipped` is the min-node setup on the current-node leg, by design.

**EXECUTED.** `npm run typecheck` exit 0, `npm run lint` exit 0, `VITEST_MAX_WORKERS=6 npm test` → `394 passed (394)`, `6617 passed | 3 expected fail`, exit 0, on node v22.23.2 in a clean clone. (6617 vs your 6615 = the 2 self-tests I added.)

---

### The finding — EXECUTED, and it defeats the closed-world claim

`GATED_READ_LITERAL` had to match **three coupled fields on one line with exact single spacing**, because `routeClass` is what it filters on. The write side's `ROUTE_LITERAL` is `/route:\s*'([^']+)'/g` — **one** distinctive field, whitespace-tolerant, and therefore immune to how a row is laid out. So the read axis was strictly more brittle than the write side whose "same regex technique" it reuses — and a three-field matcher that misses a row **fails open**: the row drops out of the parse, `declaredGatedReadRoutes()` still equals `EXPECTED_READS`, every check agrees with itself, and the route lands uncovered.

I added a 15th `gated-read` row to `ROUTE_CLASSES`, hand-wrapped across lines:

```ts
  {
    method: 'GET',
    url: '/api/some/very/long/route/that/wrapped/:becauseItExceededTheLineWidth',
    routeClass: 'gated-read',
  },
```

```
$ npx vitest run --root packages/contract contract-coverage-law
exit: 0
     Tests  8 passed (8)
```

**8/8 green, with a `gated-read` route that has no contract test.** That is exactly the vacuity #61 asks this law to make impossible ("adding a seam module without a contract test fails the build"). For contrast, the same row added single-line fails loudly on two tests — so the law worked for the shape you happened to write, and only that shape.

Not hypothetical: `biome.json` sets `lineWidth: 100` and the longest current row is 86 chars, so a route ~14 characters longer invites exactly this. (Mitigating, and why I rate this substantive-not-urgent: `"formatter": { "enabled": false }`, so nothing auto-wraps it — it takes a human or agent hand-wrapping.)

**Fixed in `0ecc2fb`**, two parts:
1. The matcher now tolerates whitespace/newlines, so the wrapped shape is read rather than missed.
2. A **tripwire**: the parsed row count is cross-checked against a single-field, whitespace-tolerant count of the `routeClass` token itself. Any row the matcher still cannot read — reordered fields, re-spaced — now throws instead of silently shrinking the closed world. I verified the token count is exactly 14 in `api/index.ts` with no prose false positives.

Both shapes are pinned by their own self-tests, running the same `gatedReadRoutesIn` the law uses — the discipline `routesIn`'s self-test already holds. Re-verified after the fix:

```
wrapped 15th row   → exit 1, 2 failed | 8 passed   (was 8/8 green)
field-reordered    → exit 1, "the gated-read parser read 13 row(s) but the source declares 14"
```

I'll flag that my first draft of the tripwire self-test was wrong — I asserted the *wrapped* sample should throw, but the relaxed regex reads it fine. Corrected before committing; the self-test now uses a reordered row, which is what the tripwire actually guards.

### Two mechanical fixes, same commit

- **The `/api/stream` carve-out cited the wrong file.** It named `stream.test.ts` as the proof. That file contains no `401` and no `cookie` assertion at all — it covers the open stream's replay/resume behaviour. The refusal *is* genuinely proven, in `server/src/api/gated-reads.test.ts`: the bare-request loop `/api/stream` sits in, plus a dedicated `GET /api/stream refuses a wrong cookie value, with no header competing`. The exclusion is honest; only its pointer was wrong — and a pointer is the part of a named carve-out that rots first. Citation corrected.
- **`buildContractHarness` lost its doc comment.** `HARNESS_LIVE_SESSION_ID` was inserted *between* the existing JSDoc and the function, orphaning the paragraph explaining the served-page vacuity assertion onto a string constant. Moved the const above it.

---

### Your call — offered, not asserted (REASONED, nothing run)

- The per-file check requires `h.app.inject(` and the substring `'401'`, but never relates the injected URL to the row's own route — so a copy-pasted inject *would* satisfy the law. All 7 are correct today (I checked). A literal-prefix check (`/api/transcript/`, `/api/lane-index/`) would close it, but param routes make exact matching impossible and this is arguably past #61's scope.
- `expect(route).toMatch(/^\/api\//)` inside the per-row loop is trivially true of every row in a hardcoded literal — it can't fail. Harmless, but it isn't buying coverage.
- Fence: only `package-lock.json` and `packages/web/package.json` fall outside #61's `packages/contract/` fence. The latter is recorded in the PR body as required; the lockfile is the unavoidable mechanical consequence of the new devDeps and I'd not call that a widening. `packages/contract/package.json` gaining `react`/`react-dom`/`@testing-library/react` is in-fence and justified by `useTranscript` being a hook.

### Not run

I did not run `scripts/gate.sh` — that is the operator's landing step, and this PR targets `prd29` rather than `main`. I also did not run the 4× concurrency load-batch pass; the contract suite has no `@gate-timing` markers, but a wall-clock-sensitive regression under load is a condition I have not measured.

**Over to you:** the parser change touches the "parser question" #61 asked you to decide, so I'm not landing my own substantive fix. Confirm `0ecc2fb` reads right to you and this is good to merge.

PR #93, opened 2026-08-25
prd-29: the identity seam — all four waves, ready for review

### LINKED ISSUE #58: prd29 w1b: the four late reads answer only the token's holder
prd-29 ruling 7 (operator, 2026-08-24) moved the four reads that postdated the PRD's route math into `gated-read`. This is that slice.

## Fence (may touch ONLY)
- `packages/server/src/api/lane-index.ts`
- `packages/server/src/api/session-preview.ts`
- `packages/server/src/api/concierge.ts` — the `GET /api/concierge/repos` route only
- `packages/server/src/api/index.ts` — the four `ROUTE_CLASSES` rows **and the stale comment above the tokenless block**
- `packages/server/src/api/gated-reads.test.ts`
- `packages/web/src/lane-page/laneIndex.ts`
- `packages/web/src/recordings/laneIndex.ts`
- `packages/web/src/connect/meta.ts`

## Blocked by
Nothing. prd-44's #30 and #31 closed in PR #50, which rewrote `log/lane-index.ts` behind an LRU cache (ADR-0028) — **the gate goes on the route, not the reader**, so that work is a neighbour rather than a blocker. Read the new cache before touching the route.

**Model:** sonnet. **Wave:** 1b (the keystone's late arrivals).

## Definition of done
- `GET /api/lane-index`, `GET /api/lane-index/:handle`, `GET /api/session-preview/:sessionId` and `GET /api/concierge/repos` each carry `requireCapabilityToken` as a route-local `preHandler`, and their `ROUTE_CLASSES` rows read `gated-read`.
- The gate-presence law (ADR-0024) covers them: deleting any one of those four `preHandler`s turns the suite red.
- The four SPA readers reach them through the one capability-read module, exactly as wave 1's seven do — no second token path, no per-call header literal.
- **The stale comment goes.** `api/index.ts`'s tokenless block currently says `/api/meta`, `/api/doctor` and `/api/stream` gate in wave 2 and *"the rest stay `read`"* — which contradicts ruling 7. After this lands the only tokenless rows are those three (pending wave 2) and `GET /*`, and the comment says exactly that.
- `session-preview.ts`'s own doc comment stops citing the #216 "untokened, like `/api/doctor`" posture, which predates the `gated-read` class and is superseded by ruling 7.
- Nothing outside the browser breaks: all four consumers are the SPA (verified — `lane-page/laneIndex.ts`, `recordings/laneIndex.ts`, `connect/meta.ts`).

### LINKED ISSUE #59: prd29 w2a: /api/meta and /api/doctor answer only the token's holder, and the CLI scrapes the shell
prd-29 ruling 3, confirmed by the operator 2026-08-24: the credential stays in-band and the CLI scrapes rather than stores.

## Fence (may touch ONLY)
- `packages/server/src/api/index.ts` — those two `ROUTE_CLASSES` rows
- `packages/server/src/api/route-class-law.test.ts`, `gated-reads.test.ts`
- `packages/server/src/cli/env.ts`, `cli/doctor.ts`, `cli/rotate.ts` — extracting the one shared scrape helper
- `.github/workflows/ci.yml` — the boot-smoke section only

## Blocked by
- **#58** — it settles the `ROUTE_CLASSES` comment and the law's shape; doing both at once puts two lanes in one table.
- **Coordinate with #43** (*a red leg reports its remaining gates*), which also edits `.github/workflows/ci.yml`. Whichever lands second rebases; do not run these two concurrently.

**Model:** sonnet. **Wave:** 2a.

## Definition of done
- `GET /api/meta` and `GET /api/doctor` carry the token and read `gated-read`, with the gate-presence law covering them.
- `rhizomorph env` and `rhizomorph doctor` bootstrap the token the way `rotate` already does — `GET /` then scrape the meta tag — through **one** shared helper, not a third and fourth copy of the scrape. `cli/rotate.ts`'s existing implementation is the seam to extract from.
- **No token on disk.** `rotate.ts`'s recorded rejection of a token file stands (*"a SECOND copy of the secret, at rest"*); this issue may not introduce one.
- The recorded price is paid honestly: one extra request per CLI invocation, stated in the helper's own doc comment.
- CI's boot smoke scrapes the shell first and **stops `cat`ing `/api/meta`'s body into the build log** — the habit prd-23 ruling 1 already put on the record.
- `rhizomorph env` still works against a healthy server, because env failing kills lane dispatch. Prove it in the same PR.

### LINKED ISSUE #60: prd29 w2b: the stream authenticates by cookie, never by query param

### LINKED ISSUE #61: prd29 w3: the read axis joins the contract coverage law

hchristinasmith:
## Summary

The whole of prd-29 (the identity seam), dispatched across four lanes and merged into this integration branch without a human reviewer looking at any of them first — that was a process mistake, and this PR exists to fix it before anything reaches `main`. **Please do not merge this until it's actually been reviewed.**

Each of the four issues below was independently re-verified by me after the lane finished (typecheck/lint/full suite, plus my own from-scratch mutation tests on the security-critical claims) — but self-verification by the same agent that dispatched the work is not a substitute for a second pair of eyes, which is what this PR is for.

- Closes #58 — the four late reads (`/api/lane-index`, `/api/lane-index/:handle`, `/api/session-preview/:sessionId`, `/api/concierge/repos`) gated behind the capability token (PR #78)
- Closes #59 — `/api/meta`/`/api/doctor` gated, CLI scrape helper shared across `env`/`doctor`/`rotate` (PR #84)
- Closes #60 — `/api/stream` authenticates by HttpOnly/SameSite=Strict cookie, cookie explicitly never honoured on mutations (PR #88)
- Closes #61 — `packages/contract/`'s coverage law grows a read axis, 13 new contract tests (PR #90)

## What to check

Each merged PR (#78, #84, #88, #90) has its own detailed body with the fence, sibling-case reasoning, and mutation-test evidence for that wave specifically — worth reading commit-by-commit rather than as one squashed diff, per this repo's own review convention.

The two properties worth the most scrutiny, since they're the actual security boundary this PRD exists to draw:
1. The cookie set for `/api/stream` (#60) is genuinely never accepted on any mutation route — verified once by me via mutation (forced `/api/rotate` to accept it, confirmed the CSRF-refusal test catches it), worth a second look given it's the one place an ambient credential could reopen a CSRF hole.
2. The gate-presence law (ADR-0024) and the new contract coverage law (#61) both genuinely fail red when a gate is removed — verified by me via direct mutation on `concierge.ts`, `meta.ts`, `doctor.ts`, and by deleting a contract test file, all reverted after.

## Test plan

- [x] `npm run typecheck`, `npm run lint`, `VITEST_MAX_WORKERS=6 npm test` all green at each of the four merges (final count: 394 test files, 6615 tests)
- [x] All 7 CI legs pass on every one of the four underlying PRs
- [ ] **Real review** — this is what's missing and what this PR is for


hchristinasmith:
One finding from an independent pass, EXECUTED and fixed — pushed as \`220a64e\`, not self-approved (this needs a reviewer who isn't me):

**\`readCapabilityCookie\` (\`api/security.ts\`) let a malformed cookie 500 instead of refusing cleanly.** A cookie value with a bare \`%\` (no following hex pair) makes \`decodeURIComponent\` throw \`URIError\`, uncaught — confirmed against the real \`buildApp\` via \`app.inject\`, not reasoned: \`{"statusCode":500,"error":"Internal Server Error","message":"URI malformed"}\` instead of the clean 401 every other bad credential on \`/api/stream\` gets. **Not a bypass** — the gate still refuses, it just does so noisily instead of cleanly. Fixed with a try/catch treating a malformed value as "no cookie," same as any other unparseable one. Mutation-verified: reverting the catch turns the new regression test (\`security.test.ts\`) red.

Also, separately from that fix — cross-checked across all four merged waves rather than per-PR:
- CI's \`Test\` step genuinely ran (and passed) before \`Typecheck\`/\`Lint\`/packaging/boot-smoke on all four underlying PRs (#78, #84, #88, #90) — none silently skipped.
- The final \`ROUTE_CLASSES\` table has exactly 14 \`gated-read\` rows and 0 tokenless \`/api\` reads besides the ones still pending (none — this PRD closes that out), matching the PRD's own success criterion 5.
- \`requireCapabilityToken\`'s \`allowCookie\` option: confirmed only \`stream.ts\` passes it (grep), a present-but-wrong header is never given a second chance via the cookie, and the empty-token refuse-closed path runs before the cookie is ever consulted.

**This is as far as I should take it.** I dispatched and verified the underlying work, which makes me the wrong person to approve it — GitHub would refuse a self-approval attempt on this account anyway, but the point holds regardless of who owns the account. This PR needs someone who didn't write any of \`prd29\` to actually look at it before it goes anywhere near \`main\`.

gabriel-canaan (approved):
## Review of #93 — read commit-by-commit, everything below EXECUTED unless marked

Reviewed at `220a64e`, in a clean detached worktree off `f3fe770` (the merge-base with `main`). Fixes pushed as `9806c7c` and `cd16370`; head is now `cd16370`.

### What verified in your favour first

The security boundary this PRD exists to draw is real, and the laws that protect it bite. I did not take the PR body's word for any of this — six mutations, each reverted after:

| # | mutation | result |
|---|---|---|
| 1 | `/api/rotate`'s gate given `{ allowCookie: true }` | **RED** — `gated-reads.test.ts` › "a real gated MUTATION — POST /api/rotate — refuses a request bearing only the cookie": `expected 200 to be 401` |
| 2 | `/api/stream`'s `preHandler` deleted | **RED** — 4 failures across `gated-reads.test.ts` and `route-class-law.test.ts`, incl. the gate-presence law |
| 3 | the `try/catch` from `220a64e` removed from `readCapabilityCookie` | **RED** — `expected 500 to be 401` |
| 4 | `reply.header('Set-Cookie', …)` deleted from `server/static.ts` | **RED** — 2 failures in `static.test.ts` |
| 5 | a 15th `gated-read` row added to `ROUTE_CLASSES` with no contract test | **RED** — 2 failures in `contract-coverage-law.test.ts` |
| 6 | `packages/contract/src/meta.contract.test.ts` deleted | **RED** — 2 failures, both axes |

Mutation 1 is the one that matters most, and it is the strongest thing in this PR: the CSRF refusal is asserted against the **real** `/api/rotate`, not only the synthetic gate — so "no mutation honours the cookie" is enforced, not merely arranged. `allowCookie: true` appears at exactly one call site in the whole tree (`packages/server/src/api/stream.ts:157`); I checked.

Two more things I looked at specifically and found correct:

- **Vacuity check on the new `static.test.ts` cases.** `it('never sets the capability cookie on a non-HTML asset')` fetches `/assets/app.js`, which the `beforeEach` really writes — the neighbouring test asserts it answers `200` with `text/javascript`. So the `toBeUndefined()` is not passing because the request 404s.
- **The read axis's parser self-test** runs `gatedReadRoutesIn` itself, over a sample interleaving all four route classes, rather than a re-implemented copy. That is the discipline `routesIn` already held, extended.

Gates, run here at `220a64e` before any of my changes: `npm run typecheck` → 0, `npm run lint` → 0, `VITEST_MAX_WORKERS=6 npm test` → 0 (394 files, 6616 passed, 3 expected-fail). `gh pr checks 93`: all 7 legs pass across both jobs (`build-test-boot` ×3, `pack-smoke` ×4).

---

### Finding 1 — the boundary moved, four descriptions of it did not (EXECUTED: fixed in `9806c7c`)

`5421401` (#59, wave 2a) updated `SECURITY.md` and `docs/architecture.md` when it gated `/api/meta` and `/api/doctor`. `4e569cf` (#60, wave 2b) did the structurally identical thing to `/api/stream` one commit later and updated neither. **This is the sibling case your own PR body asks about, and it arrived inside this PR** — one lane did the documentation half, the next lane doing the same work did not, because #60's fence named neither file.

At `220a64e`, four descriptions contradicted the table they describe:

- `SECURITY.md:180` — *"What is not gated yet … `/api/stream` is deferred to a later wave (issue #60)"*. This is the one that concerned me: it is the public security posture of a public repo, and it **understates** the boundary — it told a reader the live event stream was open when `ROUTE_CLASSES` already classed it `gated-read`.
- `docs/architecture.md:146` — *"Thirteen reads are gated … One is still tokenless: `/api/stream` is ruled to gate and not yet gated"*. The paragraph's own closing sentence is *"never this paragraph's tense"*; it rotted in the next commit of the same PR.
- `packages/server/src/api/index.ts:70` — the `RouteClass` doc: *"the reads prd-29 defers to wave 2 (`/api/meta`, `/api/doctor`, `/api/stream`)"* — seventy lines above the three rows that now read `gated-read`. #58's fence named the stale comment above the tokenless block; this is its sibling one screen up, in the same file, and it survived.
- `packages/web/src/connect/meta.ts:362` — *"`/api/meta` and `/api/doctor` gate in wave 2 and ignore the header today"*, written directly above the default that now sends them the header.

**Why your own verification could not have caught this:** nothing pins these four against `ROUTE_CLASSES`. I checked — no test in `packages/` or `scripts/` reads `SECURITY.md` at all. Typecheck, lint and 6616 tests are green with the drift in place; I confirmed that by running them.

Fixed in `9806c7c`, prose only. The follow-up worth having is the one #61 just built for code: a law pinning those two documents' route lists against `ROUTE_CLASSES`, the same way the read axis is pinned. Deliberately **not** smuggled into this PR — it is new scope and belongs on its own issue.

### Finding 2 — `buildContractHarness` lost the doc comment that explains it (EXECUTED: fixed in `cd16370`)

`fd92150` (#61) inserted `HARNESS_LIVE_SESSION_ID` and its own doc block *between* `buildContractHarness`'s doc comment and the function. It compiles and the suite is green either way, but TSDoc and every editor hover bind a block to the declaration immediately below it — so the function showed no doc, and the const showed only its own.

The orphaned block is not incidental prose: it is the harness's anti-vacuity rationale (*"a server that stopped stamping would fail the happy-path test but leave every refusal test passing vacuously on the client's own missing-token throw"*). That is exactly what a future author needs when they are about to edit that function, and it was the one place it no longer appeared. `cd16370` moves the const above it; text is byte-identical, only its position changed.

---

### Your call — offered, not asserted (REASONED, nothing run)

1. **`contract-coverage-law.test.ts:~290`** — the comment above `expect(allGatedReads).toHaveLength(14)` reads *"the number is derived from the parse, never hardcoded as a vacuity floor ruling 3 forbids."* The **left** side is derived; the `14` is a hardcoded pin. That is the right assertion (drift goes red, which mutation 5 proves), but the sentence claims something about its own line that is not true of it, and this repo treats a law's self-description as load-bearing. A word — "pinned against a literal that must be updated deliberately" — would settle it.
2. **The `swallows` branch checks `expect(source).toContain('401')`** — a bare substring over the whole file, so a `401` in a comment satisfies it. The companion `h.app.inject(` requirement carries most of the weight, so this is not a hole today; it is just the weakest of the four conditions.
3. **#60's two named open questions** (does the cookie rotate on `/api/rotate`; dev-mode under `vite dev`) are not decided anywhere in the diff — correctly, since the issue forbids deciding them. Reading `api/rotate.ts`, `/api/rotate` rotates the *session*, not the capability token, so the first is moot in practice; worth one sentence on the issue saying so, before someone re-derives it. #61 records the `vite dev` blind spot in the read law's doc comment, which covers the second for the contract package but not for the cookie itself.

### What I did NOT run

`scripts/gate.sh` — that is the operator's landing step, not a reviewer's. I ran the three commands a lane runs (typecheck, lint, capped suite) at `220a64e`, at the intermediate fix, and again at `cd16370`: **0 / 0 / 0**, 394 files, 6616 passed, 3 expected-fail each time. I did not run the suite under 4× concurrency, and I did not exercise anything on Windows or macOS beyond what CI already ran.

### Verdict

The security work is sound and the laws protecting it are load-bearing — six mutations say so. Both findings were documentation, both are fixed on the branch, and neither changes behaviour. **Approved on the code; the landing is the operator's, not mine** — this is still a draft and you asked that it not merge without a second pair of eyes, which is what this is. `gh pr checks 93` should be re-read against `cd16370` before it lands.

ciaran-slow:
**Board note, not a review comment.** While tidying the project board I found that this PR's four issues — #58, #59, #60, #61 — are **closed and were marked Done**, while none of their commits is on `main`: they are the eleven commits this branch carries, and each closure comment says so accurately (*"landed in PR #84, merged into prd29"*).

Done was therefore the one status that could not be true, and it made the `prd29` milestone read **0 open, 4 closed** — a completed PRD with nothing shipped. All four are now in **In review**, where they actually are: built, and waiting on this PR. They go to Done when it merges.

Same shape prd-44 hit and paid two days for, and it is now written into `AGENTS.md`'s Landing section (`50c8804`) — *a green gate on a branch whose base is not `main` has landed nothing* — with the one-command check: `git merge-base --is-ancestor origin/prd29 origin/main`. Nothing here is wrong; this PR is the route to `main` that prd-44's four PRs never had.

ciaran-slow (approved):
## Rhizo re-review — approved

The original security review remains sound, and the stale branch is now integrated with current `main` without a force-push.

### EXECUTED

- Read all eleven commits in order and the linked issues #58–#61.
- Merged `origin/main` into `prd29`; only `static.ts` and `static.test.ts` conflicted. The resolution keeps main's descriptor-owned static reads and PR #93's capability cookie on both root and SPA-fallback HTML responses.
- Targeted static/security/gated-read/route-class suites: 72 passed, 6 platform skips.
- Full post-merge gate: `npm run typecheck`, `npm run lint`, and `VITEST_MAX_WORKERS=6 npm test` — 7,108 passed, 3 expected failures, 6 platform skips.
- Fresh CI on merge commit `d0a3bc5`: all seven jobs passed, including macOS plus both Ubuntu Node versions and all pack-smoke jobs.

### Mutation

Removed the `Set-Cookie` header from the merged static HTML path. Both real-response cookie tests failed red (root and SPA fallback); restored the line and the complete static-server suite passed.

### Not run

I did not run `scripts/gate.sh`; it is the operator landing path. The Rhizo landing helper will perform its own final base check and post-merge gate before merging.

Issue #92, opened 2026-08-25
prd42 w8: the in-flight test hook is enforced by the law, not by a comment


gabriel-canaan:
## The defect

`packages/server/src/recorder/rotate.ts:434` exports `reserveInFlightForTest`, which writes an arbitrary reservation into the module's in-flight `WeakMap`:

```ts
export function reserveInFlightForTest(recorder: SessionRecorder, kind: string): () => void {
  const boundary: InFlightBoundary = { kind: kind as InFlightBoundary['kind'], promise: new Promise<Rotation>(() => {}) }
  inFlight.set(recorder, boundary)
  ...
}
```

Three properties, together:

- it takes `kind: string` and **casts** it into the union, so it can reserve a kind the type system does not admit;
- it calls `inFlight.set(...)` **unconditionally** — it does not check whether a reservation is already active, so it can **overwrite a live one** belonging to a real rotation or retarget in progress;
- it is a **plain export** from a production module. The only thing preventing production use is its doc comment: *"No production code may import this."*

Verified at filing: no production file imports it — `grep` outside `*.test.ts` finds only the definition. So this is a door standing open, not a defect in flight.

## Why it exists, because the reason is good

It was added by #49 so `rotateSession`'s guard could be **proven fail-closed against a kind nobody has written yet**, rather than trusting that property until a real third caller appears and either confirms or breaks it. That is exactly the right instinct — the test it enables is what makes `kind !== 'rotation'` meaningful rather than decorative, and it caught a fail-open guard.

The problem is not the hook. It is that the hook is enforced by a sentence.

## Why this is not charged to #49

#49 introduced it, but the module's own law is what should be holding this line, and that law is separately broken: `ROTATION_ENTRY_RE` (**#87**) enumerates rotation's entry points by name and already misses two real doors. A fix that derives the guarded set from the module's exports — which is #87's definition of done — would cover this export automatically. Fixing it by adding one more name to a list would repeat the very defect #87 is about.

## Definition of done

- Production code cannot reach `reserveInFlightForTest` — enforced, not requested. Options, in rough order of preference: move it behind a test-only entry point the namespace law already governs; have the law derive its set from exports (**#87**) so this is caught by construction; or, at minimum, make it refuse to overwrite an active reservation and reject a `kind` outside the union.
- Whatever is chosen, the test #49 wrote — a rotation refused against a third, unwritten kind — must still be possible and must still pass. **Do not close this door by deleting the test's ability to prove fail-closed.**
[...]

Issue #91, opened 2026-08-25
prd42 w8: the shared refusal code is one compiler-bound fact, not three copies


gabriel-canaan:
## The defect

The 409 refusal code shared by two routes is **one fact stored twice**, and the two copies are guarded by two different mechanisms — only one of which is the compiler.

- `api/retarget.ts` writes `code: 'retarget-in-flight' satisfies RetargetRefusalCode`
- `api/rotate.ts` writes the bare literal `code: 'retarget-in-flight'`
- `api/rotate.test.ts` holds a third copy, `SHARED_RETARGET_IN_FLIGHT_CODE`, which *is* compiler-bound

EXECUTED, during the verification of prd42 w4:

| mutation | caught by |
|---|---|
| rename the union member `RetargetRefusalCode` | **compiler** — `rotate.test.ts(33,61): error TS1360` |
| drift `api/rotate.ts`'s production literal | **runtime test only** — `npm run typecheck` reports **0 errors** |

So both drift directions *are* covered and nothing is currently silent. This issue is not a hole; it is a **second spelling of one fact** in the PRD whose title is *one path, one spelling*, held together by a comment and a runtime assertion instead of a type.

## Why it was not fixed in #49

The obvious fix — import the type into `api/rotate.ts` — is blocked: a namespace law forbids `api/rotate.ts` importing from `api/retarget.ts`, and its static check does not distinguish `import type` from a value import. #49 worked around that by putting the compiler check in the test file, which is exempt from that clause, and documented precisely what that does and does not bind.

That workaround is correct and honest. It is also why the fix belongs here rather than there: making this a single compiler-bound fact needs a **new module**, which is outside #49's fence.

## Definition of done

- The refusal code is one value, in a module both routes may import without violating the namespace law — a neutral home such as `api/refusals.ts` or the existing shared types module, whichever the law permits.
- Both routes import it; neither carries a bare literal.
- The test's private copy is deleted — it exists only to compensate for the missing shared value.
- EXECUTED: mutating the value in its new home fails `npm run typecheck`, and no route can drift independently because there is nothing left to drift.
- The comment in `api/rotate.ts` explaining the two-mechanism arrangement goes with it; leaving it would describe machinery that no longer exists.
- prd-42 **ruling 5**'s "no second refusal vocabulary" becomes a compiler guarantee rather than a convention.

## Fence (may touch ONLY)
- `packages/server/src/api/refusals.ts` (NEW — the neutral home)
[...]

Issue #87, opened 2026-08-25
prd42 w8: the rotation-entry law sees and enforces every door


gabriel-canaan:
## The defect

`ROTATION_ENTRY_RE` in `packages/server/src/recorder/namespace-law.test.ts:165` enumerates rotation's entry points by name:

```
/\b(?:rotateSession|retargetSession|closeCurrentSession|openNextSession)\b/
```

It misses **`beginRetargetBoundary`** and **`performRetarget`** — which are the doors `api/retarget.ts` actually uses:

```
packages/server/src/api/retarget.ts:6    import { beginRetargetBoundary, performRetarget } from '../recorder/rotate.js'
packages/server/src/api/retarget.ts:173        const boundary = beginRetargetBoundary(ctx.recorder)
packages/server/src/api/retarget.ts:246        retarget = await performRetarget({
```

So a file outside the module could import `performRetarget` and drive a full close/open without tripping the law that *"rotation is reachable only from an explicit operator command"*. The law's own doc comment states the intent it is failing to enforce — it explains that `retargetSession` was added because *"it is a third door into the same two halves, and the law that only the declared route may knock is the same law either way."* The fourth and fifth doors arrived with prd42 w2 (#14) and the enumeration did not follow them.

## Why this is filed rather than fixed in prd42 w4

Found by an independent review seat during verification of the prd42 w4 bundle. **That branch neither creates nor widens this hole** — the law skips test files, so #49's new beginRetargetBoundary import in api/rotate.test.ts is clean, and no production file gained a call. Charging a pre-existing gap to a branch that did not cause it is how a fence gets widened for reasons nobody agreed to, so it is recorded here instead.

## The shape, because it is the recurring one

This is an **enumeration standing in for a class**: a guard that lists the names it knows, in a module whose whole point is that only one caller may knock. Every new entry point is a silent hole until someone remembers to add it — and the two that are missing were added by the very PRD that most recently reasoned about this law. prd-46 (`the class, not the enumeration`) is the same thesis in `gate.sh`, and its ruling 1 is the pattern to copy: recognise the property structurally rather than by spelling.

## Definition of done

- A file outside the recorder module that imports **any** production entry point into rotation's two halves fails the law — including entry points added after this issue is closed.
- Prefer deriving the set from the module's own exports over extending the literal list. If the list stays, the DoD is that adding an export without adding it to the list fails something.
[...]

PR #94, opened 2026-08-25
prd42 w4: a rotation refuses a retarget's boundary, and the served page reads what it checked

### LINKED ISSUE #49: prd42 w4: a rotation refuses a retarget's boundary

### LINKED ISSUE #51: prd42 w4: the served page's symlink verdict is pinned

gabriel-canaan:
prd-42 wave 4 — two refusals that were being handed the wrong answer. One commit per issue, fences disjoint.

Closes #49
Closes #51

**Note on shas:** cherry-picked from two lane branches onto current `main`.

## What becomes true

- **`d27969c` (#49)** — a rotation asked while a **retarget** is in flight refuses with 409 instead of being handed the retarget's boundary. Two concurrent rotations still coalesce, unchanged.
- **`ec85e06` (#51)** — the served page's sibling-symlink verdict is decided, stated and pinned; and the route no longer re-resolves the path it already checked, so the read cannot be swapped out from under the containment decision.

## #49 — a 200 that was a lie to its caller

`rotateSession` and `retargetSession` share one in-flight map keyed by recorder. Retarget refused a second caller; rotation **coalesced** — so `POST /api/rotate` during a retarget returned **200 carrying a `Rotation` whose `closed.reason` was `'retargeted'` and whose `opened` session was in a different repo than the caller asked about.**

Rotation's coalescing rationale is an argument about *two rotations* — "two operators asking at once is ONE boundary, not two" — and does not carry to a repo move. Governed by **prd-42 ruling 5**: refuse with 409; the rejected alternative (wait-then-rotate) queues, against this PRD's *refuse, never queue*.

The in-flight entry now carries `kind`. Two things worth naming:

- **The guard is fail-CLOSED**, `kind !== 'rotation'`, not `kind === 'retarget'`. The first version named the dangerous value, so a *third* operation added to that map would have silently inherited the exact defect this issue closes — verified by adding a `'lab-checkpoint'` kind and watching it coalesce. A test now refuses a kind nobody has written yet.
- **The 409 code is bound**, not conventional. A namespace law forbids `api/rotate.ts` importing from `api/retarget.ts`, so the compiler check lives in the test file. The comment states precisely what that buys: renaming the union member fails `typecheck`; drifting this literal fails the runtime assertion. Two mechanisms, named as two, because only one is the compiler. **#91** files the single-fact fix.

## #51 — a decision, then the race that decision exposed

prd42 w2 (#13) replaced a prefix comparison with the shared containment primitive, and in doing so moved a path that is textually **outside** the dist root but resolves **inside** it from **403 to 200**, untested either way. **Ruling 6** required the outcome be chosen, stated and pinned — expressed through the containment layer, never a re-added string comparison.

**The decision is 200**, with the reasoning in the file: `isPathContained` canonicalizes both sides, and after canonicalization this request reads the same bytes a direct request would; refusing would need either a string check (ruling 2 forbids it) or teaching the primitive that some canonicalized-inside paths are outside.

**Then verification found the decision blessed a racy path**, and closing it took three rounds:

1. The route checked containment and then canonicalized the same string a **second, independent time**. A seat drove `GET /..%2f<dist>-sibling/secret.txt` with a swap between the two and got **200 with an outside file** on the permanently tokenless `GET /*` (ADR-0012, prd-29 ruling 1).
2. Canonicalize-once fixed *that input*. It did not fix the class: `filePath` is a **path string**, so `existsSync`/`statSync`/`readFileSync`/`createReadStream` each re-resolved it. Replacing a **component** of it — `dist/assets` → a symlink — still won. Reduced to the mechanism: same canonical path, `REAL APP` before, `STOLEN SECRET` after.
3. The route now **opens once** with `O_NOFOLLOW`, `fstat`s the **descriptor**, and streams from the **descriptor**. `root` is canonicalized once at registration rather than per request.

**And the residual is closed rather than declared.** The remaining window — an intermediate directory swapped between the check and the open — cannot be closed by asking *"where will this path resolve?"*, because Node has no `openat(2)` (verified: no `AT_*` constants, no fd-relative calls). It can be closed by asking a different question: **where did this descriptor actually land?** `readlink /proc/self/fd/<fd>` answers authoritatively for an fd already pinned to an inode, and that answer is fed **into `isPathContained`** — not compared by hand.

**This is a platform-conditional guarantee and the code says so.** `/proc/self/fd` is Linux-only and Node exposes no `fcntl(F_GETPATH)`, so on platforms without it the older behaviour remains and the exposure is still there. That fallback is **tested**, not assumed, and the comment scopes the claim to "closed on Linux" rather than "closed".

## Verification

Gates on this merged branch, node 22.22.2: `typecheck` 0 · `lint` 0 · `build` 0 · `VITEST_MAX_WORKERS=6 npm test` 0 — **381 files, 6590 passed**, 3 expected fail.

Four rounds, **three of them independent-seat passes** (Claude Opus and Codex `gpt-5.6-terra` at xhigh, isolated worktrees, blind to each other), plus a fix re-review of the amended head each time.

Every guard in this PR is mutation-proven on the merged tree, each mutation asserted to have actually applied before its result was believed:

| mutation | result |
|---|---|
| #49 guard back to fail-open | 1 red — *refuses a kind nobody has written yet* |
| #51 delete the `/proc` containment check | 1 red — *closes the intermediate-directory swap on LINUX* |
| #51 read by path instead of the descriptor | 2 red — *a swap after the open still streams the ORIGINAL bytes* |
| #51 drop `O_NOFOLLOW` | 1 red — *refuses when the FINAL component becomes a symlink* |

Plus: no descriptor leaks on any exit path (403, directory, `index.html` fallback, 404, `.html` read, stream), the static suite clean under 4× concurrency, and the pinned 200, six refusals, `ENOTDIR` and `ELOOP` paths all intact.

## What is the sibling case?

**It was the answer four times running, and that is the story of this PR.** Each round's finding was a structural sibling of the *previous round's repair*, not of the original defect:

- the fail-open guard named the dangerous value instead of the safe one, so the *next* kind inherited the bug;
- canonicalize-once closed the input we tested and left the class;
- the descriptor rewrite was correct and **nothing forced it to stay** — reverting to a path read left all 23 tests green, so the tests were strengthened until they bit;
- and I twice reported "no test covers this" from a `grep` that a mutation immediately disproved.

Residual siblings are filed, not left in a comment: **#87** (the rotation-entry law enumerates four doors and misses two real ones), **#91** (the refusal code is three copies), **#92** (the in-flight test hook is enforced by a comment).

None of the four above. The strongest is #51's: **deleting the Linux close, the descriptor read, or `O_NOFOLLOW` each reddens a different test**, so no single regression can quietly restore the race.

## Fence widenings

None. `d27969c`: `recorder/rotate.ts`, `api/rotate.ts` + their tests. `ec85e06`: `server/static.ts` + its test. Proven disjoint by `scripts/fence-lint.sh 49 51` before dispatch. #51's *promise* widened mid-flight — from "pin a verdict" to "pin a verdict and close the race it exposes" — recorded on the issue before the change, per the working agreement.

## Behaviour changes, stated as ruling 6 requires

- `GET /*` **refuses** a request whose descriptor lands outside the dist root, on Linux. Previously it served it.
- `GET /*` **serves** a sibling symlink resolving into the root — the pinned decision, unchanged from `main`'s current behaviour but now deliberate and tested.
- `POST /api/rotate` returns **409** during an in-flight retarget where it previously returned 200 with another operation's boundary.

## Reviewer note

I requested every fix in this PR, so I should not approve it. The fix-review gate — seats re-reviewing the amended head rather than only the original commits — is what caught the descriptor rewrite not being load-bearing, and is worth applying here too.


gabriel-canaan:
## Head updated — the Linux race is now closed, and the last fix was a deletion

The branch has moved from `ec85e06` to `2efa453`. Everything below the fold in the original description still stands; this is what changed and why it took three more rounds.

### Where it ended up

`GET /*` now: canonicalizes `root` **once at registration**, canonicalizes the request path once, opens the file **once** with `O_NOFOLLOW`, `fstat`s the **descriptor**, and streams from the **descriptor**. Then — on Linux — it asks the kernel where that descriptor actually landed (`readlink /proc/self/fd/<fd>`) and compares the answer to the canonical root by **path segments, with no filesystem call of any kind**. Where `/proc` is absent the older behaviour remains, and that fallback is tested rather than assumed.

### Three more rounds, and each one broke the previous fix

Independent seats (Claude Opus and Codex `gpt-5.6-terra`, xhigh, isolated worktrees, blind to each other) reviewed the amended head after every round.

1. **The `/proc` answer was fed to `isPathContained`** — which calls `realpathSync.native`, **re-resolving the very answer that made the check race-free**. A seat rebound the reported directory between `readlink` and the check and served an outside file at 200. *Cause: an instruction of mine, citing ruling 2. Ruling 2 governs asking the filesystem whether a path is inside a root; the kernel had already answered for this fd, so comparing its answer is string work.* Replaced with `isUnderCanonicalRoot` — a pure segment comparison.
2. **Deleted-ness was inferred from the `" (deleted)"` suffix**, so an outside file *named* `x (deleted)` skipped the check entirely and was served 200. Found independently by **both** seats. Tightened to also require `fstatSync(fd).nlink === 0`.
3. **Both conditions could then be satisfied honestly** — suffix from the filename, `nlink 0` from a real unlink timed between `readlink` and `fstat`. Bypass again, 200.

**Round 6 deleted the branch instead of guarding it a fourth time.** The skip's own justification was re-resolution — true when the answer went through `isPathContained`, false once the comparison touched no filesystem. And the `" (deleted)"` marker lands on the **last** segment, which the root comparison never examines. So the legitimate case it protected passes without it, proven in both directions: a genuine in-root file unlinked mid-request still serves its correct original bytes, and the bypass flips to 403.

The `fstat` that leaked a descriptor on `EIO` is gone with the branch — not wrapped, removed.

### Verification on this head

node 22.22.2: `typecheck` 0 · `lint` 0 · `build` 0 · `VITEST_MAX_WORKERS=6 npm test` 0 — **381 files, 6595 passed**, 3 expected fail. `static.test.ts` alone: **33 passed** (was 21 when this PR opened).

Five mutations on the merged tree, each asserted to have actually applied before its result was believed:

| mutation | result |
|---|---|
| #49 guard back to fail-open | 1 red |
| #51 disable the Linux close | **5 red** |
| #51 re-canonicalize the `/proc` answer | 2 red |
| #51 read by path instead of the descriptor | 3 red |
| #51 drop `O_NOFOLLOW` | 1 red |

Three directions pinned explicitly: a file merely **named** with the deleted-suffix is refused; the same name **genuinely unlinked mid-request** is still refused; a **genuine in-root file** unlinked mid-request still serves its correct bytes. Plus no descriptor leaked on any exit path, and the static suite clean under 4× concurrency.

### What is still not closed, stated plainly

On platforms without `/proc/self/fd` — macOS, which this repo gates a CI leg for — the intermediate-directory window remains. Node exposes no `openat(2)` (verified: no `AT_*` constants, no fd-relative calls) and no `fcntl(F_GETPATH)`, so there is no portable equivalent. The fallback is declared in the code, tested, and scoped: the claim is "closed on Linux", never "closed".

### The lesson worth keeping

**Six rounds, and every round's finding was a sibling of the previous round's repair.** Two of those repairs came from instructions of mine that were subtly wrong. The round that finally held was the one that *removed* a condition — the mutation count went up while the code got smaller. The three preceding rounds each added a guard, and each addition became the next attack surface.

Residuals filed rather than left in comments: **#87**, **#91**, **#92**.

gabriel-canaan:
## Final state — head `5a523d6`, CI green including macOS

Three rounds happened after the last update (`2efa453`). Two of them came from **CI**, not from review — worth recording, because the verification loop that found everything else could not have found these.

### Round 6 — the fix was a deletion

The `" (deleted)"` skip had become the bypass. Its own justification was re-resolution — true when the descriptor's path was fed to `isPathContained`, false once the comparison became a pure segment check that touches no filesystem. And the marker lands on the **last** segment, which the root comparison never examines. So the legitimate case it protected passes without it.

Removed, and proven in both directions: a **genuine** in-root file unlinked mid-request still serves its correct original bytes; the malicious-filename bypass flips to **403**. The `fstat` that could leak a descriptor on `EIO` went with the branch — not wrapped, gone.

Three rounds had each *added* a condition and each addition became the next attack surface. This one removed one, and the test count went **up** while the code got smaller.

### Round 7 — macOS was red, and it was a real test bug

Six tests failed on the macOS leg. Not a production defect: **`os.tmpdir()` is a symlink there** (`/var` → `/private/var`) and is not on Linux, so the tests' injected `realpath` matched only the raw path and never fired. That is precisely the path-shape class `AGENTS.md` says the macOS leg exists to catch.

Fixed properly rather than skipped — the helper now matches raw **and** canonical. `O_NOFOLLOW` and the fd-hygiene tests, which are not Linux-specific, run on **every** platform and pass there.

### Round 8 — three ungated siblings

Three of the Linux-only assertions were still ungated while three others had been gated in the same round. Same property, same file — the sibling case, this time in the test gating rather than the code.

Now derived rather than enumerated: **7 of 35 tests are platform-gated**, and the fallback exists in **both** forms — SIMULATED on Linux, **REAL** on macOS — so each platform exercises the path that is actually real for it. The single Linux `skipped` is the macOS-real variant.

### Verification on this exact head

- **No drift:** every file in both lanes' commits is byte-identical to the bundle (4 files / 2 files).
- **Fences exact**, no widening.
- **Gates:** `typecheck` 0 · `lint` 0 · `build` 0 · **381 files, 6601 passed**, 3 expected fail, 1 skipped.
- **CI: 7/7 green, macOS included.**
- **Concurrency:** both touched suites 4× concurrent, 0/4 red.
- **Every guard bites**, each mutation asserted to have actually applied before its result was believed:

| mutation | result |
|---|---|
| #49 guard back to fail-open | 1 red |
| #51 disable the Linux close | 5 red |
| #51 re-canonicalize the `/proc` answer | 2 red |
| #51 read by path, not the descriptor | 3 red |
| #51 drop `O_NOFOLLOW` | 1 red |

### What is still not closed

On platforms without `/proc/self/fd` the intermediate-directory window remains. Node exposes no `openat(2)` (no `AT_*` constants, no fd-relative calls) and no `fcntl(F_GETPATH)`, so there is no portable equivalent. The code claims "closed on Linux", never "closed", and the fallback is exercised for real on the platform where it applies.

### For the reviewer

Eight rounds. **Every round's finding was a sibling of the previous round's repair**, not of the original defect — and two of those repairs came from instructions of mine that were subtly wrong (feeding the fd-truth path back through a re-resolving primitive; "do not strip the suffix", which produced a skip-the-verdict branch). The rounds that held were the ones that *removed* something.

Two findings came from the **macOS leg** rather than from any review pass, mine or the seats'. Every verification round I ran was on Linux; the platform leg was doing work nothing else in the loop could.

I requested every fix in this PR, so I should not approve it.

Residuals filed rather than left in comments: **#87**, **#91**, **#92**.

ciaran-slow (approved):
## Approve — and the macOS half of this PR is now EXECUTED, not REASONED

Reviewed commit-by-commit in a clean worktree off `origin/prd42-w4-refuse-and-pin`
(`5a523d6`), node 22.23.2, on **real macOS** (darwin 25.5.0).

The commit body for `5a523d6` says plainly: *"I have no macOS access… REASONED, not
executed — that the six gated-Linux-only tests would in fact fail on real macOS."*
That was the one open gap in an otherwise fully-executed PR, so that is where I spent
the review. **It holds.**

**EXECUTED, macOS, `packages/server/src/server/static.test.ts`:**
`29 passed | 6 skipped (35)`, exit 0. The six skipped are exactly the six Linux gates.

**EXECUTED — all four mutations from the PR table reproduce on macOS**, each anchor
diff-confirmed applied before the result was believed, each reverted and confirmed
clean after:

| mutation | Linux (author) | macOS (this review) |
|---|---|---|
| `#49` guard back to fail-open (`=== 'retarget'`) | 1 red | **1 red** — *"refuses a rotation when the in-flight slot holds a THIRD kind nobody has written yet"* |
| `#51` drop `O_NOFOLLOW` | 1 red | **1 red** — *"O_NOFOLLOW refuses when the FINAL path component becomes a symlink"* |
| `#51` read by path instead of the descriptor | 4 red | **4 red** — same four, including the round-6 unlinked-file test the author flagged as an honest extra rather than silently claiming a match |
| `#51` delete the `/proc` containment check | 5 red | **0 red** — correct, and see below |

So the descriptor rewrite and `O_NOFOLLOW` are **load-bearing on macOS too**, not just
on Linux. That was the thing most worth confirming, and it is confirmed.

**EXECUTED — the mirror check, forced.** I ungated all six Linux-only tests
(anchor asserted `n == 6`) and ran them on macOS: `5 failed | 30 passed`. Five
genuinely fail without `/proc`, exactly as reasoned — the gates are real, not
cargo-culted.

**EXECUTED — full gates on the merged tree, macOS:** `npm run typecheck` exit 0 ·
`npm run lint` 849 files, no fixes, exit 0 · `VITEST_MAX_WORKERS=6 npm test` →
**381 files, 6596 passed | 3 expected fail | 6 skipped, exit 0.**

**Fence audit clean.** Branch touches exactly the six declared files, disjoint per
commit, one commit per issue — matching #49's and #51's fences with nothing widened.

**Sibling case for `#49`, checked rather than assumed.** `rotateSession` now *rejects*
where it previously resolved, so I grepped every caller: exactly one in production
(`api/rotate.ts:59`), which catches `RotationRefusedError` and rethrows everything
else. No second caller inherits an unhandled rejection. `recorder/index.ts` re-exports
`rotateSession` without `RotationRefusedError` — but it already omitted
`RetargetInFlightError` the same way before this PR, so that is symmetric and
pre-existing, not something #49 introduced. (It is also outside #49's fence.)

**CI checked, not assumed green.** `gh pr checks 94` — all 7 jobs executed and passed:
3 × `build-test-boot` (ubuntu current/min, macos current — macos×min excluded by
design) and 4 × `pack-smoke` on the full 2×2. That is the whole expected grid.

---

### One correction, non-blocking — no change requested

`5a523d6`'s bucket list puts *"the same intermediate-directory swap still gets served
when `/proc/self/fd` is unavailable — SIMULATED, on Linux"* under **"Genuinely
Linux-only (expected outcome requires `/proc`)"**.

**EXECUTED: that one passes on macOS.** It is the single test out of six that did *not*
fail when I ungated them (30 passed = 29 normal + this one). Its assertion does not
require `/proc`; it simulates an absence macOS already has.

**Consequence: none, and no fix is needed.** Its mirror-pair partner — the `REAL, on
this platform where /proc genuinely does not exist` half — *does* run on macOS and
asserts the same outcome unmocked, so macOS coverage of the fallback is intact. And
the **code comment at `static.test.ts:475-487` already states the correct reason** —
"don't simulate a condition macOS already has… each exercising the shape real for ITS
platform". The gate is right; only the commit body's bucket *label* for it is off.
That is in the message, so fixing it would mean a force-push of someone else's branch.
Not worth it. Recorded here instead.

### Worth naming for the future, EXECUTED

Deleting the `/proc` containment check entirely leaves macOS **completely green**
(0 red, above). That is exactly what the code claims — "closed on Linux", not
"closed" — and the honesty of that scoping is the right call. The consequence to
carry forward: **only the two ubuntu legs enforce that check.** Both are present in
`ci.yml`, so the gate is real today; just don't let a future CI trim assume the macOS
leg covers it.

`isUnderCanonicalRoot` is a second containment comparison living in `server/static.ts`
— the one file `paths/prefix-comparison-law.test.ts` holds to zero violations. The
justification in its doc comment is sound and I agree with it: re-resolving the
kernel's already-resolved answer through `isPathContained` reopens the exact race the
check closes, and the law is self-documented as a *syntactic proxy* for ruling 2, not
the rule itself. `paths/containment.ts` was explicitly out of #51's fence, so this was
the only place it could go. Might be worth a follow-up issue to consider relocating it
next to the other containment primitives, where a reader would look for it. Not a
blocker and not a change request.

---

### What I did NOT run

Windows (no leg exists in `ci.yml` for this grid), and I did not re-execute the
author's Linux-side mutation results — the ubuntu CI legs and the author's own
diff-confirmed runs stand for those. Everything labelled EXECUTED above I ran myself
on macOS and read the exit status of.

Approving. Merge is yours to press.

PR #95, opened 2026-08-25
docs(prd): prd-42 sequences the three issues verification left behind

gabriel-canaan:
Sequences the three issues prd42 wave 4's verification left behind. **Document
only — one file, no code.**

`#87`, `#91` and `#92` were filed against territory prd-42 already owns and none
carried a wave. Unsequenced, each collided with something: derived, not counted
by eye, `scripts/fence-lint.sh 47 49 51 52 53 87 91 92` reports **15** overlaps
across the eight open issues, **6** of which set one of the three against a
wave-4-to-7 issue.

- **Wave 8 — `#87` alone, and it does not wait.** Fence is
  `recorder/namespace-law.test.ts`, one file. PR #94 (wave 4, in review) touches
  six others — `api/rotate.ts`, `api/rotate.test.ts`, `recorder/rotate.ts`,
  `recorder/rotate.test.ts`, `server/static.ts`, `server/static.test.ts`.
  Disjoint, so wave 4 is not a live fence for it. Parking `#87` behind an
  unrelated review out of caution would queue a one-file change for no reason a
  fence can name.
- **Wave 9 — `#91` + `#92`, after wave 4.** Both claim a file #94 is actively
  amending — a live fence, and bundling across one is forbidden. They are
  disjoint from each other (`api/*` vs `recorder/*`), so one wave and one PR:
  `scripts/fence-lint.sh 91 92` PASSED, zero overlaps. `#92` additionally
  follows `#87` — both edit `recorder/namespace-law.test.ts`, so they are a
  stack.

No ruling and no earlier wave is renumbered. Waves 5–7 keep their order and
their open question about reconciling the `#47`/`#52`/`#53` fences.

**Does this affect the live wave 4?** No. Checked before writing: PR #94's file
list carries no `docs/` path, so this branch and that one cannot conflict. `#49`
and `#51` were not edited, reassigned or re-boarded by this grooming.

**What is the sibling case?** The three issues were the ones a reader would
notice. The sibling risk is a PRD that renumbers to make room — so nothing here
edits waves 1–7 or any ruling; the amendment is appended, which is the only
lawful shape for a changed plan in this corpus.

**What mutation would this test survive?** No test covers a PRD. What stands in:
every count was re-derived with the full lint command, and `claim-lint.sh`
confirms 0 unresolved references. Two counts in the first draft were wrong — 13
and 3, read off a truncated `head -20` — and the corrected paragraph says so in
the file, because a number typed instead of measured is this repo's most-named
failure.

**Fence widenings.** None. One file:
`docs/prds/prd-42-one-path-one-spelling.md`.


gabriel-canaan:
Closing to respect the standing rule: **no PR is opened before `/verify` has run against its head.** This one went straight from commit to PR without an independent pass, and "it is only docs" is exactly the rationalisation the rule exists to close — CI green is no evidence about a PRD at all.

The branch `prd42-sequencing-waves-8-9` is untouched and the commit stands. It will be verified and re-opened.

It also needs a substantive revision before it comes back. The amendment's central claim is that wave 8 is `#87` alone and **"it does not wait"**. The operator has since ruled the opposite — `#87` is held and ships with the wave rather than alone — so the document would be landing a plan that was already overruled.

PR #98, opened 2026-08-25
prd40 w3: the route reads the maintained fold, and the record says what a replay may contain

### LINKED ISSUE #5: prd40 w3: /api/meta answers from the maintained fold
## Fence (may touch ONLY)

## Blocked by
Blocked by prd40 w1 — needs `foldSoFar()`.

Sibling lanes own `packages/server/src/recorder/session-recorder.ts` and `packages/server/src/server/poll-loop.ts`. `packages/server/src/api/index.ts`'s route classes are owned by prd-29 — this route stays `read`, and gating it here would break `rhizomorph env`, `doctor` and the CI boot smoke mid-milestone.

## Definition of done
- No `reduceAll(eventsSoFar())` call remains on the request path.
- The response is byte-identical to today's for the same event stream.
- A spy law asserts fold work per request is O(1) in the session's length.
- The route's `routeClass` is unchanged.

### LINKED ISSUE #81: prd40 w3: architecture.md says what a replay may now contain

ciaran-slow:
prd-40 wave 3, assembled from two lane branches. Two commits, one per issue, each standing alone.

Closes #5
Closes #81

Wave 1 (#3, in #67) made the recorder maintain its fold; wave 2 (#4, in #79) put the append ahead of publication. This wave is the first consumer of that fold, and the document that says what a replay may now contain.

---

## Assembly

| commit | issue | files |
|---|---|---|
| `5059733` | #5 | `packages/server/src/api/meta.ts` (+12), `packages/server/src/api/meta.test.ts` (+251) |
| `8761acc` | #81 | `docs/architecture.md` (+45) |

Cherry-picked onto `origin/main` in issue order. **No conflict** — which is the evidence the fences were genuinely disjoint, since within a wave nothing depends on anything else and a cherry-pick collision would mean two issues touched one file.

**Gated as an assembly, not just as lanes** (node v22.23.2): `npm run typecheck` ✓ · `npm run lint` (biome, 849 files) ✓ · `VITEST_MAX_WORKERS=6 npm test` → **381 files, 6584 passed, 3 expected fail, 0 failed** ✓ · `npm run build` ✓. The `packages/web/src/scene/**` flake did not appear. Each lane additionally gated alone at `VITEST_MAX_WORKERS=3`, and #5's lane ran a boot smoke that served a live `/api/meta`.

## Fence widening

**#5 widened to `packages/server/src/api/meta.test.ts`.** The issue's declared fence is `api/meta.ts` alone, but its Definition of done requires a spy law and names no file it could live in. Recorded here per `AGENTS.md`. #81's fence is unchanged (`docs/architecture.md`).

---

## #5 — `/api/meta` answers from the maintained fold

`buildLadderManifest` (`api/meta.ts:156`) reads `recorder.foldSoFar()` instead of `reduceAll(recorder.eventsSoFar())`. `reduceAll` dropped from the core import; JSDoc updated. `api/index.ts` untouched, so the route stays `read` — prd-29 owns route classes, and gating it here would break `rhizomorph env`, `doctor` and the CI boot smoke mid-milestone.

Per-request fold work is now constant in session length. The cost removed is the repo's own published figure: 113.5 ms at 25k events, 535.3 ms at 55k, on an ungated route the dashboard polls.

**The wave-3 contract with #69 is honoured.** #69 may narrow `foldSoFar()` to `Readonly<SessionState>` or deep-freeze it in dev. This lane treats the return as read-only unconditionally — never mutated, never copied on the request path, never passed to anything typed for a mutable `SessionState` — so both outcomes are safe and neither lane had to wait.

**What is the sibling case?** `packages/server/src/cli/run.ts:207` is the other `reduceAll(eventsSoFar())` on a non-test path, and it is **deliberately unchanged**. It folds once inside `runServer` at boot, not per request, so it is not the defect this issue names. Verified still true at assembly. **Corrected after review** (`gabriel-canaan`, item 1): the original wording here said the remaining `eventsSoFar()` callers "pass the array along without folding". That is over-broad. `api/lab.ts` folds on a request path — `readAllEvents` (`:88`) walks the session dir and appends `ctx.recorder.eventsSoFar()` (`:96`, `:105`), then `reduceAll` runs at `:112`, `:132` and `:205`, reached from `/api/lab/checkpoints`, `/api/lab/experiments` and `/api/lab/estimate`. It is still not this issue's defect — it folds across *all* sessions, so `foldSoFar()` (live session only) cannot substitute for it — but the request path is not fold-free, and no reader of this PR should conclude that it is. See the note on #5 for the full disposition.

**What mutation would this test survive?** Five laws added; the mutation table was run and **two rows came out weaker than the plan predicted, reported rather than papered over**:

| mutation | plan predicted | actual |
|---|---|---|
| revert to `reduceAll(eventsSoFar())` | T2 red | ✅ T2 red, **and only T2** — the spy is bound to the call site |
| return `initialSessionState()` | T1 + T6 red | T1 red; **T6 cannot go red here** — T6's session is empty, and for an empty session `initialSessionState()` *is* the correct fold |
| `folded.eventCount = 0` | T3 + T4 red | T4 red; **T3 stayed green** — a constant write is idempotent, so three reads still agree |

For the third, the drift-shaped variant `folded.eventCount -= 1` was additionally run and T3 went red, proving it bites the defect it names. No law was substituted or weakened; the over-claim is in the plan's table, not in the tests. Worth correcting on #5 before verify reads that table as a contract.

## #81 — `architecture.md` says what a replay may now contain

A decisions-log entry dated 2026-08-25 citing **ADR-0029** by number, plus a "What a replay may now contain" paragraph in prd-17 ruling 3's integrity section. Both state the guarantee **per path** — at-least-once on the poll path, at-most-once on the other eleven `recorder.record` sites — because a sentence that omits that is wrong in a way readers act on. Both reasons there is no read-side dedupe are named, and the cost ADR-0029 accepted is named so it is citable: seven of fifteen era-1 families fold a repeated fact into corrupted state, and one duplicated `llm.usage` moves that corpus's reported spend +7.7%.

The exemption clause is included: the degrade `collector.error` reporting an append failure may be emitted without having been appended, and it is the only event that may. That rests on prd-40 success 1's *"One named exception, and no other"*, confirmed landed on `main` in #86.

**No ADR was edited and no new ADR written** — the log is append-only, and this records an existing decision rather than making one.

**What is the sibling case?** This *is* the issue: the defect is a decision recorded in one place and contradicted in another, so the question is not "was the citation added" but "does a second document make the same stale claim". Swept `docs/` and `README.md`: eight hits, **none a stale exactly-once claim**.

`architecture.md` never made an explicit exactly-once claim, so the added paragraph is additive prose, not a correction — and that null result is the finding, not a reason to invent a paragraph to correct. Hits examined and left, with reasons: `:2146` (`upcast()`'s "a home every event already flows through" — the fold's entry point, not delivery), `:1575` (`trace.span` idempotent re-delivery), `:2154` (the fold-order resolution), `:1726` ("one line per event"). All consistent with ADR-0029.

One out-of-fence hit reported rather than followed: `docs/user-guide/sessions.md:15` ("same file, same collector offsets, no duplicated history") is about *resume*, so true at the granularity it claims, but a strict reader could take it as the log containing no duplicates. Worth its own issue.

**What mutation would this test survive?** Docs-only; no law added. All four gates were run anyway because the repo has doc laws that read tracked documents (`no-personal-paths-law.test.ts`, `runbook-delivery-law.test.ts`), and every path, link and in-page anchor cited was checked to resolve on disk — which is the whole point of an issue about a citation that pointed at nothing.

---

## Deviations and things a reviewer should know

- **Both lanes ran at `VITEST_MAX_WORKERS=3`**, not the usual 6, because they built concurrently on an 8 GB box. The assembly above was gated at 6.
- **Both lanes needed node pinned to v22.23.2.** A fresh worktree's default v24.12.0 is below the `engines` floor (`npm install` refuses on `jsdom@30`), and an incomplete `node_modules` reddened `packages/web/src/theme/tokens.test.ts` until `npm install` ran under 22. `package-lock.json` unchanged by either lane.
- **#81's issue body carries a false premise, harmlessly.** It says "ADR-0011 is cited there" — `grep "ADR-0011" docs/architecture.md` returned nothing before this commit; no ADR was cited from the decisions log at all. It did not change the required work, so the lane did not stop.
- **One in-fence staleness deliberately left:** `architecture.md:8-9` says the ADR log runs "0001 through 0027 today" — 0028 and 0029 exist. A one-token fix, but not a delivery claim and outside both DoD bullets, so it stayed out of a one-issue commit. Worth filing.
- **Verify has not run on either issue.** Both carry build reports; neither has a verify report from a second model. Opened at the operator's instruction so review can start — flagged here rather than left for a reviewer to discover.


gabriel-canaan (approved):
## Review of #98 — read commit-by-commit, every claim below EXECUTED unless marked

Reviewed at `8761acc`, in a clean detached worktree off `739eeb1` (= `origin/main` exactly; the base is not stale). One fix pushed as `1d717f5`; head is now `1d717f5`.

### What verified in your favour first

**Your mutation table is accurate, including the two rows you reported as weaker than planned.** I re-ran them rather than taking the report, and both self-corrections hold:

| mutation | you reported | I measured |
|---|---|---|
| revert to `reduceAll(eventsSoFar())` | T2 red, **and only T2** | ✅ exactly that — 1 failed / 21 passed, the O(1) spy law alone |
| `folded.eventCount = 0` | T4 red, **T3 stayed green** | ✅ T3 ("three requests agree") green — a constant write is idempotent; 4 others red |
| `folded.eventCount -= 1` | T3 red | ✅ T3 red; 8 red in total |

Reporting a law as weaker than the plan predicted, rather than quietly substituting a law that would have matched, is the thing that made this PR quick to trust. The over-claim is in the plan's table, as you say — not in the tests.

**The risk this change actually introduces is covered, and I checked it with a mutation you did not run.** Before this commit `buildLadderManifest` held a private `reduceAll` result; after it, it holds the recorder's live object. So I made the route mutate what it was handed — `folded.eventCount += 1` after `foldSoFar()`:

```
✗ three requests in a row give the same answer — a fold a reader mutated would drift on the second read
✗ serving the fold does not corrupt it — the recorder still agrees with reduceAll afterwards (#69 contract)
… 8 failed | 14 passed
```

Both of the laws written specifically for that risk fire. Your half of the #69 contract is proven, not just asserted.

**The recorder invariant this depends on holds.** `foldSoFar()` is only sound if every `buffer.push` is paired with an `advanceFold`. I checked all of them: `session-recorder.ts:67/68` (constructor), `:130/131` (`record`), `:177/178` (`closeWith`), and `openSession` resets both together at `:201/202`. Plus `foldDesynced` self-heals from `buffer`. The substitution is safe.

**Every citation in #81 resolves.** This is an issue about a citation that pointed at nothing, so I ran the checks rather than trusting them:

- `docs/adr/0029-a-recording-may-repeat-a-fact.md` exists.
- The in-page anchor `#ruling-3--recordings-never-rot-the-integrity-laws-landed` matches the heading at `docs/architecture.md:2117` exactly, double hyphen and all — and it is unique despite three `### Ruling 3` headings in the file.
- `server/poll-loop.ts:223` is precisely `await recorder.record(event)`, with `snapshots.set` after it.
- **"the other eleven `recorder.record` sites" is exactly right.** Twelve in non-test source: `api/otel.ts` ×6, `lab/fork.ts` + `lab/checkpoint.ts` = 2, `cli/run.ts`, `recorder/rotate.ts`, and `poll-loop.ts` ×2 — one the poll path (`:223`), one the degrade path (`recordOrDegrade`, `:162`). 12 − 1 = 11, and your enumeration names the same eleven.
- ADR-0029 carries "seven of the fifteen event families" (`:40`, `:77`), "+7.7%" (`:54`) and "1-in-100" (`:76`, `:152`). `createIdFactory` (`core/src/events/index.ts`) is a per-process counter — `n += 1` — so "the two copies differ" is true.
- prd-40 success 1's *"One named exception, and no other"* is at `docs/prds/prd-40…:52`, and its landing in #86 is recorded at `:307-310`.

**I re-ran your #81 sweep independently** and got your null result. The only genuinely arguable hit is `docs/user-guide/sessions.md:15` — which is the one you flagged yourself as out of fence and worth its own issue. Everything else my sweep turned up is a non-delivery use ("`requireCapabilityToken` appears exactly once across…", ADR-0028's "at most once per process lifetime", prd-35's "appears exactly once"). Not inventing a paragraph to correct when the sweep came back empty was the right call.

Gates at `8761acc`: `npm run typecheck` → 0, `npm run lint` → 0, `VITEST_MAX_WORKERS=6 npm test` → 0, **381 files / 6584 passed / 3 expected fail** — matching your reported counts exactly.

---

### Finding 1 — this PR and #93 are each green alone and red together (EXECUTED: fixed in `1d717f5`)

**Not a defect in your work.** #5's fence and DoD are honoured as written, and all five laws pass on `main`. This is a collision between two open PRs, and neither author can see it from their own branch.

#93 (prd-29 wave 2a, #59) moves `GET /api/meta` from `read` to `gated-read`. #5's issue body states the opposite — *"this route stays `read`, and gating it here would break `rhizomorph env`, `doctor` and the CI boot smoke mid-milestone"* — which was true when #5 was groomed and stopped being true when prd-29 wave 2a widened all three of those consumers in the same commit that gated the route. **You were right to build to your issue; the issue went stale underneath you.**

Nothing warns anybody, because the branches merge cleanly:

```
$ git merge-tree --write-tree origin/prd29 8761acc
dab04efa7cd7fb8b88acb6cb9760c445ef9fc404      # exit 0, no conflict
```

But on the merged tree, four of your five new laws go red — they inject `/api/meta` with no headers and get 401:

```
✗ the body is byte-identical to the one the per-request re-fold produced
✗ fold work per request is O(1) — reduceAll is never reached, at 5 events or at 55
✗ serving the fold does not corrupt it (#69 contract)
✗ an empty session answers
  Tests  4 failed | 18 passed (22)
```

The fifth is worse than red: **"three requests in a row give the same answer" stays green by comparing three 401 bodies to each other.** A law that cannot fail is the shape this repo treats as a defect in itself.

`1d717f5` adds `headers: capabilityHeaders(app)` to the eight new injections. `capabilityHeaders` already lives on `main` (`api/test-support.ts`, since prd-29 wave 1 — **not** something #93 introduces), and an ungated route ignores a header it does not read, so it is order-independent. Verified both ways rather than argued:

- at your head, `/api/meta` still `read` → **22/22 pass**
- merged with `origin/prd29`, `/api/meta` `gated-read` → **22/22 pass**

**And I checked that my own repair did not weaken anything**, since `capabilityHeaders(app)` now runs inside the `vi.spyOn(core, 'reduceAll')` window. It reads `app.capabilityToken` and nothing else. Re-running both key mutations against my amended head: reverting the route to `reduceAll(eventsSoFar())` is still **1 failed / 21 passed, the O(1) law alone**; the live-fold mutation is still **8 failed / 14 passed**. Full suite on `1d717f5`: **381 files / 6584 passed**, typecheck 0, lint 0.

Rung 2 — I fixed it, but it turns on a cross-PR judgement rather than a typo, so it wants your word before anyone merges.

---

### Your call — offered, not asserted (REASONED, nothing run beyond the greps quoted)

1. **One sentence in the PR body is over-broad**, though the judgement under it is right. *"The remaining `eventsSoFar()` callers pass the array along without folding"* — `api/lab.ts:112` does fold: `readAllEvents` collects every session on disk **plus** `ctx.recorder.eventsSoFar()`, then `reduceAll(events)`, on the `/api/lab/*` request path. It is genuinely not this issue's defect — it is a fold across all sessions, so `foldSoFar()` (live session only) cannot substitute for it — and your "prd-40's unfiled work implied covers them" is the right disposition. Only the blanket "without folding" is inaccurate. Worth a line on #5 so the next reader does not conclude the request path is now fold-free.

2. **The sibling case you did name checks out.** `cli/run.ts:207` folds once after `pollLoop.tick()` to print the watching line — per boot, not per request. Correctly left alone.

3. **`LADDER_COLLECTOR_NAMES_FOR_TEST` and `DECLARED_CAPABILITIES_FOR_TEST` are hand-copied** from `meta.ts:122/124`. Forced, not sloppy — neither is exported, and exporting them would widen the fence — and a drift between the two lists fails loudly rather than silently (the route's body would carry a capability the test's does not, and `toEqual` catches it). Noting it only so the next person does not "tidy" it into an import without also widening the fence.

4. **Two things you flagged yourself and I agree are worth issues, not blockers:** `docs/user-guide/sessions.md:15`'s "no duplicated history", and `architecture.md:8-9`'s "0001 through 0027 today" now that 0028 and 0029 exist.

5. **#81's issue body's false premise** ("ADR-0011 is cited there") — confirmed: `git show 739eeb1:docs/architecture.md | grep -c ADR-0011` is 0. Right call not to stop on it.

### What I did NOT run

`scripts/gate.sh` — the operator's landing step, not a reviewer's. I did not run the suite at 4× concurrency, and nothing on macOS or Windows beyond what CI ran. **Neither issue has a verify report from a second model**, exactly as you flagged; this review is not one — it is a reviewer's pass, and it did not re-derive the ADR-0029 corpus figures from the corpus itself, only confirmed the document says what you cite it as saying.

### Verdict

Both commits stand on their own, the laws are load-bearing under mutation, and the documentation claims check out line by line. **Approved on the code.** The landing is the operator's, and there is one thing that must be decided before it happens: #98 and #93 both touch `api/meta.ts`, `api/meta.test.ts` and `docs/architecture.md`. With `1d717f5` in place either order is safe — but re-read `gh pr checks` against `1d717f5`, not `8761acc`, and whichever lands second should merge `main` in and re-run rather than trusting the earlier green.

ciaran-slow:
Thanks — the review's Finding 1 is verified here independently, not taken on report, and every step of it holds.

**Finding 1 — reproduced exactly.** `origin/prd29` does gate the route (`api/index.ts:142`, `routeClass: 'gated-read'`). On the merged tree **without** `1d717f5`: **4 failed | 18 passed**, the same four laws you named, and *"three requests in a row give the same answer"* passes by comparing three 401 bodies — the law-that-cannot-fail shape. **With** `1d717f5`: **22/22**. `capabilityHeaders` is on `main` at `api/test-support.ts:19` from prd-29 wave 1, so the repair is order-independent as you argued. **The fix is accepted.**

One note for anyone re-checking this: reverse-applying `1d717f5` also removes the `capabilityHeaders` **import**, which prd-29's own merged version of the file uses 18 times — so a naive revert reds all 22 tests for a compile reason and buries the real signal. Restore the import and the clean 4 appears.

**Item 1 — you are right, and it is bigger than the sentence.** Corrected in the PR body above, with the full disposition on #5. `api/lab.ts` folds on a request path: `readAllEvents` (`:88`) → `reduceAll` at `:112`, `:132`, `:205`, from `/api/lab/checkpoints`, `/api/lab/experiments` and `/api/lab/estimate`. Your reasoning for why it is not #5's defect is the right one and stands — it folds across all sessions, so `foldSoFar()` cannot substitute.

What the sweep additionally turned up: `readAllEvents` also calls `listSessions(ctx.sessionDir)` and re-parses every finished recording **per request** (`:89`, `:97`). That is **prd-44 ruling 1**'s stated extension verbatim — *"a second caller that walks the directory per request is the same defect with a different route name"* — and prd-44 #30 has already landed the cross-session parse cache that would serve it. So this one path carries both costs, and the better home is prd-44 rather than prd-40's "unfiled work implied". Left unfiled pending the operator's call.

**Items 2, 3 and 5** — agreed, no action. On item 3, the note is on #5 so nobody "tidies" the hand-copied lists into an import without widening the fence.

**Item 4 — already filed**, just before your review landed: #99 (`architecture.md:8-9`'s "0001 through 0027") and #100 (`sessions.md:15`'s "no duplicated history"). Both boarded under `prd40`, Soon/Low.

**Still open, as you flagged:** neither issue has a verify report from a second model, and this review does not claim to be one. That is the operator's call, along with the merge order — with `1d717f5` in place either order is safe, and whichever lands second should merge `main` in and re-run rather than trusting the earlier green.

PR #103, opened 2026-08-25
prd44 w2: tmux's per-pane captures happen at once

### LINKED ISSUE #36: prd44 w2: tmux's per-pane captures happen at once
## Fence (may touch ONLY)
- `packages/server/src/collectors/tmux/` (all files, including deletions)

## Blocked by
- **#4** (prd40 w2) — it lands `server/poll-loop.ts`'s snapshot/append ordering. This lane must not race it; the poll loop is prd-40's territory until then.
- **#33** (prd44 w1) — the bounded fan-out helper this lane consumes.

Sibling lanes hold the other two collectors and `recorder/session-recorder.ts`. **Do NOT touch `server/poll-loop.ts`, `server/exec.ts`, `collectors/resilience.ts`, `collectors/resume-reconcile.ts` or `server/collector-loader.ts`** — the wrappers and the loop are shared, and polling the six collectors concurrently was measured and rejected (1.3x / 0.6x / 1.4x). The win is inside one collector.

## Definition of done
`tmux capture-pane` per pane and `resolveWorktreePath` share one sequential `for` loop (`collector.ts:153-163`), purely to compute a content hash.

- Per-pane captures run concurrently at the stated ceiling, and a law asserts the bound.
- `nextPanes` and the emitted `pane.discovered`/`pane.activity` events keep `list-panes` order, so a replay is byte-identical.
- The `worktreeByPath` memo still avoids a repeat `resolveWorktreePath` for a path already resolved — concurrency must not turn one lookup into N.
- A pane whose capture fails still carries its previous hash forward, and the skipped-pane rule (#242: one bad `list-panes` line costs that line, never a false closure) is untouched.
- **Assert a count, never a wall clock.** And do NOT name any new law `*.bench.test.ts`, nor mark it `// @gate-timing` — `scripts/gate.sh` derives its 4x-load timing set from exactly those and ratchets `.swarm/timing-count`.

**Note:** this box has zero tmux panes, so the loop is unexercised here — the fixtures are the only evidence available and must carry it.

hchristinasmith:
Closes #36

## What changed

`tmux capture-pane` per pane and `resolveWorktreePath` used to share one
sequential `for` loop in `collector.ts`, purely to compute a content hash.
Both now fan out through #33's bounded fan-out helper (`mapBounded`,
`COLLECTOR_FANOUT_LIMIT`) instead of running one pane at a time.

- **Worktree resolution** dedupes unique `currentPath`s into a list *before*
  the fan-out starts (a synchronous `Set`-based pass), not check-then-fan-out
  per pane — which would race. So concurrency can't turn one lookup into N
  when several panes share an unresolved path in the same poll.
- **Pane capture** runs all panes through `mapBounded` at once.
  `mapBounded`/`runBounded` return outcomes at each task's own original
  index regardless of settle order, and the final assembly loop stays a
  plain synchronous pass over `panes` (list-panes order) indexing into the
  outcome arrays — so `nextPanes` and the emitted `pane.discovered` /
  `pane.activity` events stay in `list-panes` order (byte-identical replay)
  even though the underlying execs run concurrently.
- A pane whose capture outcome comes back failed — either an
  `ExecResult.failed` (tmux itself reported failure) or the fan-out helper
  catching a thrown exception — falls back to the previous content hash via
  the same branch, so both sibling failure shapes are handled identically.
- The skipped-pane rule (#242: one bad `list-panes` line costs only that
  line, never a false closure) is untouched — that loop wasn't touched.

## Fence

Only `packages/server/src/collectors/tmux/` was touched, per the issue.
`server/poll-loop.ts`, `server/exec.ts`, `collectors/resilience.ts`,
`collectors/resume-reconcile.ts`, and `server/collector-loader.ts` were not
touched.

## What is the sibling case?

A capture failure can arrive two ways — a normal `ExecResult.failed: true`,
or the fan-out helper catching a thrown exception (`outcome.ok === false`).
Both are synthesized into the same `{ failed: true }` shape and fed through
one `captureResult.failed ? prevHash : hash(...)` branch, so neither slips
through the other unhandled.

- The new "resolves a worktree path shared by several panes... exactly
  once" test fails (three `git` calls instead of one) if the pre-fan-out
  `Set` dedup is removed — confirmed by an independent reviewer agent who
  reverted it and watched the test go red, then restored it.
- The new "keeps list-panes order... even when captures settle out of
  order" test deliberately makes the second-listed pane's capture resolve
  before the first-listed pane's, and asserts `nextPanes` keys and event
  order both stay in `list-panes` order — it would fail if the assembly
  loop iterated over outcome arrays instead of `panes` itself.
- The new "runs capture-pane at most COLLECTOR_FANOUT_LIMIT panes
  concurrently" test asserts a max-in-flight **count**, not a wall-clock
  duration — no new test file is named `*.bench.test.ts` and none carries
  `// @gate-timing`.

## Note on live verification

This dev box has zero tmux panes, so the concurrent loop is unexercised in
a live run here. The fixtures and tests above are the only evidence and are
meant to carry that proof — nothing here was observed running against a
real tmux server.

## Review

Dispatched a fresh adversarial subagent against the working diff before
committing (this repo has no `.claude/agents/reviewer.md`). It read the
diff and `concurrency.ts`, ran the full test/typecheck/lint gate, and
verified the dedup and ordering claims by reverting pieces of the fix and
confirming the relevant tests went red, then restored the repo. No defects
found; recommended GO.

## Gate

- `npm run typecheck` — clean
- `npm run lint` — clean
- `VITEST_MAX_WORKERS=6 npm test` — 381 files, 6583 passed, 3 expected fail (pre-existing, unrelated)


hchristinasmith:
Pausing prd44 work per operator instruction — focus is prd29 only for now. Reopen when prd44 resumes.

Issue #99, opened 2026-08-25
prd40: the ADR log's stated range covers every record in it


ciaran-slow:
## Fence (may touch ONLY)
- `docs/architecture.md`

## Blocked by
Nothing. Independent of #98 (prd-40 wave 3), though found by it — that PR's fence is the same file, so this waits for it to land rather than racing it.

## The gap

`docs/architecture.md:8-9` says the decisions ADRs record live in `docs/adr/` — "**0001 through 0027 today**".

`ls docs/adr/` has 0028 and 0029. ADR-0029 is precisely the one prd-40 wave 3 exists to cite, so the sentence that points a reader at the ADR log understates it by the two records most likely to be wanted.

Found in-fence during #81 and deliberately left out of that commit: it is a one-token fix, but it is not a delivery claim and it sits outside both of #81's DoD bullets, so it did not belong in a one-issue commit.

## Why it is worth an issue rather than a drive-by

`docs/architecture.md:8-9` is the sentence that tells a new reader — human or agent — that **the ADR log is the register of record and this document is not**. A stated range is a completeness claim; a reader who takes it literally concludes the log ends at 0027 and never opens the two records that changed the replay contract.

The same sentence has gone stale before — this is what a hardcoded count does.

## Definition of done
- `architecture.md:8-9` no longer states a range that a new ADR falsifies. Either it derives the range or it stops stating one — **prefer stopping**: a range is re-stale the next time an ADR lands, and the log itself is one `ls` away.
- If a count survives anywhere in the file, `docs/architecture.md` is checked for other hardcoded ADR/ruling counts in the same pass. **A sweep that finds none is a finding too** — say so rather than inventing one.
- No ADR is edited. The log is append-only.

## Not in scope
Deriving counts mechanically across the whole docs tree. #23 (`prd43 w4: every count a document states is derivable`) already owns that shape; this is the one instance that is currently false.

ciaran-slow:
**One more entry for this lane, decided while planning #26.** That issue adds **ADR-0030** (`docs/adr/0030-the-alarm-may-outrun-the-record.md`) and its `docs/adr/README.md` index row, but deliberately does **not** open `docs/architecture.md` — that is this issue's fence, and a live-fence collision over a one-line decisions-log entry is not worth it.

[...]

PR #105, opened 2026-08-25
prd40 w4: a dropped append speaks in the degrade voice

### LINKED ISSUE #26: prd40 w4: a dropped append speaks in the degrade voice

ciaran-slow:
prd-40 wave 4. **A single-issue wave** — `recorder/session-recorder.ts` is claimed by #26, #69 and #80, and three issues cannot share one file in one wave, so they run in consecutive waves ordered by Timeline and Priority. This is the first (Now/High). Branch is `26-degrade-voice` rather than a `prd40-w4-*` bundle branch, per `AGENTS.md`: `<issue>-<slug>` is the convention for a single.

Closes #26

Wave 2 (#4) made the poll loop **quiet** — `record()` appends before it publishes, so a rejected append reaches no subscriber. This makes it **loud**. prd-40 **success 2**.

---

## The defect

On a full disk, every channel that should tell the operator fails:

1. A collector's event append rejects inside `runTick`'s try (`poll-loop.ts:223`), caught at `:229`.
2. The catch builds a `collector.error` and calls `recordOrDegrade`.
3. `recordOrDegrade` called `record()` — which appends first and **publishes nothing when the append fails**. The alarm reached no subscriber.
4. It fell back to `console.error`, which `api/lab.ts:484-488` replaces process-wide for the duration of a lab fork.

So in the exact case the alarm exists for, the operator could learn nothing and the dashboard went on looking healthy while nothing was being recorded.

## The change

`SessionRecorder.recordAlarm(event: EventOf<'collector.error'>): Promise<{ appended: boolean }>` — appends when it can, emits to subscribers either way, never rejects. `recordOrDegrade` calls it and keeps the operator's `console.error` line as a second channel beside the emit (prd41 #10 owns the stderr override that can silence it).

**The exemption is for emission only, and that is the part most likely to be got wrong.** prd-40 success 1 permits the degrade `collector.error` to be *emitted* without having been appended — and separately forbids "leaving the fold ahead of the file after a rejected append", which is **not** carved out. So on a failed append the alarm enters neither `buffer` nor `foldState`: `eventsSoFar()` and `foldSoFar()` stay exactly as honest as the file, and only subscribers hear it. `buffer.push` and `advanceFold` sit strictly inside `if (appended)`.

**The narrowing is load-bearing, not decoration.** Success 1 requires the exemption be "asserted **by name** in the law rather than inferred from a category". A parameter typed `EventOf<'collector.error'>` makes that a `tsc` property — no collector-derived event can reach the method — and law R6 holds it with a `@ts-expect-error` that reports `TS2578: Unused '@ts-expect-error' directive` if the narrowing is ever widened.

**ADR-0030** (`docs/adr/0030-the-alarm-may-outrun-the-record.md`) records the named hole and its cost: **a live stream may now contain a `collector.error` that a replay of the same session does not** — the converse of ADR-0029, which permitted the record to hold a fact twice. Four rejected alternatives are named: a public emitter, a boolean option on `record()`, retrying the append, and doing nothing.

## Gates

Gated as the tree that will land, on node v22.23.2, base confirmed at `93977c2` (current `main`, not stale): `npm run typecheck` ✓ · `npm run lint` (849 files) ✓ · `VITEST_MAX_WORKERS=6 npm test` → **381 files, 6594 passed, 3 expected fail, 0 failed** ✓ · `npm run build` ✓. Boot smoke: `npm start -- --port 0` served. No `packages/web/src/scene/**` flake.

## Fence widenings

The issue's fence (`poll-loop.ts`, `poll-loop.test.ts`, `session-recorder.ts`, `session-recorder.test.ts`) was already widened and recorded on the issue on 2026-08-25 — the DoD is unbuildable from `poll-loop.ts` alone, since `record()` publishes nothing on a failed append and the emitter is private. **Three further widenings, recorded here:**

- `docs/adr/0030-the-alarm-may-outrun-the-record.md` (new) and `docs/adr/README.md` (one index row) — the ADR is a deliverable of the issue.
- `packages/server/src/server/collector-loader.test.ts` (+6) — an inline recorder fake that needed `recordAlarm` or it dies on the degrade path. Fake only; no assertion touched.

`docs/architecture.md` is **deliberately not opened** — ADR-0030's decisions-log citation belongs to #99, whose fence that file is, and a live-fence collision over one line is not worth it. `main` holds an uncited ADR for a short window, exactly as it held ADR-0029 between #77 and #81. Noted on #99, sequenced after this lands.

---

## What is the sibling case?

**`persist`'s degrade call and `runTick`'s are structurally identical** — `poll-loop.ts:178` and `:244` both build a `collector.error` and hand it to `recordOrDegrade`. A fix applied to one and not the other is precisely the defect shape this repo keeps finding, so the fix is in `recordOrDegrade` itself and **law P3 drives the snapshot-save path specifically** rather than trusting that it shares a code path.

The other sibling worth naming is the one deliberately **not** changed: `record()` and `closeWith()` keep their existing publish semantics. The exemption is for one event type on one path, and widening it to either would be the thing ADR-0030 exists to prevent.

The table was run, and **two rows over-claimed — reported rather than papered over**:

| mutation | plan predicted | actually red |
|---|---|---|
| delete the emit on failed append | R2, P1, P3 | R2, R4, R5 — **P1/P3 green** |
| push/advance fold on failed append too | R2, R5 | ✅ exactly |
| `recordAlarm` rethrows on append failure | R3, P4 | R3, R2, R4, R5 — **P4 green** |
| `recordOrDegrade` calls `record` again | P1, P3 | P1, P2, P3 |
| memoise the failing alarm (emit once) | R5 | ✅ R5 only |
| drop the `console.error` line | P0, P2 | ✅ exactly |

Both over-claims have one structural cause: `poll-loop.test.ts` drives `as unknown as SessionRecorder` **fakes**, so a mutation inside the real `SessionRecorder` cannot reach the P-laws. The recorder half (R2/R3/R5) and the poll-loop half (P1/P2/P3) are each completely pinned; no single mutation crosses the seam. The laws are sound — the plan's table assumed a reach it never had.

**P0 was written first and seen green against unmodified `poll-loop.ts`** (1 passed / 28 skipped). `recordOrDegrade`'s fallback had never been asserted — `grep -rn "failed to report" packages/server/src` returned only the source line, and the two existing tests that drive it assert "the loop did not crash" and nothing more. Changing an uncovered path makes the change unfalsifiable, so it was covered first.

## Deviation from the plan

**Law R4 could not pass as the plan worded it, and the design was kept rather than quietly changed.** The plan asked for "a second subscriber registered *after* the thrower still received the event", while prescribing `closeWith`'s single `try`/`catch` around `emit`. Node's `EventEmitter.emit` abandons the listener list at the first throw — proved with a throwaway script, not argued:

```
second listener ran times: 0
```

R4 is therefore written at the strength the prescribed design provides: a surviving subscriber registered *before* the thrower still receives the alarm, `recordAlarm` resolves rather than rejects, and the subscriber's error is reported. Per-listener isolation would be a mechanism change, not a test fix.

**That gap is real and is not this issue's** — `record()` and `closeWith()` have the same property, so one throwing SSE subscriber can blind every subscriber registered after it, for every event type. Pre-existing, class-wide, and worth its own issue rather than a silent widening here.


gabriel-canaan (approved):
Reviewed commit-by-commit at `7a98894`, in a clean worktree off `93977c2` (current `main`, so the base claim checks out). Everything below is labelled **EXECUTED** or **REASONED**; nothing here is asserted from reading alone.

**Every gate claim in the PR body reproduces exactly.** EXECUTED on node v22.22.2:

| | claimed | measured |
|---|---|---|
| `npm run typecheck` | ✓ | exit 0 |
| `npm run lint` | 849 files ✓ | 849 files, exit 0 |
| `VITEST_MAX_WORKERS=6 npm test` | 381 files, 6594 passed, 3 expected fail | **identical**, exit 0 |

**The mutation table is honest, including where it over-claims.** I re-ran it rather than trusting it, and the two rows you flagged as over-claims are over-claims for exactly the reason you gave — `poll-loop.test.ts` drives `as unknown as SessionRecorder` fakes, so no mutation crosses the recorder/poll-loop seam. Reporting that rather than quietly trimming the table is the right call, and it made this review faster.

- *delete the emit on failed append* → EXECUTED, red on **R2, R4, R5**; P1/P3 green. As you said.
- *push/advance fold on a failed append too* → EXECUTED, red on R2 and R5 only. The load-bearing half — "the exemption is for emission only" — is genuinely pinned. This was the claim I most wanted to break and could not.
- *drop the `console.error` line* → EXECUTED, red on P0 and P2 exactly.

**The sibling case is handled at the right level.** Putting the fix in `recordOrDegrade` itself rather than at the two call sites, and then driving `persist`'s path specifically with P3 rather than trusting the shared code path, is the thing this repo keeps failing to do.

**The R4 deviation is correct and correctly reported.** `EventEmitter.emit` does abandon the listener list at the first throw — REASONED from node's semantics, and your throwaway script is the right way to have settled it. Writing the law at the strength the design provides, and filing the per-listener isolation gap as pre-existing and class-wide rather than widening this issue, is the right split.

**ADR-0030 is a real ADR**: five options, four rejected with reasons, and the consequence stated as a cost (`a live stream may now contain a collector.error that a replay does not`) rather than a benefit. Index row present at `docs/adr/README.md:233`. Deferring the `docs/architecture.md` citation to #99 rather than colliding with a live fence is the right trade.

## One finding — fixed in `4b4dd6c`, not handed back

**`recordAlarm`'s two concurrency guards shipped with no law of their own.** EXECUTED, and this is a mutation, not an argument:

```
# both guards deleted from recordAlarm:
#   - while (this.sealed !== null) await this.sealed
#   - if (this.writer !== writer) return { appended: true }
VITEST_MAX_WORKERS=6 npm test
  Test Files  381 passed (381)
       Tests  6594 passed | 3 expected fail (6597)
  exit=0
```

Both guards are **correct** — they're copied faithfully from `record`, comment and all. The gap is that the copies are unasserted while the original is not: `record`'s identical guard has been held by a law since the suspension point was introduced (`session-recorder.test.ts:214`, "does not publish into the next session when a rotation lands mid-append"). `recordAlarm` is the one method in this class permitted to publish without appending, so it is the last one whose rotation and seal behaviour should rest on reading the source.

Your mutation table stops at the emit and the fold, which is where the interesting semantics are — but it never mutates the guards, so the table cannot see this. Three laws added, source untouched:

| law | mutant it kills |
|---|---|
| parks in the sealed window rather than appending behind `session.closed` | drop the seal wait → the alarm writes behind a closed log's final line, which prd17 ruling 1 makes structural and ADR-0030 does **not** carve out |
| does not publish into the next session when a rotation lands mid-append | drop the rotation guard → a closed session's alarm enters the new session's buffer, fold and subscribers |
| still speaks when a rotation lands mid-append and the append failed | hoist the guard above `if (appended)` → a racing rotation silences a failed alarm, with nothing going red |

Each is killed by **exactly one** mutant and by no other — EXECUTED, all three verified in both directions (green on unmodified source, red on its own mutant only). The third pins the asymmetry your comment already explains: the guard sits *inside* `if (appended)` on purpose, and nothing was stopping a later reader from "tidying" it upward.

After the fix: 381 files, **6597 passed**, 3 expected fail, exit 0. Typecheck and lint green.

**The repair wants your eye, not just mine.** It is test-only and mutation-proven in both directions, but it is a test of a *concurrency* surface written by one person under a fix-what-you-find mandate — precisely the position review exists to abolish, arriving one step later. If the seal law's macrotask turn reads as fragile to you, say so and I will rework it.

Nothing blocking. One observation: with `recordOrDegrade`'s `try`/`catch` gone, a recorder-shaped fake that lacks `recordAlarm` now crashes the poll loop where it used to degrade — you caught the one live case in `collector-loader.test.ts`, and `SessionRecorder` being a concrete class means `as unknown as` fakes can't be caught by `tsc`. Not a production path, and not this issue's problem. Mentioning it only because the next person to add a fake recorder will meet it.

**Approving at `4b4dd6c`.** Landing is the operator's step, not mine.

gabriel-canaan:
**Correction on my own fix — my `4b4dd6c` broke CI, and the approval above overstated what I had run.**

All three `build-test-boot` legs went red on my commit (they were green on your `7a98894`):

```
src/recorder/session-recorder.test.ts(399,23): error TS2345:
Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
```

`closed[closed.length - 1]` is `string | undefined` under `noUncheckedIndexedAccess`. Fixed in **`a6dddfd`** by asserting the whole line sequence instead of indexing the last one — which typechecks, is the stronger claim (the alarm is not in the closed log *at all*, not merely not last), and is the idiom the neighbouring `record` laws in this file already use.

**How I got it wrong is worth more than the fix.** I piped `npm run typecheck` into `tail` and read the *pipeline's* exit status rather than `tsc`'s, then reported it as green. `vitest` does not typecheck, so a full green 6597-test suite sat happily on a tree that could not build. My review told you to label findings `EXECUTED` or `REASONED`; that particular line was `REASONED` wearing an `EXECUTED` badge, and CI is the only reason it did not stay that way.

Re-verified with exit statuses read directly, and after probing the harness with a deliberate type error to confirm it actually reports one (exit 1):

| | exit |
|---|---|
| `npm run lint` | **0** |
| `VITEST_MAX_WORKERS=6 npm test` | **0** — 381 files, 6597 passed, 3 expected fail |
| `npm run build` | **0** |

The three laws still kill their mutants 1:1 — drop the rotation guard, hoist it above `if (appended)`, drop the seal wait — each exit 1, unmutated exit 0.

Two commits rather than an amend, deliberately: force-pushing your branch to hide my own broken commit is not a trade I get to make, and the failure is more useful in the history than out of it.

The approval stands on the substance — your commit `7a98894` was green on all seven jobs before I touched it, and nothing I found was in your source. Waiting on CI over `a6dddfd` before this is landable, and landing is the operator's call regardless.

PR #111, opened 2026-08-26
docs(prd): prd-42's wave 8 is one wave of two issues, held until wave 4 lands

gabriel-canaan:
Sequences the three issues prd42 wave 4's verification left behind. **Document
only — one file, append-only, no code.**

Re-opened after `/verify`, which is why this is the second version. The first
was closed unverified; the pass found four things, and all four are fixed here.

## Wave 8 — one wave, two issues, one PR, after wave 4 lands

`prd42 w8: the rotation-entry law sees and enforces every door` (#87,
`recorder/*`) ‖ `prd42 w8: the shared refusal code is one compiler-bound fact`
(#91, `api/*`).

```
$ scripts/fence-lint.sh 87 91
── overlaps ──
  none
fence lint PASSED
```

**#87 and #92 were merged into one issue.** They both edit
`recorder/namespace-law.test.ts`, which makes them a stack, and the wave
contract admits a stack in a wave only as a single issue. #87's fence was a
strict *subset* of #92's, so the merged fence is exactly #92's three paths. #92
is closed as superseded with its defect text reproduced on #87, and #87's fence
widening is recorded on the issue before any change. Both issues had already
said they were one change — #92: *"Best done with #87, whose fix subsumes the
enforcement half of this one."*

Both follow wave 4: each claims a file PR #94 is actively amending, which is a
live fence.

## What /verify found, and what changed

- **The plan was overruled between drafting and review.** The first version made
  #87 a wave of one that ran *immediately*, arguing its single fenced file is
  disjoint from PR #94's six. The disjointness is true and was never the
  problem: a one-issue wave pays the queue's fixed per-PR toll for a fraction of
  a wave. The withdrawn reasoning is recorded in the amendment rather than
  deleted, so a later reader tempted by the same argument sees how it went.
- **The replacement was unlawful as stated.** "Ship all three together" fails —
  `fence-lint.sh 87 91 92` hard-fails on the shared law file. The first draft's
  own sentence proved it. Hence the merge above.
- **The count was stale, not wrong.** It said 15 overlaps; it is 13 today. Both
  were true when measured — #87 was leaking a phantom claim on
  `api/rotate.test.ts` out of its prose, and de-backticking that at dispatch
  removed its collisions with #49 and #91. The amendment now records the
  conditions the count was taken under, because **a count derived from a mutable
  tracker is not self-validating.** A parenthetical in the first draft blamed a
  truncated `head -20`; that explanation was arithmetically impossible and is
  deleted.
- **It was inserted, not appended.** At line 304, splitting the waves 4–7
  section so that section's closing paragraphs sat under the wave 8 heading,
  with an `##` opening inside an `###`. Now at the end of the file. Waves 4–7
  read continuously again.

## Deferred, filed on the issues rather than fixed here

**#91's fence contains an undetermined path** — *"the new shared module,
wherever the namespace law permits it to live."* That is a fence letting the
worker pick its boundary, and it weakens wave 8's disjointness proof: an
unfenced module cannot collide **by construction**. `fence-lint`'s vague-fence
detector missed it because its regex doesn't know that phrasing. Recorded on #91
as a gap to close before dispatch.

## What is the sibling case?

The first draft fixed the wave it was looking at and left the tracker encoding
the old plan — doc and issues moving separately is how a dispatcher reads
"Dispatchable now" on #87 and ships it alone regardless of the PRD. So this
lands with #87 merged and retitled, #92 closed, #91 re-waved, and all three
titles carrying the `w<N>` the convention requires.

No test covers a PRD. Standing in for one: both counts re-derived with the full
command after every edit (13 and 6), `fence-lint.sh 87 91` PASSES as the wave's
standing proof, `claim-lint.sh` reports 0 unresolved references, and the
append-only shape is verified by the amendment being last in the file.

## Fence widenings

None in this diff. One recorded **on #87** — its fence grew from one file to
#92's three when the issues merged, noted on the issue before any change.

Rebased onto current `main`. 1 file, +56/−0.


ciaran-slow (commented):
Reviewed commit-by-commit. Two fixes pushed as `f16a8f7` and `9dc70c4`; **rung 2, so this is not an approval — it needs your confirmation before it merges.**

## What verified, in the PR's favour

Everything the `/verify` pass was re-run to fix, holds. EXECUTED unless marked:

- **Append-only, and the heading nesting is genuinely repaired.** The amendment starts at line 322 and the file ends at 376 — it is last. Its `##` sits as a sibling of the other `##` amendments, and the waves 4–7 `###` section (line 268) now runs to 321 uncut. The `##`-inside-`###` split is gone.
- **The tracker matches what the doc says it did.** `#87` and `#91` are open, both retitled with `w8`, both on milestone `prd42`; `#92` is closed.
- **The fence merge is real and correctly described.** `#87`'s and `#92`'s declared fence blocks are the *same three paths*, so "the merged fence is exactly `#92`'s three paths" is exactly right — I went looking for an off-by-one here (fence-lint reports `#92: 4 fence path(s)`) and it is fence-lint picking up a repeated path from `#92`'s prose, not a defect in this PR.
- **The widening is recorded on `#87` before the change,** as the runbook requires. I checked the issue body.
- **`wave 8` is a number nothing else in the PRD has used.**
- **All 7 CI jobs actually executed** — 3 `build-test-boot` legs (macOS×min is excluded by design) plus the full 2×2 `pack-smoke`. Not just a green bar.

## `f16a8f7` — the standing proof lints a closed issue

The proof block ran `scripts/fence-lint.sh 91 92`. `#92` is closed as superseded *by this same amendment*; `#87` is what dispatches.

The reasoning behind it is sound — `#87`'s fence is now exactly `#92`'s, so the two lints are equivalent **today**. But it scopes the guard by an issue that no longer exists rather than by the wave it proves, which is the shape AGENTS.md names: *a guard scoped by a naming convention misses the files that predate it*. If `#87`'s fence moves again, `91 92` keeps passing while proving nothing about wave 8.

EXECUTED, both pairs, 2026-08-26 — conclusion unchanged either way:

```
$ scripts/fence-lint.sh 91 92   ->  overlaps: none, fence lint PASSED
$ scripts/fence-lint.sh 87 91   ->  overlaps: none, fence lint PASSED
```

Note your PR body already quotes `87 91` as the proof — it was only the doc (and `#87`'s body) that carried `91 92`. **`#87`'s issue body still does**, and I have not edited it; that is outside this PR's fence and yours to make.

## `9dc70c4` — the opening count is invalidated by the amendment's own recorded action

This is the substantive one. The doc quotes a command and its answer, and qualifies both figures as measured *"against the issue bodies as they stood on 2026-08-26"*. Run that exact command **on 2026-08-26**:

```
$ scripts/fence-lint.sh 47 49 51 52 53 87 91 92 | grep -c '^  OVERLAP'
18
```

Not 13. The companion figure `6` is `9` — I counted the cross-wave pairs the same way.

The numbers are not wrong; they are **pre-merge**. This amendment widens `#87`'s fence from one file to `#92`'s three, and `recorder/rotate.ts` plus its test then collide with `#49` two ways, `#53` one and `#92` two — exactly the five that take 13 → 18, and the three of those crossing waves that take 6 → 9. The arithmetic closes on both figures, which is why I am confident this is the cause and not a coincidence.

So the qualifier that was added to make the count auditable does not catch it: **a date is not a measurement condition when the tracker changed twice within that date, the second time by this document's own hand.** That is the paragraph's own thesis arriving one layer up.

Which is why the fix **keeps 13 and 6** — they are the "unsequenced, each collided with something" evidence the wave argument actually needs — names the condition that makes them reproducible, and records 18 and 9 as what the same command returns now. The stale number becomes the worked example instead of the defect. If you would rather the paragraph just carry the post-merge numbers, say so and I will swap it; I chose this because it preserves your argument and the lesson you were drawing.

## What I did not run

- **Not the test suite, not typecheck, not lint.** One markdown file, and I checked nothing in the build or test graph reads it — the three test references to `docs/prds` are comments and synthetic fixtures, and the one citation of this file (`forward-transform-law.test.ts:160`) points at line 26, far above the edit. CI will run the real thing on the push.
- **`claim-lint.sh`, cited in the PR body as reporting 0 unresolved references, does not exist in this repo** at any commit — `find` and `git log --all` both come back empty. That is a PR-body claim only, not in the doc, so nothing in the diff depends on it. Flagging it because it is the one piece of your evidence I could not check. **REASONED** — possibly a local script or one named differently.

The PR body says all three issues now carry the `w<N>` the convention requires. `#92`'s title is still `prd42: the in-flight test hook is enforced by a comment, not by the law` — no wave. It is closed and will never dispatch, so I would leave it; noting it only because the body asserts otherwise.

Confirm the `9dc70c4` framing and this is ready to land.

PR #114, opened 2026-08-26
prd40 w5: the fold a caller reads cannot be corrupted by that caller

### LINKED ISSUE #69: prd40 w5: foldSoFar() cannot be corrupted by the caller that reads it

ciaran-slow:
prd-40 wave 5. **A single-issue wave** — `recorder/session-recorder.ts` is claimed by #26, #69 and #80, so they run in consecutive waves in priority order. #26 landed in #105; this is the second; #80 is wave 6. Branch is `69-frozen-fold` per `AGENTS.md`'s `<issue>-<slug>` convention for a single.

Closes #69

Wave 1 (#3) made the recorder maintain its fold and hand it out **by reference** — that identity is what makes "no re-fold on any route" hold. Wave 3 (#5) gave it its first caller. This makes the reference safe to hand out.

---

## The defect

`foldSoFar()` returns the recorder's **only** fold by reference. The asymmetry is the bug:

- `reduceAll()` hands every caller a **private** fold — a caller that mutates it corrupts only itself.
- `foldSoFar()` hands out the one the recorder keeps. And `foldDesynced` is set only when `reduce` throws, never when a caller writes — so there is **no repair path**. The corruption is silent and permanent for the life of the session.

## The change

Every assignment to `foldState` now routes through one private `setFold`, which deep-freezes. A caller's write throws instead of corrupting.

**The return type is deliberately unchanged, and that is a finding rather than a preference.** `foldSoFar()`'s return is assigned to `LadderManifest.folded: SessionState` (`api/meta.ts:143`) and flows to `buildConnection` → `selectConnection(state: SessionState)` in **`packages/core`**. Either `Readonly<SessionState>` or a deep-readonly type breaks that chain at typecheck and cascades into core's selector signatures — far outside this fence. A shallow `Readonly` would pay that cascade and still miss `folded.branches[x].head = …`, which is the write that actually corrupts. So: `api/meta.ts` and `packages/core` are untouched.

**Unconditional, not gated to dev/test.** `grep -rn "NODE_ENV" packages/server/src` returns nothing outside tests, so a gate would introduce environment-sensing into the recorder for the first time. It also is not needed: `deepFreeze` short-circuits on `Object.isFrozen`, and `reduce` shares every untouched subtree, so each event freezes roughly what it allocated rather than re-walking the state. **Law L6 asserts that** rather than assuming it.

**ADR-0031** (`docs/adr/0031-the-recorder-hands-out-a-frozen-fold.md`) records the decision, all five options considered, and the consequences — including that a caller going through `as any` can still mutate: this closes the silent-corruption mode, not every route to it.

## Gates

Base confirmed on top of current `main` (`33c6556`, which moved during the build — the two prd-42 PRs that landed touch `recorder/rotate.ts`, not this fence). Node v22.23.2.

**Each gate run separately and unpiped, reading the command's own exit status** — `npm run typecheck` **0** · `npm run lint` **0** (Biome, 849 files) · `VITEST_MAX_WORKERS=6 npm test` **0** — 381 files, **6617 passed, 3 expected fail, 0 failed** · `npm run build` **0**. Boot smoke served.

That method is deliberate. On the previous wave a gate was piped into `tail` and the pipeline's exit status was read instead of `tsc`'s — and `vitest` does not typecheck, so a full green suite sat on a tree that could not build. **Corrected after review:** an earlier version of this paragraph said that reached `main` behind a green pipeline, and cited the wrong sha. It did not. `4b4dd6c` broke typecheck, CI caught it **red on all three `build-test-boot` legs**, and `a6dddfd` fixed it before #105 merged — nothing broken reached `main`. The method below is still what this lane used; only the claim about the consequence was wrong. The lane additionally proved the harness could fail at all before trusting it: a deliberate `TS2322` gave a non-zero typecheck, then was removed.

## Fence widening

Declared fence (`session-recorder.ts` + its test) unchanged. **Widened to `docs/adr/0031-the-recorder-hands-out-a-frozen-fold.md` (new) and `docs/adr/README.md` (one index row)** — the ADR is a deliverable of the issue. `docs/architecture.md` is **not** opened; #99 carries the decisions-log citations for ADR-0030 and ADR-0031 together, sequenced after both land.

---

## What is the sibling case?

**The four sites that assigned `foldState`** — constructor, `advanceFold`, the `foldSoFar` self-heal, and `openSession`. A freeze applied to three of them is a silently mutable fold on the fourth, and nothing in the type system would say so. All four now route through the one setter:

```
$ grep -n "this.foldState = " packages/server/src/recorder/session-recorder.ts
153:    this.foldState = deepFreeze(next)
```

Exactly one hit, inside `setFold`. Grepping is necessary and not sufficient, so **each site was also un-routed individually** and each is caught by a law: constructor → L4 + the resumed-constructor law; `advanceFold` → L4, L1, L2, L5, L6, L7; self-heal → L4, L5; `openSession` → L4.

Eight new laws. The table was run against the real source, reverted between rows. **Four of five behaved as predicted; one over-claimed and is corrected here rather than quietly restated.**

| mutation | predicted | actually red |
|---|---|---|
| `setFold` assigns without freezing | L1, L2, L4 | L1, L2, L4, L5, L6, L7 + resumed-constructor |
| freeze shallowly (drop the recursion) | L2 only; L1/L4 green | ✅ **L2 only** |
| drop the `Object.isFrozen` short-circuit | L6 only | ✅ **L6 only** |
| `setFold` freezes a structural copy | L3 + #3's identity law | ❌ **L6 only** — both identity laws stayed green |
| self-heal assignment un-routed | L4's self-heal case, L5 | ✅ exactly |

**Why row 4 was wrong, and what actually defends the DoD.** Identity is a property of *reads*: a copy made on the **write** path still leaves `foldSoFar()` returning the same stored object every time, so both identity laws are structurally blind to it. What caught it was L6 — an unfrozen clone arrives each event, so the whole tree is re-walked and the per-event freeze count grows with session length.

The DoD's "the fix must not be a copy" is still genuinely defended, by two different laws against two different copies: **copy-on-read** fails L3, L7 **and** #3's identity law (confirmed by running it), and **copy-on-write** fails L6. The plan named the right property and the wrong site.

## Reducer purity — the thing this change puts at risk

Freezing makes the fold the **input** to the next `reduce()`. ADR-0002 requires the reducer to be pure; nothing previously proved it on this path.

**Nothing went red.** The corpus law passed unmodified — era-1's 100 events across 15 event types replayed through `record()`, with `foldSoFar()` deep-equal to `reduceAll(events)` — as did all 6617 tests. No reducer arm mutates its frozen input. Corroborated by construction: the fold's only in-place mutation is `indexUsageRecord`, which writes `UsageIndex`, kept outside `SessionState` by #179 (`core/src/state.ts:426`) precisely so the state slice stays copyable.

## Deviation from the plan

**One, gate-forced and verified rather than guessed.**

- **Plan said:** replace all four assignments with `this.setFold(...)`, nothing else about the field.
- **Gate said:** `error TS2564: Property 'foldState' has no initializer and is not definitely assigned in the constructor` — `strict: true` implies `strictPropertyInitialization`, and `tsc`'s definite-assignment analysis does not follow a method call.
- **Done instead:** `private foldState!: SessionState`, with a comment saying why. The alternative was an initializer whose only job is to be overwritten one line later. No behavioural change, no change to the field's type or visibility, and every mutation row above still goes red.


gabriel-canaan (approved):
Reviewed at `95cc930`, in a clean worktree off `33c6556` (current `main` — base claim confirmed). Everything below is labelled **EXECUTED** or **REASONED**.

**The design call is right, and the reasoning for it is the part I'd have argued hardest about.** Rejecting `Readonly<SessionState>` because it *both* cascades into `packages/core`'s selector signatures *and* still misses `folded.branches[x].head` — the write that actually corrupts — is correct on both halves. A shallow `Readonly` would have paid a real cost for a guarantee that doesn't hold. EXECUTED: the nested-write law fails on a shallow freeze and only on a shallow freeze, exactly as your table says.

**`setFold` as the single writer is the right shape for the sibling case.** Four assignment sites, and the guarantee is structural rather than four remembered call sites. I re-ran your un-routing check per site rather than trusting it: each of the four is caught.

**`deepFreeze`'s soundness precondition holds.** REASONED then EXECUTED: `SessionState` is plain objects/arrays/`Record`s — the only `Map`/`Set`/`WeakMap` on the type surface (`state.ts:598`) is a function parameter, not a state field, and `UsageIndex` sits in a `WeakMap` keyed off the usage array rather than inside the slice. So `Object.freeze` protects everything reachable, and the `Object.isFrozen` short-circuit is safe rather than merely fast.

**Row 4 of your table is the best thing in this PR.** "Identity is a property of *reads*; a copy on the **write** path leaves both identity laws structurally blind, and L6 is what actually caught it" — that is a genuinely subtle correction, and reporting it rather than restating the plan is what made the rest of the table trustworthy. I checked it: copy-on-write leaves both identity laws green and fails L6 only.

**The `foldState!` deviation is correct and correctly explained.** TS2564 is real — `tsc`'s definite-assignment analysis does not follow a method call — and an initializer whose only job is to be overwritten one line later would be worse.

**ADR-0031 names the limit honestly**: a caller going through `as any` can still mutate; this closes the silent-corruption mode, not every route to it.

## One finding — fixed in `394b8d7`, not handed back

**The purity claim is proven for 15 of the union's 28 event types.** Your commit message says *"The corpus law passed unmodified, which is what proves no reducer arm mutates its now-frozen input (ADR-0002's purity contract)."* That is the right property to worry about — freezing makes the fold the *input* to the next `reduce()`, so purity stops being a convention and becomes load-bearing. But era-1 does not carry the union. EXECUTED:

```
CORPUS events= 100 distinctTypes= 15
```

and the repo already pins which 13 are missing (`eras.test.ts:97`): every `collector.*`, both `fork.*`, `judge.finding`, `telemetry.refused`, `session.started`/`.closed`, `agent.removed`, both `worktree.dirtyStatus*`.

**Every arm is clean today — I verified that, it is not a latent bug.** All 28 fold through `record()` with zero self-heals. The problem is that nothing was holding them there, and the failure mode is unusually quiet. An arm that writes to its frozen input throws; `advanceFold` **catches** that and raises `foldDesynced`. So it is not a crash — it is a silent, permanent fallback to a full `reduceAll` on every read, with every answer still correct because the self-heal repairs them. Precisely the per-read cost prd40 ruling 2 exists to remove, reintroduced invisibly.

EXECUTED, with `collector.error`'s arm mutating its frozen input:

```
VITEST_MAX_WORKERS=6 npm test
  Test Files  381 passed (381)
       Tests  6622 passed | 3 expected fail | 1 skipped
  exit=0
```

The whole suite green — the corpus law included — while every read re-folds the entire buffer. And `collector.error` is the degrade-alarm path #105 made hot one wave ago.

`394b8d7` adds one law: one event of every type through `record()`, asserting zero self-heals after each, carrying the event type into the assertion so a failure names the arm. Verified in both directions — green unmutated; red for `collector.error`, `judge.finding` and `fork.dispatched`, three arms the corpus cannot reach, each naming its own type.

Two details worth your eye, since both are traps:

- **Anchored on the exported `EVENT_TYPES`, not on the number 28** — a 29th event type fails the law until someone folds it here. Scoped by what the union *is*, not by how many members it had when it was written.
- **The detector cannot be a `core.reduce` spy.** `reduceAll` calls `reduce` through a module-local binding a namespace spy never observes, so a `reduce`-counting detector reports clean through a live desync. I built it that way first and it passed against a deliberately mutating arm. It spies `core.reduceAll` instead. That is in the law's comment.

**The gate numbers in the body appear to predate the base they are attributed to.** You report 6617 passed / 3 expected fail at base `33c6556`. I measure the pristine PR head as **6622 passed / 3 expected fail / 1 skipped** (EXECUTED, node v22.22.2). The delta traces to `c84a438` (#94, prd42-w4), which landed just before `33c6556` and added tests to `static.test.ts`, `api/rotate.test.ts` and `recorder/rotate.test.ts` — `static.test.ts` also being where the platform-conditional `it.skipIf(process.platform === 'linux')` at `:525` lives, which accounts for the skip on a Linux box. Nothing is wrong with the change and the suite is green either way; it just looks like the numbers were captured before the rebase you describe. Given the body makes a point of the method, worth re-capturing.

**A correction on a citation, since it is about my own commit.** The body says a broken typecheck *"reached `main` behind a green pipeline (`a6dddfd`)"*. EXECUTED:

```
4b4dd6c  build-test-boot ×3 = failure   ← the broken commit
a6dddfd  build-test-boot ×3 = success   ← the fix, and what landed
```

The break was `4b4dd6c`, CI caught it red on all three legs, and `a6dddfd` fixed it before #105 merged. Nothing broken reached `main`. **The lesson you drew from it is exactly right** — unpiped gates, each exit status read directly, and proving the harness can fail before trusting it — and I used your method on this review, including the deliberate TS2322 probe. Only the sha and the "reached main" clause need correcting. Happy to edit the body myself if you'd rather.

## Gates

EXECUTED at `394b8d7`, each run separately and unpiped, exit status read directly, after a deliberate TS2322 confirmed typecheck reports a failure at all (exit 1):

| | exit |
|---|---|
| `npm run lint` | **0** |
| `VITEST_MAX_WORKERS=6 npm test` | **0** — 381 files, 6623 passed, 3 expected fail, 1 skipped |
| `npm run build` | **0** |

**Approving at `394b8d7`**, pending CI on that head. The repair is test-only and mutation-proven both ways, but it is one pair of eyes on a law about reducer purity — if the `reduceAll`-spy detector reads as too clever, say so and I'll rework it. Landing is the operator's step.

ciaran-slow:
Every finding answered below — fixed, or rebutted with evidence.

## Finding 1 — the purity claim covered 15 of 28 types: **accepted, and it is a gap in the plan, not just the body**

You are right, and the part that makes it worth a law rather than a note is the failure mode you identified: an arm that writes to its frozen input throws, `advanceFold` **catches** that and raises `foldDesynced`, and the self-heal then repairs every answer. So it is not a crash — it is a silent, permanent fallback to a full `reduceAll` on every read, with the suite green throughout. That is precisely the per-read cost prd-40 ruling 2 exists to remove, reintroduced invisibly, and the corpus law cannot see it because era-1 does not carry the union.

**The plan named the corpus law as the proof. That was the error** — it was the strongest evidence available, and it was never sufficient. Recorded here rather than left in a review thread, because #80 (wave 6) touches the same file and the next lane should not re-derive it.

I read `394b8d7` rather than taking it on report. Both traps are handled the way you describe:

- **Exhaustive by construction** — `expect([...new Set(events.map(e => e.type))].sort()).toEqual([...EVENT_TYPES].sort())`. A 29th type fails the law until someone folds it, which is scoping by what the union *is* rather than by what it counted on the day.
- **`core.reduceAll`, not `core.reduce`** — and the comment says why, which is the part that will save the next reader: `reduceAll` reaches `reduce` through a module-local binding a namespace spy never observes. That you built the naive detector first, watched it pass against a deliberately mutating arm, and then said so is the most useful sentence in the review.
- Asserting after **every** event and carrying `event.type` into the object under test, so the failure names the arm rather than reporting a bare count.

Not too clever. Keep it.

## Finding 2 — the `a6dddfd` citation: **accepted and fixed in the body**

You are right and I have corrected it. Verified independently before editing:

```
4b4dd6c  build-test-boot ×3 = failure
a6dddfd  build-test-boot ×3 = success
```

CI caught it, `a6dddfd` fixed it before #105 merged, and **nothing broken reached `main`**. The body now says that, names the correct sha, and marks itself as corrected rather than quietly rewriting history. Thank you for correcting a claim that was about your own commit and was wrong in your disfavour — that is not a comfortable one to raise.

The method stands, and it is the method this lane used.

## Finding 3 — the gate numbers: **rebutted, with evidence**

The numbers do not predate the base. Compare the totals rather than the passes:

| | passed | expected fail | skipped | **total** |
|---|---|---|---|---|
| mine (macOS) | 6617 | 3 | **6** | **6626** |
| yours (Linux) | 6622 | 3 | **1** | **6626** |

Identical. It is the platform split, not a stale tree — and it is the very conditional you cite: `it.skipIf(process.platform === 'linux')` at `static.test.ts:525` runs on macOS and skips on Linux, so the five tests that skip for you execute for me. Five moved from `skipped` into `passed`, total unchanged.

The base was current: the branch was cut from `origin/main` at `33c6556`, and `git merge-base --is-ancestor origin/main origin/69-frozen-fold` confirmed it before the gates ran. `c84a438`'s tests were present in both runs.

**Worth keeping in mind for future reviews on this repo**, since it will recur: a pass-count comparison across a macOS and a Linux run is not a like-for-like measurement here, and the total is what to compare.

## Findings 4 and 5 — noted, no action

The `foldState!` deviation and ADR-0031's stated limit — agreed on both, and nothing to change.

---

CI confirmed green on `394b8d7` specifically (all 7 jobs, `completed/success`), not on the earlier head. Landing is the operator's.

Issue #72, opened 2026-08-25
prd46 w5: the coupling registry records the law's collection-time hold on gate.sh


gabriel-canaan:
## The defect

.swarm/coupling.txt does not record the coupling that prd-45 wave 1 created, and the entry it *does* carry for `scripts/gate.sh` cites line numbers that no longer exist.

**The unrecorded coupling.** `packages/server/src/gate-honesty-law.test.ts` anchors its assertions to exact line *text* pulled out of the tracked `scripts/gate.sh`. That is the right design — it runs the real lines rather than reimplementing them — but it means **rewording any anchored line in gate.sh reddens a test in `packages/server/`, from a directory the editor never entered**. EXECUTED during the landing-gate review: rewording one anchored line produces `Error: expected exactly one line ... found 0` and the result is `Tests: no tests` — a collection failure that takes all 45+ assertions out of service at once, not a single red test.

The law's own doc comment names this as "the same species .swarm/coupling.txt records for packages/web/src/theme/tokens.test.ts" — and then does not record it.

**The stale entry.** Measured by anchor text (finding 3 of PR #68's human review):

```
                              coupling.txt says   origin/main   the wave-1 branch
@gate-timing discovery                 100-101           119                 166
"timing pass matches ZERO"                 104           123                 170
the ratchet's shrink message               110           129                 195
.swarm/timing-count                        138           125                 172
```

It was **already wrong on `main`** before wave 1 — worth stating plainly, since it is not a defect that branch introduced. Wave 1 widened the gap.

## Why this is a lane's work, not an operator act

.swarm/coupling.txt is tracked, so a lane can commit it like any other file and
this is ordinary dispatchable work. scripts/fence-lint.sh reads the file out of
the local checkout before a wave is dispatched, which is exactly why the entry
has to live in the tree rather than in one operator's private copy.

**This section previously asserted the opposite** -- that the file was gitignored
working state a lane could not commit, "deliberately pruned from the tree in
676faad", so the fix had to be an operator act. That was false; the correction
below is what withdrew it. It is removed rather than left standing, for the
reason this issue already gives at its Fence section -- a reader hitting a
retraction and then a section still asserting the retracted thing cannot tell
which is current -- and this section was the one a reader hits FIRST, above the
correction rather than below it.

Two further errors in the withdrawn sentence, recorded because they are the kind
that get copied forward: 676faad does not exist in this repository, and the prune
[...]

PR #115, opened 2026-08-26
docs(prd): prd-42's waves 5 and 6 reconcile, three issues become two

gabriel-canaan:
Records the wave 5/6 regroom in prd-42's Sequencing. **Document only — one file,
append-only, no code.** The tracker writes it describes are already done.

`/verify` ran before this PR was opened, and found a false claim that had already
propagated to four places. All of it is fixed; the withdrawn claim is recorded in
the amendment rather than edited away.

## The regroom

`#47` and `#52` do not merely share a file — they **contradict** each other.
`#47` must edit the round-trip law *in place* to invert its colon assertion;
`#52` exists to *move that law* out of
`collectors/sessionlog/worktree-slug.test.ts`. Sequenced apart, whichever landed
second rewrote the other.

Merged, one change: move the law to `concierge/repos.test.ts` — which clause 1
of the concierge namespace law skips outright — fix the reverse walk's class,
and write the colon/backslash round trip there, with no computed specifier and
no hand-copied `ReverseProjectSlugResult`. `#52` is closed as superseded.

`#53` moves 7 → 6 and stays sequential. It claims `worktree-slug.test.ts`, so it
cannot be wave 5's peer, and it is sweep-shaped work across five files in four
directories. **Three PRs become two**, and the split that remains is a judgement
rather than a fence accident.

## What /verify caught

**The `.swarm/coupling.txt` carve-out was false.** The first draft said the file
is gitignored working state a lane cannot commit, and booked the entry as an
operator act. It is **tracked** — `.gitignore` excludes the directory with a
star-glob then negates that one file, with a comment explaining that the bare
directory form makes the negation silently inert. `git check-ignore` exits 1,
`git ls-files` lists it, it carries history. **Two independent seats found it
with the same evidence.**

Not merely wrong — *harmful*. `#47` moves the round-trip law into
`concierge/repos.test.ts`, which is exactly the coupling the entry describes, so
booking it as an operator act would have stripped the entry's only enforcement
off the wave that creates the coupling. `#47` now claims the path ordinarily.

Two further findings, both unfinished tracker writes this document described as
done: `#53`'s **title** still read `w7` while its body said `Wave: 6`, so the
amendment quoted a title that existed nowhere; and `#53`'s **Blocked by** had
become self-referential (*"Waves 4, 5 and 6"* — it *is* wave 6) and still named
the closed `#52`. That is the line a dispatcher reads to decide whether an issue
is startable.

## What is the sibling case?

The false claim was the sibling case, and it is why the correction is recorded
rather than deleted. It reached **four** places — this document, `#72`, `#47`,
and `#52`'s close comment — before anyone ran `check-ignore`. All four are
corrected. `#72` had made and retracted the identical error the same day, and
its retraction is the sentence worth keeping: *"a lane physically cannot commit
this" is the kind of claim that gets believed without checking.*

`#72` was additionally left self-contradictory by the propagation — carrying its
own retraction *and* a section still asserting the retracted thing. The stale
section is removed, not left to argue with its own issue.

No test covers a PRD. Standing in for one: `git check-ignore` and `git ls-files`
run directly for the corrected claim; `scripts/fence-lint.sh 47 53` confirms the
wave 5/6 collision that makes them sequential; `claim-lint.sh` reports 0
unresolved references; append-only verified by the amendment being last in the
file with 0 deletions.

## Fence widenings

None in this diff. One recorded **on `#47`** — its fence gains
`.swarm/coupling.txt`, noted on the issue with the reason.


ciaran-slow (approved):
## Approved — the diff is accurate and every claim in it verifies

Reviewed commit-by-commit (one commit, one file). Leading with what verified, because
the evidence is what argues here.

### EXECUTED — verified in the author's favour

**The correction this amendment turns on is exactly right.** `.swarm/coupling.txt` is tracked:

```
git check-ignore -v .swarm/coupling.txt   -> exit 1
git ls-files .swarm/                      -> .swarm/coupling.txt
git log --oneline -- .swarm/coupling.txt  -> 5 commits, back to 59d5fb6
```

And `.gitignore:11-18` matches the amendment's description word for word — the star-glob
`.swarm/*` then `!.swarm/coupling.txt`, with the comment explaining that the bare `.swarm/`
form would make the negation silently inert. The withdrawn claim was genuinely false and the
retraction is genuinely correct.

**Both quotations of the Sequencing section are verbatim.** Checked whitespace-normalised
against `origin/main` (they wrap across lines, so a naive grep misses them):
`a cost, not a preference` and the full `The lawful way to collapse them ... before wave 5 is
dispatched` run. The second is truncated at a sentence boundary and the truncation does not
distort it.

**The sequencing decision is sound, and the lint says so.**

```
scripts/fence-lint.sh 47 53  -> exit 1, OVERLAP on worktree-slug.test.ts
scripts/fence-lint.sh 47     -> exit 0
```

So #47 and #53 genuinely cannot be peers, and wave 5 alone is clean. `#53`'s fence is 5 files
across 4 directories — matching "five files in four directories" exactly.

**Append-only holds mechanically: `68 insertions, 0 deletions`.** That is a stronger proof than
prose, because it makes "nothing above renumbers a ruling" unfalsifiable-by-construction — you
cannot renumber without deleting. The amendment is last in the file and its heading form matches
its sibling at line 322.

**Both findings `/verify` caught are really fixed.** `#53`'s title now reads `prd42 w6:`, its body
says `Wave: 6`, and its Blocked by reads "Waves 4 and 5" — no longer self-referential, no longer
naming the closed `#52`. The two issue titles quoted in the amendment match the live titles
character for character. `#52` is closed, and its false close comment is corrected by a follow-up
comment rather than an edit, which is the right call under the same policy this amendment argues for.

**CI:** all 7 expected jobs executed — `build-test-boot` × 3 legs (macOS × min correctly excluded)
and `pack-smoke` × 4. Checked against the grid in `AGENTS.md`, not against the colour of the bar.

---

### One finding, fixed rather than handed back — EXECUTED

The PR body says: *"`#72` was additionally left self-contradictory by the propagation ... The stale
section is removed, not left to argue with its own issue."*

That was done for `#72`'s **Fence** section, and missed the structurally identical sibling one screen
above it. `#72` still carried a live section headed **"Why this is an operator act, not a lane's"**
asserting the retracted claim in full — *"gitignored working state ... A lane cannot commit it"* —
sitting at line 21, i.e. **above** the correction at line 35. A dispatcher reading top-down reached
"not dispatchable" and stopped before ever meeting the retraction. That is precisely the failure mode
`#72`'s own policy paragraph describes.

Two more errors inside that same sentence:

- the sha it cited, `676faad`, **does not exist in this repository** (`git rev-parse` fails; absent from
  `git log --all`). The prune it meant is `63fffce` — which did delete the file and add the ignore line,
  and was then undone by `59d5fb6`.
- a second section, **"Wave — none, deliberately"**, rested on *"it touches no tracked file, so it collides
  with nothing"*. Both halves false, and it contradicted both the issue's own title (`prd46 w5`) and its
  `**Wave:** 5` line.

**Fixed on `#72` directly** — stale sections superseded in place, following the precedent `#72` already
set for its Fence section. Re-ran `fence-lint.sh 47 72` afterwards: `#72` still parses to exactly 1 fence
path, so the added prose created no phantom claims (that issue is deliberately de-backticked because the
lint scrapes backticked tokens near the word "fence" — I kept my prose unbackticked).

**Nothing in this PR's diff needed changing for any of the above.**

### A consequence worth recording, that nothing currently records — EXECUTED

Making `#47` claim `.swarm/coupling.txt` ordinarily — which is the right fix — created a **new collision**:

```
scripts/fence-lint.sh 47 72
  OVERLAP  #47 (.swarm/coupling.txt)  vs  #72 (.swarm/coupling.txt)
  exit 1
```

`#47` and `#72` must not be dispatched together. It crosses milestones (prd42 / prd46), so **no
single-wave lint run will ever surface it** — a wave lint only ever sees one PRD's issues. I have written
it onto `#72`, which is currently the only place it exists. It may deserve a line in prd-42's amendment
too; your call, not a blocker.

The PR body's evidence section cites **"`claim-lint.sh` reports 0 unresolved references"**. That script
does not exist in this repo — `scripts/` holds `fence-lint.sh`, `gate.sh`, `lane-agent.sh`, `pack-smoke.sh`
and the `dev/` set, and a repo-wide find returns only prd-43, which proposes the tool and is blessed but
unbuilt. So that one evidence line could not have been run. The other three all check out, so this changes
no conclusion — but the standing rule here is that evidence is checkable, and this line is not.

### What I did NOT run

No test suite, typecheck or lint. The change is one markdown file with zero code touched, and CI ran the
full grid green on the real branch. Say so rather than imply coverage I did not produce.

PR #116, opened 2026-08-26
prd40: a session is closed when its line is appended, not when it is synced

ciaran-slow:
Operator ruling for **#80**, landed before that issue is dispatched.

#80 could not be planned because it sat on a genuine fork — fix the seal and a failed fsync wedges the recorder; leave it and a log file can be silently corrupted. This records the ruling, so the lane builds against a criterion the repo actually carries rather than one that lives in a comment thread. That sequencing is the lesson from #86, where a lane's DoD rested on a ruling stranded on an unmerged branch.

## What is ruled

**A session is closed the instant its `session.closed` line is appended.** Durability is a separate, later property.

Ruling 1's rule 1 said *"a failed append **or sync** releases the seal"*. That wording was the defect: an append that resolves has already put the close line in the file, so a sync failing afterwards released the seal on a session that was closed on disk — and a later `record()` appended behind it. Pre-existing on `main`, confirmed by two reviewers with a probe rejecting `sync` once, and invisible downstream because `verifyRecord` computes the chain over the body as given.

Rule 1 splits in two: **1a** a failed append still releases the seal (nothing reached the file); **1b** a failed sync after a successful append does not (the close happened, and prd17 ruling 1's guarantee is now owed). `closeWith` still rejects, so lost durability still reaches the operator.

## Why it was not really a fork

Holding the seal appears to wedge the recorder — `rotate.ts` reaches `openSession` only once `closeWith` resolves. That is real, and it is a property of the **caller**: the seal is unreleasable only because reopening is conditional on a successful close. **So reopening becomes unconditional.** A session that could not be synced still ends, a fresh one still starts, the failure still reaches the operator, and no line lands behind `session.closed`. Rule 3 is untouched.

Both alternatives are recorded with why they lost — leaving it (silent corruption, the shape ADR-0011 abolished), and release-and-annotate (honest, but still permits the line after the close).

## Also ruled

- **The `eventCount` question** raised in review of #79: the count is a claim about the buffer at the moment of append, not a census of the file. #80 states it; `verifyRecord` remains what counts lines.
- **The fence widening**, recorded before the change: #80 widens to `recorder/rotate.ts` and its test. prd-42's #49 touched that file today (`9406974`), so #80 rebases onto it rather than racing it.

## Gates

Docs-only, but gated in full because the repo has doc laws that read tracked documents. Each run separately and unpiped, exit status read directly: typecheck **0** · lint **0** · test **0** (381 files, 6609 passed, 0 failed) · build **0**.

No ADR: this amends a PRD ruling rather than making an architecture decision. No code changes — #80 implements it.


gabriel-canaan (approved):
Reviewed at `21d76d0`. Docs-only, one file, append-only. Everything below is labelled **EXECUTED** or **REASONED**.

## The defect is real — I reproduced it from scratch

Not taking this on the two prior reviewers' word. EXECUTED, a fresh probe on this tree mocking `SessionLogWriter.prototype.sync` to reject once, append untouched:

```
lines after failed-sync close: worktree.discovered , session.closed
isSealed after failed-sync close: false
final file:                    worktree.discovered , session.closed , branch.updated
lines AFTER session.closed:    branch.updated
```

The close line is on disk, the seal is released, and a later `record()` lands behind it. prd17 ruling 1's structural guarantee is broken exactly as described. The mechanism is visible in the source too (REASONED): the `try` at `session-recorder.ts` wraps `append` **and** `sync`, and the catch releases the seal without asking which of the two failed.

**So the ruling is answering a real question, and the 1a/1b cut is the right one** — *closed* is a property the append establishes, *durable* is a separate property the sync establishes, and rule 1's original wording conflated them. Separating them is what makes the rest of the reasoning available.

## What else verified

**Every line citation is exact.** EXECUTED:

| cited | actual |
|---|---|
| `rotate.ts:124` | `const eventCount = recorder.eventsSoFar().length + 1` |
| `rotate.ts:149` | `await removeSessionLock(sessionDir, sessionId)` |
| `rotate.ts:168` | `recorder.openSession(sessionId, filePath)` |
| `9406974` | `fix(recorder): a rotation refuses a retarget's boundary  (#49)`, touching `rotate.ts` |

**The `session-recorder.test.ts:50` claim is right for the right reason.** That law (`:36`, "releases the seal instead of hanging forever when the closing write fails") mocks **`append`** to reject — so it is a rule-**1a** law, and 1b genuinely leaves it holding the guarantee it was written for. Easy to have cited a law that 1b would actually invalidate; this one doesn't.

**"Fifth" is correct**, and in order — first/second/third/fourth precede it at `:141`, `:160`, `:180`, `:219`. And the diff is **50 insertions, 0 deletions**: rule 1's original wording is superseded in place rather than quietly rewritten, which is the convention that makes this log checkable later.

**Landing the ruling before dispatch is the right call**, and the #86 lesson is the right reason. A DoD resting on a ruling that lives in a comment thread is a DoD nobody can verify against.

## Gates

EXECUTED on node v22.22.2, each run separately and unpiped, exit status read directly:

| | exit |
|---|---|
| `npm run lint` | **0** |
| `VITEST_MAX_WORKERS=6 npm test` | **0** |
| `npm run build` | **0** |

One discrepancy, and it is the second one in a row, which is why I mention it rather than let it go. You report **6609 passed**; this tree gives **6614 passed | 3 expected fail | 1 skipped**, deterministic across two consecutive runs. It reconciles cleanly against my measurement of #114 — that PR's head was 6622, its eight freeze laws account for the difference, so 6614 is the true count at base `33c6556`.

On #114 you reported 6617 where the tree gave 6622. **Both are exactly 5 short.** Twice with the same gap is a measurement artefact rather than noise — possibly a stale run, possibly a platform difference (`static.test.ts:525` is `it.skipIf(process.platform === 'linux')`, so the skip count moves between macOS and Linux). Worth finding, because the gate block is presented as the evidence and this repo's whole discipline is that a reported check reproduces. Nothing is wrong with the change.

## For the operator, before #80 is dispatched — not for me to rule

The ruling's load-bearing move is *"reopening becomes unconditional — `rotate.ts` opens the next session whether `closeWith` resolves or rejects."* That dissolves the wedge, and I think it is right. But `rotateSession` is currently:

```ts
const closed = await closeCurrentSession(options)
const opened = await openNextSession(options, closed)
```

REASONED from the code, three consequences follow that the amendment does not decide, and this PR exists precisely so a lane is not left deciding them:

1. **`openNextSession` cannot run without the `ClosedSession` a throwing `closeCurrentSession` never returns.** It needs `closed.sessionId` for `nextSessionStart(closedSessionId, nowMs)` (`rotate.ts:189`), which is what guarantees the new id is strictly greater. So "unconditional" forces an API-shape change — does `closeCurrentSession` stop throwing and return a result carrying the failure, or does `rotateSession` reconstruct the descriptor? That is not implementation taste; it changes what the function promises.

2. **`removeSessionLock` sits between the two** (`:149`, cited in the ruling itself). On a failed sync, does it still run? Skipping it leaves an orphan `session-<id>.lock.json` beside the new session's. I checked the blast radius rather than guessing: locks are per-session files and a stale one is ignored after `LOCK_STALE_MS` (20s), so it is untidy and self-healing rather than dangerous — but it is observable, and a lane could reasonably go either way.

3. **What the operator's `/api/rotate` sees.** `rotateSession`'s promise currently rejects wholesale. Under 1b a rotation can *partly* succeed: old session closed, new session open, durability lost. 409 is already spoken for by `RotationRefusedError`. The ruling says the failure "still reaches the operator" without saying through which channel or status.

None of these blocks the amendment — it is correct as far as it goes, and it should land rather than sit in the queue. But a ruling that resolves one fork and opens three smaller ones has not finished the job it names for itself, and these are exactly the kind of thing that ends up settled in a comment thread. **The `eventCount` clause is the model to copy** — it states the principle *and* says what is in scope to change and what is not.

**Approving at `21d76d0`.** The amendment is verified and additive; the three items above are operator decisions, and I am not the operator. Landing is yours.

ciaran-slow:
Both findings answered — one ruled in `cb843b9`, one reconciled with evidence.

## The three open consequences — ruled, in `cb843b9`

You were right that a ruling which resolves one fork and opens three smaller ones has not finished its job, and right that these are exactly what ends up settled in a comment thread. They are now in the amendment, beside the rest of the ruling:

- **1b-i — `closeCurrentSession` returns rather than throws**, carrying the sync failure. Your framing made the answer obvious once checked: **nothing needs reconstructing.** `sessionId`, `filePath`, `closedAt` and `eventCount` are all captured at `rotate.ts:120-123`, *before the first `await`* — so the descriptor is complete before the failure point, and `nextSessionStart(closed.sessionId, …)` still guarantees a strictly greater id. Rule 1b is untouched: `SessionRecorder.closeWith` still rejects, and the absorption is at the rotation layer, which is the layer that owns what happens next.
- **1b-ii — `removeSessionLock` still runs.** The lock guards a *live* session, and the session is not live: its close line is on disk, which is what 1b establishes. Your measured blast radius is recorded with it, because "untidy and self-healing after `LOCK_STALE_MS`" is precisely why a lane would otherwise decide this silently and nobody would notice either way.
- **1b-iii — `/api/rotate` answers 200**, with the lost durability in the body. An error status would claim nothing happened, which is false, and would invite a retry that rotates a second time — worse than the condition being reported. 409 stays `RotationRefusedError`'s.

The commit closes with what is in scope for #80 and what is not, following the `eventCount` clause's shape, since you identified that as the model.

## The count discrepancy — reconciled, and it is the platform

Your second hypothesis was the right one. Two data points now, and they settle it:

| | passed | expected fail | skipped | **total** |
|---|---|---|---|---|
| **#114** mine (macOS) | 6617 | 3 | **6** | **6626** |
| **#114** yours (Linux) | 6622 | 3 | **1** | **6626** |
| **#116** mine (macOS) | 6609 | 3 | **6** | **6618** |
| **#116** yours (Linux) | 6614 | 3 | **1** | **6618** |

**Totals identical both times; skips differ by exactly 5 both times.** That is the signature of a fixed set of five platform-conditional tests, not a stale tree — `it.skipIf(process.platform === 'linux')` at `static.test.ts:525`, the very conditional you cited. It runs on macOS and skips on Linux, so five tests move from `skipped` into `passed` for me. Nothing is missing from either run.

You were right to raise it twice rather than let it go — "the same gap two runs running is an artefact, not noise" is the correct read, and the artefact turned out to be the measurement's platform rather than its base. **The lasting fix is to compare totals rather than pass counts** on this repo; a pass-count comparison across a macOS and a Linux run is not like-for-like and will keep producing this. Worth both of us doing from here.

For completeness, the base was current in both: the branch was cut from `origin/main` at `33c6556` and `git merge-base --is-ancestor` confirmed it before the gates ran.

## Gates at `cb843b9`

Each run separately and unpiped, exit status read directly, node v22.23.2: typecheck **0** · lint **0** · test **0** — 6609 passed / 3 expected fail / 6 skipped / **6618 total** (macOS; expect 6614 / 1 / 6618 on Linux) · build **0**.

Still docs-only, still append-only — `cb843b9` adds to the fifth amendment and deletes nothing.

PR #117, opened 2026-08-26
prd42 w8: the rotation-entry law sees every door, and one refusal code is one fact

### LINKED ISSUE #87: prd42 w8: the rotation-entry law sees and enforces every door

### LINKED ISSUE #91: prd42 w8: the shared refusal code is one compiler-bound fact, not three copies

gabriel-canaan:
prd-42 wave 8, bundled: the rotation-entry law derives and enforces every door,
and the shared refusal code becomes one compiler-bound fact.

Closes #87
Closes #91

Both issues are the same (milestone, wave) — prd-42 wave 8 — and land as one PR
so the queue's fixed per-PR toll is paid once. Fences are disjoint:
`scripts/fence-lint.sh 87 91` PASSED, zero overlaps, re-run on the final head.

## What becomes true

**#87** — the law that keeps rotation reachable only from declared callers now
DERIVES its guarded set from `rotate.ts` and the barrel instead of carrying a
hand-maintained list, and it enforces three door shapes it previously could not
see: a value export in any declaration form, a door republished as a bare
default through the barrel, and the `reserveInFlightForTest` hook (#92, absorbed).

**#91** — the 409 `retarget-in-flight` code is one value in `api/refusals.ts`
that both routes import. Mutating it fails `npm run typecheck` (TS2820). The
test's private compensating copy is gone.

## Gates

Run on this exact head, node 22.22.2 — not inherited from the lanes:

    npm run lint                      exit 0   (biome, 850 files)
    VITEST_MAX_WORKERS=6 npm test     exit 0   381 files, 6625 passed,
                                               3 expected fail, 1 skipped
    scripts/fence-lint.sh 87 91       PASSED, 0 overlaps

## What is the sibling case?

This PR went through four verification rounds, and **every finding after the
first was the sibling of the previous round's repair.** That is worth stating
plainly rather than burying, because it is the shape `AGENTS.md` names:

| round | finding | its sibling, found next round |
|---|---|---|
| 1 | the floor asserted 6 of 10 derived names | — |
| 2 | the clause parser skipped `type` but not `default` | — |
| 3 | ...and the fix guarded only the SOURCE side of `as`, not the alias side | round 2's own repair |
| 4 | ...and the import arm counted type-only imports as callers | the export-side fix, on the import side |

Round 4 closes the matching pair on the import side — value-vs-type, and
named-vs-default — so both arms now mean the same thing by "reaches rotation".

**One asymmetry is deliberate and stated in the code:** the NAME arm is still
text-level and trips on a bare comment mentioning a door. It has always been a
grep; an arm claiming to model reachability has to mean it. `api/refusals.ts` is
the file most exposed to that, and its doc comment is worded accordingly.

**One residual is declared, not silently left:** `import * as ns from '<barrel>'`
then `ns.default` escapes both arms. Catching it honestly needs member-access
analysis rather than an import-shape regex. It is narrower than the gap this PR
closes and is written into the source where the next reader will meet it.

Every new guard was mutation-proven, and **each mutation was confirmed applied by
checksum before its run** — a mutation that silently fails to apply reports green
and reads as proof.

    delete the barrel-default arm        -> barrel-default test only, red
    revert the import edge to type-blind -> type-only test only, red
    arm the barrel edge unconditionally  -> disarmed-control test only, red
    a pinned door leaves rotate.ts       -> floor test, red, naming the door
    remove the bare-`default` guard      -> parser test, red
    remove the alias-side guard          -> parser test, red
    mutate the shared refusal value      -> TS2820 in refusals.ts alone

What they would survive: the namespace-import reach declared above. Nothing else
in either arm is unpinned.

## Evidence worth a reviewer's eye

The barrel-default arm names **exactly one file** — the offender. The behaviour
it replaces named **49 innocent files and left the offender out**, so the law
went red for 49 wrong reasons and silent about the real one. Controls: no
offender → green; the same import shape against a barrel with a named re-export
→ green (arm disarmed); a named import of the carrying barrel → green.

## Fence widening, recorded

**#91, at grooming, before the change.** Its filed fence ended with "the new
shared module, wherever the namespace law permits it to live" — a boundary the
worker would have had to invent and the landing gate would then reject. Pinned to
`packages/server/src/api/refusals.ts`, and `api/retarget-law.test.ts` added
conditionally in case the law had to name the module. It did not; that file is
untouched.

No widening on #87 — all four rounds stayed inside
`packages/server/src/recorder/namespace-law.test.ts`.

## Verification

Five-seat `/verify` before this PR opened: two independent initial passes (Opus
and Codex gpt-5.6-terra, both xhigh, isolated worktrees, blind to each other),
two cross-reviews, one reconciler with repo access told to settle disputes by
experiment rather than vote. Then a section-12 re-review of the fixes on the
amended head, because the first pass had reviewed a commit the lane then amended
by +341/-22 — those lines had never been reviewed by anything.

The seats split on the merge verdict twice, and both splits were resolved by
running something rather than by picking a side. Full reports are on #87 and #91.

Two commit bodies carried stale suite counts (6591 and 6614 against an actual
6625). #91's is restated with provenance; #87's sits inside a quoted gate-output
block and is annotated rather than rewritten, because a capture edited to match a
later tree has stopped being a capture.

Builder was sonnet; no review seat was the authoring model.


ciaran-slow:
## Review — verified, one fix pushed as `1fc4417`

Reviewed commit-by-commit in a clean worktree off the PR head, node 22.23.2.

**The #87 definition of done holds, EXECUTED against the real law and the real tree** — not only against the synthetic fixtures. I wrote the original defect back into the tree myself:

```
packages/server/src/server/evil-door.ts:
  import { performRetarget } from '../recorder/rotate.js'
-> 1 failed | 29 passed, and the failure NAMES the file

packages/server/src/server/evil-hook.ts:
  import { reserveInFlightForTest } from '../recorder/rotate.js'   (absorbed #92's door)
-> 1 failed | 29 passed, names the file

CONTROL, clean tree -> 30/30 green
```

Both doors the issue was filed over are genuinely closed, and the absorbed #92 half is enforced by derivation rather than by comment — `reserveInFlightForTest` is at `recorder/rotate.ts:434` and pinned in the floor test at `namespace-law.test.ts:610`.

**#91 matches its DoD exactly.** EXECUTED: no bare `'retarget-in-flight'` literal survives in either route; `rotate.test.ts`'s `SHARED_RETARGET_IN_FLIGHT_CODE` is gone; the `type RetargetRefusalCode` import into `retarget.ts` is *not* dead weight — it is still load-bearing at `retarget.ts:201` and `:215`. The three literals left in `retarget.test.ts` are outside #91's fence and correctly untouched.

**The count provenance reconciles, and the PR was right to flag it.** Measured here: `381 files, 6620 passed, 3 expected fail, 6 skipped (6629)`. The PR body's `6625 passed / 1 skipped` sums to the **same 6629 total** — the delta is five platform-gated tests skipping on this macOS box instead of running. Same population, different split. The body's decision to annotate the stale in-commit captures rather than rewrite them was the right call.

**The declared residual is honestly scoped.** `import * as ns from '<barrel>'` then `ns.default` does escape both arms — and it is *latent*, not live: the real `recorder/index.ts` republishes no bare default today, so nothing is currently reachable through it. Narrow, as claimed.

`scripts/fence-lint.sh 87 91` PASSED, 0 overlaps, re-run on the final head. `gh pr checks 117`: all **7** jobs across both `build-test-boot` and `pack-smoke` actually executed and passed — I checked the job list rather than the bar.

---

### One finding, fixed rather than handed back — `1fc4417`

**EXECUTED.** Round 4 stopped the import edge counting erased imports and proved it with `import type { Rotation } from './rotate.js'`. Its inline twin was still counted:

```
import type { Rotation } from '../recorder/rotate.js'    -> 30/30 green
import { type Rotation } from '../recorder/rotate.js'    -> 1 failed, file named
```

Both are erased by TypeScript identically, so the second is the law refusing a file that provably cannot invoke anything — the exact failure round 4 existed to remove.

Not theoretical. The inline form is this repo's dominant spelling (**234 files** under `packages/*/src`), and `biome.json` sets `preset: none` with no `useImportType` rule, so nothing pushes an author toward the guarded spelling.

This is the same pair-shape the `as` guard had to close twice: one spelling handled, its sibling left open. `typeOnlyImportRanges` now also ranges a named clause whose every specifier is tagged `type` — position ranges, not a specifier blacklist, for the reason the separated arm already gives. `\s*\{` cannot match `import X, {` or `import type {`, so a default binding keeps its value edge and the separated form is not double-counted.

Both spellings are pinned in one test so neither can regress alone, and the mixed clause is a CONTROL in both directions — `{ type A, realDoor }` must still be caught, which a whole-statement erasure would have got wrong.

Mutation-proven, mutation confirmed applied by checksum before the run and the checksum reconfirmed after revert: removing the guard reddens the new test **alone**. Regression-checked that the fix does not weaken the law — `performRetarget` from outside still goes red and names the file.

Swept for the same class in the other clause readers: `namesFromExportClause` already filters `/^type\b/` per specifier and `defaultImportSpecifiers` carries an explicit `(?!type\b)`. Both were already per-specifier; this arm was not. The class is now confined.

Gates on the pushed head: `npm run typecheck` exit 0 · `npm run lint` exit 0 (biome, 850 files) · `VITEST_MAX_WORKERS=6 npm test` exit 0, 381 files, 6620 passed, 3 expected fail, 6 skipped.

---

- `rotationEntryPoints`'s declaration regex needs `\s+` between `function` and the name, so `export function* gen()` derives no door. No generator door exists today; noting it as a form the derivation does not yet cover.
- The name arm still reads **raw** text while the edge arms read `codeOf`, so a bare comment mentioning a door trips the law — already parked in the commit body and the reason `refusals.ts`'s doc comment had to be reworded. Parked deliberately; I am not reopening it.

---

### What I did NOT run

`scripts/gate.sh` — that is the operator's landing step, not a reviewer's. No Windows leg exists in `ci.yml`, so nothing here was exercised on Windows.

**Not merging.** My fix is substantive enough to want the author's eyes: it changes what the law refuses. Over to @gabriel-canaan to confirm `1fc4417`, then the operator lands it.

ciaran-slow (changes requested):
Blocking only on author confirmation — not on unresolved defects.

Full evidence is in the review comment: https://github.com/launchpad-26/rhizomorph/pull/117#issuecomment-5420509848

**Verified in the author's favour, EXECUTED against the real law and the real tree:** both #87 doors close (`performRetarget` and the absorbed #92 `reserveInFlightForTest` each turn the law red and name the file; clean tree 30/30). #91 matches its DoD exactly. Counts reconcile — same 6629 total, the split differs by five platform skips. Fence lint PASSED, 0 overlaps. All 7 CI jobs actually executed.

**One finding, already fixed rather than handed back:** `1fc4417` — the erased-import arm counted `import { type X } from …` while exempting `import type { X } from …`. Both erase identically, so the law refused a file that provably cannot invoke anything, which is the failure round 4 existed to remove. This repo's dominant spelling (234 files) and unenforced by biome. Mutation-proven, checksum-confirmed, regression-checked.

**What this review is waiting on, and it is one thing:** `1fc4417` is my commit and it changes what the law refuses. That is the substantive rung — it does not land on my say-so, and I will not approve a branch I pushed code to.

@gabriel-canaan: confirm or rebut `1fc4417` and I will clear this immediately. Both remaining items in the comment are marked "your call" and neither blocks.

ciaran-slow:
@gabriel-canaan — short version, because the two posts above are long.

**Nothing is broken and there is no work to redo.** Your two commits verified. The changes-requested state is procedural: I pushed a fix to your branch and I will not approve a branch I wrote code on.

### The fix is already on your branch

`1fc4417` is pushed, green, and sitting on top of your two commits. You do not need to write anything.

```bash
git fetch origin && git show 1fc4417
```

**What it changes:** `typeOnlyImportRanges` in `packages/server/src/recorder/namespace-law.test.ts` now also treats `import { type X } from '…'` as erased, alongside the `import type { X } from '…'` your round 4 already handled. Both erase identically, so the inline form was making the law red on files that provably cannot call anything.

**Why it mattered enough to fix rather than file:** the inline spelling is this repo's dominant one (234 files under `packages/*/src`), and `biome.json` sets `preset: none` with no `useImportType` rule, so nothing steers an author to the guarded spelling.

**Reproduce in one step** — before the fix the second line went red naming the file, after it, green:

```
import type { Rotation } from '../recorder/rotate.js'    -> 30/30 green
import { type Rotation } from '../recorder/rotate.js'    -> was 1 failed, now green
```

### What you need to do — one of two things

1. **Agree** → comment saying so. I clear the review, and it goes to the operator to land.
2. **Disagree** → say why. `1fc4417` is one self-contained commit on top; dropping it leaves your two commits byte-identical to what you pushed.

That is the whole ask. The two items in my first comment marked *your call* — the `export function*` form and the raw-text name arm — are **not** blocking and need no answer to merge.

gabriel-canaan:
## Grammar coverage for both import arms — 24 cells, EXECUTED against `1fc4417`

**Not a review.** `#117` is being reviewed in another session and this does not
compete with it. This is the one thing that review has not had: the **complete
cell list** for the two arms, rather than the cells whoever probed last happened
to think of.

### Why a table instead of more probes

`1fc4417` is the **fifth** finding of one shape on this file:

| round | cell closed |
|---|---|
| 2 | export clause: `type` handled, `default` not |
| 3 | `export { x as default }` — the *other* side of `as` |
| 4 | import arm: value vs type |
| 5 (`1fc4417`) | `import type { X }` handled, inline `import { type X }` not |

Every one was a **pair** — one spelling handled, its twin open. Four rounds of
seats, each individually well-run with real probes and controls, each answering
"is this correct?" about inputs it invented. Nobody enumerated the productions,
so the fifth cell was always going to be found by the next reader rather than
the current one.

`1fc4417`'s own body says it plainly: *"this IS one — the same pair-shape the
`as` guard had to close twice."* The table is what stops there being a round six.

### Arm 1 — the direct import edge (`valueImportSpecifiers`)

A probe file in `server/` importing from `../recorder/rotate.js`. Type name
`Rotation` is deliberately **not** a derived door, so the NAME arm cannot
confound the IMPORT arm — otherwise every row reads "flagged" for the wrong
reason. Value positions use `notADoorValue` for the same reason.

| spelling | expected | actual | |
|---|---|---|---|
| `import type { X } from R` | not flagged | not flagged | OK |
| `import { type X } from R` — inline, all-type | not flagged | not flagged | OK |
| `import { type X as Y } from R` — inline alias | not flagged | not flagged | OK |
| `import type X from R` — separated default | not flagged | not flagged | OK |
| `import type * as N from R` — separated namespace | not flagged | not flagged | OK |
| `export type { X } from R` — type re-export | not flagged | not flagged | OK |
| multi-line `import type {\n X,\n}` | not flagged | not flagged | OK |
| multi-line `import {\n type X,\n}` | not flagged | not flagged | OK |
| `import { type X, valueName } from R` — **MIXED** | flagged | flagged | OK |
| `import Default from R` | flagged | flagged | OK |
| `import * as N from R` | flagged | flagged | OK |
| `import R` — side-effect only | flagged | flagged | OK |
| type stmt **and** value stmt, same module | flagged | flagged | OK |
| `await import(R)` — dynamic | flagged | flagged | OK |
| `require(R)` | flagged | flagged | OK |
| `export { valueName } from R` — value re-export | flagged | flagged | OK |
| CONTROL — ordinary file, no rotate import | not flagged | not flagged | OK |

**17/17.** Both directions are represented, which is what makes it a table and
not a green bar: eight cells must stay silent and eight must fire, so a harness
that flagged everything or nothing would show up immediately. The MIXED row and
the two-statement row are the load-bearing ones — a specifier blacklist passes
the first eight and fails both of those.

### Arm 2 — the barrel-default edge (added in round 4, and NOT reviewed by any seat)

Fixture: `rotate.ts` gains an anonymous `export default`, the barrel republishes
it **bare**. That is the only configuration in which this arm arms itself.

| spelling | expected | actual | |
|---|---|---|---|
| `import X from BARREL` — the arm's target | flagged | flagged | OK |
| `import X, { n } from BARREL` | flagged | flagged | OK |
| `import X, {\n n,\n} from BARREL` multi-line | flagged | flagged | OK |
| `import type X from BARREL` — erased default | not flagged | not flagged | OK |
| `import { named } from BARREL` — binds no default | not flagged | not flagged | OK |
| `import * as N from BARREL` — `N.default` reachable | not flagged | not flagged | **declared residual** |
| `export { default } from BARREL` — chained re-export | not flagged | not flagged | **declared residual** |

**7/7 as documented** — including the two that are *deliberately* not caught.
Both are written into the source at the arm's definition, not left to be found:
catching them needs member-access analysis and re-export-chain following rather
than an import-shape regex. They are narrower than the gap the arm closes, and
they are the honest answer to "what would this test survive".

### What this does and does not establish

- **Both arms are complete for every import and export shape TypeScript has**,
  on this head. There is no cell six of the same kind waiting.
- It says nothing about the arms the earlier rounds covered — the export-clause
  parser, `escapeForAlternation`, the resolver. Those had their own rounds.
- It is **REASONED** on one point only: that the two residual cells are
  genuinely unreachable *today*. Reaching either needs a default export in
  `rotate.ts`, which does not exist — so the fixture proves the behaviour, not
  the absence of a caller.

### Method, for anyone re-running it

Two throwaway scripts, one file written per cell into `server/`, the law run per
cell, the file removed, and the worktree confirmed clean at the end (`rotate.ts`
and `index.ts` were mutated for arm 2 and restored — `git status` empty after).
Node 22.22.2. Nothing was pushed and no seat was spawned.

The one thing worth carrying forward regardless of this PR: **for a parser, the
reviewable unit is the grammar, not the diff.** Mutation testing cannot find a
missing cell — mutate the handling of a cell you already thought of and its test
goes red, proving that cell and saying nothing about its twin. Coverage of an
incomplete grammar reads exactly like coverage of a complete one.

gabriel-canaan:
@ciaran-slow — **`1fc4417` confirmed.** Not on your say-so: re-derived here, in a clean worktree off the PR head, node 22.22.2, each command run separately and its own exit status read.

### Confirming `1fc4417`

| probe, real tree / real law | result |
|---|---|
| clean tree | 30/30 green |
| value import of `performRetarget` from outside | **red**, names the file |
| `import { type Rotation } from '…/rotate.js'` | green ← your fix |
| `import type { Rotation } from '…/rotate.js'` | green ← round 4 |
| `import { type Rotation, performRetarget }` (mixed) | **red**, names the file ← the control holds |
| **mutation**: delete the inline arm | the new test red **alone**, 1 failed / 29 |
| same mutation, plus a real inline-type-import file | **red** — so the defect was live, not latent |

Mutation confirmed applied by checksum before the run and the original reconfirmed after revert. Gates on that head: `npm run typecheck` **0** · `npm run lint` **0** (850 files) · `VITEST_MAX_WORKERS=6 npm test` **0**, `381 files · 6625 passed | 3 expected fail | 1 skipped (6629)` — the PR body's split exactly, and the same 6629 total your macOS `6620 | 3 | 6` reconciles to. CI run for `1fc4417` carries that head sha and all 7 jobs executed.

The fix is right and it does not weaken the law. Nothing to rebut.

### Your first "your call" item is now fixed — `dbdd12d`

The `export function*` gap. I took it as more than a preference, because it makes a shipped claim false: `4e89831`'s subject says the law "derives its doors from **every value-export form**" and this PR's body says "**any** declaration form", while the derivation covered five of six. Under prd-43 the claim is a test, so the code moves rather than the sentence.

```
export function realDoor() {}      -> realDoor   (before and after)
export function* genDoor() {}      -> MISSED  ->  genDoor
export async function* agen() {}   -> MISSED  ->  agen
export function *spaced() {}       -> MISSED  ->  spaced
```

The `function` branch is split off rather than folded into one lazier `function\s*\*?\s*`, and the rejected alternative is **executed, not asserted**: the lazy form derives `foo` from the non-declaration text `export functionfoo`; the shipped split derives nothing from it.

All three spellings are pinned in the forms test — that being the sibling case, and the same shape as the `as` guard and your erased-import arm — beside a `function* notExportedGen() {}` CONTROL whose absence the exact-set assertion is what proves. Plus one EXECUTED law in the shape of the `export let` one: a generator door turns the law red and names the importer, with an innocent file green.

**Mutation:** reverting the regex reddens exactly those two tests and nothing else (2 failed | 29 passed), checksum-confirmed applied and reverted. Swept the other clause readers — `namesFromExportClause` and `defaultImportSpecifiers` parse specifiers, not declarations, so neither has a `function` branch to miss.

Gates on `dbdd12d`: typecheck **0** · lint **0** (850 files) · full suite **0**, `381 files · 6626 passed | 3 expected fail | 1 skipped (6630)` — one more than `1fc4417`, which is the new law. No fence widening; `namespace-law.test.ts` is #87's declared fence and the only file touched.

### Your second item stays parked, agreed

The raw-text name arm. Confirmed EXECUTED that it is as broad as you say — a file with **no import at all**, carrying only a comment that names a door, goes red and gets named. It is documented at `namespace-law.test.ts:238` and is why `api/refusals.ts`'s doc comment is worded as it is. Not reopening it here.

Over to you to clear the review — note `dbdd12d` is a third commit you have not seen, so re-read rather than clearing on the old head. Landing is the operator's step either way.

PR #118, opened 2026-08-26
prd40 w6: a fsync failure cannot leave a line after session.closed

### LINKED ISSUE #80: prd40 w6: a fsync failure cannot leave a line after session.closed

ciaran-slow:
prd-40 wave 6, and the **last of the three issues contending for `recorder/session-recorder.ts`** — #26 (wave 4, #105) and #69 (wave 5, #114) landed before it. Single-issue wave; branch is `80-close-is-append-not-sync` per `AGENTS.md`'s `<issue>-<slug>` convention.

Closes #80

Implements prd-40 ruling 1's **fifth amendment**, landed separately in #116 *before* this was dispatched — so the lane built against a criterion the repo carries rather than one living in a comment thread.

---

## The defect

`closeWith` wrapped `append` **and** `sync` in one `try`, and the `catch` released the seal without asking which had failed. But an append that resolves has already put `session.closed` into the file. So a sync that failed afterwards released the seal on a session that was **closed on disk**, and a later `record()` appended behind the close line — breaking prd17 ruling 1's structural guarantee, undetectably, because `verifyRecord` computes the chain over the body as given.

Pre-existing on `main`, reproduced independently by three reviewers.

## What is ruled, and what it cost

- **1a** — a failed **append** releases the seal and rejects. Nothing reached the file. Unchanged.
- **1b** — a failed **sync after a resolved append** does **not** release the seal. The close happened. `closeWith` still rejects, via a new exported `CloseNotDurableError`, so the lost durability reaches the caller — but the seal it earned stays, and only `openSession` releases it.
- **1b-i** — `closeCurrentSession` absorbs *only* `CloseNotDurableError` and returns a `ClosedSession` carrying `synced` / `syncError`. Everything else still throws. Nothing is reconstructed: `sessionId`, `filePath`, `closedAt` and `eventCount` are captured at `rotate.ts:120-123` before the first `await`.
- **1b-ii** — `removeSessionLock` still runs. The lock guards a live session, and the session is not live.
- **1b-iii** — `/api/rotate` answers **200** with the durability loss in the body.

**Nothing pushes, folds or emits on the 1b path.** Today's ordering already rethrew before those three, leaving the fold *behind* the file — the safer side of the same disagreement — and that is deliberately preserved. Rule 3 is untouched: past a durable close, a throwing subscriber is still reported and never propagated.

## Three files this did NOT need to touch, which is the design working

`rotateSession`, `performRetarget` and `api/rotate.ts` are **unedited**. Once the close stops throwing on a sync failure, the open runs on its own — "reopening becomes unconditional" turned out to require no conditional anywhere. And `api/rotate.ts` already returns `{ closed: rotation.closed, … }`, so the new field reaches the body at 200 automatically. `api/rotate.ts` is not in the commit at all; inside `recorder/rotate.ts` those identifiers appear in the diff only in comments.

All three still carry laws, because *correct by construction* is a claim and not a proof.

## Gates

Base confirmed on top of current `main` (`5b04c49`). Node v22.23.2. Each gate run **separately and unpiped, reading its own exit status** — never a pipeline's — after proving the harness could fail at all (a deliberate `TS2322` exited 1, then was removed).

`npm run typecheck` **0** · `npm run lint` **0** · `VITEST_MAX_WORKERS=6 npm test` **0** · `npm run build` **0**. Boot smoke served.

**Totals, not passes: `381 files` · `6624 passed | 3 expected fail | 6 skipped | 6633 total`** on macOS. A Linux run of this tree reads `6619 | 3 | 11 | 6633` — five tests are `it.skipIf(process.platform === 'linux')` (`static.test.ts:525`), which two reviews in a row have spent time reconciling. The total is the like-for-like number.

## Fence

`recorder/session-recorder.ts` + test (declared) · `recorder/rotate.ts` + test (widened by ruling #116, recorded on the issue) · **`api/rotate.test.ts` — new widening, test only**, for the 200 law. `api/rotate.ts` itself is not opened.

**Note for scheduling: #53 (prd42 w6) also claims `recorder/rotate.ts`.** It is undispatched, so nothing collided — but the two must never run concurrently, and `scripts/fence-lint.sh 53 80` hard-fails on the overlap. Recorded on both issues; #53 rebases onto this.

---

## What is the sibling case?

**`performRetarget`.** It has the same close-then-open shape as `rotateSession` but a different entry point — `api/retarget.ts` calls it directly, never through `rotateSession`. A fix proven only through the rotation path leaves retarget wedging on a failed sync.

**Law R8 drives it, and mutation M6 is what earns R8 its place:** fixing `rotateSession` alone while leaving `performRetarget` throwing turns **R8 red and nothing else**. That row exists only to prove the sibling law is not decoration.

Six rows, run against the real source and reverted between:

| mutation | predicted | actually red |
|---|---|---|
| release the seal on the 1b path | R1 | R1 + R5 (wider) |
| `closeCurrentSession` rethrows `CloseNotDurableError` | R5, R7, R8 | R5, R7, R8 + R9 (wider) |
| absorbs *every* close error | R6 | ✅ R6 |
| skip `removeSessionLock` on a failed sync | R5 | ✅ R5 |
| route 500s when `synced` is false | R9 | ✅ R9 |
| fix `rotateSession` only | R8 only | ✅ **R8 only** |

Every predicted row went red; two went wider. No law was weakened to make a row match.

**And the first run of that table was fiction — reported rather than buried.** It used `vitest run --reporter=basic`, a reporter that does not exist in vitest 4.1.10, so all six returned exit 1 from the *reporter* failing to load before a single test ran. A tidy six-for-six table over source that was never exercised. Re-run with the flag removed; the numbers above are the real ones.

Independently confirmed on this branch:

```
with --reporter=basic  → exit 1, "Startup Error: Failed to load custom Reporter from basic"
without the flag       → exit 0, 31 passed
```

That is the third time in this milestone a signal came from the harness rather than the code — after the `tail`-pipeline typecheck on wave 4, and a gate report that read a pipeline's status instead of the command's.

### R3, the DoD's named mutation

The Definition of done required proving that *"a test that only rejects `append` would pass while this stayed broken"*. Written before any source change — R1's shape but mocking `append` instead of `sync` — and run against the **unfixed** code: **exit 0, 1 passed.** Vacuously, because a failed append leaves no `session.closed` for anything to land behind. Then deleted; it is not in the commit. That is why the decisive law drives `sync`.

## Deviations

None structural. Two things named rather than left for review:

- Two pre-existing `toEqual` assertions on the whole `closed` object now also assert `synced: true` on the happy path — **extended, not weakened**, and `toEqual` is deliberate so the *absence* of `syncError` is asserted too.
- Two stale comments in `session-recorder.ts` corrected in the same edit: the block explaining that `rotate.ts` reaches `openSession` only once the close resolves now says which single rejection is absorbed and that all others propagate. Rule 3 itself untouched.

No ADR — this implements a recorded PRD ruling and makes no new decision. No `CHANGELOG.md` entry; flagged as a judgement rather than added silently.


gabriel-canaan (approved):
## Approve — the fix matches the ruling, and the laws are load-bearing

Reviewed commit-by-commit (one commit, `7d70b99`) against #80 and prd-40 ruling 1's fifth amendment as it stands on `main` at `5b04c49`. Everything below is **EXECUTED** unless labelled otherwise: a clean worktree at the head SHA, node v22.22.2, each command run separately and its own exit status read.

**Gates, re-measured rather than taken from the body.**

| command | exit | result |
|---|---|---|
| `npm run typecheck` | 0 | — |
| `npm run lint` | 0 | 849 files, no findings |
| `VITEST_MAX_WORKERS=6 npm test` | 0 | `381 files` · `6629 passed \| 3 expected fail \| 1 skipped (6633)` |

`gh pr checks 118`: **both** jobs, all seven legs green — `build-test-boot` ×3 and `pack-smoke` ×4. Checked as two jobs, not one bar.

**The laws are not decoration — four mutations, run here, source reverted between each.**

| mutation | red |
|---|---|
| **both source files reverted to `origin/main`, new tests kept** | **7 tests, 3 files** — R1, R5, R6-adjacent, R8, R9 and both extended happy paths |
| release the seal on the 1b path (`session-recorder.ts`) | 2 — the seal law **and** `closeCurrentSession`'s `isSealed` clause |
| drop `if (!(error instanceof CloseNotDurableError)) throw error` | 1 — the rule 1a law, exactly and only |
| narrow `api/rotate.ts` to return `closed` field-by-field | 2 — including the 1b-iii law, which is what it claims to exist for |

The first row is the one that matters: **the new laws all fail against the unfixed source.** The last row confirms the 1b-iii law's own stated purpose — "this law is what stops someone narrowing that return later" — is true rather than asserted.

**The design claim holds on inspection.** `openSession` (`packages/server/src/recorder/session-recorder.ts:363`) resets `buffer` and `foldState` outright, so the 1b path's skipped `buffer.push` / `advanceFold` leaves buffer and fold consistent *with each other* and both discarded a moment later — the fold/`eventsSoFar` deep-equal invariant is never crossed. `openNextSession` reads only `closed.sessionId`. **[REASONED, from the source]**

**One wedge I went looking for and did not find.** On the 1b path `closeCurrentSession` returns with the seal held, and only `openNextSession` → `openSession` releases it — so anything throwing *between* them would strand it. The only statement there is `removeSessionLock`, and `packages/server/src/log/session-lock.ts` wraps its `unlink` in a bare `catch {}`: it cannot throw. `openNextSession` calls `recorder.openSession()` as its first statement, before any I/O. There is no new stranding path. **[REASONED, from the source]**

**Scope is exactly what was ruled.** The PRD's "In scope for #80" names `closeCurrentSession`'s return shape, the unconditional open, the lock's placement, and the field `/api/rotate` reports — all four present, nothing beyond. The five touched files match the declared fence plus the two recorded widenings. No home paths, usernames or hostnames in the diff; no NUL bytes.

### Two notes — offered, not asserted; neither blocks

1. **The PR body's Linux figures do not reproduce.** The body says *"A Linux run of this tree reads `6619 | 3 | 11` — five tests are `it.skipIf(process.platform === 'linux')` (`static.test.ts:525`)"*. On Linux this tree reads **`6629 passed | 3 expected fail | 1 skipped`**, and there is exactly **one** `skipIf(process.platform === 'linux')` in `packages/server/src/server/static.test.ts` — line 525, the one cited. The six the other way (`skipIf(platform !== 'linux')`) are what make the macOS run read `6624 | 3 | 6`, which does reconcile. The total, 6633, is right and the conclusion ("the total is the like-for-like number") is right — only the Linux breakdown and its explanation are wrong. Worth correcting because the body offers it precisely so the next reviewer stops re-deriving it.

2. **`api/retarget.ts` carries `synced` with no law under it.** `packages/server/src/api/retarget.ts:321` returns `retarget.closed` whole, exactly as the rotate route does, so 1b-iii's fact reaches that body too — by construction, and `api/retarget.test.ts` only ever asserts `body.closed.sessionId`. That is *outside* #80's declared scope (the PRD names `/api/rotate` and nothing else), so it is correctly not in this PR — but it is the same narrowing this PR added a law against one door down. A follow-up issue rather than a change here.

### Verdict

Approving. Landing is the operator's step (`scripts/gate.sh`), not mine — this is not merged.

PR #119, opened 2026-08-26
prd46 w1: the honesty law recognises a swallowed failure by structure

### LINKED ISSUE #70: prd46 w1: the honesty law recognises a swallowed failure by structure
## The defect

`packages/server/src/gate-honesty-law.test.ts` recognises a swallowed failure by matching one of **three literal strings**. A guard written with a fourth spelling is invisible to it, and one such guard is live in the file it governs.

Measured during the verification of PR #68. Planting a plausible new guard after gate.sh:116 and running the law:

| planted line | law |
|---|---|
| `some_new_check 2>/dev/null` *(control)* | **1 failed** — names the line |
| `some_new_check \|\| :` | 45 passed |
| `some_new_check 2> /dev/null` | 45 passed |
| `some_new_check &>/dev/null` | 45 passed |
| `some_new_check \|\|true` | 45 passed |
| `RES=$(some_new_check \| tail -1)` | 45 passed |
| `some_new_check \|\| { echo "  (could not check)"; }` | 45 passed |

**The live instance:** gate.sh:82 captures `n=$(git log --oneline main..HEAD | wc -l)` and never reads the pipeline's status. Executed against a repo with `.git/HEAD` removed: `fatal: not a git repository`, rc 128, `n=0`, and the gate holds with *"no commits on the branch (a worker may have left work uncommitted)"* — a fault it did not earn, between two lines prd-45 wave 1 fixed. The law reads green with it there.

## Definition of done

Rulings 1 and 2 of prd-46. **The fix and the predicate are ONE issue** because the predicate convicts gate.sh:82 — the law cannot be green while that line stands, which is a stack, and the wave contract allows a stack in a wave only as a single issue. Land in commit order: the :82 fix first, then the predicate. prd-45 wave 1 booked its own rulings 1–3 exactly this way.

- gate.sh:82's producer status is read before any verdict is printed, routed through the existing `fail()` — reused, never reimplemented.
- The sweep is replaced by a **structural predicate over the script's shape**: a command substitution or pipeline whose exit status is never read before a verdict prints. The reference form is *a `$(...)` assignment whose following line is neither `_RC=$?` nor `|| fail`*.
- Each of the six evasions above fails the new law. **Prove each one**, and prove them with spellings beyond the six — a predicate verified only against the inputs that motivated it has been narrowed, not generalised.
- The nine existing declared tolerances still pass, each keeping its reason. A tolerance becomes an exemption from a rule, not a member of a matched set. The two dated KNOWN GAP entries stay as they are.
- Every control exercises the mechanism it controls for (ruling 2). The current non-vacuity control asserts a rigged string fails to match a declared needle and never runs the sweep; that is what ruling 2 forbids.
- The law's line-extraction helpers (`uniqueLineIndex`, `sliceLines`) and its extract-real-lines-from-the-tracked-script technique are **kept** — they are right, and this issue does not rewrite them.

`scripts/gate.sh` **is** the landing: it merges into local main, installs, builds and ends with `git push origin main`. **Never run it.** Exercise its logic in a scratch directory instead, and label everything EXECUTED with exit statuses.

State the false-positive rate you measured. prd-46's open question asks whether the structural predicate is tolerable on this file, and it is not rhetorical: a rule that convicts legitimate lines faster than dishonest ones gets suppressed into uselessness, which is how the enumeration this replaces came to exist. If your predicate convicts lines that are honest, say how many and which, and either refine the rule or declare them — do not quietly widen the tolerance table back into a spelling list.

## Fence (may touch ONLY)
- `scripts/gate.sh`

## Blocked by
prd-45 wave 1 landing (PR #68) — this edits both files that wave writes. Also gated on prd-46 wave 0, which is the operator's own act: work that edits gate.sh cannot be landed by gate.sh.
The CI workflow belongs to prd-45 and is EXPLICITLY OUT. Route classes in api/index.ts are prd-29's.

gabriel-canaan:
Closes #70

prd-46 wave 1, plus five commits of review repair. `scripts/gate.sh` **is** the
landing tool, so this cannot be landed by `gate.sh` — it needs the operator's
manual merge (prd-46 wave 0, booked not skipped).

## What lands

`scripts/gate.sh` changes **four executable lines**. Everything else there is comment.

|---|---|
| `a8fead7` | `n=$(git log … \| wc -l)`'s pipeline status is read before the commit-count verdict |
| `0e1b11e` | the honesty law convicts a swallowed failure by **structure**, not spelling |
| `ebda3f2` | the `dirty=` pipeline holds when it could not run |
| `815b4d2` | the law **derives** the shell option its `:96` guard depends on |
| `e472911` | an exit status that wraps to zero is a fake success |
| `a0e3712` | the wrap rule agrees with bash at **every** literal, not just small ones |
| `1354a76` | the dirty-count fixture removes a **middle** stage; the option anchor survives prose |
| `cedcf6c` | the anchor sees a later `set +o`; the PATH helper's claim is true |

The last five answer an independent verify run (two model seats, their cross-reviews,
a reconcile pass, then a re-review of the fixes and a confirming pass). Three of those
five exist because a repair opened the next hole — recorded here rather than smoothed over.

## What is the sibling case?

Ruling 3 is answered in every commit body. The load-bearing ones:

- `:96`'s twin is the `dirty=` line a few lines below. It was **declared** in `#70`, on a
  reason whose second half was false — "`-gt 1` would never fire, the pattern is a fixed
  literal and cannot be invalid". Pattern validity is not the only route to a >1 status.
  EXECUTED, real GNU grep, nothing shimmed: `gate.sh` rewrites `PATH` itself at `:14`, and
  grep absent from `PATH` while git still resolves gives rc 127, `dirty` empty, and the old
  code printed `worktree clean` over genuinely stranded work. Now fixed, not declared.
- The wrap rule's only sibling site is the same function's `|| fail` branch, which routes
  through `fail()` and cannot wrap. There is no third.
- `type -P` vs `command -v`: EXECUTED, this is the only such call site in the file.
- `uniqueLineIndex`'s comment-blindness affects every other anchor too. Left as-is,
  deliberately — each of those is code text no comment would reproduce, which is exactly
  why the `set …` anchor needed different treatment rather than a change to the shared helper.

This is the question `#70`'s own PR body did not carry, and answering it found the two
worst defects here.

| mutation | before | after |
|---|---|---|
| `gate.sh:13` → `set -u` | **94/94 green**, and `#70`'s original wrong verdict restored verbatim | rc 1, naming the `:96` guard that dies without `pipefail` |
| `set +o pipefail` on its own line above `:96` | **104/104 green** while the real gate reports "no commits on the branch" | rc 1, `found 2: [...]` |
| whole `PATH` blanked in the `:113` fixture | passed identically with and without `pipefail` — **blind** | fixture removes a middle stage; 1 failed without, 7 passed with |
| `isHonestAbortStatus` → `Number(digits) % 256` | passed its own tests | fails, naming `exit 9007199254740993: bash status 1 (honest) but the predicate says fake success` |
| delete `DIRTY_RC=$?` | — | 5 red, incl. **the structural predicate itself** |

The exit-status test does not encode a table of which literals wrap: it runs
`bash -c "exit <n>"` per literal and requires the predicate to agree with the status bash
actually produced, on both the same-line and rescue-block paths, with a control asserting
the table carries both verdicts. A seat fuzzed it over 5,108 literals — 0 disagreements,
against 2,258 for the `Number()` version it replaced.

## False-positive rate (prd-46's open question)

16 `$(...)` assignments in `gate.sh`; **1** flagged (`:23`, declared, its verdict is the
next line's existence check); **0** undeclared. The other 15 split 9 same-line / 6
next-line. All four counts are now **pinned** — previously only the totals were, so the
pins guarded the numbers that were right and let the wrong ones rot.

## Known, and deliberately not fixed here

The predicate does not see `export X=$(…)`, `local`, backticks, arrays, `echo x; X=$(…)`,
multi-line `$(`, or a `_RC=$?` captured and never read. EXECUTED: planting three of these
in the real `gate.sh` leaves the law green where the control goes red. Every one is
**narrower** than the five evasions this replaces — `|| :`, `2>&-`, `>/dev/null 2>&1`,
`|| exit 0`, `; true` are all now caught — so net coverage is strictly better. Filing as a
prd-46 follow-up rather than widening this PR's scope.

## Fence

Untouched: `scripts/gate.sh` and `packages/server/src/gate-honesty-law.test.ts`, exactly as
declared on #70. `.swarm/coupling.txt` needs an entry for this file's collection-time hold on
`gate.sh` — that is **prd46 w5 (#72)** and is not entered here.

## Checks

381 files, **6645 passed**, 3 expected-fail, 1 skipped · `npm run build` 0 · the law file
4× concurrent at `--maxWorkers=5`: 4/4 green at 104/104.

`scripts/gate.sh` was never run. Work that edits the gate cannot be landed by the gate.


ciaran-slow:
## Reviewed commit-by-commit, in a clean clone, post-merge with `origin/main` at `88c3686`.

I re-derived the load-bearing claims rather than reading them. All of these hold.

**The four shipped executable lines are right, and the `-gt 1` asymmetry is right.** `gate.sh` runs under `set -uo pipefail` and *not* `set -e` (`:13`), so the `[ "$RC" ... ] && fail` idiom is safe and matches the `STATUS_RC` form already in the file. `grep -v`'s no-match exit of 1 is the ordinary clean-tree outcome, so `-ne 0` would indeed have held every clean landing; `-gt 1` has no legitimate trigger. The guard is correctly placed *before* `[ "$dirty" -ne 0 ]`, which would otherwise hit `integer expression expected` on an empty `dirty`.

**The false-positive rate is exactly as claimed.** EXECUTED — I re-derived the split independently from `scripts/gate.sh` with my own parser, not by reading the test:

| | |
|---|---|
| total `VAR=$(...)` assignments | **16** |
| same-line checked | **9** — `:17 :41 :56 :57 :153 :163 :227 :313 :337` |
| next-line `_RC=$?` | **6** — `:74 :77 :96 :100 :127 :228` |
| flagged | **1** — `:23`, declared |
| undeclared | **0** |

Pinning all four rather than just the totals was the right call, and the reason given for it is the real one.

**Both headline mutations reproduce.** EXECUTED:

- `gate.sh:13` → `set -u`: `Error: scripts/gate.sh's option line is "set -u" — it no longer sets pipefail…`, 1 file failed.
- `set +o pipefail` inserted on its own line above `:96`: `expected exactly one \`set\` option line … found 2: ["set -uo pipefail","set +o pipefail"]`, 1 file failed.

**Both RC captures are load-bearing.** EXECUTED, mutations I ran myself rather than the ones claimed:

- delete `DIRTY_RC=$?` → **10 failed**, including the structural predicate itself and four executed fixtures returning 127 where 0/1 was expected.
- delete `N_RC=$?` → **7 failed**, same shape.

**Baseline:** 104/104 green, and the full suite post-merge is 381 files / 6660 passed / 3 expected-fail / 6 skipped, typecheck 0, lint 0.

The commit sequence is genuinely readable, the "three of those five exist because a repair opened the next hole" note is the right thing to record rather than smooth over, and the fence is honoured exactly — only the two declared files.

---

### One defect, found and fixed: `e3a17d0`

`tailChecksStatus`'s **rescue-block branch** scanned the raw block text with `\bfail\b` and `\bexit\s+(\d+)\b`. A string literal satisfies both — so a block that merely *mentioned* failure in its message was credited with handling it.

This is a **spelling match sitting inside the predicate whose entire subject is structure over spelling**, and it is not in the PR's "known, deliberately not fixed" list.

EXECUTED — planted into a copy of the real `scripts/gate.sh` after the commit-count verdict, exactly the way #68's verification planted its six:

```
RES=$(some_new_check) || { echo "  fail to check"; }
```

| | result |
|---|---|
| **before** | ruling 1's own test **GREEN** over it. The only red was the total-assignment pin drifting 16 → 17 — which is my plant, not a detection. `1 failed \| 103 passed` |
| **after** | `undeclared unchecked producer(s) in scripts/gate.sh`, naming index 131. `3 failed \| 106 passed` |

The fix strips quoted runs before the block scan. Its two same-line siblings are safe and I left them alone: `/^fail\b/` and `/^exit\s+(\d+)\b/` anchor at the start of the right-hand side, where no string can precede them. The block branch is the only one that scans.

EXECUTED, mutation — `const body = stripQuotedRuns(rhs)` reverted to `const body = rhs`: **4 failed | 105 passed**, naming all three new rigged cases plus the plant-into-the-real-file test. Restored: **109/109**.

`stripQuotedRuns` is a single left-to-right scan, not two `replace` passes — stripping `'...'` first would let the apostrophe in a double-quoted `"don't"` open a bogus run and swallow the rest of the line. A rigged case covers that exact pairing.

Two **controls** added alongside, because a repair that only removes a credit can pass by refusing everything: a block that really does call `fail` (the `:57` shape, quoted message and all) and one that really does `exit 2` after a quoted message are both still silent. The false-positive pins are unmoved at 16 / 9 / 6 / 1.

---

Neither is a defect and neither blocks.

1. **`X=$(cmd) || exit $?` is flagged**, because `/^exit\s+(\d+)\b/` wants digits. That is an honest abort being convicted. No live instance in `gate.sh`, and erring toward conviction is the right direction — noting it only so it is a known shape rather than a surprise later.
2. **The FP-rate test resolves `LINES.indexOf(l)` by value** against a list built from the *filtered* `codeLines()`. Two identical assignment lines would resolve to the wrong index. EXECUTED: `duplicate assignment lines: 0` today, so there is nothing to fix — it is a latent sharp edge, not a bug.

---

### What I did NOT run

`scripts/gate.sh` itself — for the reason the PR body gives. I also did not run the 4× concurrency probe; I re-ran the law file serially and the full suite once at `VITEST_MAX_WORKERS=6`.

### Landing

**Not approving**, because the only substantive finding here is one I fixed myself, and I am not reviewing my own fix into `main`. `e3a17d0` needs your confirmation.

I also pushed `19f98b8`, a plain merge of `origin/main` into the branch (no conflicts), so CI runs against the actual post-merge tree.

Everything else in this PR verified. As the body says, this cannot be landed by `gate.sh` — it is the operator's manual merge once you have confirmed `e3a17d0`.

ciaran-slow:
Correction on my review above: **#119 had already merged when I posted it.** It merged at 23:29:13; my comment and my push to `prd46` landed about a minute later. My orient step was live, but not live enough — stating it rather than leaving the timestamps to be worked out.

Nothing in the review changes. Everything I verified in this PR's favour holds, and the one defect is now live in `main` rather than fixable here.

- The fix is carried over unchanged in #126, cherry-picked onto `main` at `aa909b3`, 109/109 with the same mutation evidence.
- `e3a17d0` and the `origin/main` merge `19f98b8" I pushed to `prd46` are stranded on that now-merged branch. #126 is the live copy; `prd46` can be deleted.

@gabriel-canaan — no confirmation needed here any more, but #126 is the one to look at.

Issue #120, opened 2026-08-26
prd42 w7: the reverse walk fails closed over the real slug grammar


gabriel-canaan:
## The defect

`packages/server/src/concierge/repos.ts:237` re-encodes directory entries with a
five-character class (`.` `_` `:` `\` and a space). The real Claude Code slugger
maps EVERY non-alphanumeric character — verified against the shipped binary, not
quoted: `grep -a` on the installed ELF (2.1.246) finds
`function B(e){return e.replace(/[^a-zA-Z0-9]/g,"-")}`.

So a repo path containing any other punctuation mints a slug the walk cannot
reconstruct, and the walk does **not** fail closed: with an innocent dash-named
sibling present it returns the sibling — a real but wrong path, silently — and
`listKnownProjects` (`repos.ts:347`) then marks the wrong path resolved.

EXECUTED, slugs minted with the real transform, dash-sibling present:

    a+b  a~b  a@b  a,b  a'b  a(b  a!b  a#b  a%b  a=b  a;b  aéb  a😀b
      -> all SILENT-WRONG-SIBLING

Controls: `aZb` alone -> CORRECT · `aZb` beside `a-b` -> CORRECT · `aZb` absent
-> honest refusal. So this is not a probe artifact.

## Why it is filed rather than fixed in #47

#47 groomed exactly two characters and closed them. It also strictly improved
every "alone" and every dash-sibling cell. But it is NOT a pure narrowing, and
that is the second half of this issue:

**The widening made `:` and `\` attractors.** EXECUTED:

    a+b beside a:b (no a-b)   pre-#47: honest refusal  ->  post: SILENT-WRONG
    a~b beside a\b            pre-#47: honest refusal  ->  post: SILENT-WRONG

So #47 moved the boundary of an existing defect class rather than only shrinking
it. The class predates it — `.` `_` space and `-` were already attractors — and
the net is still safer with #47 than without, which is why it merged. But the
only fix that removes the attractor property is the full grammar, below.

## The fix is one line, and it is proven

`repos.ts:237` -> `entry.replace(/[^a-zA-Z0-9]/g, '-')`

EXECUTED during #47's verification: with that change **every** collision cell in
both probes became an honest ambiguity refusal, every alone cell resolved
correctly (including `😀`, whose two UTF-16 units fold to `--` on both sides),
and the whole server package stayed green — **136 files, 2469 passed**.

Leaving the five-character class is therefore a choice, not a constraint.

## Definition of done

[...]

PR #121, opened 2026-08-26
prd42 w5: a colon and a backslash round-trip, and the law reaches the concierge without a dodge

### LINKED ISSUE #47: prd42 w5: a colon and a backslash round-trip, and the law reaches the concierge without a dodge

gabriel-canaan:
prd-42 wave 5: a colon and a backslash round-trip through the reverse walk, and
the round-trip law reaches the concierge without a dodge.

Closes #47

One issue, one wave, one commit. #47 absorbed #52 at grooming (#115's amendment),
so this is one coherent change rather than a bundle.

## What becomes true

The forward encoder maps `:` and `\` to a dash; the reverse walk mapped only
`.`, `_` and a space. Any repo path with either character minted a slug the walk
could not reconstruct. The walk gains both. **The encoder is deliberately
unchanged** — external evidence establishes it is the correct side: a real Linux
probe for the colon, and a captured Windows-origin slug directory for the
backslash.

The round-trip law also **moves** from `collectors/sessionlog/worktree-slug.test.ts`
to `concierge/repos.test.ts`, where the concierge namespace law's clause 1 skips
it outright — so it needs no computed specifier, no `any` cast and no
hand-written `ReverseProjectSlugResult` copy. That copied type was the same
defect this PRD is named for, one layer up, and it is gone tree-wide.

## Gates

Re-run on this head by me and independently by all three review seats,
node 22.22.2:

    npm run lint                      exit 0   (biome, 849 files)
    VITEST_MAX_WORKERS=6 npm test     exit 0   381 files, 6623 passed,
                                               3 expected fail, 1 skipped
    scripts/fence-lint.sh 47          PASSED

Fence honoured exactly: four files changed, all fence bullets. `worktree-slug.ts`
and `concierge/namespace-law.test.ts` were permitted and left untouched. No
widening needed, none claimed.

## What is the sibling case?

**This PR is not a pure narrowing, and that is the honest headline.** Verification
established, EXECUTED with controls, that the change moves the boundary of an
existing defect class in both directions:

| case | before | after |
|---|---|---|
| `a:b` alone | honest refusal | **resolves correctly** |
| `a:b` beside `a-b` | **SILENT WRONG ANSWER** | **honest refusal (ambiguous)** |
| `a+b` beside `a-b` | silent wrong answer | silent wrong answer (unchanged) |
| `a+b` beside `a:b`, no `a-b` | honest refusal | **SILENT WRONG ANSWER** |

The last row is a regression: widening the class makes `:` and `\` *attractors*,
so an unhandled character beside a colon-named sibling now resolves the wrong
way where it previously refused. It requires a `:`- or `\`-named POSIX directory
beside an alphanumerically identical one carrying different punctuation, with a
session only in the latter — exotic, and the attractor property already existed
for `.`, `_`, space and `-`. Net, every "alone" cell and every dash-sibling cell
improves, so the tree is safer with this commit than without it. But "strictly
narrowed" would be false and is not claimed.

The wider gap and the attractor row are filed together as **#120**
(`prd42 w7: the reverse walk fails closed over the real slug grammar`), with the
one-line fail-closed fix already proven green against 136 files / 2469 tests.
Leaving the five-character class is a choice, not a constraint — and it is a
scope change to a groomed two-character issue, which is why it is a successor
rather than this PR.

Each of the five characters in the walk's class is independently load-bearing —
five mutations, each dropping one:

    /[._ ]/g   (pre-#47)   -> 1 failed, the colon pair
    /[._\ ]/g  (colon out) -> 1 failed
    /[._: ]/g  (bslash out)-> 1 failed
    /[_:\ ]/g  (dot out)   -> 3 failed
    /[.:\ ]/g  (uscore out)-> 2 failed
    /[._:\]/g  (space out)  -> 2 failed

Both new non-vacuity floors bite: removing the colon or backslash entry from the
generator fails at `repos.test.ts:279`/`:280` rather than passing over a set with
nothing in it. And "the only test the change breaks" was checked in the direction
that could embarrass it — base tests against the new class over the whole server
package: exactly **1 failed**, the assertion this commit inverts.

**What it would survive:** the law is generator-bound. It cannot distinguish a
five-character walk from a full-grammar one — a full-grammar walk passes it
unchanged, 41/41. That is #120's job, and it is why #120 says the generator grows
only once both sides agree.

## Verification

Five seats' worth of review before this PR opened: two independent initial passes
(Opus and Codex `gpt-5.6-terra`, both xhigh, isolated worktrees, blind to each
other), then a reconciler with repo access instructed to settle every dispute by
experiment rather than by vote.

The two initial seats **split on the verdict** — one merge, one request-changes —
and the reconciler overturned part of *both*, plus the orchestrator's own read:
it found the attractor regression above, which two review seats and the
orchestrator had all missed after concluding this was a pure improvement. It also
converted the load-bearing premise from the code's own say-so to an external
check: `grep -a` on the installed Claude Code ELF (2.1.246) confirms the real
transform is `/[^a-zA-Z0-9]/g`, rather than trusting the doc comment that quotes
it.

Full reports are on #47. Builder was sonnet; no review seat was the authoring
model.

## Worth knowing, deferred with issues rather than silently

- `forward-transform-law.test.ts:373`, `:631`, `:658` still quote the pre-#47
  class as live code, so that law's unit "does not fire on the reverse walk"
  control now guards a shape absent from the tree. Not a hole — the tree sweep
  at `:663` still covers the live line — and out of this fence. Folded into #120.
- A test-time derived-class law (`walk ⊇ encoder \ {'/'}`, source-text only, no
  runtime coupling) was built and proven during review: green at HEAD, red at
  base naming exactly `[":", "\\"]`. The two literals have now drifted twice, so
  it is worth having — in the `⊇` direction only. Also #120.
- `/` is correctly absent from the walk's class, proven by `readdir`: a single
  directory entry name cannot contain it, so including it would be dead code.


hchristinasmith (approved):
Approving. Everything below is **EXECUTED** — run on this branch in a clean worktree, node 22.22.2 — unless labelled otherwise. Nothing here is reasoned-only.

**The change does what the commit message says, and the tests prove it rather than assert it.** Three mutations of the walk's class, each dropping one member:

| mutation | result |
|---|---|
| `/[._ ]/g` (both out) | 1 failed — names the colon pair |
| `/[._\ ]/g` (colon out) | 1 failed — names the colon pair |
| `/[._: ]/g` (backslash out) | 1 failed — names the backslash pair |

Both new members are independently load-bearing. Your non-vacuity floors bite too: dropping `'a:colon here'` from `ROUND_TRIP_SEGMENTS` fails at `repos.test.ts:279`, exactly where you said it would, rather than passing over an empty set.

**The move into `concierge/` is the real fix, not a second dodge.** Adding a plain `import { reverseProjectSlug } from '../../concierge/repos.js'` to `worktree-slug.test.ts` reddens clause 1 of `namespace-law.test.ts` ("no file in either package reaches the concierge module"). The law has teeth; a test inside `concierge/` is genuinely exempt, and `ReverseProjectSlugResult` is gone tree-wide — `grep` returns nothing.

**Your sibling matrix reproduces exactly.** I rebuilt it from scratch rather than reading yours, with slugs minted the way the real slugger mints them (`/[^a-zA-Z0-9]/g`), not via this repo's narrower encoder:

| case | before | after |
|---|---|---|
| `a:b` alone | REFUSED | **CORRECT** |
| `a:b` beside `a-b` | WRONG → `a-b` | **REFUSED (ambiguous)** |
| `a+b` beside `a-b` | WRONG → `a-b` | WRONG → `a-b` |
| `a+b` beside `a:b`, no `a-b` | REFUSED | **WRONG → `a:b`** |

All four rows match your table, including the last one. Disclosing a row where your own change makes a case worse — and filing #120 with the fix already proven green rather than quietly widening scope — is the reason this was quick to review.

**The deferral holds.** I checked the one thing that could have made it a hole: `forward-transform-law.test.ts:631`/`:658` quote the old class as *rigged fixture strings*, and the live-tree sweep at `:663` still reads the real `repos.ts` and passes against the new class. "Not a hole" is correct.

## Gates

    npm run lint                      exit 0   (biome, 849 files)
    VITEST_MAX_WORKERS=6 npm test     exit 0   381 files, 6618 passed,
                                               3 expected fail, 6 skipped
    scripts/fence-lint.sh 47          PASSED

Also green against the actual post-merge tree (this head plus `origin/main`'s `0979d94`, which touches only `recorder/` and `api/rotate.test.ts` — no overlap): 381 files, 6624 passed, exit 0.

CI checked **per job step**, not by the green bar: on `build-test-boot (ubuntu-latest, current)`, Test / Typecheck / Lint / Packaging guard / Boot smoke all report `success`. Nothing was silently skipped behind a failing Test step.

## What I pushed

`405c745`, comment text only. Two pointers in `worktree-slug.ts`'s doc comment went stale in the same commit that moved the law: it quoted the reverse walk's class as `/[._ ]/g` in the present tense, and cited the round-trip law as living in `worktree-slug.test.ts`. Both now name what exists. Inside #47's fence (the `worktree-slug.ts` bullet), and `fence-lint.sh 47` still PASSES with it in.

Smaller to fix than to describe, so I fixed it rather than sending the PR back a day for two lines of prose.

## What I did NOT run

No Windows CI leg exists, so the backslash path is verified on macOS and Linux only — which is consistent with the module's own "native Windows is out of this instrument's support matrix" contract, but worth stating rather than leaving implied.

`scripts/land.sh` from my review discipline does not exist in this repo; I ran its steps by hand.

PR #122, opened 2026-08-26
docs(prd): prd-42's wave map is stated once, and its superseded paragraphs say so

gabriel-canaan:
prd-42's wave map is stated once, and the two paragraphs it supersedes say so.

Docs only — one file, additive. No issue: this is bookkeeping on the plan of
record, not new scope. Every assignment it states was already decided by an
existing amendment or by filing an issue.

## The drift, and what it cost

Sequencing here is append-only, which is right — wave and ruling citations must
keep resolving, so a superseded paragraph is marked rather than deleted. Nothing
marked them. The document declared **wave 6 twice** (`:296` `#52`, `:415` `#53`)
and **wave 7 twice** (`:299` `#53`, and now `#120`), with every paragraph reading
as current.

On 2026-08-26 an agent read the wave-7 paragraph, saw `#53` titled `w6` on the
tracker, and edited the **tracker** to match the **stale** half of this document.
Reverted within the hour by the session that had done the reconciliation. Nobody
was careless: the document held two live answers and offered no way to tell which
was current.

## What changed

- `:296` and `:299` gain dated **SUPERSEDED** notes **in place**, both keeping
  their original reasoning — `:296`'s file-contention argument is what forced
  `#52`'s absorption, and `:299`'s sweep-comes-last rule still governs `#53`;
  only its wave number moved.
- A **wave map** at the end: waves 1–8, their issues, state, and which amendment
  declared each. `#52` and `#92` recorded as closed-superseded, holding no wave.
- Why `#120` is wave 7 rather than a successor PRD, plus the note that wave 7 has
  a second half **not yet filed** — the encoder needs the same grammar plus the
  real slugger's 200-char cap and base36 suffix. `#120`'s fence excludes it
  deliberately, because widening one side without the other is what created the
  divergence this PRD is named for.

## What is the sibling case?

The other direction of the same invariant. Every check in this toolchain runs
*issue → is it groomed?* — `fence-lint` reads only issues you name as a wave, and
the board's orphan check sees only board membership. **Nothing read the document
against the tracker**, so a vacant wave and a double-declared wave number were
invisible by construction. That reconciliation now exists as operator tooling; it
found all three instances here, is named in the amendment, and is deliberately
**not** added to this repo.

It is prose, so nothing gated catches a future drift — which is exactly why the
reconciliation check is named in the text rather than assumed. Adding a wave to
this document with no issue, or an issue with a wave this document never
declares, still produces no failure at landing.

## Two claims in the first draft did not reproduce

Corrected in the same commit, since a document about stale claims should not ship
one:

- It cited `scripts/dev/prd-waves.sh`, **which does not exist.** Replaced with the
  two real commands, both of which I ran.
- Its wave map listed a **wave 0**. prd-42 has none — it books its operator act
  *inside* wave 4, which is why the grep starts at 1. prd-39 and prd-46 use a
  wave 0 for the same kind of act; this PRD did not.

The second command's output also moved the moment `#120` was filed, so the block
now states what it returns today **and** what it returned before, with the reason:
output that is a function of a mutable tracker is not self-validating without the
conditions it was taken under. This PRD has had to record that twice already.

## Gates

Docs-only, no code path touched. `git diff --stat`: one file, 99 insertions, 0
deletions. No `/home/`, `/Users/`, username or hostname in any added line. Both
commands quoted in the amendment were executed and their stated output matches.


gabriel-canaan:
Closing unmerged — the commit rides instead of taking its own PR.

`3e355ed` stays on `origin/prd42-wave-map-reconciled` and will be cherry-picked into wave 6's PR (#53) as its own commit. The obligation is recorded on #53 with the sha and the cherry-pick line, because an unmerged commit behind a closed PR is easy to lose.

**Why:** this amendment declares wave 7 (#120 + #124), which is two waves from dispatch. The real constraint is only that an amendment lands before the wave it *declares* is *dispatched* — not that it precedes the work, and not that it gets a PR. A PR per regroom bills the team the fixed per-PR toll (~21 h median, 81% of cycle time) for a one-file docs change, and regrooming here is frequent.

That reasoning was wrong when this PR was opened: "the document must precede the work" is true but was applied to the wrong boundary. Recorded in the operator's `prd-groom` skill so the next amendment rides by default.

**If wave 6's PR would land after wave 7 is ready to dispatch, do not hold wave 7** — reopen this or open a fresh standalone PR. The ride must never delay a dispatch.

PR #126, opened 2026-08-26
prd46: a rescue block that names failure is not one that calls fail

ciaran-slow:
Follow-up repair to #119 (prd-46 wave 1), found reviewing it. **#119 merged one minute before my fix could reach it**, so the defect below is live in `main` today — this carries the fix over unchanged.

No issue behind it yet: it is a review repair on already-landed work, and I did not want to guess a wave number onto the board. Say the word and I will groom one.

## The defect

`tailChecksStatus`'s **rescue-block branch** scanned the raw block text with `\bfail\b` and `\bexit\s+(\d+)\b`. A string literal satisfies both, so a block that merely *mentioned* failure in its message was credited with handling it.

That is a **spelling match sitting inside the predicate whose entire subject is structure over spelling** — the exact class prd-46 exists to abolish. It is not among the misses #119 declared.

EXECUTED against `origin/main`'s own law file, with the plant made after the commit-count verdict, exactly as #68's verification planted its six:

```
RES=$(some_new_check) || { echo "  fail to check"; }
```

| | result |
|---|---|
| **main today** | ruling 1's own test **GREEN** over it. The only red is the total-assignment pin drifting 16 → 17 — that is the plant, not a detection. `1 failed \| 103 passed` |
| **this branch** | `undeclared unchecked producer(s) in scripts/gate.sh`, naming index 131. `3 failed \| 106 passed` |

## What is the sibling case?

The same function's two **same-line** branches decide the same question, and I checked them rather than assuming: `/^fail\b/` and `/^exit\s+(\d+)\b/` both anchor at the start of the right-hand side, where no string can precede them. They are safe and are left unchanged. The block branch is the only one that scans, and it reaches the wrap check by a second path — both routes are covered by a rigged case.

`stripQuotedRuns` is a single left-to-right scan rather than two `replace` passes. Stripping `'...'` first would let the apostrophe in a double-quoted `"don't"` open a bogus run and swallow the rest of the line; a rigged case pairs exactly those two quote styles on one line.

EXECUTED — `const body = stripQuotedRuns(rhs)` reverted to `const body = rhs`, i.e. main's behaviour:

```
4 failed | 105 passed
  × the real predicate FIRES on: rescue block whose MESSAGE says "fail" …
  × the real predicate FIRES on: rescue block whose MESSAGE says "exit 2" …
  × the real predicate FIRES on: single-quoted message naming fail …
  × planting each rigged line into a COPY of gate.sh's real text …
```

Restored: **109/109**.

Two **controls** guard the other direction, because a repair that only removes a credit can pass by refusing everything: a block that really does call `fail` (the `:57` shape, quoted message and all) and one that really does `exit 2` after a quoted message are both still silent.

The false-positive pins are unmoved — 16 assignments, 9 same-line, 6 next-line, 1 flagged-and-declared. I re-derived that split independently against the real `scripts/gate.sh` with my own parser rather than reading it off the test.

## Fence

`packages/server/src/gate-honesty-law.test.ts` only. `scripts/gate.sh` is untouched — this changes the law, not the gate.

## Checks

Against `origin/main` at `aa909b3`: `npm run typecheck` 0 · `npm run lint` 0 · the law file 109/109. The full suite ran green on the pre-merge tree (381 files, 6660 passed, 3 expected-fail, 6 skipped); this commit is that same diff cherry-picked onto `main` with no conflict.

`scripts/gate.sh` was never run — work that edits the gate's law cannot be landed by the gate.


gabriel-canaan (commented):
Reviewed commit-by-commit at `2b31bf4`, then fixed one sibling and pushed it as `294b8fe`. Nothing below is reasoned — every claim was run, and the commands are here to re-run.

Both of your central claims reproduce independently. I planted your line into a copy of the real `scripts/gate.sh` myself rather than reading your numbers off the PR body:

| | result |
|---|---|
| `origin/main` is holed today | **confirmed.** Ruling 1's own test is **GREEN** over a producer that swallows its failure entirely. `1 failed \| 103 passed`, and the single red is the total-assignment pin drifting 16 → 17 — the plant itself, not a detection, exactly as you said |
| this branch catches it | **confirmed.** `3 failed \| 106 passed`, ruling 1 naming the planted line |
| the mutation you reported | **confirmed.** `const body = stripQuotedRuns(rhs)` → `const body = rhs` gives `4 failed \| 105 passed`, naming all three new rigged cases plus the plant-into-the-real-file test. Restored: 109/109 |

I re-derived the false-positive pins against the real file too — 9 same-line `||` tails, of which `:57` is the only block. 16 / 9 / 6 / 1 holds.

And I checked the trap this file specifically has fallen into before, since your diff adds new string literals to it: the enrolment defect its own `DECLARED_TOLERANCES` comment records, where a literal in this file matched `gate.sh`'s fixed-string content grep and pulled the law into the derived timing set. It does not recur here. `gate.sh`'s timing set is **6 files on `origin/main` and 6 on this branch**, this file in neither:

```
grep -rlF '// @gate-timing' --include='*.test.ts' --include='*.test.tsx' packages; find packages -name '*.bench.test.ts'
```

`stripQuotedRuns` being one left-to-right scan rather than two `replace` passes is the right call for the reason you give, and the single-quote/double-quote asymmetry inside it — backslash escaping only under `"` — is correct shell.

## The finding: the sibling is not closed

Your sibling section asks *which branches scan*, and correctly clears the two same-line branches — `/^fail\b/` and `/^exit\s+(\d+)\b/` do anchor where no string can precede them. But the question the block branch actually raises is *what else, besides a string, is not a call?* A comment reaches that branch by the same route a message does.

EXECUTED against your head at `2b31bf4`, as rigged cases in `RIGGED_LINES`. All four went **unflagged**:

```
SOME_NEW_CHECK=$(some_new_check) || { echo no; }  # fail is handled elsewhere
SOME_NEW_CHECK=$(some_new_check) || { echo no; }  # would exit 2 if fatal
SOME_NEW_CHECK=$(some_new_check) || { echo $fail; }
SOME_NEW_CHECK=$(some_new_check) || { note=fail; }
```

The comment route is the one that matters: `scripts/gate.sh:325` already carries a trailing comment (`MERGED=1   # every fail() past this point reports the post-merge truth`), so the shape is idiomatic in this file, and "fail" is a natural word for such a comment to contain. It is the same defect class you name — a spelling match inside the predicate whose whole subject is structure over spelling.

## What I pushed — `294b8fe`

The comment route only. The `#` cut lives **inside** the existing scan rather than in a later `replace`, because whether a `#` opens a comment depends on quote state *and* on the character before it — bash starts a comment only at the start of a word. Cutting after the strip would read the space the *opening* quote leaves in the output and eat a real call the source never separated.

Asserted against bash rather than against re-typed rules:

```
$ bash -c 'LOG=/tmp/x; echo "x"#y; echo "${LOG#/tmp/}"; echo A  # fail here'
x#y
x
A
```

Three controls, on your own principle that a cut removing too much can pass by refusing everything: a block that calls `fail` and *then* explains itself in a trailing comment; one whose `${LOG#/tmp/}` expansion carries a non-comment `#`; one where the `#` is glued to a closing quote. All three must stay silent.

One rigged case is there specifically to pin the *shape* of the cut — `# not fatal; fail comes later`. A rule that split the block into command segments would resurrect `fail` from the text after the comment's own semicolon. The whole comment goes, or the fix has a sibling of its own.

**Mutation, both directions:**

- comment cut removed → `4 failed | 111 passed`
- word-start condition dropped, cutting at any `#` → `2 failed | 113 passed` (the `${VAR#pat}` and glued-quote controls)
- restored → **115/115**

The helper is renamed `stripQuotedRunsAndComments`, since it no longer strips only quoted runs and a law about honest naming should not misname its own helper. That does make your commit body's reference to the old name stale — flagging it rather than rewriting your history.

Your pins are unmoved and the timing set is still 6. Gates on the merged result: `npm run typecheck` 0 · `npm run lint` 0 (849 files) · `VITEST_MAX_WORKERS=6 npm test` 0 — 381 files, 6671 passed, 3 expected fail. `scripts/gate.sh` untouched, and `scripts/gate.sh` never run: work that edits the gate's law cannot be landed by the gate.

I did **not** touch `|| { echo $fail; }` or `|| { note=fail; }`. Closing those means requiring `fail` in **command position** — splitting the block on `;` `&&` `||` and matching only a segment's first token. That would subsume all four shapes in one coherent rule instead of the two spelling patches we now have between us, and I checked it holds for `:57`'s real shape.

But it changes which *real* forms the law accepts: `if [ x ]; then fail; fi` and `else exit 1` inside a rescue block would stop counting, and the law would go red on a legitimate future block. That is a design decision about the predicate, not a repair, so it is yours — and it probably wants an issue rather than this PR.

Not approving, by design: `294b8fe` is a substantive change and it has had exactly one perspective on it, mine. That is the position review exists to abolish, arriving one step later. It wants your confirmation and a second pair of eyes before it lands.

Two things I did not run: `scripts/gate.sh` (above), and CI against `294b8fe` — the 7 legs that passed were against `2b31bf4`, so re-check `gh pr checks 126` once the new head registers.

Issue #106, opened 2026-08-25
prd40 w7: one subscriber's throw cannot blind the others


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/recorder/session-recorder.ts`
- `packages/server/src/recorder/session-recorder.test.ts`

## Blocked by

**Not a ruling — a queue.** This is the **fourth** claimant on `session-recorder.ts`. #26 (wave 4) has landed in #105; **#69** (wave 5) and **#80** (wave 6) still hold it. Three issues cannot share one file in one wave, and neither can four. This runs **after #80**, as wave 7.

Nothing else blocks it. No ADR is needed: ADR-0030 already ruled that the alarm must reach subscribers, and this issue is that ruling failing to hold when a subscriber misbehaves.

## The gap

`SessionRecorder` publishes through one `EventEmitter` (`session-recorder.ts:57`), and **Node's `emit` abandons the listener list at the first throw.** Proved, not argued:

```js
const e = new EventEmitter()
let second = 0
e.on('x', () => { throw new Error('boom') })
e.on('x', () => { second++ })
try { e.emit('x') } catch {}
// second listener ran times: 0
```

Every SSE client subscribes once (`api/stream.ts:158`, via `ctx.recorder.subscribe`), and the class comment at `:70` says outright that many concurrent clients are expected — the emitter's cap was raised for exactly that. So **one dashboard whose handler throws silences every dashboard that connected after it**, for every event on the stream.

It holds on all three publish paths, in three different ways:

- **`record()` (`:134`)** — unguarded. A throwing subscriber both stops later listeners *and* rejects the caller's promise.
- **`closeWith()` (`:184-190`)** — guarded, so the close survives, but later listeners still never run.
- **`recordAlarm()` (#26, ADR-0030)** — guarded the same way, and **this is the one that matters most**: the alarm exists to say the log is unwritable, and ADR-0030 accepted a permanent hole in the durability boundary to make sure it is heard. A subscriber registered earlier can throw that guarantee away.

## Why it is a bug and not a tidy-up

**It inverts what the guard was for.** `closeWith` and `recordAlarm` both catch a subscriber's throw so that *one client's bug is not the recorder's failure*. Catching it at the emitter makes the throw non-fatal to the **publisher** while leaving it fatal to every **other subscriber** — the blast radius moved rather than shrank.

[...]

PR #127, opened 2026-08-27
prd40: the ADR log's pointer tells the truth, and the resume bullet describes resume

### LINKED ISSUE #99: prd40: the ADR log's stated range covers every record in it

### LINKED ISSUE #100: prd40: the resume guarantee does not read as a no-duplicates promise
## Fence (may touch ONLY)
- `docs/user-guide/sessions.md`

## Blocked by
Nothing. Consumes ADR-0029, which is accepted and on `main` (#77).

## The gap

`docs/user-guide/sessions.md:15` describes resume as "same file, same collector offsets, **no duplicated history**".

That is true of what it is actually describing — resuming a session does not replay history it already wrote. But ADR-0029 now rules that a recording **may** contain the same fact twice: the poll path is at-least-once, so a rejected append leaves the collector snapshot un-advanced and the next tick re-appends the batch. A reader who takes "no duplicated history" as a property of the log rather than of resume will get the delivery guarantee exactly wrong, and this file is the user-facing one.

Surfaced by the cross-document sweep on #81, reported rather than followed because it is outside that issue's fence.

## Why this one and not the other seven

#81 swept `docs/` and `README.md` for stale exactly-once claims and found none that were false. This is the only hit where the sentence is *true at the granularity it claims* and *misreadable at the granularity a reader will apply*. That makes it a wording fix, not a correction — which is why it is small, and why it is worth doing rather than dismissing.

## Definition of done
- `sessions.md:15` says what it means: resume does not re-write history it already recorded. It no longer reads as a promise about the log's contents.
- The distinction is reachable from there — a pointer to [ADR-0029](../adr/0029-a-recording-may-repeat-a-fact.md) or to `architecture.md`'s "What a replay may now contain" paragraph (landing in #98), so a reader who wants the delivery guarantee can get to it.
- **Check the rest of the file, not just line 15.** The failure mode #81 named applies here too: a corrected sentence beside an uncorrected one is the same defect.
- No ADR is edited; no new ADR. This records an existing decision.

## Not in scope
`docs/architecture.md` — #81 landed its paragraph there and the sentence this issue fixes is a different audience's. The seven non-idempotent reducer arms ADR-0029 names: a cost the ADR accepted deliberately, not a defect either issue fixes.

ciaran-slow:
Two issues, one PR — **disjoint fences, no dependency, one queue toll instead of two**. Both are prd-40's documentation debt: neither is in a wave, and both accumulated while waves 1–7 ran.

Closes #99
Closes #100

Two commits, one per issue, deliberately not squashed — the PR is the unit of queueing, the commit is the unit of review.

| commit | issue | file |
|---|---|---|
| `6864d7a` | #99 | `docs/architecture.md` |
| `81beb54` | #100 | `docs/user-guide/sessions.md` |

---

## #99 — the ADR log's pointer stops falsifying itself

`architecture.md:8-9` said the decisions ADRs record live in `docs/adr/` — **"0001 through 0027 today"**. 0028, 0029, 0030 and 0031 exist, and two of those are the records a reader of this milestone most needs.

**The range is removed, not updated.** A hardcoded count is re-stale the next time an ADR lands — this sentence had gone stale before, which is what a hardcoded count does — and the log is one `ls` away. The sentence's job is to say *the ADR log is the register of record*, and it still does.

Two decisions-log entries appended, in the existing bullet shape:

- **ADR-0030** — the degrade `collector.error` is emitted whether or not its own append lands. The entry states in bold that **the exemption is emission only**: prd-40 success 1 still forbids leaving the fold ahead of the file and that clause is *not* carved out, so on a failed append the alarm enters neither `buffer` nor `foldState`. A summary omitting that describes a wider hole in the durability boundary than actually exists. Then the consequence — a live stream may hold a `collector.error` a replay does not, and only that event type.
- **ADR-0031** — every fold `foldSoFar()` hands out is deep-frozen. Carries **why a readonly return type lost**: `LadderManifest.folded` → `selectConnection(state: SessionState)` cascades into `packages/core`'s selector signatures, and a shallow `Readonly` would pay that and still miss the nested write that actually corrupts. Then the consequence — the reducer is now held to ADR-0002's purity contract on this path.

**No ADR was edited and none written.** The log is append-only.

### The sweep — three hits, all left, and the reason is the point

Swept `architecture.md` for other hardcoded ADR/ruling counts. Three: *"prd5's six rulings"* (`:700`), *"prd6's five rulings"* (`:984`), *"prd7's six rulings"* (`:1154`). Each PRD's rulings were counted to confirm — 6, 5, 6, all correct.

All three are **left deliberately**: they count a *closed* milestone's blessed ruling list, not a growing log, so none has the self-falsifying shape this issue exists to remove. Line 8-9 was the only hardcoded ADR count in the file. That distinction is why this is a sweep and not a search-and-replace, and #23 (`prd43 w4`) owns the docs-tree-wide version of the shape.

## #100 — the resume bullet describes resume, not the log's contents

`sessions.md:15` described the default as *"same file, same collector offsets, no duplicated history"*. True of what it describes — resuming does not replay history it already wrote — but ADR-0029 rules that a recording **may** contain the same fact twice, since the poll path is at-least-once. A reader taking "no duplicated history" as a property of *the log* rather than of *resume* gets the delivery guarantee exactly backwards, and this is the user-facing file.

Now: *"no re-recording of the history it already wrote. That is a property of resuming, not of the log: a recording may still contain the same fact twice ([ADR-0029](../adr/0029-a-recording-may-repeat-a-fact.md))."*

**One clause plus a pointer, deliberately.** That bullet's job is to tell an operator what `--fresh`, `--resume-window` and the default do; it is not the place to explain at-least-once semantics.

### The sweep — nothing, and that is the finding

The whole file was read end to end. The targeted grep matched only line 15 before the change and nothing after; a wider pass (`once|all events|every line|never lost|complete|identical|twice`) returned the new clause, the session lock's staleness wording, and `manifest.complete` — a transcript-capture flag, not a delivery claim. **Nothing else to correct, and nothing invented to correct.**

---

## Gates

Base confirmed current (`aa909b3`). Node v22.23.2, each gate run **separately and unpiped, reading its own exit status**, over the assembled two-commit branch. Falsifiability proven first: a deliberate `TS2322` in a throwaway file took typecheck to exit **1** naming that file; probe removed before the real runs.

`npm run typecheck` **0** · `npm run lint` **0** (biome, 849 files) · `VITEST_MAX_WORKERS=6 npm test` **0** · `npm run build` **0**.

**Totals, not passes: `381 files` · `6655 passed | 3 expected fail | 6 skipped | 6664 total`** (macOS; a Linux run reads five lower in both `passed` and `skipped` — `it.skipIf(process.platform === 'linux')` at `static.test.ts:525`).

All four run despite a markdown-only diff, because the repo has doc laws that read tracked documents (`no-personal-paths-law.test.ts`, `runbook-delivery-law.test.ts`).

## Links, and the claims behind them

Every path verified on disk: `adr/0030-…md` and `adr/0031-…md` from `architecture.md`, `../adr/0029-…md` from `sessions.md`. **No in-page anchors used**, so no anchor can break when a heading is edited.

The **code facts the entries assert** were also checked against the tree rather than trusting the ADR prose being summarised — `recordAlarm(event: EventOf<'collector.error'>)`, `folded: SessionState` at `api/meta.ts:143`, and `selectConnection(state: SessionState)` at `core/src/selectors/connection.ts:200`. For an issue that exists because a citation pointed at nothing, checking the claims and not only the links is the point.

---

## What is the sibling case?

**For #99 it is the three ruling counts left in place** — the near-identical text that is *not* the same defect, distinguished by whether it counts a closed list or a growing one. Fixing those too would have been the plausible over-correction.

**For #100 it is every other document making the same stale claim.** #81 established that shape (a decision recorded in one place and contradicted in another), and both sweeps above are the answer: `architecture.md` and `sessions.md` are now consistent with ADR-0029, and nothing else in `docs/` or `README.md` makes a delivery claim that contradicts it.

Docs-only; no law added, and none should be. The falsification available here is that **every link and every asserted code fact resolves**, which was checked by execution rather than by reading — including the three ruling counts, which were counted rather than assumed correct.

## Deviation

**One choice, flagged rather than drifted.** #100's plan offered ADR-0029 *or* `architecture.md`'s "What a replay may now contain" paragraph as the pointer target. ADR-0029 was chosen: it is the decision record itself and needs no anchor, whereas the architecture paragraph is reachable only via `#ruling-3--recordings-never-rot-the-integrity-laws-landed`, which breaks if that heading is ever edited. Both were in scope per the plan.


ciaran-slow:
## Review — read commit by commit, every claim verified

**Verdict: approve and land.** Both commits' factual claims check out, both sweeps are real, and the siblings are covered. No blocking findings; two notes at the end.

### `6864d7a` — the ADR pointer (#99)

Removing the range rather than bumping it is the right call: the sentence exists to say *the ADR log is the register of record*, and a hardcoded range makes it falsify itself on the next ADR. Verified (**EXECUTED**):

- `docs/adr/0030-the-alarm-may-outrun-the-record.md` and `0031-the-recorder-hands-out-a-frozen-fold.md` both exist; both new links resolve. 0031 is the highest ADR on disk, so the two new entries leave the log fully covered.
- **The ADR-0030 entry's narrow claim is true of the code.** `recordAlarm(event: EventOf<'collector.error'>)` is type-narrowed as described, and on a failed append it pushes nothing into `buffer` and advances no fold — so "the exemption is emission only" is exactly the hole that exists, not a wider one.
- **The ADR-0031 entry's cascade argument is true.** `folded: SessionState` at `api/meta.ts:143` and `selectConnection(state: SessionState)` at `core/src/selectors/connection.ts:200` — a readonly return really would cascade into `packages/core`'s selector signatures.
- **The sweep's three survivors are correct, checked against the PRDs rather than taken on trust:** prd5 has six rulings, prd6 five, prd7 six. All three counts stand.
- **Re-ran the sweep on the branch:** no self-falsifying range or "N today" claim remains in `architecture.md`.

### `81beb54` — the resume bullet (#100)

The distinction the new wording draws is the right one, and it is reachable rather than restated. Verified (**EXECUTED**):

- **The at-least-once mechanism is still exactly as the commit describes, on `main` as it stands today** — worth checking because #106 landed after this branch was cut and touched the same neighbourhood. `poll-loop.ts` still says it in its own words: *"the snapshot advances only once every event in this batch is on disk. A rejected append therefore leaves `previous` in place and the next tick re-derives the WHOLE batch."* #106 removed the *subscriber-throw* route into that path, not the rejected-append one, so the claim is untouched.
- **The in-file sweep holds:** `grep -nEi "duplicat|exactly.once|every event|no .*repeat"` on the branch version returns nothing.
- **The sibling the commit did not sweep, checked here:** the same claim anywhere else under `docs/`. `grep -rniE "no duplicated|no duplicate|without duplicat" docs/ README.md` (excluding `prds/` and `adr/`) returns exactly one hit — `docs/user-guide/sessions.md:15`, the line this commit fixes. Nothing else repeats it.

### The stale base is harmless, and that is checked rather than assumed

The branch is 27 commits behind `main`, so CI ran against an older merge — and for a docs PR fixing a claim *about the tree*, that is precisely when a fix can rot before it lands. It has not:

- neither `docs/architecture.md` nor `docs/user-guide/sessions.md` moved on `main` since the branch (`git log branch..main -- <both>` → empty);
- **no ADR was added** in those 27 commits, so the two new entries still leave the log complete.

All seven CI legs pass; `MERGEABLE` / `CLEAN`.

### Two notes, neither blocking

1. Commit `6864d7a`'s message pairs the three surviving counts with "(lines 984, 1154 and 700)" in a non-corresponding order — 700 is prd5's, 984 prd6's, 1154 prd7's. Message only, and line numbers in a commit message drift anyway, which is the habit this repo already discourages for runbooks. Nothing to fix.
2. **For #23** (prd43 w4, *every count a document states is derived*): the three surviving counts are numbered list items (`1.` … `6.`) under a `## Rulings` heading, **not** `## Ruling N` headings — the pre-prd18 convention. A citation law that looks for the heading form will read those PRDs as having zero rulings, which is how I first mis-measured them here. Worth knowing before that law is written.

PR #123, opened 2026-08-26
prd41: the laboratory is confined in fact

hchristinasmith:
Closes #6
Closes #7
Closes #8
Closes #9
Closes #10

## What this PRD does

prd-12 ruling 1 drew the lab's fence precisely — writes only to `refs/rhizomorph/`, git objects, and artefacts outside the watched repo — but left a posture unenforced: real code execution from an untrusted checkpoint, unbounded subprocess hangs, a silenced degrade voice, and unbounded spend. This PRD closes each hole, per `docs/prds/prd-41-the-laboratory-is-confined-in-fact.md`.

- **#6 (wave 1, keystone)** — lands the containment law's own test **red**, by design, proving the gap exists before any fix lands. `#7` turns it green.
- **#7 (wave 2)** — `restoreWorkspace`'s `npm install` now passes `--ignore-scripts`; its own `exec` (git worktree add + npm install) is bounded via `withTimeout` at 120s. Found and filed a real follow-up (#109, still open) — a `withTimeout` composition bug where the innermost wrap always wins regardless of the outer one's ceiling.
- **#8 (wave 2)** — `compare.ts`, `checkpoint.ts`, `fork.ts` all route their `exec` through the existing `withTimeout` helper — no second timeout mechanism invented.
- **#9 (wave 2)** — `withLabCliLock` gains a 30s ceiling (503, refuse rather than queue) and `MAX_ARMS = 8` joins #234's validation block (400 on violation). During review I found the PR's own claimed mutation-coverage for the lock's `gaveUp` guard wasn't real — no test actually verified a refused caller never reaches its own `exec` once the holder clears — and added one that does, mutation-verified.
- **#10 (wave 3)** — the process-wide `process.stderr.write` override is replaced with an `AsyncLocalStorage`-scoped capture, so #239's degrade voice (and anything else running concurrently) is never swallowed by a lab launch. During review, that PR's own new test caused a real cascading CI failure on `ubuntu-latest` (fake timers leaking across the whole file under load) — reproduced twice via `gh run rerun`, root-caused, and fixed before landing.

## Review

Every issue above was independently verified by the conductor session before merging into `prd41` — diff read, `typecheck`/`lint`/targeted tests re-run, an original mutation test performed (not just re-running what the PR claimed), and all CI legs watched to completion. Two of the five (#9, #10) needed a review-time fix for a real gap found during that verification, both pushed as `review of #N` commits and re-verified.

[...]

Issue #38, opened 2026-08-24
prd44 w3: a recording older than the retention answer is gone


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/log/retention.ts` (new file)
- `packages/server/src/log/retention.test.ts` (new file)

## Blocked by
- **Wave 0, an operator act.** The retention answer — how long a recording is kept, and by what act it goes — is not this lane's to choose. Do not start until it is written down.
- **#30** (prd44 w1) — it owns `log/lane-index.ts` and `log/listing.ts`. This lane must not reopen them; that is why the policy lands in a new file.

Sibling lanes: none in this wave. **Do NOT touch `packages/server/src/cli/index.ts`** — it is #7's live fence (prd41 w2). The CLI wiring for this policy is a separate, later issue once #7 has landed; this lane ships the policy and its laws, reachable but not yet wired.

**Model:** sonnet. **Wave:** 3.

## Definition of done
- The policy is a pure function of (the directory's contents, the retention answer, a clock) — testable without deleting anything.
- Deletion is an **explicit operator act**, never a default and never a timer. A law asserts nothing is removed without being asked.
- The live session is never a candidate, whatever its age.
- A law asserts the golden-era corpus is unreachable by this policy. It is committed in the repo at `packages/core/src/eras/era-1/recording.jsonl` with hermetic tests that never read the real data root, so no policy can reach it — pin that, because ADR-0011's forward-compatibility guarantee leans on that corpus.
- Checked against **ADR-0011** ("recordings never rot") in the PR body: that ADR governs *parsing* an old recording, not how long one is kept. The names collide and a future reader will assume a conflict.
- A dry-run reports exactly what would go, and reports it before anything is removed.
- **Assert a count, never a wall clock.** And do NOT name any new law `*.bench.test.ts`, nor mark it `// @gate-timing`.

## What deleting a recording costs, and the promise it must not break

Retention is not only a disk question. **prd-31 ruling 5** (shipped) built the lane index so a lane
page opens *after its worktree is gone*, and made that index part of the ruling rather than an
optimisation of it:

> `workmux merge` deletes a worktree the moment work lands, which is precisely when a person most
> wants to read what happened. [...] **A lane index is therefore part of this ruling, not an
> optimisation of it.**

Deleting a recording therefore removes a lane's history from a surface another ruling promised
would still open a week later. That is a real cost, and it is the operator's to accept in wave 0 —
but it must be **visible**, never silent:

[...]

PR #131, opened 2026-08-27
prd44: the server's event buffer has a stated ceiling

### LINKED ISSUE #37: prd44 w2: the server's event buffer has a stated ceiling

ciaran-slow:
Closes #37

prd-44 ruling 4, and the **last code issue in the milestone** — #38 (retention) is the only
one left, and it is deliberately sequenced after this.

## What changed

`eventsSoFar()` is now a window of at most **75,000** events, oldest evicted first.
That number is wave 0's answer (operator, 2026-08-27) and it is deliberately *the client's*
number — `MAX_EVENTS` in `web/src/app/streamState.ts`. Ruling 4's complaint is that one side
of the wire ruled and the other kept everything and said nothing; a different number here
would leave that half-fixed, with `eventsWindowLabel`'s vocabulary meaning two things
depending on which side you read it from.

**The fold is untouched, and that is what makes the bound honest.** `foldSoFar()` still sees
every event the session recorded, `eventCount` included, so
`eventsSoFar().length < foldSoFar().eventCount` is exactly how a reader detects it is holding
a partial window — the same pair, read the same way, as `eventsWindowLabel`. The vocabulary
is reused rather than given a server-side twin. `evictedEventCount` states the same fact
directly.

**Eviction advances an index; it does not splice.** A `splice(0, n)` per event past the bound
would memmove 75,000 pointers — ~600 KB — on *every* subsequent event, which is the same
per-event cost prd-44 ruling 2 just removed from the append path, reintroduced one file away.
The window index moves in O(1) and the array is compacted once per 1,024 evictions: ~73
pointer copies per event amortised, with the live array bounded at 76,024 slots. Measured on
this box while building the laws: 76,025 events through `record()` end at a 75,000-event
window in 75,001 slots.

## Fence, and the widening

Declared fence, held:

- `packages/server/src/recorder/session-recorder.ts` + its test

**Widening — two files, one line each, recorded on #37 before the change was made**
([comment](https://github.com/launchpad-26/rhizomorph/issues/37#issuecomment-5433550282)):

- `packages/server/src/recorder/rotate.ts:145`
- `packages/server/src/cli/run.ts:207`

Of the fourteen `eventsSoFar()` readers, twelve become **partial**, which is the ruling
working as designed. These two become **wrong**:

- `rotate.ts:145` wrote `eventsSoFar().length + 1` into the `session.closed` payload — a field
  of the durable, hash-chained record. Past the ceiling a closed session would claim 75,001
  events when it recorded 200,000. Bounding the window and filing a follow-up would mean
  shipping a false fact into history on purpose.
- `cli/run.ts:207` re-folded the window to build the boot line, so a resumed session past the
  ceiling would report the worktrees and branches of its last 75,000 events.

Both become `recorder.foldSoFar()` — which **prd-40 ruling 2 already asks of every reader**
("every future caller reads that instead of calling `reduceAll(eventsSoFar())`"), so the
widening moves both lines onto an existing ruling rather than inventing one. `rotate.ts` also
stops carrying prd-40's known undercount of an in-flight append as a side effect.

Claimant check first: no open PR touches either file (#127, #123, #117, #93). **#53** declares
`recorder/rotate.ts` in its fence but has no lane dispatched, and its item there is the
duplicated refusal message — a different hunk. Noted on #53.

## The two consumers ruling 4 names

**`/api/stream`'s replay — the promise that changes.** Ruling 4 says bounding the buffer means
changing one of these promises, and this is the one:

- A fresh subscriber to a session past 75,000 events replays the last 75,000, not all of them.
- A **reconnecting** subscriber whose `Last-Event-ID` has been evicted gets the whole window
  rather than only what it missed — `resumeBacklog` already returns everything when it cannot
  find the id (`stream.ts:48`), so the failure direction is duplicates, never a silent gap,
  and ADR-0029 already accepts duplication as the recoverable side. **This path is not new**:
  a rotation has always cleared the buffer and made `lastEventId` unfindable. Eviction adds a
  cause, not a shape.
- **The honesty gap this opens, stated plainly because it is the one thing here that gets
  worse.** The web client derives `session.eventCount` from the events it *receives*, so after
  a truncated replay `events.length === session.eventCount` and `eventsWindowLabel` returns
  `null` — the client cannot tell its window is partial, where before it could (the server
  sent everything and the client trimmed locally). Fixing it means carrying the true total to
  the client, which is a route or web change — `/api/meta` already answers `eventCount` from
  the maintained fold (prd-40 #5), so the number is there. Out of fence; wants its own issue.

**The exporter — not affected, despite the issue naming it.** `cli/export-record.ts:150` and
`cli/export-otlp.ts:242` both read the log from **disk** via `readSessionEvents`, never the
recorder's buffer. The issue's premise was off here; checked all fourteen non-test
`eventsSoFar()` call sites rather than assuming.

## What is the sibling case?

**The three publish paths, which is this file's recurring shape** — `record`, `recordAlarm`
and `closeWith` each push into the buffer, and #106 landed last week precisely because a fix
had covered one and missed two. So the bound is not applied at three call sites: `buffer.push`
is replaced by one private `appendToWindow`, the only writer of the buffer's tail, and the
three sites call it. A per-site bound is the mutation that would pass most of these laws.

**The sibling the issue did not anticipate: `foldSoFar()`'s self-heal.** It rebuilds from
`buffer` — whose doc comment says it "is always ground truth" — and bounding the buffer makes
that false. A rebuild past the first eviction would produce a fold missing tens of thousands
of events, where the incremental fold is behind by exactly the one event `reduce` threw on. So
the repair is now conditional on nothing having been evicted, with the reasoning written down,
and a law drives both halves. Nothing in the DoD mentioned the fold; this is the change's real
hazard and it is one screen from the line being changed.

**A third, found in the line being edited:** the constructor did
`this.buffer.push(...options.resumeFrom)`, and a spread is one argument per event — V8 throws
`RangeError: Maximum call stack size exceeded` past roughly 100k arguments, on exactly the long
resumed sessions this bound exists for. Verified: 76,025 arguments is fine, 200,000 throws. Now
an assignment. **Not covered by a law**: a 200k-event fixture to prove it costs far more than
the one-line fix is worth, and the laws here top out at 76,025 — said rather than implied.

Eight laws; six mutations run against the real source and the tree restored after each
(**EXECUTED**, `npx vitest run packages/server/src/recorder/session-recorder.test.ts`):

| mutation | red |
|---|---|
| no eviction at all — the defect | **7 of 8** (the constant law stays green, correctly: it is about the number, not the behaviour) |
| keep the oldest, drop the newest | the oldest-first law, **alone** |
| never compact the evicted slots | the slots law, **alone** |
| a rotation carries the eviction count across | both rotation laws |
| the fold repair ignores eviction | the repair law, **alone** |
| the resumed fold is taken over the window, not every event | fold-completeness + repair |

No law is redundant and each mutation's red set is precise. The three that go red **alone** are
the ones worth reading: they are the properties a plausible-looking implementation gets wrong.

**They run at the real 75,000, not an injected bound.** A configurable ceiling exercised only
at 4 would never test the default it exists to protect — AGENTS.md's own second-worst defect
shape. It is affordable because `resumeFrom` seeds a whole session in one constructor call
(~130 ms for 76,025 events) and because the one law that must drive `record()` mocks the
writer's append: the bound is a memory property, so 75,003 disk writes would buy nothing.

One thing the event *type* forced, worth knowing for the next large-session law here:
`worktree.discovered` accumulates a record per worktree, so folding 76,025 of them copies a
growing map 76,025 times and takes **minutes** — the first probe hung on it. `collector.error`
folds into one entry plus an errors list capped at `MAX_ERRORS`, and the same 76,025 events fold
in 36 ms. The laws say so where they build their fixtures.

**No new law is named `*.bench.test.ts` and none carries `// @gate-timing`** — every assertion
is a count (events, evictions, slots, `reduce`/`reduceAll` calls). `.swarm/timing-count` does not
move.

## Not proven here

- The `MAX_BUFFERED_EVENTS === 75_000` law pins the operator's number but cannot see the
  client's. The stronger form is a seam test on the **web** side importing both constants, the
  way `web/src/app/boot-reason-seam.test.ts` already reads `SESSION_BOOT_REASONS` from server
  source. Out of fence.
- `rotate.ts:147` also hands `eventsSoFar()` to `captureSessionTranscripts`. Past the ceiling a
  lane appearing only in evicted events would be missing from the capture. Not a one-line fix —
  it needs the session from disk — so it is named, not widened into.

## Gate (lane checks)

Node v22.23.2, in the lane's own worktree:

- `npm run typecheck` — **0**
- `npm run lint` — clean, 850 files
- `VITEST_MAX_WORKERS=6 npm test` — **382 files, 6706 passed, 3 expected fail, 6 skipped**

Landing (`scripts/gate.sh`, with `[load-batches]` since this branch touches tests) is the
operator's step, not run here.


ciaran-slow:
**Both follow-ups this PR names are now filed and boarded**, so neither is a comment-only recommendation (the failure mode #130 exists to break):

- **#132** — `stream honesty: the client can tell a live window is partial, whatever the server had room to send`. The honesty gap in the "two consumers" section. Fenced to `web/src/app/StreamContext.tsx` + test, deliberately **not** `api/meta.ts` (the number is already served there at `:253`, and PR #93 holds that file) and **not** `streamState.ts` (the label's rule is right; its inputs are what changed). Only the live branch is affected — `StreamContext.tsx:186`; the replay and fixture branches fold a whole recording.
- **#133** — `transcript capture: a rotation captures every lane the session named, not every lane still in the window`. The `rotate.ts:147` case from "Not proven here". Fenced to `log/transcript-capture.ts` + test plus the one call site, with both candidate shapes (read the closed log from disk vs derive from the maintained fold) written down and the choice left to the plan.

Both are **blocked by this PR** — neither defect exists until the window is bounded — and both carry a DoD line requiring the law to fail on today's code, so neither can be closed by a test that was already green.

Boarded Soon / High / Bug.

PR #140, opened 2026-08-27
docs: prd-47 — the answering hand, and the input-latency audit it stands on

KelliherL:
Two documents, paper only.

**The audit** (`docs/review/2026-08-24-input-latency-audit.md`): six read-only lanes over the drag path at `4140f6b`, re-verified at `48c3476`. The operator symptom — drag latency and stutter in the packaged shell — attributed: a pure camera change re-runs the whole model+tessellation though the camera is one uniform; the worst frame runs 1.9–3.3x the median at pinned inputs and no table prints the tail; FleetContext re-renders five surfaces per SSE event against its own stated beat; a paused, provably identical picture is rebuilt at 60 Hz. Six tempting fixes refuted with numbers so nobody re-proposes them. One verification finding for prd-44's authors: **#32's committed after-table does not reproduce on this box** (persistent200 paint 17–19.5 ms at each of three post-#32 commits vs 2.58 ms committed; the two newest scene commits exonerated by bisect) — a cache-hit counting law would settle it in one run.

**The PRD** (`docs/prds/prd-47-the-answering-hand.md`, **proposed** — blessing sought at the meet): four rulings (repaint-not-rebuild · identical-frame skip · the beat honoured · variance measured before fixed), sequencing in four waves with no live-fence collision (checked against prd-42/43/46 lanes, prd-33's model stage, prd-44's shipped territory). Mechanisms are marked candidates, not rulings — prd-44's own closeout lesson, applied.

No milestone is created and no issue is filed until blessing (bless before backlog).


ciaran-slow:
prd-44's author here. I built the counting law your addendum asked for and ran the bisect behind it. Short version: **you were right that something was wrong, right that the palette carrier and the quality dials are innocent — and neither of the two causes you offered is it.** There are two discontinuities, one of them is benign, and the other one is mine.

Everything below is `EXECUTED` on macOS, each cell the suite's own `retired-field persistent200` line, filtered (`-t "200 retired"`) so it is comparable within itself, endpoints re-measured minutes apart in a throwaway worktree.

## 1. The cache hits. 600 of 600, zero misses

Counted on the real `persistent200` field, not timed:

```
frame 0: settledRibbons=600  misses=600  (cold fill)
frame 1: settledRibbons=600  misses=0
frame 2: settledRibbons=600  misses=0
frame 3: settledRibbons=600  misses=0
```

So *"the cache does not hit on a second box's run shape"* is dead, and the mechanism has no box-dependent term in it. Filed as #146, with a second arm that moves one settled ribbon's paint and requires exactly one miss — because `misses === 0` alone would pass just as happily over an empty population.

## 2. The ~7x is a stage substitution, not a regression

The premise needs one correction, and it is the load-bearing one: **the after-table is not #32's.** Those rows were last written by `da9d10a` on 2026-08-05 — #178's before/after for the geometry and marks caches. `b89ca23` (#32) touched `batch.ts`, `frame.ts` and `frame.test.ts`, and never opened `perf.test.ts`. The 2.58 ms figure predates the tessellation cache by three weeks and was never a claim about it.

The tell you spotted and could not place is that the table's *before* column (2.93 ms) is exactly as unreproducible as its after. Here is why. At `da9d10a` the paint stage timed `paint({ ctx: draw.ctx, … })` — the canvas-2D painter issuing calls into `perf.test.ts`'s own recording stub, whose methods are *empty bodies* incrementing counters, because jsdom has no 2D context at all (ADR-0006). `8686f24` (#578) replaced that with `buildFrame`, which really tessellates, under the same jsdom:

| --- | --- | --- |
| `8686f24^` — 2D calls into a stub | 1.638 ms | 0.235 ms |
| `8686f24` — #578, real tessellation | 12.543 ms | 1.104 ms |

2.58 ms and 17–19.5 ms are not the same quantity measured twice. Nothing regressed there; a stage started measuring work it had never been able to reach. This is your first candidate — *"the table was measured under conditions the suite does not reproduce"* — and it is correct, but the condition is a painter swap rather than a run shape.

## 3. Why your three probes read flat, and what they bracketed

Worth reconciling before this doc lands: your addendum labels `6b2256c1`, `1de8a870` and `609935cd` as *"each of three post-#32 commits"*, and on this history all three are **ancestors** of #32 (`git merge-base --is-ancestor <sha> b89ca23` for each). `6b2256c` is #579 itself — the first scene commit after `8686f24` — and #32 is `b89ca23`, dated 2026-08-26, after all three.

That is not a wasted bisect, it is a load-bearing one: three flat readings in a plateau bounded below by #578 and above by #32, which is exactly why the palette carrier and the quality dials came back exonerated. My own bisect agrees with yours cell for cell across that window. The window simply did not contain either thing that moved. It also explains the absolute gap between our boxes without appealing to WSL2 — your own run-shape note (full suite ~2x the filtered cells) covers 13.4 → ~18 ms.

## 4. #32 hits perfectly and is a 1.84x regression on the cell it claimed to fix

Bisecting the 31 scene commits from `8686f24` to `48c3476` puts the whole remaining rise on one commit, mine, with the living-only ceiling flat across it:

| --- | --- | --- |
| `c31d04b` (mid-range) | 12.539 ms | 1.095 ms |
| `d6bbf78` (#32's grandparent) | 13.297 ms | 1.879 ms |
| `f60e0af` (#32's parent) | 13.392 ms | 1.904 ms |
| **`b89ca23` (#32)** | **24.682 ms** | 1.888 ms |

Both things are true at once: it warms perfectly and it costs nearly double. `digestRibbon` walks every point of every settled outline into a string every frame — a few hundred points x 600 ribbons x 60 Hz — and the walk costs more than the tessellation it saves. Isolated at `b89ca23`:

| configuration | paint (persistent200) |
| --- | --- |
| `isSettledRibbon` forced `false` (the pre-#32 path) | 13.312 ms |
| as shipped (content digest over every outline point) | 24.682 ms |
| slot keyed on the `outline` array's own identity | **10.309 ms** |

The saving is real and worth ~23% of the stage. The key is what turns it into an 84% loss — and the discipline it broke already existed one stage up, where `geometry.ts` and `marks/thread.ts` both key on array identity. Filed as #147, blocked on #146.

## What this means for prd-47

Finding 4's sentence — *"prd-44 ruling 5's shipped form (the per-mark digest cache) does not reach this"* — is right, and now more so: the shipped form is currently a net loss, so nothing here should be inherited as prior art or as a measured baseline. Your own framing already anticipated it (*"was tried and rejected on its own measurement"*); that warning turns out to apply to a mechanism still sitting in the tree.

Two things I'd suggest for this PR rather than a follow-up, both small: the sha ordering in §3, and the addendum's disjunction, which can now be replaced with the answer. I'm happy to push either as a commit here if you'd rather not carry it.

## The part worth saying plainly

The bug was mine, the table that hid it was mine, and the reason it was findable at all is that your addendum refused to guess between two hypotheses a wall clock could not separate and asked for a count instead. A timing would have shown me 24 ms and I would have blamed the box. Thanks for filing it as a verification finding rather than assuming either way.

PR #141, opened 2026-08-27
docs: prd-48 — the shared record, its research notes, and the metamorphosis design restored

KelliherL:
Paper only, three commits.

**The restore** (`docs/metamorphosis/system-design.md`): the 2026-08-13 multiplayer direction — observatory = log-shipper to one team server, two credential planes, SQLite-first behind a storage interface with Postgres the growth step, one self-migrating container, P2P rejected by name — recovered byte-for-byte from pre-deletion commit `7541e3d4`. prd-37 cites it three times; until this PR those citations resolved to nothing on the trunk.

**The research notes** (`docs/research/2026-08-24-shared-record-*`): the design brief maps the VPS/team-server ask onto the organs that already exist (the hash-chained ledger, `mergeRecords` with zero callers, the reserved `signature`, /connect's three-state discipline) and argues the one decision everything hangs on — ship the ledger up and read the team from the server, versus a replicated database on every machine. Amended on landing for prd-44's shipping: retention is ruled (**no default age, ever**), so the brief's proposed defaults are recast as org ceilings composed with `log/retention.ts`'s grammar. The spike plan turns the unknowns into eight bounded spikes — corpus census, shipper chaos round-trip, local store, schema at 10M rows, borrowed identity, the veil proven on the wire, retention/archive mechanics, a timed VPS install — each with the falsifier that would change its verdict.

**The PRD** (`docs/prds/prd-48-the-shared-record.md`, **proposed** — blessing sought at the meet): a research PRD in the house's research → verify → build order. Its waves produce notes, not code; wave 0 books the leads' rulings on the brief's §10; the build PRD ("the split") and the successor ADR to 0009 are the programme's output. Touches no live fence by construction (`docs/` and out-of-tree throwaways only).

No milestone is created and no issue is filed until blessing (bless before backlog).


KelliherL:
The preliminary spikes ran tonight — seven of eight, executed, one box. This branch now carries the notes (`docs/research/2026-08-28-shared-record-s*.md`), **the proposed system architecture** (`…-architecture.md`), and an amendment re-cutting prd-48's waves to what remains. Headlines: event ids are NOT unique (the `mergeRecords` dedup key would silently drop real events — 74.5% loss under the assumed key), an accept-fast 202 loses exactly one batch unless the ack follows a durable journal write, the wire must carry re-serialized lines (old-era `pane.activity.preview` — real terminal content — crossed otherwise; chain digests proven identical either way), 4 of 5 org questions answer in 0.19–452 ms at 10M rows and the fifth needs a rollup (7,700×), DROP PARTITION beats DELETE 3.5 s vs 17.6 s, archive invariant crash-drilled 15/15, compose-to-green in 14.2 s. Accept it, re-cut it, or refute any line with a better number.

hchristinasmith (approved):
Approving. Paper only, and the paper holds up. Everything below is **EXECUTED** — run on `060092d` in a clean worktree, node 22.22.2 — unless labelled otherwise. One open question at the end that I do not think should hold the merge.

**Both defects the amendment reports are real, and I reproduced one against the shipped code rather than reading S2's account of it.** Built the exact shape S2 measured — one actor whose `evt-NNNNNN` counter restarts mid-log — and ran the real `mergeRecords`:

```
INPUT  events=6  distinctIds=3  [evt-000001 evt-000002 evt-000003 evt-000001 evt-000002 evt-000003]
MERGED=6  EXPECTED_IF_LOSSLESS=9  DROPPED=3
```

Three real events gone, silently. And `docs/record-format.md:251`'s "an event id is only ever unique within one actor's own log" is contradicted by S2's own measurement — 16,218 unique ids across 63,653 lines in one session log — which my probe reproduces from a different direction. Both filings are correct, and characterising them as latent is correct too: `mergeRecords` has zero production callers, only its own test file imports it. I checked rather than taking the brief's word for it.

**Every organ the brief maps the ask onto is really there.** Hash-chained ledger at `record/build.ts:18-22` (`prevHash`, `chainDigest`); the reserved `signature` at `record/schema.ts:42-49`, genuinely `z.null()` with the "reserved for a future signature over `chainDigest`" comment the brief cites; /connect's three-state discipline at `connect/index.tsx:158`, named as the #367 law. The brief is describing the tree, not a remembered version of it.

**Two things I suspected and cleared by checking, worth recording so nobody re-raises them:**

- S6's absence is not a silent gap — the amendment says "seven of the eight" outright, and S6 survives as a named wave 3 item (`prd48 w3: the veil is proven on the wire`) with its residue transferred to the build PRD. Coherent with `587f7e5`'s "S2 shipper+veil".
- `packages/core/src/fence126.ts` in the S4 note is a **synthetic** path from the 10M-row generated corpus, not a claim about the tree. The note discloses the corpus is generated (§"Generated corpus") and names its own payload-realism ceiling. Every other path-like reference across the twelve files resolves, once the `log/retention.ts`-style shorthands are expanded to their package roots.

## Gates

    npm run lint                      exit 0
    VITEST_MAX_WORKERS=6 npm test     exit 0   383 files, 6731 passed,
                                               3 expected fail

`main` is **26 commits** ahead of this branch's merge base, so I built the actual post-merge tree (`origin/main` plus these twelve files — no path overlap at all) and ran it there too: 383 files, 6766 passed, exit 0. That was the run I most wanted, since a doc law landing on `main` in the last day is the one thing that could redden paper. Nothing did.

CI checked per job step rather than by the green bar: Test, Typecheck, Lint, Packaging guard and Boot smoke all `success` on `build-test-boot (ubuntu-latest, current)`.

## The open question — not blocking, but it should not stay unanswered

**`7541e3d4` is not reachable from this repository.** `git cat-file -t 7541e3d4` fails, `git fetch origin 7541e3d4` fails, and the only commit in the entire history touching `docs/metamorphosis/system-design.md` is this PR's own `3b6909c`. So "recovered byte-for-byte" cannot be checked by anyone reading the trunk.

`3b6909c`'s message explains why — it "survived the repo's loss only on a contributor's local branch" — so I read this as honest and unverifiable rather than wrong, which is why it is a question and not a change request. But it is not only in a commit message: the design brief cites the sha twice in its body, once with a line range (`` `7541e3d4:92,110-118` ``, at `2026-08-24-shared-record-design-brief.md:180`), which reads as a checkable citation and is not one.

There is an irony worth naming, given what this PR is for: it restores a document *because* prd-37's three citations resolved to nothing on the trunk, and it lands citing a sha that resolves to nothing. Pushing the object, or one line in the brief saying where it lives, closes it permanently.

I could not fix this myself — I do not have the object, and the rest of it lives in commit messages I would have to force-push to change, which I will not do on your branch.

## Two smaller things

- `3b6909c` says prd-37 "cites this document three times". By exact path (`docs/metamorphosis/system-design.md`) it is **twice**, at `prd-37:13` and `prd-37:42`; there are ten references to "the metamorphosis" more loosely. The substantive point — that prd-37 depends on this document and the dependency dangled — is unaffected.
- `060092d` re-cuts the whole wave plan after four assumptions were overturned by measurement, and carries **no body at all**. It is the most decision-relevant commit in the PR, in a repo where every other commit argues for itself. The PRD's own Amendment section does the explaining, so nothing is lost to a reader who gets there — but the commit does not point at it.

## What I did NOT run

I did not verify the S1/S3/S5/S7/S8 measurements — single-box, single-day numbers with stated loadavgs are not reproducible from here, and the notes grade their own claims [EXECUTED]/[REASONED], which is the right form. I read them for internal consistency against the PRD's waves and the architecture synthesis, not for replication. I also did not re-derive S4's 10M-row timings.

The Sequencing section still describes the pre-amendment plan with no forward pointer to the Amendment that re-cuts it. Offered, not asserted: a "superseded — see Amendment" line would save the next reader. Your house style on PRD amendments may well be to leave the original text standing, and if so, ignore this.

Issue #143, opened 2026-08-27
prd42 w9: the forward-transform law sees the negated-class shape


gabriel-canaan:
## The defect

`forward-transform-law.test.ts`'s check 3 (`findHandRolledForwardTransforms`,
`:386-404`) exists to catch a second, competing path-to-slug implementation. Its
signature is *"a positive character class that folds `/`"* — reasonable when the
canonical transform was a positive class listing `/`. Since #124 the canonical
transform is `/[^a-zA-Z0-9]/g`, a **negated** class, and a negated class never
contains a literal `/`. So the most likely form of duplication — copy-pasting
today's canonical implementation — is invisible to the check that exists to find
it.

EXECUTED. A second implementation staged as a real tracked file
(`collectors/sessionlog/zz-second-impl.ts`), same name, same call shape, only the
class differing:

    return p.replace(/[^a-zA-Z0-9]/g, '-')     ->  law GREEN, 32/32   (undetected)
    return p.replace(/[/_.\\: ]/g, '-')        ->  law rc=1           (caught)

## The one-line fix is a NO-OP — proven, so nobody spends an afternoon on it

The obvious repair is to delete the two negated-class exclusions at `:393` and
`:401` (`if (body.startsWith('[^')) continue`). **That changes nothing.**
EXECUTED: with BOTH exclusions removed, the law still passes 32/32, and the
negated-class duplicate above is STILL undetected — while the positive-class
control is still caught.

The reason is the next guard, `:394` / `:402`: `if (!body.includes('/')) continue`.
A negated class does not list `/`, so it is filtered one line later regardless of
the exclusion. The exclusion is not what hides the shape, and an earlier review
report that said it was is wrong on this point.

## What the work actually is

A new detection signature for the negated-class shape — which means deciding what
now counts as a competing implementation, since *"folds `/`"* no longer
distinguishes one. Candidates, none ruled:

- match the canonical body (`[^a-zA-Z0-9]`) specifically, and allow it only in
  the one sanctioned module;
- key on the call shape rather than the class (a `replace` whose result is joined
  under a projects root — check 1 already does this for the join case, `:466`);
- key on the function's role (a single-argument path-to-name transform in
  `collectors/sessionlog/` other than the canonical export).

## Definition of done

- A verbatim copy of today's canonical implementation, in a tracked file under
  `packages/server/src`, fails the law. EXECUTED with the staged-file method
  above, and with a CONTROL proving the law still passes on the unmodified tree.
[...]

Issue #142, opened 2026-08-27
prd42 w9: an ambiguous slug is decided by the transcript's own cwd, not by guessing


gabriel-canaan:
## The defect

`reverseProjectSlug` (`packages/server/src/concierge/repos.ts:237-270`) walks a
slug back to a real path one directory hop at a time, and among matching entries
the LONGEST encoded form wins. That rule refuses a **tie** at the longest length
(`foo-bar` and `foo.bar`, both encoding `foo-bar`) — but it silently commits to
the longest when two candidates of DIFFERENT lengths each lead to a real path.
Two different real directories can encode to one slug, and the walk answers with
one of them and no sign it was ever in doubt.

EXECUTED, and it needs no exotic punctuation — only `/` folding and a literal
dash, which is a plausible monorepo layout:

    /repo/packages/web   ─┐
                          ├─ both encode to  -repo-packages-web
    /repo/packages-web   ─┘

    reverseProjectSlug('-repo-packages-web')  ->  '/repo/packages-web'   (silent)

CONTROLS: with the walk regressed to its **pre-#120** five-character class the
same shape still answers silently (`/repo/a.b/c` vs `/repo/a.b-c` ->
`/repo/a.b-c`), so this is **pre-existing** and not a regression from #120. A
same-segment tie refuses correctly. A single real candidate resolves correctly.

`listKnownProjects` (`packages/server/src/concierge/repos.ts:347`) then marks that path resolved, so the repo
picker offers the operator a directory the session never ran in.

## The ruling — operator, 2026-08-28, in session

**On a slug that more than one real path can produce, the walk decides from
evidence and refuses when evidence is absent. It never guesses.**

The evidence is already on disk: the transcript under that slug records its own
`cwd`. Measured on a real session log — **940 of 1252 records carry a `cwd`
field**, holding the real absolute path (`/home/operator/launchpad/rhizomorph`).
So the ambiguity is decidable rather than a coin toss.

1. When two or more decompositions each resolve to a real directory, read the
   slug's transcript and take its recorded `cwd`. If that cwd is one of the
   candidates, it IS the answer.
2. **Fallback: refuse.** No record carries a `cwd` (312 of 1252 did not here),
   the file is unreadable or malformed, or the cwd names none of the candidates
   — return `path: null` with a reason naming every candidate. An honest refusal,
   never the longest.

*Rejected alternative — keep preferring the longest.* It shows a wrong path with
[...]

PR #144, opened 2026-08-27
prd42 w7: a path means the same thing on both sides of the transform

### LINKED ISSUE #120: prd42 w7: the reverse walk fails closed over the real slug grammar

### LINKED ISSUE #124: prd42 w7: the forward transform mints the slug Claude Code actually mints
## The defect

`worktreePathToProjectSlug` maps six characters — `/` `_` `.` `\` `:` and a
space. Claude Code's real slugger maps **every non-alphanumeric character**, then
caps the result at 200 characters and appends a base36 hash past that.

Verified against the shipped binary rather than quoted: `grep -a` on the
installed Claude Code ELF (2.1.246) finds
`function B(e){return e.replace(/[^a-zA-Z0-9]/g,"-")}`. The encoder's own doc
comment records the same transform from 2.1.241 under a different minified name,
and records the length cap and hash suffix beside it — so this gap is documented,
measured, and has simply never been closed.

Consequence: for any repo path containing punctuation outside those six
characters, this function mints a slug that is **not** the directory Claude Code
actually created. The transcript is under the real slug; we look under ours; the
paths never meet.

## Why now, and why separately fenced from its wave peer

Its peer is the reverse-walk half. Widening one side without the other is what
created the divergence this PRD is named for — three times now: the space
(ruling 1), then the colon and backslash (#47), and now the rest. So both sides
move in the same wave, as two issues, each owning one side.

The fences are disjoint by design: this issue owns the forward transform and its
test, the peer owns the walk, its test, the forward-transform law and the
coupling registry.

## The ordering question, UNVERIFIED — check it first

The two halves are believed independent, so either may land first:

- **walk widens alone** — verified during #47's review: encoder-minted slugs with
  an unhandled character stop reversing (honest refusal), which is harmless today
  because nothing composes encoder to walk and `reverseProjectSlug` has exactly
  one caller.
- **encoder widens alone** — **NOT verified.** The claim is that the round-trip
  law does not break, because its generator only produces characters both sides
  already handle. **Test this before writing any code.** If it does break, the
  law lives in the peer's fence and this becomes a stack rather than a peer pair;
  say so on the issue rather than reaching across.

The round-trip law's generator must **not** grow here. It grows only once both
sides agree, and it lives in a file this fence does not claim.

## Definition of done

- The forward transform maps the real grammar, and the length cap and hash suffix
  are either implemented or their absence is stated in the doc comment as a named,
  separate gap — not left implied.
- EXECUTED, with a CONTROL: a path containing a character outside the old six
  mints the slug Claude Code actually mints. The control is a path using only the
  old six, whose slug must not change.
- The doc comment stops describing a six-character class, and its three
  present-tense passages about the old class are corrected.
- `npm run typecheck`, `npm run lint` and the full suite green, and the
  round-trip law green WITHOUT its generator being touched. If green requires
  touching it, STOP and report — that is the stack signal above.
- The eight call sites downstream of the encoder, counted but never individually
  tested in ruling 1, are enumerated in the commit body with a note on whether any
  caches a slug across the change.

## Fence (may touch ONLY)
- `packages/server/src/collectors/sessionlog/worktree-slug.ts`
- `packages/server/src/collectors/sessionlog/worktree-slug.test.ts`

EXPLICITLY OUT, and owned by this wave's peer: the concierge reverse walk, its
test, the forward-transform law, and the coupling registry. Also out: the
round-trip law's generator, wherever it lives.

## Blocked by
**prd42 w5 landing (PR #121).** That PR moves the round-trip law out of
worktree-slug.test.ts, so this issue's fence describes a file whose contents
change first. Do not dispatch before it merges.

Peer in the same wave: the reverse-walk half. Fences verified disjoint.

## Provenance
Named as unfiled work in prd-42's wave-map amendment: *"wave 7 also has a second
half that is not yet an issue: the encoder needs the same grammar, plus the
200-character cap and base36 hash suffix the real slugger applies past it."*
This is that issue.

gabriel-canaan:
Closes #120
Closes #124

prd-42 wave 7: a path means the same thing on both sides of the transform. The
walk (#120) and the encoder (#124) each move to Claude Code's real grammar —
`/[^a-zA-Z0-9]/g`, read out of the shipped binary — and a third commit adds the
law that pins their agreement, which neither issue could write alone.

|---|---|---|
| `9ed40ad` | #120 | the reverse walk re-encodes real directory entries with the full grammar, so a same-segment collision refuses instead of returning a sibling |
| `9ae3705` | #124 | the forward transform mints the slug Claude Code actually mints |
| `4e791b2` | #120 | the round-trip law generates the characters the wave widened, so neither half can narrow back in silence |

The third commit closes no issue of its own: it discharges #120's own done-when
clause, which says the law's generator grows "only once both sides agree". The
merge of these two commits is that event, which is why it could not be written in
either lane.

## The premise, verified rather than quoted

`grep -a` on the installed Claude Code ELF (2.1.247) returns, byte-exact:

```
function B(e){return e.replace(/[^a-zA-Z0-9]/g,"-")}
function J(e){let n=B(e);if(n.length<=v)return n;return`${n.slice(0,v)}-${tn(e)}`}
function L(e){return g(k(),G(e))}      // G = override ?? J
```

with `v=200`. CONTROL: the same grep with `[^a-zA-Z0-8]` returns 0 hits. So `B`
is the class both halves now use, and `J` is the cap-and-hash wrapper #124
deliberately does NOT implement — `L` proves the project directory name is the
capped slug, so that gap is real, named in the doc comment, and fails in the safe
direction (an over-200 path mints a 245-char uncapped slug, which resolves to
nothing rather than to something wrong).

## What is the sibling case?

**The other side of the transform.** Every prior fix here moved one side and left
the other: prd-42 ruling 1 widened the encoder for the space, #47 for the colon
and backslash, and each time the walk or the law stayed where it was. That is the
defect this PRD is named for, and it is why these two commits are peers in one
wave rather than a stack — and why the third commit exists, because a law that
pinned only the forward transform would leave the walk free to narrow alone.

The round-trip law, as it stood, survived **all three** of these. It now catches
each — EXECUTED, with the law as the only judge:

```
encoder narrowed to the pre-#124 six      -> law rc=1  CAUGHT
walk narrowed to the pre-#120 five        -> law rc=1  CAUGHT
encoder narrowed to /[/_.\: +~(),'@-]/g   -> law rc=1  CAUGHT
```

CONTROL: green 57/57 against the unmutated tree. That third mutation is the one
worth naming — a class fitted to exactly the sample `worktree-slug.test.ts`
asserts, so it passes that file 12/12 while still mishandling `&`, `é` and `😀`.
#124's commit body originally claimed its probe test would redden "any subset";
that claim was false, and `4e791b2` is what makes it true.

It would still survive a change to a character the generator does not name, which
is why three assertions pin the generator's own coverage rather than a comment
asking future readers to remember.

## Fence widening

**None.** Exactly the six files the two issues fence between them.

## Scope corrected before merge, not after

#120's done-when said "every collision cell fails CLOSED". As written that was
unmeetable by a one-line change, and it is now scoped to **same-segment** on the
issue and in `9ed40ad`'s body. A CROSS-segment collision (`a+b` holding `c`
beside a sibling `a+b-c`) is still answered by the longest-match rule. That is
pre-existing — regressing the line to the old five-character class answers the
same shape silently too, and the purest instance needs no character this wave
widened (`/repo/packages/web` vs `/repo/packages-web` collide on `/` and a
literal dash). `repos.ts:188-197` only ever promised to refuse a tie at the
longest match, and that promise is kept. Filed as #142 with an operator ruling.

## Verification

Two independent seats (Opus, and Codex `gpt-5.6-terra`) plus the orchestrator's
own executed probes. Both seats found the cross-segment shape; Codex's claim that
this wave *introduced* it did not survive being checked against the pre-#120
code. Every Codex finding was re-run against the real tree, because its own repo
commands died at sandbox init (`bwrap: loopback: Failed RTM_NEWADDR`) — its
labels rested on standalone reimplementations.

Findings fixed here: the unpinned law (`4e791b2`), five stale doc passages that
each became false only in the merge, and two test names that claimed more than
they asserted. Findings filed instead: #142 and #143 — #143 records that the
obvious one-line fix for it is a **no-op**, proven, so nobody spends an afternoon
deleting an exclusion that changes nothing.

Gates run by the reviewer, exit statuses read: typecheck `rc=0`, lint `rc=0`,
`VITEST_MAX_WORKERS=6 npm test` `rc=0` — 383/383 files, 6756 passed.

**For the landing operator:** #120's lane saw two failures under 6-worker load
that did not reproduce here, and one of them carries `@gate-timing`. Give
`scripts/gate.sh` its load-batch count, or the serial timing pass does not run
and the landing goes green on the friendlier condition.


ciaran-slow (approved):
## Approving. Everything load-bearing here re-derived independently, and the premise verified on a different platform than the author's.

Reviewed commit-by-commit in a clean clone, node v22.23.2. Every label below is
**EXECUTED** unless it says REASONED.

---

### The premise holds, and on stronger evidence than the PR claimed

The whole wave rests on `/[^a-zA-Z0-9]/g` being Claude Code's real project-slug
class. The PR evidences that from a Linux ELF at 2.1.246/2.1.247. I checked a
**macOS binary at 2.1.241** — different platform, different version, entirely
different minified names — and traced the full call chain rather than grepping
for the class in isolation:

```
function jS(){return H1.join(Cn(),"projects")}
function s$(e){return BOu()??G7(e)}
function DM(e){return H1.join(jS(), s$(e))}
function G7(e){let t=fNr(e);if(t.length<=Hae)return t;return`${t.slice(0,Hae)}-${fJ_(e)}`}
function fNr(e){return e.replace(/[^a-zA-Z0-9]/g,"-")}
```

`DM(path)` is the projects directory, and it bottoms out in `fNr` — the exact
class both halves now use. `Hae=200`, confirming the cap. Author's `B`/`J`/`G`/`L`
map onto my `fNr`/`G7`/`s$`/`DM`.

**This mattered.** The binary contains at least ten distinct dash-folding
classes, six of which preserve `-` and/or `_` and would give a different answer:

```
8  replace(/[^a-zA-Z0-9\-_]/g,"-")
6  replace(/[^a-zA-Z0-9_-]/g,"-")
4  replace(/[^a-zA-Z0-9]/g,"-")     <- the one reached from "projects"
4  replace(/[^A-Za-z0-9_-]/g,"-")
...
```

Picking the wrong one was a live risk and the author picked the right one.
CONTROL: the same grep with `[^a-zA-Z0-8]` returns 0 hits, so the match is not a
pattern-matching artifact.

### `4e791b2` is the commit that earns this PR

The claim is that the round-trip law could not fail for the reason it existed. I
tested that directly — restored the pre-`4e791b2` law and ran all three
mutations against it, with the law as the only judge:

| mutation | OLD law | NEW law |
|---|---|---|
| encoder → pre-#124 six | **rc=0 (survived)** | rc=1 CAUGHT |
| walk → pre-#120 five | **rc=0 (survived)** | rc=1 CAUGHT |
| encoder → class fitted to the sample | **rc=0 (survived)** | rc=1 CAUGHT |

CONTROL: unmutated, both green. The old law was green against every one — a
textbook "test that cannot fail for the reason it claims," and this commit
genuinely closes it. The third-commit-closes-no-issue structure is right, and
#120's done-when ("the generator grows only once both sides agree") is what
licenses it.

I also confirmed the PR's own self-correction: the fitted-class mutant passes
`worktree-slug.test.ts` **12/12** while the law catches it. So #124's original
"reddens any subset" claim was indeed false, exactly as the PR body admits.
Self-reported false claims that check out are worth more than clean ones.

### Behaviour, against a live filesystem

Not the fixture `DiscoveryFs` — real `mkdtemp` directories, composing the real
encoder into the real walk:

```
13/13 collision cells (+ ~ @ , ' ( ! # % = ; é 😀)  -> REFUSED (ambiguous)
control  aZb alone                                  -> CORRECT
control  aZb beside decoy a-b                       -> CORRECT
4/4 round trips (wide+punct~at@x, quote'comma,paren(x), accentéhere, emoji😀here) -> CORRECT
```

The collapse mutant `/[^a-zA-Z0-9]+/g` reddens **5** tests — 4 pre-existing plus
the renamed one, matching `9ae3705`'s honest "local guard rather than the only
one" note down to the count.

### Gates, on the actual post-merge tree

Merged `origin/main` in first (clean, no overlap with the six files):

```
typecheck                 rc=0
lint                      rc=0   853 files
VITEST_MAX_WORKERS=6 test rc=0   383/383 files, 6763 passed, 3 expected fail, 6 skipped
build                     rc=0
```

**On the landing note in the PR body:** the two failures #120's lane saw did not
reproduce. I ran the full `@gate-timing` + bench set serially — all 11 files
(`fanout`, `concurrency`, `session-recorder`, `session-log`,
`transcript-capture`, `retention`, `perf`, `marks`, `reduce.bench`,
`spend-cursor.bench`, `replayFold.bench`) — **rc=0**, `session-recorder.test.ts`
included. The serial-timing risk you flagged looks clear on this box; the
load-batch argument is still the right call at landing.

`gh pr checks 144`: all **7** jobs actually executed and passed — 3
`build-test-boot` legs (macOS×min correctly excluded) and the full 2×2
`pack-smoke`. Not just a green bar.

### Fence and hygiene

Exactly six files, the precise union of the two issues' fences — no widening, as
claimed. Both issues are prd42 **w7**, so one PR = one (milestone, wave). Only
synthetic paths in the diff (`/Users/dev`, `/home/operator`); nothing naming a
real machine.

Sibling sweep: only two path-slug sites exist in the tree, and both are now on
the full grammar. `snapshot-store.ts:42` and `log/paths.ts:22` are name
sanitisers, a different concern. No narrow-class sibling left behind.

I also checked the one call site whose output crosses the change — `cutSession`'s
`sessionFile` is persisted absolute into `fork.checkpoint` and read back
literally by `verifySessionPrefix`, never re-derived. #124's caching analysis is
accurate. (Bonus: pre-change checkpoints under punctuated paths were already
broken, so this is strictly an improvement there.)

---

### One thing I got wrong, since it cuts both ways

My first live probe hardcoded `a-b` as the dash-sibling for every character and
reported the emoji cell **resolving** instead of refusing. That was my bug, not
yours: `a😀b` folds to `a--b` (two UTF-16 units, two dashes), so `a-b` is the
wrong sibling and there is genuinely no collision to refuse. Your test *derives*
it — `const sibling = encoded` — which is why it is correct and my hand-rolled
version was not. Re-run with the derived sibling: 13/13.

### Non-blocking, filed as #148 — not a defect in this PR

The law's discriminator excludes negated classes (`if (body.startsWith('[^'))
continue`). As of #124 the *sanctioned* encoder **is** a negated class, so the
law named "worktreePathToProjectSlug is the only path-to-slug implementation"
can no longer see a second implementation created the most likely way — by
copying the real function. EXECUTED, against the real detector:

```
copy of the CURRENT sanctioned encoder  -> []                        NOT DETECTED
the PRE-#124 shape                      -> [".replace(/[/_. ]/g, '-')"]  DETECTED
```

**Explicitly not held against this PR.** The gap is pre-existing and the
discriminator's own doc comment already names it ("#47's own real shape ... is a
known gap"), and neither #120 nor #124 set any criterion about detector
coverage. What the wave changed is reachability — from a gap you'd have to invent
to one you hit by doing the obvious thing. That is the PRD's own
"boundary moved, not only narrowed" shape, this time applied to the law rather
than the code it polices, which is why it deserved an issue rather than a
shrug. It is also not a reviewer one-liner: deleting the exclusion falsely
convicts the two legitimate name-sanitiser sites, so it needs a real
discriminator. Hence #148 (prd42, soon/medium/task) rather than a commit here.

### What I did NOT run

- **`scripts/gate.sh`** — that is the operator's step, and running it *is* the
  landing.
- **Any Windows leg.** There is no Windows runner in `ci.yml` at all (only
  `desktop.yml:40`, packaging), so the `C--Users-...` slug path is covered by
  unit assertions and nothing executed. Pre-existing gap, prd-25's territory —
  and `reverseProjectSlug` refuses drive-rooted slugs by design, so the blast
  radius is a refusal, not a wrong answer.
- **A real `claude -p` session probe** in a punctuated directory. I verified the
  binary's actual code path instead, which is the fact that probe was evidence
  for.
- **`zft`**, the hash behind the base36 suffix. Irrelevant to merge — #124
  correctly declines to guess it, and the named gap fails safe: an over-200 path
  mints an uncapped slug that resolves to nothing rather than to something
  wrong.

No blocking findings. The commit bodies are unusually accurate — every EXECUTED
claim I re-derived came back as stated, including the ones that admitted
something was wrong.

**Not merging** — merges on this repo are manual. Ready when you are.

Issue #147, opened 2026-08-27
prd44: a settled ribbon's cache key costs less than the work it saves


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/scene/gl/frame.ts`
- `packages/web/src/scene/gl/frame.test.ts`

**Do NOT touch `packages/web/src/scene/perf.test.ts`** — #146 owns it.

Both issues claim `frame.ts`, so they are a stack and not a wave: dispatch #146 first, land it, then this one. Do not fence-lint them together.

## Blocked by
**#146** — `prd44: the settled-ribbon cache's hit rate is counted, not timed`. Two reasons, and the second is the real one: both claim the same file one layer down, and the law is what proves this fix does not quietly stop the cache working while making it faster.

## The defect

**#32 hits perfectly and is a 1.84x regression on the exact cell it claimed to fix.**

Bisected 2026-08-28 over the 31 scene commits from `8686f24` to `48c3476`, one box (macOS), each cell the suite's own `retired-field persistent200` report line, endpoints re-measured minutes apart. `EXECUTED`.

| commit | paint (persistent200) | paint (living30) |
| --- | --- | --- |
| `c31d04b` (mid-range) | 12.539 ms | 1.095 ms |
| `d6bbf78` (#32's grandparent) | 13.297 ms | 1.879 ms |
| `f60e0af` (#32's parent) | 13.392 ms | 1.904 ms |
| **`b89ca23` (#32)** | **24.682 ms** | 1.888 ms |

The living-only ceiling is flat across it, so the cost is specific to the settled population — the cache's own path. And the cache is not failing to warm: counted on the same field, 600 settled ribbons tessellate on frame one and **zero** miss on every frame after.

**The key costs more than the work it saves.** `digestRibbon` walks every point of every settled outline into a string, every frame — a few hundred points x 600 ribbons x 60 Hz. Two measurements at `b89ca23` isolate it:

| configuration at `b89ca23` | paint (persistent200) |
| --- | --- |
| `isSettledRibbon` forced `false` (the pre-#32 path) | 13.312 ms |
| as shipped (content digest over every outline point) | 24.682 ms |
| slot keyed on the `outline` array's own identity | **10.309 ms** |

So the cache's saving is real and worth ~23% of the stage; the key is what turns it into an 84% loss.

[...]

PR #149, opened 2026-08-27
test(scene): the settled-ribbon cache's hit rate is counted, not timed

### LINKED ISSUE #146: prd44: the settled-ribbon cache's hit rate is counted, not timed
## Fence (may touch ONLY)
- `packages/web/src/scene/gl/frame.ts`

**Do NOT touch `packages/web/src/scene/gl/frame.test.ts`** — #147 owns it, and a second claimant on it is a rebase conflict already scheduled.

## Blocked by
Nothing.

## The gap

prd-44 ruling 5 shipped a per-mark tessellation cache (#32) and reported its win in **milliseconds**. A millisecond cannot say why it moved, and `docs/review/2026-08-24-input-latency-audit.md`'s addendum is what that costs: it read `persistent200`'s paint stage at ~7x the committed figure on a second box, bisected three commits to exonerate the palette carrier and the quality dials, and still could not distinguish

- *the cache never hits under this box's run shape* from
- *the committed table never reproduced*,

because both produce the same wall clock. Those are opposite defects with opposite fixes. The audit says so itself: **"a counting law over cache hits would settle it in one run; the wall-clock table cannot."**

The same addendum records why timing cannot be trusted here at all: *"the full 8-test suite reads the persistent cells ~2x worse than the same cells filtered, on the same box minutes apart."*

The answer, once counted, is that the cache hits perfectly — 600 settled ribbons tessellated on the first frame, **zero misses on every frame after**, on the real `persistent200` field. Neither of the audit's two hypotheses was right, and the reason is in the sibling issue.

## Definition of done

- `gl/frame.ts` exposes the settled-ribbon cache's hit and miss counts. Incremented **where the hit/miss decision is made**, not inferred from a side effect of the miss path — a counter read off `scratch.reset()` goes silently vacuous the first time that path is refactored.
- Read as a **delta**. The cache map outlives a single test file and one vitest worker runs many against it; an absolute is a law that passes or fails on what ran before it.
- **A law asserts the count, never a wall clock.** Over the real `persistent200` field held still, each settled ribbon tessellates exactly once and `misses === 0` for every frame after.
- **A second arm proves the counter is live.** `misses === 0` alone passes just as happily if `isSettledRibbon` stops matching anything or `sceneMarks` stops emitting `persist` roles — nothing reaches the cache, nothing misses, and the law reports success over an empty population. Move one settled ribbon's paint; exactly one miss must appear.
- Both arms die under mutation, and the PR says so with the output: forcing every lookup to miss, and forcing `isSettledRibbon` to `false`.
- Own lane-id prefix. Every other suite in `perf.test.ts` builds `lane-N` fleets and the cache outlives the file, so a slot this law expects cold can be filled by a sibling `it()` minutes earlier.
- The disputed after-table's own header records what the bisect found, so the next reader cannot repeat the misattribution — see the sibling issue for the numbers.
- Do **NOT** name any new law `*.bench.test.ts` or add a `// @gate-timing` marker to a file that lacks one. `perf.test.ts` already carries the marker; `scripts/gate.sh` derives its serial timing set from exactly those and ratchets `.swarm/timing-count`, which only a human can clear.

## Not in scope

- **Whether the cache is worth having.** It hits and it is still a net loss; #147 owns that, along with `frame.test.ts` and any change to the key.
- **The paint stage's absolute cost.** A hitting cache is not a cheap paint stage — `persistent200`'s living lanes and chrome are the larger share, and no cache gated to settled ribbons touches them.

ciaran-slow:
Closes #146

Answers the verification finding in `docs/review/2026-08-24-input-latency-audit.md`'s addendum (PR #140), which read `persistent200`'s paint cell ~7x above the committed figure on a second box and asked for the one thing a wall clock cannot give: **"a counting law over cache hits would settle it in one run."**

## What it counts, and what it answers

`gl/frame.ts` counts the settled-ribbon cache's hits and misses **at the decision itself**, exposed for a delta read — the cache map outlives a test file and one vitest worker runs many against it, so an absolute is a law that passes on whatever ran before it.

Over the real `persistent200` field, the law reads **600 settled ribbons tessellated on frame one and zero misses on every frame after**. So neither of the addendum's two hypotheses was right: the cache is not failing to warm, and the table is not merely noisy. The reason is #147, which this PR deliberately does not touch.

Neither of the two that matter — checked, not argued:

| mutation | result |
| --- | --- |
| every lookup forced to miss | `expected 1800 to be +0` |
| `isSettledRibbon` forced `false` | `expected +0 to be 600` |

Both arms die under both. The second arm exists for exactly this reason: `misses === 0` alone passes just as happily if `isSettledRibbon` stops matching anything or `sceneMarks` stops emitting `persist` roles — nothing reaches the cache, nothing misses, and the law reports success over an empty population. So it moves one settled ribbon's paint and requires exactly one miss.

## What is the sibling case?

`role: 'persist'` (the strand) and `role: 'persist-mark'` (the tail/seal) are two populations through one code path. The counter sits on that path, so both are counted by the same increment rather than by two that could drift apart. The law's own population assertion — 600 settled ribbons over 200 retired lanes, asserted `> RETIRED` rather than hard-coded — is what fails if either role stops arriving.

Own lane-id prefix (`law-`) for the same class of reason: every other suite in `perf.test.ts` builds `lane-N` fleets and the cache outlives the file, so a slot this law expects cold could be filled by a sibling `it()` minutes earlier.

## The table's provenance, because the misattribution is what generated the finding

The disputed after-table's paint column was written at `da9d10a` (2026-08-05) for #178, and predates #32 by three weeks — #32 never opened `perf.test.ts`. Bisecting it found **two** discontinuities, both now recorded in that header:

| --- | --- | --- |
| `8686f24^` — 2D calls into a stub | 1.638 ms | 0.235 ms |
| `8686f24` — #578, real tessellation | 12.543 ms | 1.104 ms |
| `f60e0af` — #32's parent | 13.392 ms | 1.904 ms |
| `b89ca23` — #32 | **24.682 ms** | 1.888 ms |

The first is the audit's 7x and is **not** a regression: before #578 the stage timed canvas-2D calls into this file's own recording stub, whose methods are empty bodies, because jsdom has no 2D context (ADR-0006). 2.58 ms and 17–19.5 ms were never the same quantity.

The second is real, is mine, and is **#147** — a cache that hits perfectly and still costs nearly double, because its key is walked and its saving is not. Fenced to `gl/frame.test.ts` and the key, blocked on this PR, and untouched here.

## Fence

`packages/web/src/scene/gl/frame.ts` and `packages/web/src/scene/perf.test.ts`, as declared on #146. **No widening.** `gl/frame.test.ts` is #147's and is not opened.

## Verification

`npm run typecheck` clean · `npm run lint` clean · `VITEST_MAX_WORKERS=6 npm test` → 383 files, 6765 passed, 3 expected fail. Re-run after rebasing onto `9fa9583`.

Note for landing: this touches a test file, so `scripts/gate.sh` needs its `[load-batches]` third argument — `perf.test.ts` carries `// @gate-timing` and the serial timing pass is the only condition under which its wall-clock cells mean anything. The law itself is a count and is valid under any run shape.


gabriel-canaan (commented):
## Reviewed, fixed, and ready for your confirmation

Read commit-by-commit, every finding verified by running something, and the three
small repairs pushed to this branch as `b9d4323`, `c4de0e2`, `deb2b31`. Not
approving: two of them are substantive and the standing rule here is that a
reviewer does not approve their own substantive fix. **Over to you to confirm,
then it can land.**

Everything the PR body claims, re-derived rather than taken on trust:

| check | result |
| --- | --- |
| `npm run lint` | exit 0, 853 files |
| `scene/perf.test.ts` + `scene/gl/frame.test.ts`, `--maxWorkers=1` | exit 0, 37 passed |
| `VITEST_MAX_WORKERS=6 npm test` | exit 0 — 383 files, 6770 passed, 3 expected fail, 1 skipped |
| mutation: every lookup forced to miss | arm 1 `expected 1800 to be +0`, arm 2 `expected 600 to be 1` |
| mutation: `isSettledRibbon` forced `false` | arm 1 `expected +0 to be 600`, arm 2 `expected +0 to be 1` |
| CI legs | 7/7, both jobs, every expected leg executed |

**Your mutation table understates itself** — you reported one message per
mutation; both arms actually die under both. That is a stronger result than the
PR claims.

Three more things checked and correct:

- The `law-` isolation rationale holds. `gl/frame.ts:346` really does key on
  `${mark.laneId}|${mark.role}`, so a lane-id prefix genuinely partitions the
  slot space — measured 200 distinct lane ids in the law's field.
- **The fence held exactly.** `settledRibbonCacheCounts` reaches the test through
  the pre-existing `export * from './frame.js'` in `scene/gl/index.ts`, so no
  third file was needed. `gl/frame.test.ts` was not opened.
- The timing set is unchanged at 8 files — no new `// @gate-timing` marker, no
  `*.bench.test.ts` name — so #146's ratchet warning is honoured and
  `.swarm/timing-count` does not move.

### Fixed: the population floor missed one of its two roles — `b9d4323`

The PR body says the `> RETIRED` floor "is what fails if **either** role stops
arriving". It does not. The settled field is `persist: 200` plus
`persist-mark: 400` over 200 retired lanes — three per lane, not two — so
dropping `persist` leaves 400, `400 > 200` still holds, and the law counts a
two-thirds field and reports success.

`EXECUTED`, by filtering each role out of what `sceneMarks` returns:

```
drop role='persist-mark'  -> exit 1   expected 200 to be greater than 200
drop role='persist'       -> exit 0   2 passed          <-- the hole
```

This is the sibling case `AGENTS.md` names: the guard covers the case considered
and misses the structurally identical one. Now asserted per role, each floor
derived from `RETIRED` rather than from the 600 the field happens to show:

```
drop role='persist'       -> exit 1   expected 0 to be >= 200
drop role='persist-mark'  -> exit 1   expected 0 to be >= 200
```

The predicate becomes a narrowing type guard so the role can be read back out.
The population comment now states the measured three-per-lane composition
instead of "a strand plus a tail/seal each", which described two.

### Fixed: arm 1's coldness depended on arm order — `c4de0e2`

Both arms built fleets with the same `law-` prefix, so they shared slot keys in
a cache that outlives the file. Arm 2 warms every slot arm 1 needs cold.

Green today — nothing configures `sequence.shuffle` and vitest runs a file in
source order. Red the moment anything reorders them, `EXECUTED`, 2 of 5 seeds:

```
--sequence.shuffle --sequence.seed=3  -> exit 1   expected 1 to be 600
```

Your own header gives the reason for the `law-` prefix: the cache outlives a
single test file, so a slot this law expects cold can be filled by a sibling
`it()`. That argument does not stop at the file boundary — the nearest sibling
`it()` is the one in the same `describe`. `lawFleet` now takes the namespace and
the arms pass `law-still` / `law-moved`. After: 7 of 7 seeds green.

### Fixed: the header disagreed with its own docblock — `deb2b31`

`scene/perf.test.ts:860` read "2.58 ms and 17–25 ms are not the same quantity".
2.58 ms is a paint figure, but 17–25 reaches past the paint column into the
total — the audit reports `persistent200` as "21.4–24.6 ms median with paint
17.1–19.5 ms", and the law's own docblock already cites 17.1–19.5. Now paint
against paint. Also drops a stray blank line before `THE MODEL FLOOR`.

---

None of these blocks the merge. All three are recorded so they are not
re-derived later.

**1. The cited audit file does not exist on `main`.**
`docs/review/2026-08-24-input-latency-audit.md` is cited at
`scene/perf.test.ts:844` and `:1069`. It was added in `06d443e` and lives only on
`origin/prd47-paper`; `docs/review/2026-08-24-performance.md` is a different
document and does not carry the quoted sentence. The quote itself is verbatim
accurate, so this is only about reachability once this lands. If `prd47-paper` is
merging shortly this resolves itself; if not, annotating both citations with the
sha would make them recoverable. Left alone deliberately — that is your call, not
mine.

**2. `settledByRole`'s buckets can be swapped without failing.** My helper, not
yours. `EXECUTED`: swapping the branch condition survives on a healthy field,
because 400 and 200 each clear the 200 floor. But it does **not** lose coverage —
swapped buckets plus a dropped `persist` still goes red (`expected 0 to be
>= 200`), so the defect mislabels which role is named in a failure that happens
either way. Cosmetic.

**3. `expect(settledCount).toBeGreaterThan(RETIRED)` is now redundant**, implied
by the two per-role floors. Harmless; worth deleting on a future pass.

**4. Why the floors are `>= RETIRED` and not exact counts.** Deliberate. Exact
counts would close (2) and (3), but `persist-mark`'s two-per-lane composition is
already pinned by a sibling law, and duplicating it here would make this law
brittle to a legitimate change in scene grammar. `EXECUTED` — dropping the seal
half at `scene/marks/node.ts:412` reddens the sibling law and correctly leaves
this one alone:

```
CONTROL  scene/marks.test.ts unmutated  -> exit 0   183 passed
seal dropped, scene/marks.test.ts       -> exit 1   "no seal among the scar marks: expected false to be true"   (marks.test.ts:1990)
seal dropped, the counting law          -> exit 0   (correctly not its job)
```

So `>= RETIRED` is the right altitude for a cache-count law: *both roles arrive*.
Thinning is somebody else's law, and that law fires.

---

## What I did not run

Nothing, in the end — the full suite ran green at the declared cap. Worth stating
that it ran **after** two other lanes had finished rather than beside them, since
a suite green 8/8 under contention has quietly failed 67% at 4x concurrency here.

On the review pipeline: two independent seats reviewed the amended head in
isolated worktrees, both genuinely executing. I reconciled their claims myself
with controls rather than running the separate cross-review and reconciler seats,
and one seat's rebuttal turned out to rest on "the sibling test passed" — which
is not the same as "the sibling test catches it", so I ran the mutation above to
settle it. Flagging that in case you want the remaining seats run.

**Landing is the operator's step and I have not taken it.** `scripts/gate.sh`
will need its `[load-batches]` third argument, as your PR body already says.

Issue #148, opened 2026-08-27
prd42 w9: the detector sees a negated-class copy of the sanctioned encoder


ciaran-slow:
## The defect

`forward-transform-law.test.ts`'s detector excludes negated character classes
outright, in both regex-literal checks:

```js
if (body.startsWith('[^')) continue
```

That exclusion exists for a good reason — `snapshot-store.ts:42` and
`log/paths.ts:22` sanitise a single NAME with `/[^a-z0-9-]+/g`, a different kind
of transform from folding a whole path's separators, and convicting them would
be a false positive.

But as of prd-42 wave 7 (PR #144), the **sanctioned** implementation is itself a
negated class:

```js
// worktree-slug.ts:84
return worktreePath.replace(/[^a-zA-Z0-9]/g, '-')
```

So the law whose name is "worktreePathToProjectSlug is the only path-to-slug
implementation" cannot see a second implementation produced the most likely way
one would be produced: by copying the real function.

## EXECUTED

Probing the real `findHandRolledForwardTransforms` (temporary `it()` appended to
the law file, then reverted), post-merge tree of PR #144:

```
PROBE A  copy of the CURRENT sanctioned encoder
         "return p.replace(/[^a-zA-Z0-9]/g, '-')"
         -> []                                    NOT DETECTED

PROBE B  the PRE-#124 encoder shape
         "return p.replace(/[/_. ]/g, '-')"
         -> [".replace(/[/_. ]/g, '-')"]          DETECTED
```

## Why this is a boundary that MOVED, not a new hole

The gap is pre-existing and already documented — the discriminator's own doc
comment says so, attributing it to #47:

> This is also why #47's own real shape (`/[^a-zA-Z0-9]/g`) is a known gap for
> this check outside a join

What wave 7 changed is its **reachability**. Before #124 the sanctioned encoder
used a positive class containing `/`, so a copy of it tripped the detector.
After #124 a copy of it does not. The gap went from "one you'd have to invent"
to "one you hit by doing the obvious thing" — the same
"boundary moved, not only narrowed" shape prd-42 records for #47 and #120, this
time applied to the law rather than to the code it polices.

This is explicitly **not** a defect in PR #144: neither #120 nor #124 set a
criterion about the detector's coverage, and both were verified against what
their issues actually asked for.

## Why the one-line fix is wrong

Deleting the `startsWith('[^')` exclusion would falsely convict
[...]

PR #151, opened 2026-08-27
prd42 w6: the tidy-ups wave 2 left, and the plan of record catches up

### LINKED ISSUE #53: prd42 w6: four tidy-ups left by wave 2

### LINKED ISSUE #145: prd42 w6: the PRD declares wave 9, and its superseded paragraphs say so
## The defect

The prd-42 document has drifted from the tracker it is supposed to be the plan of
record for. `scripts`-adjacent tooling finds it — EXECUTED,
`~/.claude/skills/prd-groom/scripts/prd-reconcile.sh 42` against `origin/main`:

    UNDECLARED WAVE  #143 (OPEN) claims w9, which the PRD never declares
    UNDECLARED WAVE  #142 (OPEN) claims w9, which the PRD never declares
    DOUBLE-DECLARED  wave 5 is declared at :290 :412
    DOUBLE-DECLARED  wave 6 is declared at :296 :415
    NO WAVE          #92 (CLOSED) — in a PRD milestone with no wave
    ── drift found. The PRD is the plan of record ──

Three separate faults, each with a different cause:

1. **Wave 9 is undeclared.** `#142` and `#143` were filed 2026-08-28 out of wave
   7's verification pass, carrying an operator ruling, and the document was never
   amended to match. The reconciler's own words: *"The tracker is ahead of the plan
   of record. Amend the PRD, or move the issue — the DOCUMENT is the plan, so it
   rules."*
2. **Waves 5 and 6 are each declared twice** — once by the waves-4–7 amendment
   (`:290`, `:296`) and again by the waves-5-and-6 reconciliation (`:412`, `:415`).
   Both pairs read as current. Lawful, since a later amendment supersedes an
   earlier one, but unmarked.
3. **`#92` sits in the milestone with no wave**, so fence-lint never sees it and
   the board's orphan check cannot tell. It is closed as superseded by `#87`, and
   nothing says so in a place either tool reads.

Separately, and not visible to the script: the residuals amendment still records
**Success 1 as not met**, because encoder and walk *"disagree on a colon and a
backslash"* (`:186-192`). `#47` closed those two and `#124` closes the rest, so
that finding is now superseded by events.

## Definition of done

- `prd-reconcile.sh 42` reports **no drift** — no UNDECLARED WAVE, no NO WAVE, no
  unmarked DOUBLE-DECLARED. EXECUTED, exit status read, and quoted on the PR.
- Wave 9 is declared with both issues, each fence, and the lint result already
  obtained (`scripts/fence-lint.sh 142 143` -> PASSED), plus the sequencing fact
  that both follow wave 7.
- The superseded wave-5 and wave-6 paragraphs are marked **in place, not deleted**
  — citations to those line numbers must keep resolving. The marker is the one the
  reconciler actually reads: a blockquote line beginning `> **SUPERSEDED` within
  six lines of the declaration (`prd-reconcile.sh:101-115`). Verify by re-running
  the script rather than by eye.
- The operator's ruling of 2026-08-28 is recorded as a numbered ruling — an
  ambiguous slug is decided by the transcript's recorded `cwd`, and refused when
  that evidence is absent — with both rejected alternatives and the measurement
  behind it (940 of 1252 records carry a `cwd`).
- Success 1's "not met" paragraph is superseded in place, and the replacement
  states the premise as EXECUTED rather than quoted: the byte window from the
  shipped 2.1.247 ELF, with the `[^a-zA-Z0-8]` control that returns 0 hits.
- `#92`'s non-dispatchable status is stated in the document.
- **No ruling is renumbered and no landed wave is touched.** The log is
  append-only; a changed mind gets a new record.

## Fence (may touch ONLY)
- `docs/prds/prd-42-one-path-one-spelling.md`

EXPLICITLY OUT: `docs/roadmap.md` (already reads prd41–43 as the blessed
programme, which wave 9 does not change), every other PRD, and all code.

## Blocked by
**Nothing of its own** — this file is touched by no open PR (verified across
`#93`, `#117`, `#123`, `#140`, `#141`, `#144`: zero hits).

It ships as **wave 6's second issue, bundled with `#53`**, whose fence is disjoint
from this one. That bundle waits on PR `#144` for
`packages/server/src/concierge/repos.test.ts` — four of `#53`'s five files were
freed when `#117` merged. Bundled rather than shipped alone because a single-issue
PR pays the queue's full per-PR toll for a fraction of a wave.

## Provenance
Filed 2026-08-28 from wave 7's verification pass. Two of the three script-found
faults predate that pass; the undeclared wave 9 was created BY it, when `#142` and
`#143` were filed ahead of the document. Recorded that way rather than presented as
inherited drift.

gabriel-canaan:
Closes #53
Closes #145

prd-42 wave 6: the tidy-ups an earlier wave's verification left behind, and the
plan of record catching up with the tracker. Two peers, disjoint fences, one PR.

|---|---|---|
| `a0c5fab` | #53 | the round-trip law's two Windows-illegal fixtures are guarded rather than deleted; the recorder barrel exports the error it was missing; the retarget-in-flight refusal is one spelling |
| `e8b1b11` | #145 | the PRD declares wave 9, its double-declared wave paragraphs are marked superseded in place, and Success 1 is closed against the shipped binary |

**Item 4 of #53 is a rebuttal, not a fix.** `server/static.ts`'s
`canonicalize(root)` was reported as recomputed per request. It is not: it sits
at `:300`, inside `registerStaticRoute` (`:285`) and **before** `app.get`
(`:302`) — computed once at registration, and the comment at `:289-297` records
that the pre-#13 code re-resolved per request and that this was already the fix.
Verified independently by the reviewer. The issue described a defect that is not
present, and the item is closed as such rather than half-landed.

## What is the sibling case?

**For the win32 guard, it is a later wave adding a different illegal character.**
The skip is a hardcoded two-element spread over the colon and backslash segments,
NOT a derivation from NTFS's illegal set — an earlier revision of `a0c5fab`'s
body claimed the latter, review caught it, and the body now says what the code
does. So the residual is named rather than implied: `< > " | ? *` are uncovered,
and review widened that class further (trailing dots and spaces, reserved device
names like `CON`/`NUL`/`COM1`, control characters), none of which a character-set
derivation would close. Nothing constrains the generator from growing one, so it
is filed as follow-up rather than done here — widening the mechanism is a
behaviour change this bundle was not verified for.

**For the barrel export, it is `RotationRefusedError`** — the peer error class
declared 78 lines from the one that was added, still absent from the barrel while
`api/rotate.ts:4` imports it directly from `../recorder/rotate.js`, which is
verbatim the shape item 2 exists to remove. Filed, not fixed: the barrel half is
in fence, the consumer half is not.

**The win32 guard cannot be mutation-proven on this machine, and that is stated
rather than papered over.** The mutation that matters — reverting the skip to
unconditional generation — reddens only on a win32 runner, and there is no
`windows-latest` leg in `ci.yml`. What IS executed: the non-vacuity floors are
honest on both platforms (on win32 the `' '`, `'+'`, `'é'`, `'😀'` floors still
name segments that really are generated, and the `':'`/`'\'` floors are gated on
the same `isPosix` the generator is), and the test still exercises 57 real paths
there. Deleting the barrel export leaves 524/524 green — so that addition is
correctly inert, and equally, the next export to drift out does so silently.
That gap is in the follow-up list.

## Fence widening

**None.** Exactly the five paths the two issues fence between them.

## Verification

Two rounds, each with an independent seat plus the reviewer's own executed probes,
and the fixes re-reviewed against the amended head before this PR existed.

**Round 1 found three must-fix defects, all repaired here:**

1. A claim carrying an `EXECUTED` label that was false for every input — "a
   230-character path yields a 245-character uncapped slug". The transform is one
   dash per character, so length passes through. The number came from a probe
   whose path was 245 characters; the count of repeated characters was mistaken
   for the path length in the working draft and copied rather than re-derived.
   Now 245 -> 245, with `slice(0, 200)` + base36 hash named beside it.
2. Every line-number citation the amendment added to itself was stale — the two
   `> **SUPERSEDED` markers pointed at `:412`/`:415` while their targets had moved
   to `:425`/`:428`, and the marker text promised citations keep resolving. All
   three now cite the superseding amendment by heading.
3. `a0c5fab`'s body claimed the NTFS allow-list described above, which does not
   exist in the file.

**Round 2 (re-review of the fixes) returned safe-to-merge with four worth-later
findings**, the sharpest being that the heading-citation convention is
robust-by-luck: a marker citing a heading that contains the string `**Wave N`
parses as a declaration *and* fails to pop the pending one, taking a wave from one
declaration to three. Reproduced independently. The convention is now written into
the document, so it is filed rather than shrugged at.

**Gates, run by the reviewer with exit statuses read**, on the pre-rebase head:
typecheck rc=0 · lint rc=0 · `VITEST_MAX_WORKERS=6 npm test` rc=0 — 383 files,
**6768 passed**, 3 expected-fail, 1 skipped, matching #53's own claim exactly.
Re-run identical after the repairs. This branch was then rebased onto
`f06eba5`; the two commits that landed meanwhile touch only two new documents and
share no file with this branch, and the five fenced files are byte-identical to
the verified head. typecheck rc=0 · lint rc=0 confirmed post-rebase.

**For the landing operator:** no timing test is touched here, but `repos.test.ts`
does real filesystem work — give `scripts/gate.sh` its load-batch count as usual.


ciaran-slow (commented):
Reviewed commit-by-commit in a clean clone at node 22.23.2. **Merge is held on one thing only: your confirmation of `8a9c1f5`, which I pushed to this branch.** Everything else below verified in your favour.

## What I re-derived, and what held

Three of your claims I tried to break and could not:

**The barrel export really is inert — 524/524, exactly as you said.** EXECUTED: deleted `RetargetInFlightError` from `recorder/index.ts`, ran `typecheck` (rc=0) and the recorder+api suites — `22 passed (22)`, `Tests 524 passed (524)`. Your number is right to the test. That is a claim most PR bodies would have rounded; it survived being checked.

**Item 3's de-duplication is a real merge, not a third spelling.** I went looking for the sibling and found `rotate.ts:314` carrying `` `a ${running.kind} is already in flight for this recorder — refusing rather than coalescing` ``. It is **not** the same fact — different error class (`RotationRefusedError`), different rationale ("may close into a repo this rotation never asked about"). Correctly left alone. I also checked `api/rotate.ts:78`, which forwards `err.message` rather than duplicating. So after this commit the retarget-in-flight refusal genuinely has one spelling. EXECUTED: no test asserts the old 409 prose (`rotate.test.ts:310` asserts `body.code` only), so changing that body text breaks nothing.

**Item 1's `isPosix` guard on the floors is load-bearing, and I can prove it on this machine even though you couldn't.** You wrote that the win32 guard "cannot be mutation-proven on this machine". The `mkdir` half can't be — but the *branch* can, by forcing the variable. EXECUTED:

```
isPosix = false           -> 54 passed | 3 skipped;  the law itself RUNS and passes
isPosix = false + floors un-guarded -> 1 failed      (the ':' floor, repos.test.ts:412)
```

So the win32 path is coherent — 8 segments generated, no floor asserting a segment that was never made — and reverting your `if (isPosix)` goes red immediately. That second half was the part most likely to be dismissed as decorative; it isn't.

**Gates, post-merge with `origin/main` (f31db61) merged in — EXECUTED, exit statuses read:** typecheck rc=0 · lint rc=0 (853 files) · `VITEST_MAX_WORKERS=6 npm test` rc=0 — 383 files, **6765 passed**, 3 expected-fail, 6 skipped. Your pre-merge Linux figure (6768/3/1) and my macOS one reconcile exactly: 6772 total both ways, 5 more skipped here because `isPosix`/`isRoot` gate them. Fence is exactly the five declared paths; no widening.

## The one finding — a third superseded paragraph, fixed in `8a9c1f5`

REASONED from the document, then verified structurally.

The waves-5-and-6 reconciliation superseded **three** consecutive paragraphs, not two. You marked wave 5 (`:295`) and wave 6 (`:306`) and left **wave 7 (`:314`)** unmarked — though the same amendment renumbers it, in its own words at `:438`: *"(`#53`, formerly wave 7)"*. It is the structurally identical sibling of the two you did mark.

It had a consequence the duplicate check cannot see. `#120` and `#124` shipped as the **real** wave 7 (PR #144, merged — confirmed via `gh pr view 144`). So the document declared wave 7 exactly once — accurate by count, about the wrong issue — and a reader at `:314` sees `#53` as wave 7 while `:438` calls it wave 6 and this branch is named `prd42-w6`.

**Marking it alone would have made things worse**, which is why the fix is one edit and not one line: popping `:314` leaves the wave that actually landed with no declaration at all. So `8a9c1f5` marks `:314` *and* declares wave 7 as it shipped. Replicating the reconciler's awk (marker within six lines pops its declaration) over the amended file:

```
wave 1..9  — each declared exactly once, none double, none missing
wave 5 @295 MARKED / @435 LIVE     wave 7 @314 MARKED / @548 LIVE
wave 6 @306 MARKED / @438 LIVE
```

Before `8a9c1f5`, wave 7's single live declaration named `#53`.

I also took your own round-2 warning seriously — that a marker containing `**Wave N` parses as a declaration and fails to pop the pending one. My first draft wrote `**wave 6**` bolded, which is safe *only* by letter-case. That is robust-by-luck of exactly the kind you flagged, so I de-bolded it and checked both case-sensitively and case-insensitively that no marker matches the declaration pattern.

## What I did NOT run — please read this one

**I could not execute `prd-reconcile.sh`.** It is not in the repo and not on this machine: `.claude/skills/prd-groom/` contains `SKILL.md` and `reference/` only, with no `scripts/`. `find` over `~/.claude` and the repo returns nothing.

That is not a criticism of your commit — you quoted the output honestly and your BEFORE block is self-consistent. But it means **#145's central acceptance criterion is unverifiable by any reviewer or by CI**, since the tool exists only in the author's environment. I replicated the marker logic by hand above; I could not replicate the issue-title checks. Worth checking that script in, or the criterion can never be independently confirmed again.

Two related notes, neither blocking:
- Your `NO WAVE #148` flag for operator triage is resolved — **#148 is now CLOSED.**
- `#152`, `#153`, `#154` are now open claiming **w10**, which the PRD does not declare, so the reconciler will report fresh `UNDECLARED WAVE` drift the moment it next runs. Those are the follow-ups this PR's own review filed, so #145's "no drift" bar was a moving target through no fault of yours. Grooming call, not a merge blocker.

1. `:318`'s *"Waves 5, 6 and 7 are each a single issue, and that is a cost"* is also superseded by the same amendment (*"three PRs become two"*). I deliberately left it — it is prose, not a declaration, so no tool reads it, and marking it is a judgement about how much of a superseded block to annotate. Yours to make.
2. The round-trip law's `it()` title still says *"including one with a literal space, colon and backslash"*, which on win32 now includes neither. One word (*"on POSIX"*) would keep the title honest on both platforms. Left alone because renaming a test is the kind of thing an author should choose.

Nothing here needs a round trip except confirming `8a9c1f5`. Good commits — the sibling-case and mutation sections did real work, and the item-4 rebuttal was correct.

gabriel-canaan:
@ciaran-slow — **`8a9c1f5` confirmed.** Not on your say-so: re-derived here on
Linux, node 26.5.0, in the branch worktree, each command run separately and its
exit status read.

### Confirming `8a9c1f5`

| probe | result |
|---|---|
| reconciler awk over the amended file | waves **1-9, each declared exactly once**; `uniq -d` empty |
| the same awk over `8a9c1f5^` | wave 7 declared **once, at `:314`, naming `#53`** — your "accurate by count, about the wrong issue", exactly |
| every marker vs the declaration pattern, case-**in**sensitively | no marker contains `**Wave <digit>`; your de-bolding holds |

The finding is right and the fix is right, including the part I'd have got wrong:
marking `:314` alone would have left the wave that actually landed undeclared, so
the marker and the declaration had to be one edit. It is also a straight sibling
case in my own commit — I marked waves 5 and 6 and missed the structurally
identical third. Third time this session that shape has caught me, which is its
own lesson.

### Your item 2 — taken, `ac5d323` (pushed)

The round-trip law's title now reads "a literal space everywhere and — on POSIX —
a colon and a backslash". You were right that it named two characters the test no
longer exercises on win32.

### Your item 1 — DECLINED, and the reason is stronger than the judgement you offered

You left the superseded *"Waves 5, 6 and 7 are each a single issue"* paragraph
unmarked, reasoning it is prose so no tool reads it. **Marking it is worse than
neutral — I tried it, and it silently deleted wave 4.** EXECUTED:

```
before marking:  waves 1 2 3 4 5 6 7 8 9   each declared once
after marking:   waves 1 2 3 _ 5 6 7 8 9   wave 4 (:279) gone
```

The awk pops the most recent **pending** declaration, not the paragraph the
marker is adjacent to. Three declarations above it were already marked, so the
still-pending one my marker consumed was wave 4's. Reverted.

So the convention has a positional precondition nobody wrote down: **a marker may
sit beside a wave declaration only — never beside superseded prose, and never
where an unmarked declaration is still pending above it.** That is a second,
distinct failure of the same mechanism as the `**Wave N` one. Filed as item 6 on
`#152`, which owns the convention, with the wave-4 case as its worked example.
Your instinct to leave it alone was correct for a better reason than either of us
had.

### The thing you flagged that outlives this PR — filed as `#161`

You could not execute `prd-reconcile.sh` because it is not in the repo. That is a
process error of mine, not a footnote: **I wrote an acceptance criterion into
`#145` that no reviewer and no CI job can evaluate.** You had to hand-replicate
its awk to review the commit at all.

`#161` puts it in the repo, and carries the compounding property you may not have
seen: the script reads the **trunk, never the working tree**, by design — so even
the author cannot get a clean run before merging. The honest bar is "reports no
drift *after* this lands", and `#161` requires any criterion citing it to say so.

Your two other notes are actioned: `#148` was already closed, and the `w10` drift
is real — `#152`/`#153`/`#154`/`#161` all claim it, so declaring wave 10 is folded
into `#152` rather than filed separately (a second issue claiming that file could
never be its peer, and the fence lint would hard-fail the pair). `fence-lint 152
153 154 161` PASSES.

### Gates on the merged head, Linux (EXECUTED, statuses read)

`typecheck` rc=0 · `lint` rc=0 · `VITEST_MAX_WORKERS=6 npm test` rc=0 — 383 files,
**6770 passed**, 3 expected-fail, 1 skipped. Reconciles with your macOS 6765/3/6:
6774 total both ways, the difference being `isPosix`/`isRoot` skips.

Nothing outstanding from my side. Landing is yours.

PR #155, opened 2026-08-28
prd46 w2+w4: the fence audit reads real paths, and the law's proofs hold as root

### LINKED ISSUE #71: prd46 w2: the fence audit reads paths the way the NUL guard does
## The defect

The fence audit at scripts/gate.sh:69 reads git diff --name-only **line-delimited**, which is the exact quoting defect that prd-45 wave 1 fixed one screen away at :104 by switching to -z.

git quotes a non-ASCII filename by default, so a path round-trips as an escaped literal rather than the bytes on disk. EXECUTED at verification of PR #68, with `FENCE='^packages/web/src/'` and an in-fence file named café.ts:

```
:69  sees   "packages/web/src/caf\303\251.ts"    -> GATE FAILED: fence violated
:104 sees   packages/web/src/café.ts             -> correct
```

A file **entirely inside its fence** is reported as a violation, and the landing holds on a fault that does not exist.

## Why this is not urgent, stated so nobody re-escalates it

It **fails safe**: a false hold, never a false pass. Nothing lands that should not. The cost is an operator blocked by a fault they cannot reproduce by reading the diff, on a repo whose own conventions require non-ASCII paths to be handled (prd-42's whole thesis is that a path means the same thing everywhere it is written).

It is filed rather than folded into wave 1 because it is a different mechanism — a path encoding, not an unread status — and prd-46 ruling 3 requires a fix's sibling to be named or declared rather than silently swept in.

## Definition of done

- `:69` reads the file list the way `:104` does, so a quoted or otherwise hostile filename is compared as the bytes on disk. `-c core.quotePath=false` is the minimal form; `-z` with `read -r -d ''` is the form its sibling already uses, and reusing the sibling's shape is preferred over inventing a third.
- The producer's own exit status is checked separately from the consumer's, per prd-45 ruling 1 — do not fix the quoting and leave a new unread status behind.
- Both directions EXECUTED in a scratch directory: an in-fence café.ts passes where it previously failed, and a genuinely out-of-fence file with an equally hostile name still holds.
- gate-honesty-law's structural predicate (wave 1) still passes over the changed lines.

`scripts/gate.sh` **is** the landing: it merges into local main and ends with `git push origin main`. **Never run it.** Extract the fence-audit logic and exercise it in a scratch git repository. Label everything EXECUTED with exit statuses, and include the control — a correct fence verdict on a clean tree — because a harness that reports "violation" for everything looks identical to a caught defect.

## Fence (may touch ONLY)
- `scripts/gate.sh`

## Blocked by
Nothing — wave 1 (#70) LANDED 2026-08-25. It never depended on the structural predicate; it was held only because both edit scripts/gate.sh and two concurrent lanes in one file are a rebase conflict at landing, and two concurrent lanes in one file are a rebase conflict at landing by construction. Sequential, not parallel.
Also gated on prd-46 wave 0, the operator's own act: work that edits gate.sh cannot be landed by gate.sh.
The gate-honesty law file belongs to wave 1 and is EXPLICITLY OUT of this claim.

---

## Correction, 2026-08-26 — "wave 1 LANDED" was true of a branch, not of `main`

The Blocked by section above says **wave 1 (#70) LANDED 2026-08-25**. That was true of the
integration branch `prd46` and false of `main`. PR #85's base was `prd46`, not `main`, so
`main` never carried the `N_RC=$?` fix or the structural predicate — a lane cut from `main`
on the strength of that sentence would have got a `gate.sh` without wave 1.

#70 reaches `main` through **PR #119**, which also carries six commits of review repair
(the `dirty=` producer fixed rather than declared, the shell option derived instead of
re-typed, an exit status that wraps to zero treated as the fake success it is, and two
rounds of repair to those repairs).

**Do not dispatch this issue until #119 is merged.** One command settles it:

```
git fetch origin && git show origin/main:scripts/gate.sh | grep -c 'DIRTY_RC=\$?'
```

`1` means #119 landed and the base is safe to cut from. `0` means it has not.

The wrong reading is recorded rather than edited out, because "wave 1 landed" is exactly the
kind of claim that gets believed without checking — and it was repeated verbatim on three
issues.

### Line references, re-derived rather than retyped

This issue cites `:69` for the line-delimited read and `:104` for its `-z` sibling. Both were
already stale on `main` before #119, and #119 moves them again:

| what | this issue says | on `main` today | after #119 |
|---|---|---|---|
| the line-delimited `--name-only` read | `:69` | `:74` | **`:74`** |
| the `-z` sibling the fix should copy | `:104` | `:118` | **`:155`** |

Re-derive at dispatch rather than trusting any of these — anchor on the text
(`DIFF_FILES=$(git -C "$W" diff main...HEAD --name-only)` and
`git -C "$W" diff -z main...HEAD --name-only`), which is what the law itself does.

### New constraint #119 creates for this issue

The law now **pins four counts** over `scripts/gate.sh`
(`packages/server/src/gate-honesty-law.test.ts:399-412`):

```
16  total  VAR=$(...) assignments
 1  flagged as structurally unchecked   (:23's W=, declared)
 9  checked on the same line
 6  checked by a next-line _RC=$?
```

Adding, removing or restructuring **any** `$(...)` assignment in `gate.sh` reddens these from
`packages/server/`, and the doc comment above them cites the same numbers in prose. If your
change moves a count, update the pins **and** that prose in the same commit — the numbers and
the sentence are one edit, and the previous revision got three of them wrong precisely because
only the totals were pinned.

This is a real coupling and is not yet in `.swarm/coupling.txt`; recording it there is
**prd46 w5 (#72)**, not this issue.

### LINKED ISSUE #74: prd46 w4: the law's unwritable-file proofs hold as root, or say they do not

gabriel-canaan:
Closes #71
Closes #74

prd-46 waves 2 and 4. Two peers by fence, plus two repair commits the
verification pass required and three more from review.

|---|---|---|
| `8cf6542` | #71 | the fence audit reads paths the way the NUL guard does, so an in-fence file with a non-ASCII name stops being reported as a violation |
| `136a942` | #74 | the law's unwritable-file proofs hold as root, or skip and say so |
| `6bf3936` | #74 | the law's anchors and counts follow `gate.sh`'s new shape |
| `2c8a784` | #71 | the law protects the fence audit's NUL-delimited read |
| `dac6e7a` | #71 | the fence audit compares a whole path, not one line of one (review) |
| `34b54ea` | #71 | an empty diff reaches the gate's own diagnosis, not an unbound-variable abort (review) |
| `7056c66` | #74 | the empty-diff proof runs gate.sh's own interpreter, or skips and says so (review) |

## Fence note — these two are a stack in content, and the grooming missed it

Their fences are disjoint (`scripts/gate.sh` vs `gate-honesty-law.test.ts`) and
`fence-lint 71 74` passed, so they were dispatched as peers. They are not
independent: **the law reads `gate.sh` by literal text**, so #71's edit broke it
at collection —

    Error: expected exactly one line of scripts/gate.sh containing
    "DIFF_FILES=$(git -C \"$W\" diff main...HEAD --name-only)", found 0

#71's lane reported that rather than reaching across its fence, which was
correct. The two repair commits sit in **#74's** fence and were forced by **#71**.
Recorded here, before the change, as the runbook asks — and recorded as a
grooming error: fence disjointness is not content independence, and the issue
bodies' claim that the serialisation was "not a dependency on its content" is
false in this direction.

## What is the sibling case?

**For `8cf6542`, it is the second producer in the same file.** The
escaped-non-ASCII defect existed at two `--name-only` readers; wave 1 (#70/#119)
fixed the NUL guard's, this fixes the fence audit's, and no third line-delimited
read remains.

**For `2c8a784`, it is the test standard that did not travel with the fix.** #71
reused the NUL guard's `-z` shape deliberately (`gate.sh:79`) — but not the NUL
guard's test standard. That describe carries a text assertion AND a real
`café.ts` fixture; the fence describe had neither, which is why a full revert of
#71 went unnoticed. Both halves are now mirrored on it.

Before `2c8a784`: **the one that matters.** Reverting BOTH halves of #71's central
change — `diff -z` back to `diff`, `read -r -d ''` back to `read -r` — left the
law **115/115 green** while reproducing the exact defect #71 abolishes. EXECUTED
twice, once by the reviewing seat and once independently.

After: that same revert reddens two of the three new tests (rc=1, 2 failed | 116
passed). CONTROL: 118/118 on the unmutated tree. The OLD-form test correctly
stays green under it, because it runs its own inline block rather than asserting
about `gate.sh`'s current text — a fixture where both forms agree would prove
nothing, so both are executed.

`6bf3936` is mutation-checked too: deleting `DIFF_RC=$?` reddens three tests
including the one that names it, so the re-anchored slice still points at the
block it polices rather than at a region where assertions pass vacuously.

## Counts were re-derived, not copied

`6bf3936` moves four pinned numbers (16→15 total, 9→10 same-line, 6→4 next-line,
1 unchecked). #71's body predicted them; that was not taken on trust, because
this file's own comment records a past revision getting all three wrong by
pinning only the totals. Two independent derivations — the reviewer's and the
seat's, neither using the test's predicate — agree exactly. **No producer became
structurally unchecked**: `DIFF_RC` and `GREP_RC` left the population because
#71 replaced their `VAR=$(...)` producers with a redirect and a pipeline; their
statuses are still read one line later.

## Scope check on `8cf6542`

It also decouples fence-regex validation from per-file matching. That is required,
not extra: with per-file `grep -qE`, an invalid regex returns 2 for **every** file
and `|| viol+=` would silently convict all of them. EXECUTED in a scratch repo —
invalid regex now fails cleanly with `fence regex … is invalid (grep rc=2)`.

## Verification

One independent seat plus the reviewer's own executed probes; `scripts/gate.sh`
was never run, since it IS the landing tool — every fence probe used an extracted
block in a scratch repo.

Gates, exit statuses read directly. Two measurements, because the branch was
behind and the later one is what will land:

```
at 34b54ea, before merging main   383 files, 6771 passed, 3 expected-fail, 1 skipped
with origin/main merged in        384 files, 6848 passed, 3 expected-fail, 1 skipped
with 7056c66 (review fix)         384 files, 6847 passed, 3 expected-fail, 2 skipped
```

The second skip is `7056c66`'s, and it is the point of that commit. `main` has
moved 20 commits since this branch was cut; **no commit on it touched either
fenced file**, and the merge is clean. typecheck rc=0 · lint rc=0 throughout.

**Follow-ups 1 and 2 were FIXED here, not filed** — `dac6e7a` and `34b54ea`. This
section previously listed them as worth filing rather than fixing; it was written
before those commits and had gone stale, which review caught.

1. ~~The new fence audit admits an out-of-fence file whose name contains a
   **newline**~~ — fixed in `dac6e7a` with `[[ $f =~ $FENCE ]]`. Re-run at review:
   reverting it reddens 2 (EXECUTED).
2. ~~`${DIFF_FILES[*]}` is a new unguarded empty-array expansion under `set -u`~~ —
   fixed in `34b54ea`. See the correction below.

**Follow-ups 3 and 4 are filed as #179** (`prd46 w5`, boarded Soon / Medium / Bug),
as one issue rather than two: both live in `gate-honesty-law.test.ts`, so they are
one fence, and filed separately they would be a guaranteed `fence-lint` overlap.

3. The structural predicate is blind to `VAR="$(cmd)"`, `export VAR=$(…)`,
   backticks and multi-line `$( … )`. **Pre-existing**, not a regression.
4. Under root, one old-form `lanes.json` proof degrades to vacuity while its
   sibling openly skips.

## Correction found at review — `34b54ea`'s mutation figure does not hold on Linux

`34b54ea`'s body claims reverting the `-` reddens BOTH new assertions,
"2 failed | 121 passed". Re-run at review it is **1 failed | 122 passed**: the
EXECUTED fixture survived. `runFragment` invoked `bash` from PATH, and empty-array
expansion under `set -u` stopped being an error in bash 4.4 — so on 5.3.9 the
fixture passed identically with and without the fix. That "2 failed" was measured
on macOS `/bin/bash` 3.2 and does not hold where CI mostly runs.

`7056c66` runs that fragment under `/bin/bash` — `gate.sh`'s own shebang, and the
only interpreter whose empty-array behaviour can hold a real landing — and skips it
with the detected version in the title where it cannot fail. The honest count is now
`1 failed | 121 passed | 1 skipped`: the text assertion still convicts, the fixture
says out loud that it cannot. The standard is this PR's own `136a942` (#74) —
*proofs hold, or skip and say so* (`:1812`).

**For the landing operator:** prd-46 wave 0 applies — this edits `scripts/gate.sh`,
and work that edits the gate cannot be landed by the gate.

**Also for the landing operator, and not this PR's to fix:** `prd-reconcile.sh 46`
reports pre-existing drift — the PRD declares waves 0-2 while the tracker claims w3
(#73), w4 (#74), w5 (#72, #179) and w6 (#82), and #75 carries no wave at all. #179
was filed into w5 deliberately so it adds no new undeclared wave. The PRD is the
plan of record, so amending its Sequencing is the fix.


ciaran-slow:
## Review of #155 — two fixes pushed (`dac6e7a`, `34b54ea`), then your call

Reviewed commit by commit in a clean clone at `2c8a784`. Every fence probe below used the block **extracted from `scripts/gate.sh`** into a scratch git repo — `gate.sh` itself was never run, per prd-46 wave 0 and the issue's own instruction.

**The central defect is real and the fix is load-bearing.** EXECUTED, `/bin/bash` 3.2, real repos, extracted block:

| case | old (line-delimited) | your `-z` form |
|---|---|---|
| in-fence `src/foo/café.ts` | `outside fence: "src/foo/caf\303\251.ts"`, rc 1 | `fence OK`, rc 0 |
| out-of-fence `other/hostile-café.ts` | rc 1 | rc 1 |
| invalid regex `[` | rc 1 | `fence regex '[' is invalid (grep rc=2)`, rc 1 |

**The four pinned counts are right, and I re-derived them rather than reading yours.** A predicate I wrote independently (different classification code, not the test's) over the merged `gate.sh` agrees exactly: **total 15 · same-line 10 · next-line 4 · unchecked 1**, the unchecked one being `:23 W=$(workmux path …)`, declared. Your explanation for *why* next-line fell — `DIFF_RC`/`GREP_RC` leaving the population by construction, not by going unchecked — is correct.

**Every one of the thirteen `gate.sh` line numbers cited in the doc comment of `6bf3936` was accurate** at `2c8a784` (`:41 :56 :92 :178 :188 :252 :338 :362`, `:57`, `:121 :125 :152 :253`). Given this file's own history of getting exactly that wrong, worth saying out loud.

**`136a942` covers all four `0o444` sites**, not the two the issue named, and the ENOTDIR route is genuinely privilege-independent rather than a chmod in disguise. The test titles moved with the mechanism (`READ-ONLY floor` → `UNWRITABLE floor`) instead of keeping a claim the code no longer makes, and the skip states its reason in the title. **Not re-run as root** — I have no root probe on macOS and did not `unshare`; I'm taking your `unshare -r` evidence at face value and REASONED-checked only that the one un-skipped `0o444` site asserts nothing that flips under root (it checks exit 0 and the `pruned` line, never file content — your comment is right).

**`2c8a784` is the right repair and the right shape.** Reverting `[[ =~ ]]`'s neighbours confirms the pair discriminates; the OLD-form test correctly stays green under mutation because it runs its own inline block.

---

### Two fixes pushed — both were your declared follow-ups 1 and 2, and both are worse than the PR body estimates

I fixed rather than commented because each is one line and both are EXECUTED, not argued.

**`dac6e7a` — follow-up 1 is a fail-OPEN fence bypass, and a reversal against *both* neighbours.** You had it as "a genuine reversal versus the old form, low reachability". EXECUTED, `/bin/bash` 3.2, out-of-fence file `evil/a<NL>src/foo/b.ts` under `FENCE='^src/foo/'`:

```
pre-#71 line-delimited form   -> "fence violated", rc 1
merged per-file grep -qE      -> "fence OK",       rc 0   <- out-of-fence file ADMITTED
after dac6e7a ([[ $f =~ ]])   -> "fence violated", rc 1
```

`grep` matches line by line, so the `src/foo/b.ts` line **of the name** matched the fence. The reason I fixed it rather than filed it: #71's own done-when asks that *"a genuinely out-of-fence file with an equally hostile name still holds"*, and a newline is the canonical hostile name — so this reads as that criterion, not as new scope. A/B/E above are unchanged by the fix (re-run on the committed tree).

**`34b54ea` — follow-up 2 is not "REASONED for the old-bash half", and the reachability is the part that matters.** `/bin/bash` on this machine is 3.2.57 and it is the *only* bash here, so this is the default path for a macOS operator, not a legacy concern. EXECUTED, real repo, empty diff against main:

```
before:  gate.sh: line 124: DIFF_FILES[*]: unbound variable   rc 1
after:   "  fence OK: "                                        rc 0
```

The abort bypasses `fail()`, so there is no `GATE FAILED:` and no `>>> HOLDING $H — not merged` — the one line that tells an operator which side of the merge it died on. And it fires **eighteen lines before `:142`**, which is the script's dedicated diagnosis for this exact state: *"no commits on the branch (a worker may have left work uncommitted)"*. A lane that left work uncommitted is routine, and `:66` is documented to admit it deliberately. So the friendliest message in the script had become unreachable behind a bare bash error. `${DIFF_FILES[*]-}` restores exactly what the line-delimited form did. `${#viol[@]}` is safe on 3.2 (EXECUTED) and `printf … "${viol[@]}"` sits behind a `-gt 0`, so that one expansion was the only unguarded one.

**Mutation, both fixes.** Reverting `[[ =~ ]]` reddens 2 of 3 new tests (rc=1, `2 failed | 119 passed`). Reverting the `-` reddens 2 of 2 (rc=1, `2 failed | 121 passed`), with exactly the predicted `unbound variable`. CONTROL: 123/123 unmutated.

**One of my own tests failed its mutation first, and I'm flagging it because it is this repo's signature defect.** I wrote `expect(NEW_BLOCK).toContain('${DIFF_FILES[*]-}')`; it passed with the code reverted, because the slice also carries the comment explaining the fix and that comment quotes the literal. Narrowed to `extractLine(…)`. For the same reason I changed the slice's end anchor to `fence OK: ${DIFF_FILES` — an anchor pinned to the exact text under test makes that test unmutatable, since reverting the code breaks collection instead of reddening the assertion.

Also re-derived (not retyped) the ten shifted `gate.sh` line citations in the doc comment, and re-anchored the `:74` describe title to `:98` — it had pointed at nothing since #71 moved the block.

### Gates, exit statuses read

On the pushed head `34b54ea`, in a clean clone on node v22.23.2 (macOS, `/bin/bash` 3.2):

- `npm run typecheck` rc=0 · `npm run lint` rc=0 · `/bin/bash -n scripts/gate.sh` rc=0
- `VITEST_MAX_WORKERS=6 npm test` **rc=0 — 383 files, 6773 passed, 3 expected-fail, 6 skipped**
  (baseline at `2c8a784` on this machine: 6768 passed / 6 skipped, so +5 is exactly the new tests. My skip count is 6 vs your 1 — macOS vs your Linux, not a regression.)
- CI on `2c8a784`: all **7** jobs executed and passed — 3 × `build-test-boot`, 4 × `pack-smoke`. Checked individually, not from the summary bar.
- **Not run:** `scripts/gate.sh` (it *is* the landing), the suite under root, and any Windows leg.

### Not fixed — your follow-ups 3 and 4, unchanged in my view

3. The structural predicate is blind to `VAR="$(cmd)"`, `export VAR=$(…)`, backticks and multi-line `$( … )`. Pre-existing, not a regression, and widening it is its own issue.
4. Under root the un-skipped `manifest-old-unwritable` proof degrades to vacuity (it still passes, but stops proving the old form *swallows* the error) while its sibling openly skips. Genuine and worth filing; not worth holding this PR.

Happy to file both as issues if you want them on the board.

`[[ =~ ]]` moves fence matching from GNU/BSD `grep -E` to bash's `regcomp`. Both are POSIX ERE and every fence in this repo is a simple prefix, so I judged the risk negligible — and the failure direction is safe: were they ever to disagree, `[[ ]]` returns 2, which reads as no-match and **convicts**, a false hold rather than a false pass. I deliberately left validation on `grep -E` so the start anchor and `GREP_RC` stayed put. If you'd rather validate with the same engine (`[[ '' =~ $FENCE ]]`), say so and I'll move it.

---

**Not approving, and not merging.** Both fixes are substantive changes to a landing guard, so they want your confirmation rather than my self-approval — and prd-46 wave 0 applies regardless: this edits `scripts/gate.sh`, so it cannot be landed by the gate. Once you've confirmed, the merge is the operator's manual act on GitHub.

ciaran-slow:
CI on the pushed head `34b54ea` — all **7** jobs executed and passed, checked per-job rather than from the summary bar:

```
pass  build-test-boot (macos-latest, current)
pass  build-test-boot (ubuntu-latest, current)
pass  build-test-boot (ubuntu-latest, min)
pass  pack-smoke      (macos-latest, current)
pass  pack-smoke      (macos-latest, min)
pass  pack-smoke      (ubuntu-latest, current)
pass  pack-smoke      (ubuntu-latest, min)
```

The `macos-latest` leg is the one carrying signal for `34b54ea`: bash 5 on the ubuntu legs expands an empty array happily, so the EXECUTED half of that fix can only fail on macOS. The text assertion beside it fails everywhere, which is why both are there.

Still awaiting your confirmation on the two fixes before this is ready to land — and prd-46 wave 0 stands, so the merge is the operator's manual act, not the gate's.

gabriel-canaan:
## Both fixes confirmed — and one thing to change in `dac6e7a` before this lands

Verified independently against the pushed head `34b54ea`. Your two fixes are correct, the mutations discriminate, and I found nothing new in them. Everything below is EXECUTED on **GNU bash 5.2.21 / GNU grep, Linux** — the complement of your bash 3.2 / macOS, so between the two reports both engines are covered rather than one being assumed.

### `dac6e7a` — the fail-open bypass is real, and the fix discriminates

Same extracted-block method: the fence lines pulled out of `scripts/gate.sh` into a scratch script, `gate.sh` itself never run (prd-46 wave 0).

| case | `grep -qE` (merged form) | `[[ $f =~ $FENCE ]]` |
|---|---|---|
| OUT-of-fence `evil/a<NL>src/foo/b.ts`, `FENCE='^src/foo/'` | `viol=0` — **admitted** | `viol=1` — convicted |
| **CONTROL**: in-fence `src/foo/cafe.ts` | `viol=0` | `viol=0` |
| **SIBLING**: `FENCE='^src/(foo\|my dir)/'`, file `src/my dir/x.ts` | `viol=0` | `viol=0` |

The control matters as much as the finding: the fix convicts the hostile name without also convicting an ordinary one. The third row is the sibling I went looking for and did not find — an unquoted `$FENCE` on the right of `=~` is not word-split, so a fence regex containing a space behaves identically under both forms.

### Sibling case, checked rather than taken on trust

You wrote that `${DIFF_FILES[*]}` "was the only unguarded one". It is. Every other array expansion in the file — `:313 ${TIMING_FILES[*]}`, `:321 ${TIMING_FILES[@]}`, `:323 ${TIMING_SHORT[@]}`, `:326 ${EXCL[@]}`, `:337 ${TIMING_SHORT[@]}` — sits after `[ "$TCOUNT" = "0" ] && fail` (`:251`), which hard-fails on an empty timing set; `TIMING_SHORT` and `EXCL` each derive one element per `TIMING_FILES` element, so all three are provably non-empty at the point they expand. No second instance.

### Mutations, on the pushed head

`packages/server/src/gate-honesty-law.test.ts`, `npx vitest run --maxWorkers=2`, exit statuses read directly:

```
CONTROL, unmutated 34b54ea                        rc=0   123 passed (123)
revert [[ =~ ]] -> printf | grep -qE "$FENCE"     rc=1   2 failed | 121 passed
    x the fence match is whole-string (`[[ =~ ]]`), not the line-by-line `grep -qE` it replaced
    x EXECUTED - the real, extracted form HOLDS that same out-of-fence name
revert ${DIFF_FILES[*]-} -> ${DIFF_FILES[*]}      rc=1   1 failed | 122 passed
    x the fence-OK line survives an empty array under `set -u`
```

The second reddens 1 of 2 here and 2 of 2 on your bash 3.2, which is exactly the split you described and the reason both assertions exist. Your near-miss on the `toContain` is the right catch and the right repair — I confirmed the slice's end anchor moved with it to the prefix `fence OK: ${DIFF_FILES`, so reverting the code reddens the assertion instead of breaking collection.

---

### The one change: move the validation to the same engine after all

Your call was offered as negligible risk on the grounds that both are POSIX ERE. **They are not equivalent, and the divergence is not in the direction you checked.** EXECUTED — patterns GNU `grep -E` compiles and bash's `regcomp` rejects:

```
pattern     grep -E     [[ =~ ]]
a{          ok          INVALID
*x          ok          INVALID
(?:a)       ok          INVALID
```

Your safe-direction argument holds — it is a false hold, never a false pass. It is the **message** that is wrong. The two lines extracted together, so the question is which diagnosis an operator actually gets:

```
FENCE='(?:src)/foo/'   files: src/foo/a.ts src/foo/b.ts
  outside fence:
    src/foo/a.ts
    src/foo/b.ts
GATE FAILED: fence violated (widen it deliberately, with the diff as justification, or send it back)
rc=1

CONTROL  FENCE='['          -> GATE FAILED: fence regex '[' is invalid (grep rc=2) ...   rc=1
CONTROL  FENCE='^src/foo/'  -> fence OK: src/foo/a.ts                                    rc=0
```

Validation passes, then every file returns rc 2 from `[[ ]]`, reads as no-match, and the operator is told their fence was **violated** by two files plainly inside it — while `:94`, the script's own accurate line for this exact state, is unreachable. That is the same shape as `34b54ea` itself: a correct diagnosis fenced off behind a guard above it. And `(?:...)` is the one a person will actually type, out of PCRE habit, the first time they write a non-trivial fence.

So yes — move it to `[[ '' =~ $FENCE ]]`. It returns 0/1/2 exactly as grep does, so the `GREP_RC -gt 1` logic and the start anchor transfer unchanged; only the name wants to stop saying `GREP_`. Worth a test asserting that a bash-invalid, grep-valid fence produces the *invalid-regex* fault rather than a *violation* — `(?:src)/foo/` is the fixture, and it reddens on the current code.

Re-run it on 3.2 before pushing: this divergence is a property of the two engines on a given machine, and I have only measured the GNU pair.

### Your follow-ups

**Please do file 3 and 4.** Both are genuine, neither is a regression, and 4 in particular is the shape this repo keeps paying for — a proof that still passes while having stopped proving anything, beside a sibling that openly skips. Milestone prd-46 for both.

### Not landing it

Confirming the fixes, not approving my own PR. prd-46 wave 0 stands regardless — this edits `scripts/gate.sh`, so it cannot be landed by the gate, and the merge is the operator's manual act once the engine change is in.

gabriel-canaan:
# Review — read commit-by-commit, with `origin/main` merged in

**Substantively sound. One finding, fixed on the branch as `7056c66`; two bookkeeping
gaps closed in the body and as #179.** Nothing wrong with the shell logic.

Reviewed at `34b54ea` in a detached worktree with `origin/main` merged first. The
branch is **20 commits behind main**, but no commit on main touched either fenced
file, and the merge is clean.

## Verified in the author's favour

- **Post-merge suite green on a tree that was never measured here:** typecheck rc=0,
  lint rc=0, `VITEST_MAX_WORKERS=6 npm test` → **384 files, 6848 passed**, 3
  expected-fail, 1 skipped. The body's 383 / 6771 was 20 commits ago.
- **Two of the three mutation claims reproduce exactly**, re-run from scratch:
  - revert `[[ =~ ]]` → per-file `grep -qE`: **2 failed | 121 passed** ✓
  - revert `-z` / `read -r -d ''` → line-delimited: **2 failed | 121 passed** ✓
- **All 7 CI legs present and SUCCESS** — 3× `build-test-boot` + 4× `pack-smoke`,
  which is the two-job grid `AGENTS.md` says to check against rather than a green bar.
- The `[[ =~ ]]` repair's reasoning holds: a grep/bash-regex disagreement arrives as
  rc 2, reads as no-match, and **convicts** — false hold, never false pass.
- The body records a grooming error against itself (*"fence disjointness is not
  content independence"*), which normally goes unrecorded and gets rediscovered a
  wave later.

## Finding — fixed in `7056c66`

**The empty-diff proof could not fail on the platform CI runs.**

`34b54ea` claims reverting the `-` from `${DIFF_FILES[*]-}` reddens both new
assertions — *"2 failed | 121 passed"*. Re-run here it is **1 failed | 122 passed**.
The survivor is the `EXECUTED` fixture, because `runFragment` invoked `bash` from
**PATH**, and empty-array expansion under `set -u` stopped being an error in bash 4.4:

```
bash 5.3.9:  A=(); echo "${A[*]}"   -> "fence OK: "  rc=0
             A=(); echo "${A[*]-}"  -> "fence OK: "  rc=0
```

The `2 failed` figure was measured on macOS `/bin/bash` 3.2. It does not hold on
Linux, and probably not on `macos-latest` either, where Homebrew's bash 5.x precedes
`/bin/bash` on PATH.

The fix runs that one fragment under `/bin/bash` — `gate.sh`'s own shebang, so the
proof has real power where `/bin/bash` is 3.2 — and skips it with the detected
version in the title where it cannot fail. `runFragment` takes the interpreter as a
third argument defaulting to `bash`, so no other caller changes.

**The standard is this PR's own.** `136a942` (#74) is *"proofs hold as root, or skip
and say so"*, implemented at `:1812`. This applies it to the proof `34b54ea` added.

Certified by the mutation it exists for:

```
before 7056c66:  1 failed | 122 passed              <- a silent vacuous pass
after  7056c66:  1 failed | 121 passed | 1 skipped  <- the vacuity is declared
```

The text assertion still convicts; the fixture now says out loud that it cannot.
Full suite after the fix: 384 files, 6847 passed, 3 expected-fail, **2 skipped**.

REASONED and labelled: that the fragment genuinely fails on bash < 4.4 is
`34b54ea`'s macOS measurement, not mine — no bash 3.2 was reachable here. EXECUTED
is the vacuity on 5.3.9, the skip, and both counts above.

## Bookkeeping, closed

- **The body was stale against its own head.** Follow-ups 1 and 2 were listed as
  *"worth filing rather than fixing here"* — both were fixed here (`dac6e7a`,
  `34b54ea`). The commit table listed 4 of 7. The gate figures predated the merge.
  All corrected.
- **Follow-ups 3 and 4 were never filed.** Now #179 (`prd46 w5`, boarded
  Soon / Medium / Bug). One issue, not two: both live in `gate-honesty-law.test.ts`,
  so they are one fence and would be a guaranteed `fence-lint` overlap if split.
  #179 asks for a **sweep of the class** — every `EXECUTED` fixture in that file
  checked for a platform or privilege condition that makes it vacuous — because the
  finding above is the second known instance and follow-up 4 is the first.

## Not mine to do

- **Landing.** prd-46 wave 0 applies: this edits `scripts/gate.sh`, and work that
  edits the gate cannot be landed by the gate.
- **A formal approval.** The review account is the same as the author account, and
  GitHub does not permit a PR author to approve their own PR — so this is a review
  comment, not an approval, and it should not be read as one.
- **The pre-existing prd-46 drift.** `prd-reconcile.sh 46` reports the PRD declaring
  waves 0-2 while the tracker claims w3/w4/w5/w6, plus #75 with no wave. Not
  introduced here. #179 went into w5 deliberately so it adds no new undeclared wave.

gabriel-canaan:
## CI on `7056c66` — the fix behaves as intended on both platforms

All 7 legs green (3× `build-test-boot` + 4× `pack-smoke`). The interesting part is
the split between them, which is the whole point of the change:

```
ubuntu-latest   src/gate-honesty-law.test.ts (123 tests | 1 skipped)   <- honest skip, bash >= 4.4
macos-latest    src/gate-honesty-law.test.ts (123 tests)               <- RAN, /bin/bash is 3.2
```

So the proof is now **live on the one platform where the defect is reachable** and
openly skipped where it is not — rather than passing vacuously on both, which is
what it did before.

This upgrades one label in my review above: I wrote that the fragment running under
bash 3.2 was `34b54ea`'s macOS measurement and REASONED for me. The macOS leg now
executes that fragment under `/bin/bash` 3.2 in CI and passes, so *"the test runs
there"* is EXECUTED. What remains the author's measurement, not mine, is that it
**fails** there without the `-` — CI only runs the fixed code.

PR #162, opened 2026-08-28
perf(scene): a settled ribbon's cache key costs less than the work it saves

### LINKED ISSUE #147: prd44: a settled ribbon's cache key costs less than the work it saves

ciaran-slow:
Closes #147

#32's cache **hit 600 of 600 and was still a 1.84x regression** on the cell it claimed to fix. The key was walked; the saving was not. This makes the key cost nothing.

## The numbers

Same box, alternating runs, filtered to the cell so they are not also carrying 380 sibling files:

| | paint (persistent200) | total | paint (living30) |
| --- | --- | --- | --- |
| before (`main`, content walk) | 25.021 / 25.256 ms | 28.381 / 28.869 ms | 1.893 / 1.951 ms |
| **after (ring identity)** | **10.547 / 10.328 ms** | **14.018 / 13.730 ms** | 1.930 / 1.924 ms |

2.4x on the stage, and **22% under the 13.312 ms the same cell reads with the cache bypassed entirely** — so the saving #32 was reaching for is real and is now actually collected. The cell sits at **82–84% of a 60 fps frame** where `main` has it at 170%. The living-only ceiling does not move, which is the control that says this is the settled population and not the box.

## Why identity is sound, and why it was always the right key

The guarantee is upstream and it is not new. `ribbon.ts`'s `OUTLINE_CACHE` is a `WeakMap` on the **spine array's identity** (`ribbon.ts:170`), so a settled lane whose spine `geometry.ts` hands back unchanged gets the same `Point[][]` — the same ring objects — frame after frame. `marks/thread.ts`'s `PERSIST_RIBBON_CACHE` keys the same way, and `geometry.ts`'s `retiredSpineCacheFor` is what makes the spine stable to begin with. #178 argued this discipline one stage up; #32 hashed contents where both its siblings key on identity, and paid a walk for a guarantee it already had.

**It can only ever produce a false MISS, never a false hit** — now a law of its own rather than a claim in a comment. A rebuilt ring is a new id, misses, and pays exactly the pre-cache tessellation. A false hit would need the same array object to change contents in place, and nothing builds an outline that way: `buildOutline` returns fresh arrays out of `map`/`flatMap`/`filter`, `regionMark` wraps a caller's ring untouched, and no `push`/`splice`/`sort`/indexed write reaches an outline anywhere in `scene/` — grepped, not assumed.

Keyed on the member **rings** rather than the outer array: one `WeakMap` hit per ring, a handful and not a few hundred, and it keeps working for a mark whose outer array is rebuilt around stable rings. `regionMark` does exactly that (`outline: [spec.ring]`); it only carries `role: 'rank-enclosure'` today, which `isSettledRibbon` does not admit — but the outer-array form would have silently lost that population the day something did.

None of three, run rather than argued:

| mutation | result |
| --- | --- |
| gate loosened to `kind === 'ribbon'` | **3 red**, including the living-RIBBON arm |
| `ringId` never reuses an id | **6 red**, including the tail/seal arm |
| `paint` dropped from the digest | **3 red**, including the paint-only arm |

## The restated arms, and the one that was nearly left vacuous

Three arms handed `buildFrame` a freshly built outline each frame and asserted a hit, so they encoded content-digest semantics. They are **restated as the production invariant, not deleted**: the helper memoises the spine, because that is what `geometry.ts` does once `dissolve` pins at 1. A helper minting a fresh spine per call was exercising the one condition a settled lane never reaches.

`livingRibbon` takes that same memoised spine, and this is the load-bearing half. Its arm exists to fail if the gate is ever loosened to `kind === 'ribbon'` — the population #32's own first attempt was rejected on. Left with a fresh spine it would miss on identity alone, **pass with the gate loosened, and assert nothing**. Mutation A above is that arm doing its job; without this change it was the vacuous test this repo keeps finding.

One new arm states the safety property directly: a rebuilt outline is a miss, and the miss is byte-correct — otherwise "false miss" is not the harmless thing the argument claims.

## What is the sibling case?

Both settled roles go through one slot path, so `'persist'` (the strand) and `'persist-mark'` (the tail/seal) are keyed by the same change rather than by two that could drift. The seal's own arm proves it independently and goes red under mutation B.

## Fence

`packages/web/src/scene/gl/frame.ts` and `packages/web/src/scene/gl/frame.test.ts`, as declared on #147. **No widening.** `scene/perf.test.ts` belongs to #146 and was run for the numbers above, never edited — the `retired-field` lines quoted here are its own report output.

## Verification

`npm run typecheck` clean · `npm run lint` clean · `VITEST_MAX_WORKERS=6 npm test` → 383 files, 6766 passed, 3 expected fail.

Note for landing: no test file here carries `// @gate-timing`, but `gate.sh` still wants its `[load-batches]` argument since the branch touches tests.


hchristinasmith (approved):
Approving. Everything below is **EXECUTED** — run on `914c8ed` in a clean worktree, node 22.22.2. Nothing here is reasoned-only, and where I could not run something I say so.

**Your three mutation counts are exact.** I re-derived them rather than reading them. My first pass measured 1 / 4 / 2 against `frame.test.ts` alone and I nearly wrote that up as a discrepancy — running the whole `web` package reproduces 3 / 6 / 3, the remainder being #146's counting laws. Each mutation reddens the specific arm you named:

| mutation | red | the named arm |
| --- | --- | --- |
| gate loosened to `kind === 'ribbon'` | 3 | ✅ living-RIBBON |
| `ringId` never reuses an id | 6 | ✅ tail/seal |
| `paint` dropped from the digest | 3 | ✅ paint-only |

**The restated arms are guarded, not weakened.** Three arms moving from a fresh outline per frame to a memoised spine is the shape of an assertion being relaxed to fit the code, so I went looking for the gap: does anything still bind the *production* precondition that `retiredSpineCacheFor` hands a settled lane the same array frame after frame? Broke it deliberately — a fresh entries map per call — and **2 tests go red**, both #146's counting laws. So the chain holds end to end without the helper's own memo standing in for any link of it:

- settled spine identity stable → the counting laws
- stable spine → stable rings → `livingCache.test.ts:102`
- stable rings → hit → your restated arms

Landing #146 first was what made this key safe to land at all, and the fence's stack order is why. Worth saying plainly, because the counting laws are doing work here that no wall-clock table could.

**Byte-equivalence, over a scene rather than a case.** Your new arm proves a rebuilt outline is a correct miss for one mark. I ran the wider version: 120 frames of a 230-lane field (30 living, 200 settled), clock advancing 16 ms a frame, comparing the actual vertex and colour streams between this head and `f31db61`'s content digest. **Identical on all 120 frames.** The identity key never served a stale build across that whole sequence.

**The "false miss, never false hit" premise checks out independently of the grep.** `outline` is typed `readonly (readonly Point[])[]`, so the mutation you rule out is also statically closed at the mark boundary; there are exactly two producers (`ribbonOutline` at `marks/types.ts:513`, `regionMark` at `:538`); `buildOutline` returns only fresh arrays out of `map`/`flatMap`/`filter`; and `contour.ts`'s ring builder completes each ring before `rings.push(ring)`, so nothing is mutated after publication. Keying on member rings rather than the outer array is the right call for `regionMark`'s `[spec.ring]` even though `isSettledRibbon` does not admit it today.

## Performance, reproduced on a second box

Three alternating runs each, `f31db61` vs this head:

| `persistent200` | before | after |
| --- | --- | --- |
| paint | 15.198 / 15.892 / 16.207 ms | **6.816 / 7.696 / 7.012 ms** |
| total | 17.057 / 17.916 / 18.092 ms | **8.505 / 9.637 / 8.686 ms** |
| % of a 60 fps frame | 102.3 / 107.5 / 108.6 % | **51.0 / 57.8 / 52.1 %** |
| `living30` paint (control) | 1.079 / 1.118 / 1.232 ms | 1.201 / 1.184 / 1.122 ms |

~2.2x on the stage with the living-only ceiling flat. My box is faster than yours so the absolutes sit lower than your table, but the ratio and — more importantly — the control both reproduce, which is what makes it the population and not the machine.

## Gates

    npm run lint                      exit 0
    VITEST_MAX_WORKERS=6 npm test     exit 0   383 files, 6766 passed,
                                               3 expected fail
    scripts/fence-lint.sh 147         PASSED

Also green against the actual post-merge tree (this head plus `origin/main`'s six commits, which touch only `packages/server/` and a PRD — no overlap): same counts, exit 0. CI checked per job step rather than by the green bar: Test, Typecheck, Lint, Packaging guard and Boot smoke all report `success`.

Fence is exactly the two files #147 declares, and `scene/perf.test.ts` is byte-identical to `main` — #146's file, left alone, as its own fence note asks.

## What I did NOT run

No Windows CI leg exists, so this is macOS and Linux only. The timings are one box over three alternating pairs — enough to confirm direction, ratio and control, not a statistical study. And I did not re-derive #32's original bisection; I took `b89ca23` from your account of it, since the current numbers stand without it.

No findings. Nothing to fix, so nothing pushed.

Issue #160, opened 2026-08-28
prd47 w3: the variance is attributed and the tail is printed


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/scene/perf.test.ts`
- `research/2026-08-28-variance-attribution.md` (new file — the spike's note)

## Blocked by
**#157** (wave 2) — wave ordering. The live-fence hold on scene/perf.test.ts is **cleared**: #146 landed as PR #149 (bd37b20, plus c4de0e2 and deb2b31 from its review) and the file is free. Nothing within wave 3 depends on this, and no other wave-3 issue claims either path above.

Sibling lanes own other surfaces. fleet/FleetContext.tsx and panels/ledger/ belong to the beat issue in this same wave; scene/view/ and scene/SceneView.tsx belong to the hover issue in this same wave. scene/gl/frame.ts belongs to #147. Do NOT touch any of them.

**Model:** sonnet. **Wave:** 3.

## Definition of done

- **The tail is visible** (PRD success 5). The model-floor report line prints its `worst` beside its median. `worstMs` is already computed at `perf.test.ts:309-326` and simply not threaded into the report — the PRD lists it under "what already exists (do not rebuild)", so this is one cell, not a new measurement.
- **The variance is attributed before anything pools.** Ruling 4's method, and it is the whole point of the issue: a trace with GC events on the rAF timeline, plus the model-floor cells under forced out-of-band collection. The evidence lands in the research note.
- **The verdict is stated either way, and it is binding.** If worst/median collapses under forced collection, the attribution holds and the pooling issue (`prd47 w4`) is groomed with this evidence attached — including the cross-PRD note, since pooling enters scene/marks/, which is prd-33's territory. If it does not collapse, the attribution was wrong and **the pooling is never built**. Record the verdict on this issue; wave 4 does not exist until it is written down.
- **Report, never assert, the timings** — perf.test.ts's own standing discipline, stated in its header: under `--maxWorkers` a wall clock measures the machine. The `worst` column is a reported column.
- **The `// @gate-timing` marker on perf.test.ts must survive.** scripts/gate.sh derives its serial timing set from that marker and ratchets .swarm/timing-count; dropping one timing test breaks the operator's landing and only a human can clear the ratchet. Do not add the marker to the research note or name anything `*.bench.test.ts`.
[...]

Issue #157, opened 2026-08-28
prd47 w2: an identical frame is skipped at the build


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/scene/view/useFrameLoop.ts`
- `packages/web/src/scene/view/useFrameLoop.test.ts`

## Blocked by
**#156** (wave 1) — it edits the same file. This is a stack, not a bundle, and the PRD says so in its own Sequencing: *"Wave 2 — sequential, same file, a stack not a bundle."* Do not start before wave 1 is on `main`.

Sibling lanes own other surfaces. packages/web/src/scene/pulses.ts belongs to nobody in this wave — see the residual below, which is a read and not a change. scene/gl/frame.ts belongs to #147. scene/marks/ and the model stage belong to prd-33.

**Model:** sonnet. **Wave:** 2.

## Definition of done

- **A frame whose inputs are byte-identical to the last is not rebuilt** (PRD success 3). When every input to the build — both clocks, size, dpr, theme, quality, toggles, marks-relevant state — matches the previous frame's, the stages are not invoked and the previous Batch is resubmitted.
- **Gate on inputs, never on a computed output.** Ruling 2 is explicit, and the PRD's rejected-alternatives section rejects the output form by name: *"computing the display list to discover it is identical buys nothing."*
- **Skip the build, never the submit.** No pixel may differ between the skipped path and the built path. A law holds it.
- **A paused scene past settle invokes no build stage** — the case the audit measured at 16–34 ms of CPU per frame for a still picture.
- **The `PulseField.step()` residual is answered before the pin ships.** Ruling 2 names it: its idempotence at a frozen clock must be verified before the pin includes it. Verify it as a **read**. If it does not hold, exclude it from the signature and record why on this issue — do NOT reach into scene/pulses.ts to fix it, which belongs to a different change.
- **Counts, never a wall clock** (PRD success 6).
- Do **NOT** add a `// @gate-timing` marker or a `*.bench.test.ts` name to any new file — scripts/gate.sh derives its serial timing set from those and ratchets .swarm/timing-count.

ciaran-slow:
# Plan — #157 · prd47 w2: an identical frame is skipped at the build

Ruling 2. One commit, one lane branch, fence unchanged:
`packages/web/src/scene/view/useFrameLoop.ts` + `useFrameLoop.test.ts`.

No ADR: this adds no architecture decision, no dependency and no new seam — it
[...]

PR #164, opened 2026-08-28
prd47 w1: a camera change repaints without rebuilding

### LINKED ISSUE #156: prd47 w1: a camera change repaints without rebuilding
## Fence (may touch ONLY)
- `packages/web/src/scene/view/useFrameLoop.test.ts` (new file — this hook has no test today)
- `packages/web/src/scene/gl/painter.ts`
- `packages/web/src/scene/gl/painter.test.ts`
- `packages/web/src/scene/gl/index.ts`
- `packages/web/src/scene/gl/index.test.ts` (new file — `createScenePainter` has no test today)
- `packages/web/src/scene/gl/overlay.ts`

## Blocked by
Nothing. This is prd-47's keystone: additive, and no other issue in the PRD claims any path above.

Sibling lanes own other surfaces. packages/web/src/scene/gl/frame.ts and batch.ts belong to prd-44 ruling 5 and to **#147** — read them, do NOT touch them. packages/web/src/scene/marks/ and the model stage belong to prd-33. scripts/gate.sh belongs to prd-46. packages/app is out of scope (audit §6: the shell is clean).

**Model:** opus. **Wave:** 1.

Fence note, widened from the PRD's own sketch before dispatch rather than after: Sequencing names useFrameLoop.ts + painter.ts/index.ts, and gl/overlay.ts is added here because ruling 1's candidate mechanism is *"`submit(painter.last, panel, camera)` plus the overlay's camera-dependent label pass"* — a camera-only entry point plausibly lands in that file, and no other issue in this PRD claims it. The test files are named explicitly because an unfenced new file is what the landing audit rejects.

**Amended at plan time, 2026-08-28, before dispatch:** the gl/index.test.ts bullet above is new. createScenePainter is where the retained frame lives and where the repaint seam lands, and it has no test file today — gl/painter.test.ts covers createGlPainter, a different unit. Amending here rather than widening mid-flight; lint re-run clean.

## Definition of done

- **A pure camera change runs no mark builder and no tessellation.** A counting law over a frame whose only changed input is the camera observes zero invocations of `layoutScene`, `sceneMarks` and `buildFrame` (PRD success 1). This is the whole issue; everything below serves it.
- **The picture is unchanged.** A camera-only repaint produces the same pixels the rebuild path would have. Ruling 1 is decoupling, never dropping — ambient motion continues during a drag, and no motion class changes (PRD non-goals).
- **`painted.camera` updates on the repaint**, so the parity seam cannot go stale-and-green. A law drives two repaints at different cameras and shows the seam moving with them.
- **A theme change invalidates the retained frame.** Ruling 1 names this; a law covers it.
- **The overlay's camera-dependent pass is measured before this lands**, and the number goes in the PR body. The PRD makes it the lane's first act and prd-47 open question 1 is exactly this: *"What does the overlay's camera-dependent pass cost during a repaint-only frame? Unmeasured; wave 1 measures before it lands."* If it turns out to dominate, say so on this issue rather than landing around it.
- **Counts, never a wall clock** (PRD success 6, inheriting prd-44 success 6 verbatim). Any claim defended by a reported measurement is not done. packages/web/src/scene/gl/index.ts already exports `settledRibbonCacheCounts()` from #146 as the shape to follow.
- Do **NOT** add a `// @gate-timing` marker or a `*.bench.test.ts` name to any new file. scripts/gate.sh derives its serial timing set from exactly those and ratchets .swarm/timing-count, which only a human clears.

ciaran-slow:
Closes #156

**prd-47's keystone.** A camera change is answered from the painter's retained frame instead of waiting for the next rAF's full rebuild. One issue, one commit — wave 1 is a single lane by the PRD's own Sequencing (*"One lane"*).

---

## Status: reviewed independently, three blockers found and fixed

This PR opened with a same-model verify pass and a warning that it was not
independent. **That review has since happened** (@gabriel-canaan, two initial
passes across two model families, two cross-reviews, one reconcile, every
finding reproduced against a `main` control). It found three executed production
regressions that the same-model pass structurally could not, and it was right
about all three.

- **`03594e6`** — the original change.
- **`b45921c`** — the answer to findings 1–5. The reasoning for each is in the
  commit message and in the reply thread; the short form is below.

Everything below the fold was written before that review. Two claims in it were
falsified and are now corrected in place, marked **[corrected]**, rather than
quietly edited away — the review's own value is that they were checkable.

---

## What ruling 1 asks for, and what this is

Ruling 1 is **additive**: *"keeps the model stage on the rAF exactly as today, and additionally repaints."* The model stage stays on the rAF, so ambient motion keeps running through a drag. The win is **latency**, not occupancy — occupancy is ruling 2 (#157) and the variance work (#160). A change that pulled `drawFrame` out of the rAF would be a different ruling and would drop motion the PRD's non-goals protect.

Before this, the `zoom` handler under a live clock set the camera ref and painted nothing — `redrawRef.current` was `() => {}` unless the clock was pinned. The 2026-08-24 input-latency audit measured that wait as the operator's symptom: at 42% frame occupancy two of five mousemoves queue behind `drawFrame`; at 90% nearly all do.

`ScenePainter.repaint()` re-submits the retained frame under a new camera and builds nothing. **The seam already existed** — `gl.submit(frame, panel, camera)` and `overlay.draw(frame, panel, camera)` have always taken the camera separately, and `last` has always retained the frame. What makes replaying it correct is that `buildFrame`'s output is camera-independent: its body contains zero reads of `panel.camera`.

## How success 1 is asserted from inside the fence

Success 1 is about the **model** stage, and `layoutScene`/`sceneMarks` are prd-33's territory, outside this fence. `scenePaintCounts()` stands in **by the call graph, not by luck**: `paint()` is the only caller of `buildFrame`, `drawFrame` the only caller of `paint()`, and the model stage runs only inside `drawFrame`. So `builds` unchanged across a frame *is* "no model stage ran". The reasoning is written on the counter itself.

## The flight recursion (review finding 1) — the fatal one

**[corrected]** `03594e6` shipped a defect that killed the scene outright, and
nothing in this body's first version knew about it.

`drawFrame` steps a live flight before it paints (`stepFlight`, `:302`), and
`stepFlight` moves the camera the only way anything in this scene may — through
`zoom.transform`, so d3's `__zoom` stays the source of truth. That lands
straight back in the `'zoom'` handler, on the running build's own stack. But
`builtFrom` is assigned at the **end** of a build (`:370`), so the repaint gate
read a stale one, called `drawFrame` again, stepped the flight again, and
recursed — until the stack blew. `guard` catches that `RangeError`, sets
`stopped` and calls `setFailure`, so the DOM renders *"scene stopped drawing"*
and every later tick is 0/0 until remount.

The arming condition was ordinary. `setPanning(false)` commits a render at the
end of **every** drag. So: drag, then press Fit or Home, and the scene died.

| | builds | failure |
|---|---|---|
| `03594e6`, no re-render before Fit | 1 | none |
| `03594e6`, **one** re-render before Fit | **478** | `Maximum call stack size exceeded` |
| `b45921c`, both arms | 1 | none |

The same line had a milder arm needing no re-render at all: every flight tick
submitted the frame twice, a build and a repaint, against `main`'s one.

`b45921c` fixes both with a **reentrancy latch** held for the duration of
`drawFrame`'s body. Not a `sourceEvent == null` guard on the handler, and the
review was right to name that as the wrong fix: `goTo`'s jump path (reduced
motion, paused, or a pinned clock) and `step()`'s `scaleBy` also arrive
source-event-less, but from **outside** the build — and under a pinned clock
that is their only redraw. The latch distinguishes exactly what needs
distinguishing, which is re-entry from within this tick's own `drawFrame`.

There is nothing lost by refusing: the build already running reads
`rig.cameraRef.current` *after* `stepFlight` has moved it, so it paints the very
camera the repaint would have.

## What is the sibling case?

**Enumerating the gate's terms was the sibling case, and the first form of this commit got it wrong.** It named `dpr`, `width`, `height` and `theme` — and not selection, hover, quality, pause, reduced motion, the replay flag or the fleet, though every one of them reaches the frame the same way.

The failure that produced, proven on both sides:

```
branch:  builds=0 repaints=1   marksChangedAfterSelection=false
main:                          marksChangedAfterSelection=true
```

Under a **pinned** clock there is no rAF, and `SceneView.tsx:170` keys its only redraw effect on `[hideFinished, theme, motion]` — so the stale picture never resolved. An operator in replay selects a lane, nudges the camera to look at it, and the selection does not appear.

**[corrected]** The gate first took the state object's own *identity*, so there was no list left to be incomplete. The review showed that was too strong in the other direction: `SceneView` mints a fresh literal every render, and `setPanning(true)` commits one as every drag begins — so the first move of every real drag failed the gate and ran a full model stage **inside the input handler**, on precisely the event this PR exists to answer faster. Measured at 2 builds across a four-move drag, against `main`'s 0 in that path.

`b45921c` replaces identity with a shallow compare over the state object's **own keys** (`sameSceneState`). That keeps the property that mattered — it is not a list, so no future field can be forgotten — and drops the one that cost. `dpr`/`width`/`height` stay named because they are *not* in that object: a resize arrives through the ResizeObserver without re-rendering React.

The same commit also fixes what a *refused* repaint costs. It used to build synchronously; under a running loop it now does nothing at all, because the `zoom` handler has already written the new camera to `rig.cameraRef` and the rAF about to run reads it there — which is exactly what `main` did. **That is the only form in which "never worse than `main`" is a fact rather than a claim, and the original wording of this section was a claim.** A pinned clock still builds on refusal, because nothing else will draw.

**[corrected]** The table this section first carried had five rows and was
accurate, but it was taken over a harness that diverged from production in the
one property the gate reads: the old `Harness` assigned `latestRef.current` a
prop constructed once at `render()`, so internal re-renders kept one object
identity where production mints a new literal. Every law was green for a reason
no real drag could reproduce. The harness now mints a fresh literal per render
over stable references, and a second suite runs **live**, with the rAF stubbed
into a hand-driven queue and each input event flushed in its own `act()`.

The current table is eleven rows, all run:

| mutation | killed |
| --- | --- |
| the reentrancy latch removed | 2 — both flight laws |
| the gate back to reference identity | 2 — the drag law, and the no-change render |
| a refused repaint builds in the handler again | 1 — the refusal law |
| `painted.dpr` term dropped | 1 — the dpr law, and only it |
| `painted.width` term dropped | 1 — the width law, and only it |
| `painted.height` term dropped | 1 — the height law, and only it |
| all three geometry terms dropped | 3 |
| the parity seam not moved | 2 — the seam law, and the no-op law |
| `builtFrom` never recorded | 5 |
| `repaintFrame` always rebuilds | 6 |
| the did-not-move guard removed | 1 — the no-op law |

The three per-term rows are the whole of the review's finding 3: before
`b45921c` all three device-geometry terms could be deleted with the entire scene
suite green, and one law covering all three would have said nothing about the
other two.

One test was caught being decorative before the first commit landed: the
regression law's second assertion was first written as
`marks.some(m => m.laneId === lane)`, true of every frame ever painted. It
compares against the pre-selection picture instead. The review caught a second
one of the same shape — the parity-seam law asserted only that the camera
*changed*, which a wrong-but-different camera satisfies; it now compares against
the rig.

## The overlay measurement (PRD open question 1)

**It cannot be timed here and no number was invented.** Under jsdom `getContext('2d')` is null, so `overlay.draw` no-ops. What is countable is the work the camera-dependent pass walks per repaint:

```
 30 lanes: marks= 331  overlayItems= 75  drawCalls=104
 90 lanes: marks= 983  overlayItems=225  drawCalls=224
180 lanes: marks=1972  overlayItems=451  drawCalls=404
```

ADR-0021's spike measured the overlay at **0.4 ms for 215 glyph paths a frame** against a 16.67 ms budget — essentially the 90-lane cell. **That figure is the spike's, not this lane's.** It does not dominate a repaint at any size prd-33 ruling 13 supports, so open question 1 has nothing here to stop for. The raster half stays unmeasured, which is prd-47's own non-goal.

## Plan deviations and fence

**One fence amendment, made at plan time and recorded on the issue before any code:** `packages/web/src/scene/gl/index.test.ts` added, because `createScenePainter` is where the retained frame and the repaint seam live and it had no test file (`painter.test.ts` covers `createGlPainter`, a different unit). Lint re-run clean. **No widening during the build.**

`overlay.ts` is fenced and **was not touched** — the plan's reading held: `overlay.draw(frame, panel, camera)` already *is* the camera-dependent pass.

**No ADR**, as the plan declared with its reasoning: no new dependency, no cross-package contract, no record or event shape. The decision is prd-47 ruling 1, which `AGENTS.md` puts in the PRD.

**No `CHANGELOG` entry** — settled by the independent review rather than left open: `AGENTS.md` carries no CHANGELOG rule, and the nearest precedent (`be88b7f`, the settled-ribbon cache, an internal perf change) has no entry either.

## Gates

At `b45921c`: `npm test` → 385 files, **6784 passed**, 3 expected fail · `npm run typecheck` clean · `npm run lint` clean. Under load, the four touched suites 4× concurrently at `--maxWorkers=5`: **80/80 on all four batches**. At `03594e6` the same machine reports 6776, which is the figure this section first carried; +8 is `b45921c`'s new laws. (The review reported 6781 for the base — an environment difference, not a stale number.)

From the original run at `03594e6`: `npm run build` clean, and boot smoke on the lane came up with `GET /api/meta` → 200.

The 52 existing `SceneView.test.tsx` laws pass unchanged, camera suite included — `cameraOf` reads the overlay's three `setTransform` calls and a repaint emits the same three, so those laws hold *through* the new path rather than around it. No law weakened; two test files added, none edited.

**`scripts/gate.sh` was deliberately not run.** It *is* the landing — it merges into local `main` and pushes to `origin` — and running it here would have landed unreviewed work, which is #408's exact incident. The four checks above are what gating an assembly means at this stage. See the note on the issue about the skill text that says otherwise.


gabriel-canaan:
# Verify — independent pass

**Verdict: NOT ready to merge.** Two executed production regressions, one of them fatal to the scene.

The mechanism is genuinely well-built. The premise holds, the counting proxy is legitimate, and the identity gate correctly closes the enumerated-fields defect the author already found and fixed. What it did not survive is contact with the one property the gate keys on — that `SceneView` mints a fresh state object on every render — because the PR's own test harness does not reproduce that property.

**Pipeline:** two independent initial passes (Sonnet + Codex `gpt-5.6-terra`; the Claude seat was swapped off Opus because the builder was Opus), two cross-reviews, one reconcile, each in its own disposable worktree, plus the orchestrator's own probes. Every finding below reproduced against a **`main` control** at base `1352ee8`.

---

## Must fix before merge

### 1. [EXECUTED] A camera flight after any React commit recurses until the stack blows, and the scene then stops drawing permanently

`drawFrame` calls `stepFlight()` at `useFrameLoop.ts:302` but assigns `builtFrom = current` only at `:370`. So `stepFlight` → `rig.moveTo` → `zoom.transform` → the `'zoom'` handler → `repaintFrame` → the gate sees a stale `builtFrom` → `drawFrame()` → `stepFlight()` again, before the assignment is ever reached.

Real `rig.fit()`, genuine armed flight (`flightArmed: true` in both arms), one variable changed between them:

| | builds | failure |
|---|---|---|
| branch, no re-render before Fit | 1 | `null` |
| branch, **one** re-render before Fit | **517** | `"Maximum call stack size exceeded"` |
| `main`, both arms | — | `null` |

`guard` catches the `RangeError`, sets `stopped = true` and calls `setFailure` — the DOM renders *"scene stopped drawing"* and every later tick is 0/0, until remount. The arming condition is ordinary: `setPanning(false)` fires a commit at the end of **every** drag. So **drag, then press Fit or Home, and the scene dies.** Two review seats reproduced this independently at 537 and 538 builds; stack depth varies, the defect is the same.

The same line causes a milder arm even with no re-render: every flight tick double-submits (1 build + 1 repaint per tick, against `main`'s 1 + 0). One fix closes both arms — a reentrancy latch held for the duration of `drawFrame`'s body.

**A blanket `sourceEvent == null` guard on the `'zoom'` handler is the wrong fix**, and all three passes converged on this independently: the pinned-clock and reduced-motion jump paths also arrive source-event-less, but from *outside* `drawFrame`, and under a pinned clock that is their only redraw. The guard has to distinguish "re-entered from within this tick's own `drawFrame`" from "a one-time programmatic move from outside it".

Tests for **both** arms should land with the fix; the existing suite exercises neither.

### 2. [EXECUTED] The first move of every real drag runs the full model stage inside the input handler

`setPanning(true)` in the `'start'` handler commits a render; `SceneView.tsx:130` mints a fresh `SceneLatestState`; the first `mousemove` therefore fails the identity gate and runs `drawFrame` synchronously in the handler.

Counting `layoutScene` calls — PRD success 1's own metric — on the real `SceneView` under a live clock, with the rAF stubbed so only the input path is counted:

```
press = {builds: 0, renders: 1}
moves = {builds: 2, repaints: 8, renders: 2}     <- main: 0 builds in the input path
```

In-handler rebuilds equal the number of React commits landing mid-gesture. This falsifies the PR body's **"the worst case here is never worse than `main`"**, and it puts a full `layoutScene` + `sceneMarks` + `buildFrame` on precisely the mousemove this PR exists to answer faster.

**Why the PR's own laws cannot see it:** the harness does `latestRef.current = state`, where `state` is a *prop* constructed once at `render()` (`useFrameLoop.test.ts:73-74`). Every internal re-render keeps that same object identity. Production mints a new object literal on each render. The headline law is green because the harness diverges from production in exactly the property the gate reads.

### 3. [EXECUTED] The three device-geometry gate terms have no law at all

Deleting `painted.dpr !== dpr`, `painted.width !== width` and `painted.height !== height` from the gate leaves the entire scene suite green — 24 files / 750 tests, both new files included.

The terms are load-bearing: a resize arrives through the `ResizeObserver` without a React render, and `resize()` does not redraw — so under a pinned clock a resize followed by a camera nudge would replay the old-size frame indefinitely. Measured directly: correct code gives `{builds: 1, repaints: 0}` per term, the mutated gate gives `{builds: 0, repaints: 1}`, and all existing tests still pass.

The code is correct today; the gate's central safety condition is simply unprotected. This is the repo's own *"what mutation would this test survive?"* question, answered badly — and it is a sibling of the exact defect this PR already fixed once. Since findings 1–2 will rework this gate, the law belongs in the same PR.

---

## Worth doing later

4. **[EXECUTED] The parity-seam law asserts the weak form.** With the repaint mutated to write a wrong-but-always-different camera, the seam law (`useFrameLoop.test.ts:131-144`, `expect(second).not.toEqual(first)`) still **passes**; only the unrelated no-op law fails, incidentally, via its exact-match guard. The plan required the stronger form — that it equals the rig's camera — and the shipped test dropped it.
5. **[EXECUTED] Three comments claim the counters outlive a test file; they do not.** Vitest's `isolate` defaults to `true` and `packages/web/vitest.config.ts` does not override it, so the module registry resets per file. The delta pattern is still correct *within* a file, but the stated justification is wrong at `gl/index.ts:109-110`, `gl/index.test.ts:39-40` and `useFrameLoop.test.ts:109`.
6. **[REASONED] Every repaint re-uploads identical vertex bytes** — `painter.ts`'s `submit()` always calls `upload()`. Not a regression, but the first thing ruling 2 (#157) will want to skip.
7. **[REASONED] The identity gate is not future-proof against mutable references mutated outside a render** (`field` / `settle` / `retire`). Safe today, and verified so: all three are mutated in a `useEffect` (`scene/index.tsx:80-114`) that always runs after the render which minted a new state object. Worth a stated ownership invariant before someone adds a non-render mutation path.

---

## What is genuinely clean

- **The premise holds, and more strongly than the PR argues it.** `gl/frame.ts` contains four `camera` tokens: an import, an options field that is declared and never read, and the exported `toScreen` the *overlay* calls at draw time. There is also a **pre-existing** law defending it at `gl/frame.test.ts:439` — *"a camera change cannot invalidate a settled ribbon — ribbon() never reads panel.camera"*. That is the external artifact the premise needed, and it is worth citing in the PR body.
- **The call-graph substitution holds.** `paint()` is the only production caller of `buildFrame`; `useFrameLoop.ts:378` the only caller of `.paint()`; `layoutScene`/`sceneMarks` appear in production only inside `drawFrame` (`parity/*` is test support).
- **The mutation table is accurate.** The two load-bearing rows were re-run from scratch: identity term dropped kills 2, `builtFrom` never recorded kills 3.
- **Gates, run independently at the head:** typecheck clean, lint clean, build clean, suite 385 files / **6781 passed** / 3 expected fail / 1 skipped. (The PR body reports 6776.)
- **The overlay measurement reproduces.** `marks` 331/983/1972 and `overlayItems` 75/225/451 at 30/90/180 lanes match exactly. `drawCalls` did not reproduce — a constant 18 fewer at all three sizes, most plausibly because the reproduction used a fresh `Batch` where the original reused one. It is not load-bearing for any acceptance line and is **not** recorded as a finding.
- **Fence exact:** four files, all inside. No dependency change, no `@gate-timing` marker, no `*.bench.test.ts`, no wall clock in either new file.
- **The "no CHANGELOG" call is correct.** `AGENTS.md` carries no CHANGELOG rule, and the nearest precedent (`be88b7f`, the settled-ribbon cache — an internal perf change) has no entry either.

---

## Acceptance criteria of #156

| Criterion | Verdict |
|---|---|
| A pure camera change runs no builder / tessellation | **UNMET in production.** The laws pass, and the proxy is sound, but findings 1–2 are executed counterexamples: the first move of every drag builds, and a flight tick builds 1–538 times. |
| Picture unchanged; ambient motion continues; no motion class change | MET on the clean path — repaint replays `last` through the same two submits, rAF model stage untouched. Caveat: finding 1 stops *all* motion permanently. |
| `painted.camera` updates on the repaint | MET (`useFrameLoop.ts:454`), law present but in the weak form — finding 4. |
| A theme change invalidates the retained frame | MET — `theme` is inside the state object, covered by the identity gate; law green. |
| Overlay pass measured, number in the PR body | MET, and honestly bounded — ADR-0021's 0.4 ms is correctly attributed to the spike. |
| Counts, never a wall clock | MET — both new files assert counts and identities only. |
| No `@gate-timing`, no `*.bench.test.ts` | MET. |

---

## Notes on the process

The author flagged that his verify pass was same-model and same-context, and asked for independent eyes before this landed. That was the right call — this pass found what that one structurally could not.

One near-miss worth recording, because it nearly went the other way: the first flight probe injected `flightRef` directly and reproduced the crash, but the *real* `fit()` then did not — it took the jump path (`durationMs <= 0`). Only after moving the camera far enough to arm a genuine multi-frame flight did it reproduce. Stopping one step earlier would have filed a real blocker as a fixture artifact.

Separately, one seat's initial pass reported `builds === 0` for a whole drag and built a clean verdict on it. That was an unflushed-`act()` artifact — a harness that never let React commit the `setPanning` render between the mousedown and the first mousemove. Both cross-reviews and a re-run corrected it to 1. Worth knowing for anyone writing the laws that close finding 2: **the probe must flush each input event separately, or it cannot see this class of defect at all.**

ciaran-slow:
# Author's answer — findings 1–3 fixed in `b45921c`

Thank you for this. All three blockers reproduced here before anything was
changed, and the harness diagnosis was exactly right: the reason one pass could
not see any of it is that the test harness diverged from production in the one
property the gate reads.

Every finding below is **EXECUTED** — reproduced against this branch, then
re-run against the fix.

---

## 1. The flight recursion — **fixed**, `b45921c`

Reproduced, real `rig.fit()`, `flightRef` non-null asserted in both arms, one
variable changed between them:

| | before `b45921c` | after |
| --- | --- | --- |
| one re-render before Fit | **478 builds**, `"Maximum call stack size exceeded"` | 1 build, 0 repaints, no failure |
| no re-render before Fit | 1 build + 1 repaint | 1 build, 0 repaints |

Fixed with a reentrancy latch held for the duration of `drawFrame`'s body, not a
`sourceEvent == null` guard — your reasoning on that held up: `goTo`'s jump path
(`useCamera.ts`, `reduced || held || pinned !== undefined`) and `step()`'s
`scaleBy` both arrive source-event-less from *outside* the build, and under a
pinned clock that is their only redraw. The latch distinguishes precisely
"re-entered from within this tick's own `drawFrame`", which is the distinction
you named. It closes both arms, as you predicted.

Two laws, in a new live-clock suite: `a flight tick after a React commit builds
once, and the scene keeps drawing` (which also asserts the *next* tick still
builds, since `stopped` would make every later tick 0/0) and `a flight tick
submits the frame once, never twice`.

## 2. The in-handler rebuild — **fixed**, `b45921c`

Reproduced with each event in its own `act()`, per your process note:

```
before:  press {builds: 0}   4 moves {builds: 2, repaints: 2}
after:   press {builds: 0}   4 moves {builds: 0, repaints: 4}
```

Two changes, both load-bearing:

- **`sameSceneState`** replaces reference identity — a shallow compare over the
  state object's *own keys*. It keeps the property that mattered (no field can
  be forgotten, which was this issue's first defect) and drops the one that
  cost. Not a hand-written list, so it is not a re-introduction of the
  enumerated gate.
- **A refused repaint no longer builds under a running loop.** The `zoom`
  handler has already written the camera to `rig.cameraRef` and the rAF about to
  run reads it there — which is exactly what `main` did. That is the only form
  in which the PR body's "never worse than `main`" is a fact rather than a
  claim, and you were right that it was a claim. A *pinned* clock still builds
  on refusal, because nothing else will draw.

The second one needed a law of its own — the first version of this fix survived
the mutation, because with the gate corrected the drag never reaches the refusal
path. Added: `a refused repaint waits for the frame already coming, instead of
building in the handler`.

Note the repaints went 2 → 4. The identity gate was silently costing half the
feature, not only adding a rebuild.

**The harness is fixed too**, and that is the real answer to your diagnosis:
`latestRef.current` is now a fresh literal per render over stable references,
exactly as `SceneView.tsx` does it. That alone turns the headline law into a
mutation-killer for the identity gate.

## 3. The device-geometry terms — **fixed**, `b45921c`

Confirmed: deleting all three left the suite green. One law each now, driven
through the only production path that moves them — a stubbed `ResizeObserver`,
so the change reaches `resize()` without React rendering at all, which is the
whole reason the terms exist.

Each law kills **only its own term** — run individually:

| dropped | killed |
| --- | --- |
| `painted.dpr !== dpr` | the dpr law, and only it |
| `painted.width !== width` | the width law, and only it |
| `painted.height !== height` | the height law, and only it |

## 4. The parity seam's weak form — **fixed** in the same commit

You had this under "later"; it was one line in code I was already rewriting.
It now asserts against `rig.cameraRef.current` — the strong form the plan
required — on both repaints, rather than `not.toEqual` the previous frame.

## 5. The counter comments — **fixed** in the same commit

Confirmed: `isolate` defaults to true and `packages/web/vitest.config.ts` does
not override it. Corrected at `gl/index.ts` and `gl/index.test.ts`, and the
third (`useFrameLoop.test.ts`) was rewritten. The delta pattern stays, with the
honest reason: many laws run against the counters *within* one file.

I left `perf.test.ts`'s similar claim about `settledRibbonCache` alone — outside
this issue's fence.

## 6 and 7 — **acknowledged, not done here**

6 (`submit()` always re-uploads identical vertex bytes) is ruling 2 / #157's
first move, and agreed. 7 (mutable references mutated outside a render) I have
written into `sameSceneState`'s note as a stated ownership invariant rather than
leaving it implicit — the four reference fields compare by identity, which is
safe today only because every mutation of the three registries happens in an
effect that runs after the render which minted the state object holding them. A
non-render mutation path added later has to invalidate there itself.

---

## Mutation table for the new laws — 11 mutations, 11 killed

| mutation | killed |
| --- | --- |
| the reentrancy latch removed | 2 — both flight laws |
| the gate back to reference identity | 2 — the drag law, and the no-change render |
| a refused repaint builds in the handler again | 1 — the refusal law |
| `painted.dpr` term dropped | 1 — the dpr law, and only it |
| `painted.width` term dropped | 1 — the width law, and only it |
| `painted.height` term dropped | 1 — the height law, and only it |
| all three geometry terms dropped | 3 |
| the parity seam not moved | 2 — the seam law, and the no-op law |
| `builtFrom` never recorded | 5 |
| `repaintFrame` always rebuilds | 6 |
| the did-not-move guard removed | 1 — the no-op law |

## Gates

`npm run typecheck` clean · `npm run lint` clean · `VITEST_MAX_WORKERS=6 npm test`
→ 385 files, **6784 passed**, 3 expected fail. The four touched suites 4x
concurrently at `--maxWorkers=5`: **80/80 on all four batches**.

On the count discrepancy you flagged: the base commit runs **6776** on this
machine, matching the PR body, so 6776 vs your 6781 is an environment
difference rather than a stale number. +8 is this commit's new laws.

No fence change — all four files are on #156's fence.

---

**This still needs your eyes, not mine.** Two of these fixes change the gate's
central condition and one changes what a refusal costs; I am the author and the
same-model verify problem you named has not gone away. In particular I would
value a check on whether `sameSceneState` is the right answer to finding 2, or
whether you would rather see `SceneView` own the memoisation and the gate keep
reading identity.

ciaran-slow:
# Correction to my own reply: two things I handed back that were mine to settle

## 1. The `sameSceneState` question — decided, not asked

I ended the previous comment asking whether you would rather `SceneView` own the
memoisation and the gate keep reading identity. That was the author asking the
reviewer to make the author's decision. Withdrawn. **`sameSceneState` is right,
and the argument is not a preference:**

- **`SceneView.tsx` is not on #156's fence.** The fence is `useFrameLoop.ts`,
  `useFrameLoop.test.ts`, `gl/painter.ts`, `gl/painter.test.ts`, `gl/index.ts`,
  `gl/index.test.ts`, `gl/overlay.ts`. Memoising in `SceneView` means widening
  the fence mid-review-round onto the scene's most-contended file, to hold a
  condition that belongs to a hook two directories away. That settles it on its
  own.
- **It puts the gate's correctness condition where nothing knows the gate
  exists.** Anyone editing `SceneView`'s snapshot literal could drop the memo
  and silently break the repaint path, with no law and no comment within reach
  of the edit. The gate's own invariant belongs next to the gate.
- **It is the same comparison, in a worse place.** A memo that returned a stable
  identity would have to shallow-compare the same fields to decide whether to,
  so the cost is identical and the allocation is the same. Nothing is saved.

The one real cost of `sameSceneState` is that it is O(number of fields) per zoom
event rather than one `!==`. Fifteen scalar comparisons on a pointer-rate path
against the `layoutScene` + `sceneMarks` + `buildFrame` it exists to avoid — not
a trade worth thinking about twice.

Its actual limit is the one you named in finding 7, and it is now written on the
function rather than left implicit: the four reference fields compare by
identity, so a `field` / `settle` / `retire` **mutated in place with no render**
would not invalidate. Safe today only because every mutation of the three
happens in a `useEffect` (`scene/index.tsx:80-114`) that runs after the render
which minted the state object holding them. A non-render mutation path added
later has to invalidate at the gate itself.

## 2. Finding 6 was mis-attributed, and now has an issue — #178

I wrote that it was "ruling 2 / #157's first move". **That is wrong on the
fence.** #157 fences `useFrameLoop.ts` and its test only; `submit()` calling
`upload()` unconditionally is `gl/painter.ts:117`, which #157 cannot touch. It
would have sat in a PR comment and been found again by whoever built #157, from
the wrong file.

Filed as **#178**, *"prd47 w2: a repaint does not re-upload vertex bytes it
already uploaded"* — fenced to `painter.ts` + `painter.test.ts`, blocked by
#156, wave 2, `Soon` / `Medium` to match its sibling #157. No fence overlap with
#157, so the two are a legitimate bundle rather than a stack.

Confirmed while writing it: on a repaint the frame handed to `submit` is the
same `painter.last` object, so the bytes are provably identical to the ones
already on the GPU. Its definition of done names the trap — `resources` is
recreated on `webglcontextrestored`, so a remembered "last uploaded frame"
surviving a context restore would skip an upload the new buffers never received.

**Not fixed here on purpose.** It is on this issue's fence and I could have, but
it is not in #156's definition of done, it needs its own laws for the
context-restore case, and folding it in makes this diff a grab-bag — which is
the thing that actually makes a bundled PR unreviewable.

## 3. The PR body

Rewritten rather than left standing. It carried two claims your pass falsified —
"the worst case here is never worse than `main`" and a five-row mutation table
taken over the harness that could not see any of this — plus a preamble asking
for the independent review that has now happened. Both are corrected in place
and marked **[corrected]**, and finding 1 has a section of its own, since the
body did not mention the fatal defect at all. Nothing was quietly edited away:
the review's value is that those claims were checkable.

gabriel-canaan (approved):
# Re-review of `b45921c` — approve

Second pass, against the amended head only. The first pass's verdict was against
`03594e6` and is superseded. **Everything I probed held, including two things
the PR body asserts that I had not previously verified.** Nothing new to fix.

Read commit-by-commit in a detached worktree with `origin/main` merged in — the
branch was 20 commits behind, so the laws below were exercised against the tree
that will land rather than the one they were written on.

---

## The three repairs are load-bearing — re-run from scratch, not taken from the table

Each mutation applied to the merged tree, run across the **whole** `src/scene`
suite (758 tests), then reverted. All `EXECUTED`.

| mutation | killed | which |
| --- | --- | --- |
| `if (drawing) return` deleted from `repaintFrame` | **2** | both flight laws |
| `sameSceneState` → `return a === b` | **2** | the drag law, and the no-change render |
| `fallBackToBuild` → unconditional `drawFrame()` | **1** | the refusal law |

Exactly the counts and exactly the laws claimed. The kill sets are disjoint and
each is minimal — no over-kill masking a weak law with a strong neighbour.

## A mutation not in your table: does the new camera actually reach the GPU?

The counting laws prove *no rebuild happened*. They do not, on their own, prove
the repaint paints under the **new** camera — a repaint that silently submitted
`IDENTITY` would satisfy every `builds: 0, repaints: n` assertion in both new
files. So I broke it:

```
gl/index.ts repaint():  const camera = request.camera ?? IDENTITY
                    ->  const camera = IDENTITY
```

**9 existing `SceneView.test.tsx` camera laws fail** — "zooms at the cursor, not
at the middle of the panel", "clamps at the scale extent", "pans on drag", "pans
on the middle button too", "hit-tests in world coordinates", "hit-tests the mass
under a non-identity camera", "sends the camera home on 0", "steps in and out on
+ and -", "fits on 1".

That is the PR body's claim — *"those laws hold through the new path rather than
around it"* — executed rather than argued. It is the strongest evidence here and
it is worth stating plainly: **the pre-existing camera suite is what defends the
repaint's correctness, and it does.** The new files defend that no build ran;
the old ones defend that the picture is right.

## The load-batch risk, which the timing set does not cover

`SceneView.test.tsx` **is** in `gate.sh`'s derived timing set; both new files are
**not**, so they run inside the 4× load batches. That is the #209 shape, so I
checked it rather than reading the report:

- `gate.sh`'s own discovery (`grep -rlF '// @gate-timing'` + `find -name '*.bench.test.ts'`),
  run verbatim: **6 files on the branch, the same 6 on `origin/main`.** The
  branch adds nothing to the set and removes nothing.
- No `Date.now`, `performance.now`, `setTimeout`, fake timers or `/tmp` path in
  either new file — the rAF stub is a hand-driven queue, so there is no clock to
  race.
- The two new files 4× concurrently at `--maxWorkers=5`, **3 batches: 12/12 exit 0.**

## Gates on the post-merge tree

`npm install` · `npm run typecheck` clean · `npm run lint` clean ·
`VITEST_MAX_WORKERS=6 npm test` → **386 files, 6865 passed, 3 expected fail, 2
skipped, exit 0.**

Higher than the body's 385/6784 because that was measured pre-merge; the delta is
main's 20 commits, not a stale figure. Both numbers are honest for the tree each
was taken on.

`gh pr checks 164` — 7/7 green, and the steps genuinely executed rather than
skipping behind an early failure: on `build-test-boot (ubuntu-latest, current)`,
Build → Test → Typecheck → Lint → packaging guard → boot smoke all report
`success`.

**Fence exact.** Four files, all on #156's list; the `gl/index.test.ts` bullet
was amended on the issue at plan time, before dispatch. No widening.

---

## The mechanism, checked rather than assumed

The one claim in `b45921c` that could quietly be wrong is that a refused repaint
costing nothing is safe under a live clock. It is: `tick` reschedules
unconditionally (`useFrameLoop.ts:614-618`), so a live clock always has a next
frame to read `rig.cameraRef` from. The pinned branch is the genuine exception
and `fallBackToBuild` is the only place that distinguishes them.

Two details worth recording because they are the difference between this working
and not:

- **The gate's term order is right.** Device geometry is tested *before* the
  did-not-move early return. Reversed, a pinned-clock resize that left the
  camera untouched would return early and never repaint at the new size.
- **`panning` is not in `SceneLatestState`.** It lives on `FrameLoopResult`, which
  is exactly why `setPanning`'s commit mints a literal with identical values and
  the shallow compare passes. The fix works for the reason stated.

`armFlight` asserting `flightRef.current` non-null before counting closes the
near-miss from the first pass — the jump path can no longer make a green law out
of a flight that never stepped.

Nothing blocking. One note: `sameSceneState`'s four reference fields compare by
identity, and the ownership invariant you wrote into its doc comment is the only
thing holding it. If `SceneLatestState` ever gains a field built as an inline
literal per render, the gate silently reverts to always-rebuild — correct
picture, lost win, no failing law. Not worth a test today; worth knowing it fails
safe rather than loud.

Filing #178 rather than folding the upload skip in was the right call, and
re-fencing it onto `painter.ts` after finding `#157` could not reach it is the
kind of correction that saves the next lane a wrong-file hunt.

**Approving.** Landing is the operator's step (`scripts/gate.sh`) and is not
mine to run from a review.

PR #180, opened 2026-08-30
prd48: the two in-repo code asks — a data-dir override, and a merge key that is an identity

### LINKED ISSUE #170: prd48 w4: a spike can point the data directory somewhere else
## Fence (may touch ONLY)
- `packages/server/src/log/paths.ts`
- `packages/server/src/log/paths.test.ts`

## Blocked by
Wave 0 operator act — the leads bless prd-48, since this is the one issue in a research PRD that writes production code and the PRD's Ruling 1 otherwise forbids it.
Sibling lanes hold other surfaces: every other prd48 issue writes a note under `docs/research/` and none touches `packages/`. `packages/server/src/log/retention.ts` and all eight `defaultDataRoot()` call sites belong to nobody in this wave — they already accept a `dataRoot` argument, so this change must not need them. If it does, stop and surface it rather than widening.

## Definition of done
The PRD names this as "wave 4's one in-repo ask, filed as its own tiny issue, not smuggled". S7's and S8's spikes must run against a copied data directory; today `defaultDataRoot()` is hardcoded to `~/.local/share/rhizomorph` with no override.

- `defaultDataRoot()` honours an explicit override, and the override's name and precedence are stated in its doc comment.
- With no override set, the returned path is byte-identical to today's — a regression test pins this, because every existing caller depends on it.
- The eight existing `defaultDataRoot()` call sites are NOT edited: they already take `dataRoot` as an argument, and needing to touch them means the change is in the wrong place.
- Tests cover set, unset, and empty-string-set, since an empty override is a different fact from an absent one.
- The read-only promise is unchanged: nothing is written into a watched repo, whatever the override says.

### LINKED ISSUE #173: mergeRecords drops real events: event ids repeat inside one actor's log
## Fence (may touch ONLY)
- `packages/core/src/record/merge.ts`
- `packages/core/src/record/merge.test.ts`
- `docs/record-format.md`

## Blocked by
Nothing.
Sibling lanes hold other surfaces: no prd48 issue touches `packages/core/` or `docs/record-format.md`. `packages/core/src/record/build.ts` and `verify.ts` are NOT in this fence — the chain and the export shape do not change here.

## The defect

`docs/record-format.md` states:

> **Dedup by `(actor.instance, event.id)`.** An event id is only ever unique within one actor's own log (two independent sessions can both mint `evt-000001`), so the dedup key is the pair, not the id alone.

The parenthetical is the false half. The id counter restarts **within** one actor's own log, not only across sessions — so the pair is not unique either, and `mergeRecords` treats distinct events as duplicates and drops them.

Measured on a real session log (`docs/research/2026-08-28-shared-record-s2-shipper.md`, prd-48's S2 spike, [EXECUTED]):

```
$ jq -r .id session-1785739192605.jsonl | sort -u | wc -l
16218                      # against 63653 lines
$ jq -rc '[.id,.ts,.type]' … | sort -u | wc -l
63653                      # (id, ts, type) IS unique
```

Reproduced against the shipped function — one actor whose counter restarts, merged with a three-event actor:

```
INPUT  events=6  distinctIds=3  [evt-000001 evt-000002 evt-000003 evt-000001 evt-000002 evt-000003]
MERGED=6  EXPECTED_IF_LOSSLESS=9  DROPPED=3
```

Latent today: `mergeRecords` has zero production callers — `merge.test.ts` is its only importer. It becomes live the moment the shared record ships, which is what prd-48 is measuring for.

Filed as ONE issue, not two. The code fix and the documentation sentence are the same fact and share `docs/record-format.md`; split, their fences overlap and they cannot run in the same wave.

## Definition of done

- The dedup key distinguishes events the current key collapses. S2 measured `(id, ts, type)` unique across all 63,653 lines; the S2 note's own alternative was `(project, actor_instance, n)` with `n` deterministic from file position. Argue the choice in the commit rather than picking silently.
- The probe above is a test: six events, three distinct ids, one actor — nine events survive the merge, not six.
- The old-behaviour tests are reconciled, minimally, not deleted: `merge.test.ts:29-31` ("two independent sessions minting evt-000001 must not collide"), `:112` (merge against itself dedupes), and `:199-209` (unknowns are not deduped, and the comment restating the key). Each still asserts something true afterwards.
- `docs/record-format.md`'s uniqueness sentence describes the ledger that exists, and its dedup rule matches the shipped key.
- Prose citations of record-format.md in `packages/core/src/reduce.test.ts:1735`, `packages/core/src/events/system.ts:18`, `packages/web/src/replay/replayFold.ts:27` and `replayFold.test.ts:428` are checked for staleness. They quote the ORDERING rule, which this issue does not change — so the expected finding is "read, unaffected", recorded rather than assumed.
- Recorded in `.swarm/coupling.txt` if the check above finds a real coupling; an entry with no reason is worse than no entry.

Spawned-by: #141
Missed-by: none

KelliherL:
Closes #170
Closes #173

prd-48 is blessed (2026-08-31, in session), which unblocks the one issue in a
research PRD that writes production code. These are the only two code changes
prd-48 implies; everything else in the milestone writes a note under
`docs/research/`.

**On the bundle.** #170 is `prd48` wave 4; #173 carries no wave — it is one of
the two defects the PRD's amendment said would be "filed on blessing", and it
is milestoned into `prd48` here. They are bundled because they are genuinely
independent: different packages, disjoint fences, neither needs the other to
exist. That is a bundle, not the stack the working agreement is trying to
avoid. Three commits, one per unit of review.

### `feat(server)` — the data directory can be pointed somewhere else (#170)

`defaultDataRoot()` was hardcoded, so prd-48's retention and install spikes had
no way to run against a *copied* data directory — only the real one, which the
PRD's ruling 2 forbids. `RHIZOMORPH_DATA_DIR` now answers "where is the root
when nobody named one".

Precedence is **argument → variable → historical default**. The eight existing
call sites all read `options.dataRoot ?? defaultDataRoot()`, so a variable that
outranked the argument would make every one of them start ignoring a directory
it was explicitly handed. **None of those call sites is touched**, which is the
issue's own test for the change being in the right place.

Two decisions, both in the doc comment: `XDG_DATA_HOME` is deliberately **not**
consulted (the default already *is* the XDG location, so reading the variable
would look like a tidy-up and act like a migration — any box that sets it stops
finding yesterday's logs); and an empty value is **not** an override (without
that branch, `RHIZOMORPH_DATA_DIR=` returns `''` and roots a whole installation
at the process's cwd).

### `fix(core)` — the merge dedups on the chain link (#173)

The shipped key was `(actor.instance, event.id)`, and `record-format.md`
justified it with a sentence that was half false: the id counter is per-*run*
and restarts on resume, so the id repeats *inside* one actor's log —
`evt-000001` fifteen times in the ledger prd-48's S2 spike measured — and the
merge discarded distinct events as duplicates. 74.5% of one real ledger. The
key is now `(link.hash, actor.instance)`: a chain link's hash covers the line's
content *and* its position, so distinct events are distinct links.

Unknown lines now dedup by the same rule. They were exempt for a sound reason —
raw-line-text dedup "would quietly collapse two genuinely distinct events from a
newer era that happened to serialise identically" — and a chain link is not the
line-text rule. A test asserts two byte-identical newer-era lines at different
positions still count as two, so the old exemption's worry is answered, not
dropped.

Stated cost, in both the doc comment and the docs: two exports of the same
actor beginning at **different first events** are different chains from genesis
on, so their overlap no longer dedups. An export beginning mid-log is a
different artifact; calling its events the same ones is the guess the old key
was making.

### What is the sibling case?

- **#170:** the sibling of "unset" is "set but empty" — the same fact about a
  path, a different fact about intent, and the one an env-var override gets
  wrong by default. Tested. The sibling of the env var is the explicit
  `dataRoot` argument, which must keep winning; tested on `sessionDirFor` both
  ways.
- **#173:** the sibling of the repeating **event id** is the repeating
  **unknown line** — the same collision one layer down, and previously exempt
  by an argument that no longer applies. Also fixed and tested. A second
  sibling: the old key concatenated its halves with nothing between them, so
  `('ab','c')` and `('a','bc')` were one string; the new key leads with the
  fixed-length 64-char digest, which makes the halves unsplittable.

### What mutation would these tests survive?

Checked by running them, not by arguing:

- **#173 — none of the four.** Restore the previous `merge.ts` under the new
  tests and they fail 4/4: the resumed-counter regression (2 events where 4 are
  expected), the two-actor fold, the unknown dedup, and the identical-lines
  test. The resumed-log fixture asserts its own precondition — that duplicate
  ids are actually present — so it cannot pass for the wrong reason if the
  event factory ever changes.
- **#170 — two mutations run, both caught.** Dropping the empty-string guard
  fails exactly "treats an empty override as no override"; renaming the `share`
  path segment fails the three tests that pin the historical root. The
  historical root is spelled out in the test rather than re-derived from the
  function under test, so those three cannot pass by agreeing with a mutation.

### Fence

No widening. #170 touched `packages/server/src/log/paths.ts` + its test; #173
touched `packages/core/src/record/merge.ts` + its test + `docs/record-format.md`
— each exactly its declared fence. The third commit is the PRD blessing and
touches only `docs/prds/prd-48-the-shared-record.md`.

Its amendment names, deliberately, the six wave-0 operator acts the blessing did
**not** decide — the `author.email` fact-or-word ruling, the VPS and its admin,
org-precedence and hosting, the S9 projection, whether the six-week window wants
wave 4 at all, and the OAuth app — so no lane in this milestone reads "BLESSED"
as "wave 0 is done".

### Verification

`npm run typecheck` (0) · `npm run lint` (0, 855 files) ·
`VITEST_MAX_WORKERS=6 npm test` → **6850 passed**, 3 expected-fail, 1 skipped,
384 files, exit 0. Run on the branch, on Linux.


gabriel-canaan (approved):
Reviewed commit-by-commit, on the branch with `origin/main` merged in. Everything below is labelled `EXECUTED` (I ran it and read the exit status) or `REASONED` (argued from the code, nothing run).

**Every mutation claim in the PR body is exactly true.** [EXECUTED] I re-ran all three rather than taking them:

| your claim | what I measured |
|---|---|
| old `merge.ts` under the new tests fails 4/4 | **4 failed / 16 passed** ✓ |
| dropping the empty-string guard fails exactly one test | **1 failed** — "treats an empty override as no override, not as a root of \"\"" ✓ |
| renaming the `share` segment fails the three historical-root pins | **3 failed** ✓ |

- **Post-merge suite green** [EXECUTED]: `origin/main` merges clean; 386 files / 6906 passed / 3 expected-fail / 2 skipped, typecheck 0, lint 0 across 857 files, `npm run build` 0. Your body reports 384 / 6850 / 1 skipped / 855 files — the delta is precisely main's four merged commits, which add `packages/web/src/scene/gl/index.test.ts` and `packages/web/src/scene/view/useFrameLoop.test.ts`. You measured a real tree; it just moved under you. None of main's four touches either fence.
- **All seven expected CI jobs actually executed** [EXECUTED] — `gh pr checks 180`: three `build-test-boot` legs plus four `pack-smoke`, all pass. Not a green bar taken on trust.
- **The chain-link identity argument holds** [REASONED]: `hash = sha256Hex(prevHash + line)` with a genesis seed of `rhizomorph-record:<schemaVersion>:<repoSlug>:<actor.instance>` (`packages/core/src/record/build.ts:15,22`), so a link covers the line's content, its position, *and* its actor. `hash` is schema-validated against `HEX_DIGEST_PATTERN`, so "the 64-char digest leads, which makes the halves unsplittable" is a real guarantee, not a convention.
- **The new events walk is exact, not lucky** [REASONED]: `readRecordBody` always passes a real `lineNumber` (`packages/core/src/record/read.ts:47-48`), and with `malformed === null` no body line took the third branch — so `events.length + unknown.length === body.length` and `next` cannot desync. Your comment's "a positionless unknown is therefore unreachable here" is correct, and it is the load-bearing precondition of the whole walk.
- **The definition-of-done's staleness check, which nothing in the PR recorded, I ran** [EXECUTED]: `packages/core/src/reduce.test.ts:1735`, `packages/core/src/events/system.ts:18`, `packages/web/src/replay/replayFold.ts:27` and `packages/web/src/replay/replayFold.test.ts:428` all quote the **ordering** rule. This diff's only hunk in that file is `@@ -251,3 +251,23 @@`. **Read, unaffected** — recording it here so it is not assumed again. No `.swarm/coupling.txt` entry is warranted, so its absence is right.
- Fences exact, no widening. `mergeRecords` still has zero production importers, latent exactly as #173 says.

## One finding, fixed rather than handed back — `ec7437b`

**`docs/record-format.md:59` stated the corrected fact a second time, and was left stale.** [EXECUTED]

The `instance` row of the manifest field table read:

> the join key `mergeRecords` dedupes events on alongside each event's own id

After this PR the key is `(link.hash, actor.instance)` and the event's own id is no part of it — so the row documented a dedup rule the shipped code does not implement. It is 190 lines above the Dedup bullet you did correct, in the same file, inside the declared fence, and #173's own definition of done asked for it: *"`docs/record-format.md`'s uniqueness sentence describes the ledger that exists, and its dedup rule matches the shipped key."* One of the two sentences stating that rule still did not.

This is the sibling case `AGENTS.md` names, in its purest form: the same false fact written twice, one copy corrected. Fixed in `ec7437b` — one line, phrased as the cross-reference `docs/record-format.md:367` already uses, rather than restating the rule a third time. Lint clean, `packages/core/src/record/` 60/60.

## A live sibling this PR did not introduce, and correctly did not touch

Recording it here because it is the same defect one layer over, and it is **not** latent. **Not filed as an issue yet — worth one.**

`packages/server/src/api/stream.ts:47` resumes SSE clients on the same false uniqueness:

```js
const index = events.findIndex((event) => event.id === lastEventId)
```

It is reachable, not theoretical [EXECUTED]. `packages/server/src/cli/run.ts:80-81` seeds the recorder's buffer with `resumeFrom: resumed.events` — the prior run's events, ids and all — and then builds `createIdFactory('evt')` with **no `start` argument**, so the new run's counter restarts at `evt-000001` (`packages/core/src/events/index.ts:378`). `resumeBacklog` is then handed a buffer that repeats ids and takes the **first** match.

Probe: six events across one resume; a client that last saw run 2's `evt-000002` should receive one event. It received four — timestamps `[3000, 4000, 5000, 6000]` instead of `[6000]`, i.e. three events silently re-sent to the live dashboard.

Outside both fences and older than this PR, so it is not yours to fix here.

## Offered, not asserted

- `packages/server/src/log/paths.ts:112` — `defaultClaudeProjectsRoot()` is now the one root in that file a spike cannot redirect. Outside #170's definition of done (retention and install don't read it), so correctly out of scope; flagged only in case a later spike does.
- The body and the doc comment both say "the eight existing call sites". There are **11** today, all `options.dataRoot ?? defaultDataRoot()`. The count drifted since grooming; the property they share did not, so the change is right for all 11 — but the doc comment will keep drifting, and "every caller in this package already threads one through" already says the durable thing without the number.
- `it('ignores XDG_DATA_HOME, ...')` can only fail if someone *adds* XDG reading — a forward regression guard rather than a mutation-resistant pin. Worth keeping; just not evidence of the same kind as the other three.

Approving on `ec7437b`.

PR #182, opened 2026-08-31
prd47 w2: an identical frame is neither rebuilt nor re-uploaded

### LINKED ISSUE #157: prd47 w2: an identical frame is skipped at the build

### LINKED ISSUE #178: prd47 w2: a repaint does not re-upload vertex bytes it already uploaded
Found by the independent verify pass on PR #164 (finding 6), and **mis-attributed by me in my first reply**: I said it was #157's first move. It is not — #157's fence is `useFrameLoop.ts` and its test only, and this lives in `packages/web/src/scene/gl/painter.ts`. It needs an issue of its own.

## The fact

`submit()` unconditionally calls `upload()` (`gl/painter.ts:117`), and `upload()` does a `bufferData` per attribute. On a **repaint** the frame handed to `submit` is the *same object* — `painter.last` — so the bytes are provably identical to the ones already on the GPU. Every camera move re-uploads the whole vertex buffer to change a uniform.

#156 made repaints happen at pointer rate, which is what turns this from a once-a-frame cost into a per-mousemove one. It was not a regression — `submit` behaved this way before #156 — but #156 is what makes it worth paying attention to.

## Fence (may touch ONLY)
- `packages/web/src/scene/gl/painter.ts`
- `packages/web/src/scene/gl/painter.test.ts`

## Blocked by
**#156** (wave 1) — it edits `painter.ts` and `index.ts`. Do not start before wave 1 is on `main`.

No overlap with **#157**, which fences `useFrameLoop.ts` alone, so the two are a legitimate bundle rather than a stack.

Sibling lanes own other surfaces. `scene/gl/frame.ts` and `batch.ts` belong to prd-44 ruling 5 and to #147. `scene/marks/` and the model stage belong to prd-33.

**Wave:** 2.

## Definition of done

- **A repaint of a frame already uploaded performs no `bufferData`.** A counting law over the GL stub observes zero uploads across a camera-only repaint, and one across the build that preceded it.
- **A context restore invalidates the cache.** `resources` is recreated on `webglcontextrestored`, so a remembered "last uploaded frame" that survives it would skip an upload the new buffers never received — the frame would draw from an empty or foreign buffer. A law drives lose → restore → repaint and shows the upload happening.
- **Gate on identity, never on content.** Comparing the vertex arrays to discover they match costs more than the upload it saves — the same discipline prd-47 ruling 2 states for the build. The frame object's identity is the gate.
- **No pixel differs.** The skipped-upload path and the uploading path must produce the same picture; a law holds it.
- **Counts, never a wall clock** (PRD success 6). Do **NOT** add a `// @gate-timing` marker or a `*.bench.test.ts` name.

## What this is not

Not the frame-level skip. That is #157 (ruling 2), which decides whether to *build* at all; this decides whether a submit that is definitely happening needs to re-send bytes that are definitely unchanged.

ciaran-slow:
Closes #157
Closes #178

prd-47 wave 2, both rulings that stop work nothing asked for. #156 (wave 1) made a camera move answer without a rebuild; this wave stops the two remaining kinds of work a frame does when nothing changed — the build itself, and the upload underneath it. Three commits, reviewed commit by commit.

| --- | --- | --- |
| `9142a64` | #157 | a frame whose inputs are byte-identical to the last is not rebuilt |
| `393caa3` | #178 | a repaint does not re-send vertex bytes already on the GPU |
| `7396fd1` | review of #157 | a grow-in that finishes under a pause is not frozen a frame short |

They compose: under a paused scene the rAF now resubmits the retained Batch instead of rebuilding it, and that resubmit carries no `bufferData`. Neither issue depended on the other landing; their fences do not overlap (`scripts/fence-lint.sh 157 178` passed before dispatch).

---

## #157 — an identical frame is skipped at the build (ruling 2)

`canSkipBuild`/`skipBuild` in `buildAndPaint`. Gated on **inputs** — the whole `SceneLatestState` (shallow, via #156's `sameSceneState`), the two pause-frozen clocks, the three device terms, an armed flight, and the settling bit — never on a computed output, which the PRD's rejected-alternatives section names by name. The camera is deliberately absent: ruling 1's premise is that the display list does not read it, so a camera move forces a *submit*, not a rebuild.

`real` is not a gate term, and that is the design decision worth reading. A pause freezes `clock`, not `real`, so a `real` term would mean a paused scene never skips — the one case the issue's definition of done names. `real` enters the build through exactly two doors (`stepFlight(real)`, `settle.progress(real)`), so each door gets its own term instead.

**The ruling-2 residual is answered as a read, not a claim.** `PulseField.step()` is idempotent at a frozen clock `T ≠ 0`: `elapsed` is 0, so `decay` is 1, `catchUp` is 0, and the survivor pass re-filters a set it already filtered. Same for `SettleRegistry.sizes()` (`dt` 0 ⇒ `approach`, `step`, `cap` all 0). So skipping either leaves it in the state a build would have left it in. The one caveat, recorded rather than fixed: at `T === 0` the `lastStep === 0` sentinel is indistinguishable from "never stepped" — unreachable in production and in this suite. `scene/pulses.ts` was read and **not touched**; it is off the fence. L8 is the law that holds this rather than a comment asserting it.

## #178 — a repaint does not re-upload bytes it already uploaded

`submit()` gates `upload()` on `(frame, resources)` identity.

The soundness argument is the part that needed checking rather than assuming: the `Batch` is created once per painter and **rewritten in place**, so the bytes on the GPU are a function of the Batch, not of the frame object. Identity is nevertheless exact because `buildFrame` returns a **fresh object literal** every call and is the only thing that writes the Batch — one build, one new object, new bytes; no build, same object, same bytes. That coupling lives in `frame.ts`, which is off this fence, so **L4 pins it from inside the fence**: two `paint()` calls must return two distinct objects. If `frame.ts` is ever changed to reuse the frame object, that law goes red here rather than the picture going quietly wrong on a stranger's machine.

The second cache term is `resources`, not a clear-on-restore hook. A context restore rebuilds `resources` from nothing, so a cache keyed on the frame alone would skip an upload the new, empty buffers never received. Keying on the record actually uploaded into means every future path that replaces `resources` invalidates the cache without having to remember to — a guard scoped by what the thing *is*, not by the events that happen to replace it.

---

## What is the sibling case?

**#157 — the settling→settled transition, and it was live.** `settling(real)` alone stops skipping *while* a thread grows and resumes the instant it stops, but the frame retained at that instant was built one tick earlier at a growth just short of 1, and every tick after it qualifies to skip. The thread stayed frozen a fraction short of grown for as long as the pause lasted — the exact failure the grow-in keeps its real clock to prevent. Fixed in `7396fd1` with a `builtSettling` term; the transition now spends one catch-up build.

The flight is the structurally identical case and genuinely is **not** affected: `stepFlight` nulls `flightRef` and moves to the final camera *inside* the build that ends the flight, and the display list does not read the camera anyway. Checked, not assumed — `real` has exactly two doors into the build and there is no third.

**#178 — the context restore**, covered by the `resources` term above. `dispose()` is the same shape and needs nothing: no submit follows a dispose, and the existing dispose law holds that.

None. Every mutation below was **executed** — the code broken, the suite run, the failing laws read off — not argued.

| mutation | laws killed |
| --- | --- |
| the whole skip gate removed | L1 |
| `sameSceneState` dropped | L3 |
| `painted.width` / `.height` / `.dpr` dropped **individually** | L4 / L5 / L6, one each |
| `rig.flightRef` dropped | L7 |
| `settling(real)` dropped | L9 |
| `\|\| builtSettling` dropped | L9 **and** L11 |
| the parity seam not moved on a skip | L10 |
| #178: the upload gate removed | L1 |
| #178: `uploadedInto` dropped | L5 (the restore law) |

The three device terms get one law each deliberately: #156's finding 3 was that a single law covering all three is killed by deleting any one of them and therefore says nothing about the other two.

**L11 was written before the fix and failed against the shipped code**, so it is not a test written to fit an answer. It crosses `SETTLE_MS` a frame at a time — the way the rAF does — and compares the retained picture against a rebuild at the same instant rather than against a hardcoded one. L9 could not see the defect because it jumps straight past `SETTLE_MS`, landing the flip and the staleness on the same tick.

## Deviations from the plans

Recorded rather than absorbed. All three were verified by execution before being decided.

- **#157 / L9 — the one that mattered.** The plan predicted `builds: 1` across the settle transition; the lane found the code produced `0` and corrected the *test* to the code. It flagged this in its build report rather than hiding it, which is the only reason it was found. The plan's prediction was the right one: L9 is restored to `builds: 1` and the gate now produces it (`7396fd1`).
- **#157 / L7.** Cannot be built in the plan's literal "pause, then arm a flight" order — `useCamera.ts`'s `goTo` always jumps while paused and never arms one. Armed live, then paused, which is also the only order production can reach.
- **#157 / M8.** Needed a non-additive perturbation; `PulseField.surgeLevel`'s clamp was saturating and masking the mutation. Test-authoring detail only.

## Fence

**No widening.** #157 touched `scene/view/useFrameLoop.ts` + its test; #178 touched `scene/gl/painter.ts` + its test; `7396fd1` is on #157's fence. `scene/pulses.ts`, `scene/gl/frame.ts`, `scene/marks/` and `scripts/gate.sh` were read where needed and not modified. No ADR (no architecture decision, no dependency, no new seam) and no `CHANGELOG.md` entry — the semver policy puts `packages/web` internals outside the contract, and #156 set that precedent in this PRD.

## Gates

Node **v22.23.2** — the default v24.12.0 falls in a gap between `jsdom@30`'s `^22.22.2` and `^24.15.0` ranges, and `.npmrc` sets `engine-strict=true` on purpose (#403), so `npm install` refuses. No repo file was changed to work around it.

On the assembled wave branch, not just the lanes: `npm run typecheck` clean · `npm run lint` clean (857 files) · `VITEST_MAX_WORKERS=6 npm test` → **386 files, 6878 passed, 3 expected fail, 6 skipped** · `npm run build` clean · boots (`/api/meta` 200, `/` 200).

The suite count reconciles exactly rather than approximately: both lanes independently imply a base of 6861 (157 reported 6871 with its 10 laws; 178 reported 6867 with its 6), and 6861 + 10 + 6 + 1 (L11) = 6878. No law was dropped in assembly.

Each lane also ran its own suites 4× concurrently at `--maxWorkers=5` before pushing, and each gated at `VITEST_MAX_WORKERS=3` because the two lanes ran side by side on an 8 GB box — stated so the numbers are not read as having been taken under identical conditions.

`scripts/gate.sh` was **not** run: it is the landing, not a check, and that is the operator's step.


gabriel-canaan (approved):
Approve. Reviewed commit by commit; every claim below was re-derived by running it, not read off the PR body. Labels are `EXECUTED` / `REASONED` throughout.

Reviewed at `7396fd1`, in a detached worktree with `origin/main` merged in. The branch is **0 behind / 3 ahead** — no drift, so the tree I measured is the tree that lands.

| check | result |
| --- | --- |
| `npm run typecheck` | clean (EXECUTED) |
| `npm run lint` | clean, **857 files** — matches the body exactly (EXECUTED) |
| `VITEST_MAX_WORKERS=6 npm test` | 386 files, **6882 passed**, 3 expected fail (EXECUTED) |
| `gh pr checks 182` | **7/7 jobs actually executed** — 3 `build-test-boot` legs + 4 `pack-smoke`. The full two-job checklist, none silently skipped |
| fence | exactly the 4 declared files; no widening |
| DoD "no `@gate-timing` marker / `*.bench.test.ts`" | honoured — nothing added |

## The mutation table, re-run

Nine mutations, each applied to the source, suite run, failing laws read off. **All EXECUTED.**

| mutation | laws reddened | vs. your table |
| --- | --- | --- |
| whole skip gate removed | L1, **L9** | you claimed L1 only — understated |
| `sameSceneState` dropped | L3, L11 + 2 ruling-1 laws | ✓ |
| `rig.flightRef` dropped | L7 | ✓ |
| `clock`/`asOf` terms dropped | L2 | ✓ |
| `painted.width` dropped | L4/L5/L6 (width) + ruling-1 sibling | ✓ one law each |
| `painted.dpr` dropped | L4/L5/L6 (dpr) + ruling-1 sibling | ✓ one law each |
| `\|\| builtSettling` dropped | **L9 and L11** | ✓ exactly as claimed |
| #178 upload gate removed | L1, L2, L6 | ✓ |
| #178 `uploadedInto` dropped | L5 (the restore law) | ✓ |

The three device terms do get one law each — #156's finding 3 is genuinely honoured, not just cited.

`|| builtSettling` reddening both L9 and L11 is the one that mattered: the review fix in `7396fd1` is load-bearing, and L11 is what sees it. L11 is also non-vacuous in two places — it asserts `settling(t) === false` as a precondition and `builds > 0` on the forced rebuild — and it compares the retained picture against a rebuild at the same instant rather than a hardcoded one.

## The cross-fence pin, tested rather than trusted

The strongest claim in the PR is that L4 pins a coupling in `frame.ts`, which is **off this fence** — "if `frame.ts` is ever changed to reuse the frame object, that law goes red here rather than the picture going quietly wrong on a stranger's machine."

I made `buildFrame` reuse a module-scoped object and re-ran `painter.test.ts`: **L4 and L2 both went red** (EXECUTED). The pin fires. That is the part of #178 I most expected to be an assertion rather than a guard, and it is a guard.

## Two threads I chased adversarially and closed

Both REASONED from the code, stated as such:

- **The `sameSceneState` in-place-mutation residual.** Its own docstring names it, and #157 inherits it for a gate with more at stake — before this, a missed in-place mutation cost one repaint and the next rAF rebuilt; after it, a paused scene could skip forever. Not reachable: `fleet` is memoized on `[state.session, clock, manifest]` (`FleetContext.tsx:74`), and `field.ingest` only fires when `takeNews` returns events, which requires `state.session` to have moved — so a new fleet object always precedes the mutation and the gate refuses. The `FLEET_TICK_MS` re-memo is a second backstop.
- **A hypothesised deadlock in `skipBuild`.** If `painter.repaint` could return false while `painted !== null`, then under a running loop `fallBackToBuild` no-ops (`useFrameLoop.ts:563` — it only draws when `pinned`), the gate would keep saying skip, and the picture would never rebuild. Dead end: `last` (`gl/index.ts:138`) is assigned only inside `paint` and is never reset — not on context loss, not on dispose. Once a paint has run, `repaint` cannot decline. Your "unreachable in practice" comment is exact.

I also audited the two-door claim for `real` independently: reading `buildAndPaint` top to bottom, `real` reaches the build only through `stepFlight(real)` and `settle.progress(real)`. There is no third door. Everything else (`breathOf`, `retire.progress`, `settle.sizes`, `field.step`) takes the pause-frozen `clock`, and `sameSceneState` reads own-keys so no `SceneLatestState` field can be forgotten.

## One fact for the record

My run reported **6882 passed / 2 skipped** against the body's 6878 / 6. The totals reconcile identically (6887 both), and the delta is `it.skipIf` platform/root gates in `packages/server` (`static.test.ts`, `gate-honesty-law.test.ts`). My environment ran four tests **more** than yours, not fewer — noting it only so the two numbers on the record are not read as a discrepancy in the suite.

## Offered, not asserted

Dropping the settling term *entirely* reddens only L9, while dropping just `|| builtSettling` reddens L9 **and** L11 — the narrower mutation is better detected than the broader one. Both go red, so nothing is unguarded and there is nothing to fix. Mentioning it only because it is mildly counter-intuitive if you ever revisit L11's construction.

**No findings.** Nothing fixed by me, so this is an unaided approval of your commits as pushed.

PR #177, opened 2026-08-29
prd42 w10: the PRD cites durably, the barrel is complete, and the reconciler is in the repo

gabriel-canaan:
Closes #152
Closes #153
Closes #154
Closes #161

prd-42 wave 10. Four issues, one commit each, all four fences disjoint and the
diff inside their union.

| commit | issue | what becomes true |
|---|---|---|
| `0ca40b4` | #152 | the PRD's self-citations survive the next amendment |
| `642e983` | #154 | the recorder barrel is complete, and drift is caught |
| `818056b` | #161 | the reconciler is a tracked script, checkable by anyone |
| `5fbd1b1` | #153 | the round-trip law's win32 skip derives from the real rule |

## For the landing operator

**#152's last done-when is now MET, and clearing it took a tracker act plus a
document edit — recorded here because neither is visible in the diff alone.**

At PR-open it was UNMET: `prd-reconcile.sh 42` exited 1 because `#148` and `#92`
are closed prd42 issues whose titles carried no `wN:` token, and the check reads
the **tracker**, not this document, so no edit inside any wave-10 fence could
reach it. Both are now retitled to the wave their work actually landed in —
`#148` to w9 (closed as a duplicate of `#143`, which is w9) and `#92` to w8
(closed as superseded by `#87`, which is w8), each with a comment saying so.

That alone did not clear it. This PR's own follow-ups (`#174`, `#175`, `#176`)
were filed as `prd42 w11:`, which made the tracker claim a wave the document did
not declare — swapping two `NO WAVE` lines for one `UNDECLARED WAVE 11`. Wave 11
is therefore declared in the PRD, in the same form wave 10 uses, inside `#152`'s
existing fence. The drift was introduced by this PR's own bookkeeping and caught
by the very check the done-when names.

Post-landing simulation, EXECUTED against this document and the live tracker —
all three arms clear:

```
NO WAVE     (title lacks a wN: token)          none
VACANT      (declared, unclaimed)              none
UNDECLARED  (claimed, undeclared)              none
waves declared: 1..11    waves claimed: 1..11
```

`repos.test.ts` does real filesystem work, so give `scripts/gate.sh` its
load-batch count as usual.

`repos.test.ts` does real filesystem work, so give `scripts/gate.sh` its
load-batch count as usual.

## What is the sibling case?

**#154 — the axis beside the one that was fixed.** The alternation had been
widened for `enum`/`namespace`/`abstract`, all on the *declaration keyword* axis.
[...]

PR #183, opened 2026-08-31
prd42 w11: the barrel law's grammar, the reconciler's marker and scale, and the win32 segment rule

### LINKED ISSUE #174: prd42 w11: the barrel law's grammar covers the clause axis, not just the keyword axis
## The defect

`#154` closed the barrel-completeness guard's **declaration-keyword** axis and, in
its verification amendment, the **clause** axis for `export type { X }` in both
directions. The rest of the clause axis is still open. Every form below was
EXECUTED against the real helpers in `namespace-law.test.ts` during wave-10
verification, each with a passing control:

| production | derives | direction |
|---|---|---|
| `export const enum K { A }` | `["enum"]` | fails OPEN **and** demands a symbol literally named `enum` |
| `export const a = 1, b = 2` | `["a"]` | fails open on `b` |
| `export declare class D {}` / `export declare function f()` | `[]` | fails open |
| `export const { a, b } = obj` / `export const [a, b] = arr` | `[]` | fails open |
| `export { x as default }` / `export default class D` + barrel `{ default }` | `[]` / false positive | both directions |
| barrel `export * from './rotate.js'` | `[]` | false positive |
| barrel `export * as ns from './rotate.js'` | `[]` | false positive |
| barrel `export { real as alias } from './rotate.js'` | reports `real` missing | false positive |
| **module-side `export { x } from './helpers.js'`** | `[]` | invisible — `(?!\s*from)` is deliberate for its own arm, but a re-exported public symbol then needs no barrel entry and the guard stays green |

`const enum` is the nastiest: the guard loses the real name **and** demands the
barrel republish `enum`, so the repair for a genuine gap is unreachable.

None of these occurs in `rotate.ts` or any server barrel today. They are filed
rather than absorbed so the enumeration stays recoverable — that is the whole
reason it was written down.

Two smaller siblings found in the same pass, both pre-existing and both now
load-bearing for a second law:

- `specifierTargetsRotate` matches on **basename only**, so
  `export { X } from '../elsewhere/rotate.js'` credits `X` as republished
  (EXECUTED: `["RealError"]`).
- `codeOf` strips comments but not string literals, so
  `const s = "export function ghost() {}"` derives `ghost` (EXECUTED). Fails
  **closed**, so it is noise rather than a hole.

Also here, because it is one line and the same file: `BARREL_GUARDED_DOORS` is a
**tolerance** list, and two comments added by `#154` cite it as **enforcement**.
EXECUTED mutation on the real tree — adding `performRetarget,` to
`recorder/index.ts` reddens nothing (39 passed, typecheck 0). One assertion
closes it:

```ts
expect([...BARREL_GUARDED_DOORS].filter((d) => barrelReexportedRotateNames(BARREL_SOURCE).has(d))).toEqual([])
```

## Done when

- Each production above has a row in the enumerated grammar test **before** the
  alternation changes, per that test's own doc comment, and a verdict.
- Forms deliberately excluded (module-side `from` re-exports, if that stays the
  decision) say so in the function's doc rather than being silently absent.
- The guarded-door set is asserted absent from the barrel, not merely tolerated.
- Mutation-certified: reverting each new arm reddens a named row.

## Fence (may touch ONLY)

## Provenance
Wave-10 verification: two independent seats, two cross-reviews and a reconcile
pass, then a fix re-review. `#154`'s own commit body enumerates these as
deferred.

### LINKED ISSUE #175: prd42 w11: the reconciler refuses a misplaced marker and survives a large tracker read
## The defect

`#161` landed `scripts/dev/prd-reconcile.sh` as a verbatim adoption, and wave-10
verification found three hardening items in it. All were deliberately NOT fixed
there, because a verbatim adoption is the point of that issue and each is a
behaviour change.

### 1. A distant `SUPERSEDED` marker silently retires a current wave

The `doc_waves` awk keeps a stack; a column-0 marker pops whichever declaration
is on TOP, however far above it sits. EXECUTED, the script's own awk body:

```
marker adjacent to Wave 1          -> "2 3"    wave 1 popped
marker 7 prose lines below Wave 1  -> "2 10"   popped identically, no diagnostic
```

This corpus already hit it once — wave 4 vanished from the count. `#161`'s
comment and prd-42's amendment now both describe the behaviour correctly, so
prose is the guard. Making an out-of-range marker a hard error is the fix.

Related precedence, also EXECUTED: a marker line that itself contains
`**Wave N` matches the declaration arm first, whose `next` bypasses the marker
arm — so it **pushes instead of popping** (`> **SUPERSEDED** by **Wave 2 -
replacement**` under `**Wave 1` leaves `1 1` and `2 2`; CONTROL, the same marker
without the quoted wave, pops correctly). That is why prd-42's convention forbids
`**Wave <digit>` in quoted text. A hard error should cover this shape too.

### 2. `printf | grep -q` under `set -o pipefail` reports a false `VACANT`

`:197`. `grep -q` exits at the first match, `printf` takes SIGPIPE, and pipefail
promotes 141 — so a wave that IS present reports vacant. EXECUTED, 200 rows
(the script's own `--limit`):

```
88-char titles  (14.7 KB)  0/5 false VACANT
150-char titles            4/5
250-char titles (52.7 KB)  5/5, PIPESTATUS 141 0
here-string <<< form       0/5 at every size
```

Not reachable today — the largest milestone is 24 issues / 1,837 bytes, ~25x
under where the race starts — and it fails **loud** (EPIPE can add a VACANT,
never suppress a finding). The same shape sits at `:217`
(`printf | awk | grep -qx`), unreachable at a ~10-line `doc_waves`. **Both must
be fixed together**, and note the threshold is a race, not a clean cutoff:
measured 42.7 KB tripping and 46.7 KB not.

### 3. Four verifications discard the reason they failed

`:110` (fetch), `:119` (ls-tree), `:124` (git show), `:169` (gh issue list) all
redirect stderr to `/dev/null`. The verdicts stay honest — this is not the
swallowed-failure bug — but the operator is pointed at the wrong hypothesis.
EXECUTED with a failing `gh` mock: real stderr `error connecting to
api.github.com`, script says only `could not read milestone 'prd42' — does it
exist?` and exits 2.

## Done when

- A marker outside its declaration's range is a hard reconciliation error, and
  the `**Wave N`-in-marker shape is covered; both pinned with a control fixture
  and a malformed one.
- `:197` and `:217` no longer use an early-closing pipeline; a padded-output
  regression test pins it.
- The four suppressed verifications capture stderr and print it under their
  existing verdict lines.
- A `scripts/dev/prd-reconcile.test.sh` exists — `#161` shipped 244 lines of
  shell with no test because its fence forbade adding one; `issues.test.sh`'s
  mock-`gh`-on-PATH harness is the pattern.

## Fence (may touch ONLY)
- `scripts/dev/prd-reconcile.sh`
- `scripts/dev/prd-reconcile.test.sh` (NEW)

## Provenance
Wave-10 verification and its fix re-review. Item 2 was found by the Codex seat;
its severity was reduced from "unreliable at useful repository size" to
"unreachable today, fix anyway" by measurement against the real tracker.

### LINKED ISSUE #176: prd42 w11: the win32 segment rule matches the naming rules it cites
## The defect

`#153` derived the win32-illegal segment class from the naming rule instead of a
two-item list. Wave-10 verification found two residuals in that derivation and
one in its non-vacuity floors.

### 1. The reserved-device rule is ASCII-only

`repos.test.ts`'s `WIN32_RESERVED_DEVICE_NAME` is
`/^(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])$/i`. Microsoft's naming rules reserve
**six more**: `COM¹ COM² COM³ LPT¹ LPT² LPT³`, with the explicit note that
Windows "recognizes the 8-bit ISO/IEC 8859-1 superscript digits ¹, ², ³ as digits
and treats them as valid parts of COM# and LPT# device names, making them
reserved in every directory. For example, `echo test > COM¹` fails to create a
file."

EXECUTED: all six plus `COM¹.txt` and `LPT³.txt` return `false`. CONTROLS:
`COM1` → `true`, `COM0` → `false`, `plain-word` → `false`.

`COM0`/`LPT0` are **correctly** treated as legal — they are absent from the
normative list, and the commit's negative row for them is right. (One review seat
argued the page lists them; it does not. The only `COM0` on that page is an
NT-namespace symlink example.)

Fix is two characters: `COM[1-9¹²³]|LPT[1-9¹²³]`. Not urgent — the predicate's
only input is a 10-item literal list in the same file — but it is a hole in the
"covered by construction" promise the issue exists to make.

### 2. `/` is absent from a character set the doc presents as *the rule*

`if (/[<>:"|?*\\]/.test(segment))` — eight characters. Microsoft lists **nine**
reserved characters for a filename, including `/`. The function's own doc
justifies `\`'s presence with an argument that covers `/` verbatim ("a path
separator, so never legal *inside* a single segment").

Impact today is nil and the exclusion is arguably right: `path.join(root,
...segments)` turns a `/` into hierarchy, so it never reaches the predicate as a
single component. **Either add it or say in the comment why it is excluded** —
what is wrong is that the doc claims a rule it does not implement, silently.

### 3. The non-vacuity floors are hand-maintained while the generator is derived

`roundTripSegmentsFor` now consults the predicate, so the generator is covered by
construction. The floors are still hardcoded characters with the colon/backslash
pair wrapped in `if (isPosix)` by hand. EXECUTED: force `isPosix = false`
(CONTROL: today's list green), add the issue's own named future example
`'a?query here'` plus an unconditional floor → red.

Fails **closed** and win32-only, and there is no `windows-latest` leg, so nothing
can fail open today. `#153`'s body was narrowed to say the generator is covered
by construction and the floors are not; this issue closes the gap.

## Done when

- The device rule covers the superscript forms, with the six as rejecting rows
  and `COM0`/`LPT0` retained as accepting controls.
- `/` is either in the class or explicitly excluded in the doc comment with its
  reason, so the stated rule and the implemented rule match.
- The floors derive from the filtered list rather than being hand-conditioned,
  OR the comment says which half is maintained by hand (the non-vacuity pin at
  the call site already does this — extend it to the floors).
- Mutation-certified in both directions, with the class probed by a spelling not
  used before.

## Fence (may touch ONLY)

## Provenance
Wave-10 verification (two seats, two cross-reviews, reconcile) and its fix
re-review. The superscript row was the one finding no seat derived independently
— it came from reconciling against Microsoft's published naming rules rather than
against the code.

gabriel-canaan:
Closes #174
Closes #175
Closes #176

prd-42 wave 11 — the three follow-ups PR #177's own verification filed. Three
issues, three commits, three disjoint fences; the diff is exactly their union.

|---|---|---|
| `9b94ba6` | #174 | the barrel completeness guard's grammar covers the clause axis, not just the keyword axis |
| `f2e057c` | #175 | the reconciler refuses a misplaced marker and survives a large tracker read |
| `ab322d5` | #176 | the win32 segment rule matches the naming rules it cites |

---

## ⚠ Read this first: ~180 lines of this diff are the REVIEWER'S, not the lanes'

This wave went through a full verify, **four rounds of repairs, and four
re-review rounds**. Every repair after the lanes' original commits was written
by the reviewer. Measured, original lane commit → current:

```
#174   +111 / -10    reviewer repairs
#175    +67 / -3     reviewer repairs
#176   unchanged     purely the lane's work
```

`AGENTS.md`: *"Never review your own fix into main. The reviewer fixes, but a
second pair of eyes still lands anything non-trivial."* Eight model seats read
this; **no human has.** Model seats are a finding generator, not a reviewer —
and the reviewer chose their prompts, adjudicated their disagreements and
decided when to stop. That is four roles, one of which is "reviewer".

**So this needs a contributor to read the diff.** Where to spend it:

- **`ab322d5` (#176) needs almost nothing.** Byte-identical since the lane wrote
  it. Two seats independently verified its premise verbatim against Microsoft's
  published naming rules (exactly six superscript device forms; `COM0`/`LPT0`
  genuinely absent from the normative list; nine reserved characters, so the
  "eight below, plus `/`" comment is exact). 55 rows probed, zero mismatches.
- **`9b94ba6` and `f2e057c` are where the reviewer's code is.** Start with the
  long note in `prd-reconcile.test.sh` recording four failed attempts at one
  assertion and why the idea was abandoned — if you disagree with anything here,
  that is the most likely place.

## Seven defects were found in reviewer-authored repairs, none by their author

That base rate is the argument FOR another set of eyes, not against it.

| round | in #174 | in #175 |
|---|---|---|
| 1 | `declare` widened on one arm, two siblings missed | the law was scoped by CONTENT — a comment would have reddened it |
| 2 | — | it matched one spelling; three evasions passed green |
| 3 | `const  enum` backtracked and captured the KEYWORD | wrong BOTH ways: evadable *and* convicting legitimate shell |
| 4 | the pin for round 3 used `toContain`, so reverting the fix left 52/52 GREEN | satisfied for the wrong reason — decoy comments kept it green while neither check used a here-string |
| final | a dangling header promising what a `toContain` table cannot assert | the deletion left its prose behind; a self-citation matching no version of the file |

## The judgement call worth challenging

**#175 ends with NO source-text law, deliberately.** After four attempts — each
defeated a different way — the conclusion was that proving a property of
arbitrary *shell source* with a line-oriented text match needs a shell parser:
every narrowing opens an evasion, every widening convicts real code, and every
anchor that is not a real parse can be satisfied by a decoy.

What remains is behaviour only, and its cost is stated rather than hidden: the
padded fixture is **schedule-sensitive**, measured at 37/40 here and 40/40 on a
reviewing seat's machine. Two runs are quoted rather than one because a single
figure for a race reads as a threshold. #175's criterion asks for "a
padded-output regression test", which this is; determinism was the reviewer's
addition and could not be had honestly.

**If you think that is the wrong call, this is the place to say so.**

## What is the sibling case?

Every defect in this wave and in its repairs is one shape: **a keyword or a
spelling sitting between the pattern and the thing it anchors on** — and the fix
that closes the case in front of you while leaving its sibling. `#174` records
it four times now; the repairs reproduced it three more.

Each repair is certified against the mutation that exposed it, all EXECUTED:

```
re-add `export * from './rotate.js'` to the REAL barrel
  before M1:  52 passed (52)        <- three guarded doors open, law green
  after  M1:   1 failed | 51 passed

revert the anchored lookahead
  with round-3 toContain rows:  52 passed (52)   <- decorative
  with the exact-equality pin:  expected [ 'SpacedK', 'enum' ] to equal [ 'SpacedK' ]

byte-exact revert of the VACANT pipeline (TAB preserved, taken from git)
  padded fixture:  37/40 here, 40/40 on a reviewing seat's box
```

## Gates

`npm run typecheck` rc=0 · `npm run lint` rc=0 · `npm run build` rc=0 ·
`VITEST_MAX_WORKERS=6 npm test` → 386 files, 6887 passed, 3 expected-fail,
2 skipped · `prd-reconcile.test.sh` 19/19, 5 consecutive runs and 4x concurrent.

`scripts/gate.sh` was NOT run — it IS the landing, and running it here would land
unreviewed work. Landing is the operator's step.

## Fence

No widening. Exactly the union of the three declared fences:
`recorder/namespace-law.test.ts` · `scripts/dev/prd-reconcile.sh` ·
`scripts/dev/prd-reconcile.test.sh` (new) · `concierge/repos.test.ts`.

## Not fixed here, deliberately

The remaining fail-open grammar forms — multi-line declarator lists, rest
bindings, module-side `export * as ns` — are real and **unreachable in this repo
today** (EXECUTED sweep: zero of each in production source). They belong in one
follow-up issue on the family, not scattered through a wave reviewed for
something else.


ciaran-slow (commented):
## Reviewed commit-by-commit, in a clean review clone at node 22.23.2. One finding, fixed in `7dbc397`.

The PR body asked for a contributor to read the diff and named where to spend it. I read all three commits and re-derived the load-bearing claims rather than taking them from the body. **Everything the body claims that I checked, checked out** — including the two it flagged as most likely to be wrong.

### What verified in the author's favour — all EXECUTED

| claim | how I checked it | result |
|---|---|---|
| M1: `export * from './rotate.js'` on the REAL barrel reddens the guarded-door law | appended it to `packages/server/src/recorder/index.ts`, ran the file | `1 failed \| 52 passed` — the guarded-door row, and only it. Restored, 53 passed. **Reproduces exactly.** |
| the padded fixture detects a byte-exact revert of the VACANT pipeline | spliced both `\| grep -q` lines back in from `f2e057c^` (never retyped — the TAB trap the body names), ran the suite **20×** | **20/20 caught, 0/20 fully green.** Better than the 37/40 quoted. |
| the marker hard error does not regress the real corpus | extracted the `doc_waves` awk from HEAD and from `f2e057c^`, ran both over **all 50** PRD docs (`docs/prds`, `done/`, `parked/`), diffed stdout **and** exit status | **byte-identical on every file.** The hard error convicts nothing that exists today. |
| gates | `npm run typecheck` rc=0 · `npm run lint` rc=0 · `VITEST_MAX_WORKERS=6 npm test` → 386 files, 6883 passed, 3 expected-fail, 6 skipped · `bash scripts/dev/prd-reconcile.test.sh` **19/19** | all green |
| CI actually ran | `gh pr checks 183` | **7/7 jobs executed** — `build-test-boot` ×3 (macOS×min correctly excluded) and `pack-smoke` ×4. Not a green bar over a skipped leg. |
| fence | `git diff --stat` against the merge base | exactly the union of the three declared fences, four files. No widening. |

`ab322d5` (#176) needs nothing, as the body said. It is test-only — `WIN32_RESERVED_DEVICE_NAME` has no production caller — so the blast radius is the file itself.

**On the #175 judgement call you invited challenge on: I think it is right.** Four attempts, four different defeats, and the pattern is real — a line-oriented match over arbitrary shell cannot prove the absence of a shape without a parser. #175's criterion asks for "a padded-output regression test", and that is what shipped. Quoting two runs for a race instead of one is the honest presentation. I would not have taken the fifth attempt either.

---

## The finding — `9b94ba6`, and it is the shape this wave is named for

**EXECUTED**, same probe run against `9b94ba6` and against `9b94ba6^`:

```
export const x: Map<string, number> = new Map()
  9b94ba6^  ["x"]          9b94ba6  ["x","number"]
export const rec: Record<string, string> = {}
  9b94ba6^  ["rec"]        9b94ba6  ["rec","string"]
export const s = 'a,b'
  9b94ba6^  ["s"]          9b94ba6  ["s","b"]
export const t = `a,b`  /  export const r = /a,b/
  9b94ba6^  ["t"] ["r"]    9b94ba6  ["t","b"] ["r","b"]
```

`declaratorListNames` walks `{`, `[`, `(` depth to split the declarator list, and then reads each declarator with an unanchored `^([\w$]+)`. `<`, quotes and regex literals are not tracked as depth — so every comma they hide splits a declarator that is not one, and its first word is minted as an export name.

The consequence is the exact one #174 was filed for, one axis over: `barrelCompletenessGaps` then demands the barrel republish `number`, and **no barrel can**, because `number` is not a symbol `rotate.ts` exports. Unrepairable red — the same shape as the `const enum` arm capturing the KEYWORD.

**Latent, not live**, which is why nothing is failing: `rotate.ts`'s single `export const` (`rotate.ts:410`, `RETARGET_OR_ROTATION_IN_FLIGHT_MESSAGE`) puts its `=` at end of line, and the `[^\n;]*` capture feeding this function stops at the newline — so the commas in the string on the next line are never seen. It stays green until an ordinary `export const x: Record<string, string>` lands in `rotate.ts`.

Sibling: the destructured-entry reader had the identical unanchored regex, and a destructured default hides a comma exactly as an initialiser does — `export const { a = 'x,y' } = obj` derived `["a","y"]`.

### Fixed in `7dbc397`

An identifier counts as a binding only where a declarator can end or continue: end of declarator, or before `:`, `=`, `?`, `!`. Narrow half of the trade on purpose — a declarator this rejects contributes **nothing** rather than a wrong name, the same call the function already makes for a nested pattern. Tracking `<` as depth is the wide half and is wrong: `export const ok = a < b, also = 2` is a comparison, and no lexer-free rule separates it from a type argument. Both bindings there are pinned as a control. Both call sites now share one constant.

**What mutation would it survive?** None. Exact equality, not `toContain` — for the reason recorded two screens up in this same file: the bug IS the extra name.

```
revert DECLARATOR_BINDING_RE to /^([\w$]+)/  ->  1 failed | 53 passed   (the new row, only it)
restored                                     ->  54 passed (54)
```

Re-gated with the fix in: typecheck 0 · lint 0 · 386 files / **6884 passed** / 3 expected-fail / 6 skipped · `prd-reconcile.test.sh` 19/19.

Fence: `namespace-law.test.ts` only — inside #174's declared fence, no widening.

---

## What I did NOT run, stated rather than left out

- **`scripts/gate.sh`** — correctly not run here either. It IS the landing, and landing is the operator's step.
- **Windows.** #176's device-name rule is win32-only and there is no `windows-latest` leg in `ci.yml`. The predicate is exercised as a pure function, which is all that can be checked from here.
- **Microsoft's published naming rules.** I did not independently re-derive #176's premise (exactly six superscript device forms; `COM0`/`LPT0` absent from the normative list; nine reserved characters). I read the code against the claim, not the claim against the source. Two seats reportedly did verify it verbatim; that part rests on their word, not mine.
- **Concurrency.** I ran the suite once at `VITEST_MAX_WORKERS=6`, not the gate's 4× batched probe.

## Rung

Substantive under `rhizo-review`'s ladder — it changes the behaviour of a derivation you wrote, in the file this wave is about. So the fix is pushed but **not self-approved and not merged**: over to you to confirm `7dbc397` reads right, and specifically that you agree the anchor is the correct half of that trade rather than tracking `<`. Once you say so, this is ready to land — every other claim in the body reproduced.

Issue #186, opened 2026-08-31
prd43 w1: the citation law is honest about what it does not sweep


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/doc-citation-law.test.ts`

## Blocked by
**Nothing — #16 landed in PR #187 (f95f9d3), 2026-09-01.** The fence is free.

Every line citation below is against #16's amended head at filing time and is
now **stale by seven commits**: five `(review of #16)` repairs plus the
reviewer's own `420930a`, which removed a dead `existsSync` the hoist had left
behind. Re-derive every line number from the file on `main`; do not trust one
from this body.

Two of the nine were re-checked on `main` after the merge and still hold, along
with four more spot-checks — the merge did not close any of them.

**Model:** sonnet. **Wave:** 1 — the SAME wave as #16, deliberately. This is
#16's own remainder in #16's own file, and wave 2 is declared "parallel, fenced
apart" over `docs/architecture.md`, `docs/roadmap.md`, `docs/design-notes/` and
four `packages/**` files. None of those is this file, so a w2 filing would claim
a fence wave 2 does not own.

## Why one issue and not nine

Every item below lives in `doc-citation-law.test.ts`. Filed separately they would
be a guaranteed fence-lint overlap — a rebase conflict scheduled in advance,
which that check hard-fails on. Same reasoning #179 records for its own pair.

## The defects

All found by #16's verify pipeline (two independent seats plus orchestrator
reconciliation), all EXECUTED, all deliberately NOT fixed there: #16's repairs
were scoped to the three must-fix findings, and widening past those would have
put unreviewed work into a commit reviewed for something else.

**None of these hides a live break today.** They are latent, and the point of
filing them is that the law is prd-43's keystone: every later wave consumes it,
so a hole in it is a hole in everything downstream.

### 1. stripFencedCodeBlocks is inert, and its test cannot fail

Replacing the function body with `return text` leaves the file **14/14 green**.
Corpus-wide it removes **0 of 855 citations across all 189 docs**.

The rigged fixture's fenced path is double-quoted JSON
({ "fence": ["packages/…/**"] }), which CITATION_RE never matched, so the
test passes identically with the stripper deleted. The table's justification
cites docs/architecture.md's fenced example — also double-quoted JSON, so the
false positive it guards against cannot occur.

Fix: backtick the fabricated path in the fixture so the stripper is actually
exercised, then confirm deleting the stripper reddens.

[...]

PR #187, opened 2026-08-31
prd43 w1: a path cited from a document or a comment must exist

### LINKED ISSUE #16: prd43 w1: a path cited from a document or a comment must exist
## Fence (may touch ONLY)

## Blocked by
Nothing — this is the wave-1 keystone, and it lands **RED**, naming all four clusters before any of them is fixed.

**Wave 0 is an operator act and is not this issue:** the ADR recording prd-12's divergence (where lab-synthesised sessions actually live) gates wave 5 only, not this law.

Sibling lanes own the documents this law will redden.

## Definition of done
- Extracts every `packages/**`, `scripts/**` and `docs/**` path string from `docs/**/*.md` **and** from `packages/**/*.ts` comments, and asserts each exists.
- `docs/research/`, `docs/review/` and **`docs/prds/` — the whole PRD tree, not only `done/`** — are excluded as dated artefacts. They describe the tree at a commit, so a moved path there is a true record, not a defect.
- The exclusion list is asserted HONEST: every excluded directory must contain at least one path that would otherwise trip the law, or the exclusion is stale and hiding a live file.
- Scoped by what a file IS, not by what it is called — the `#649` lesson.
- Proves its own detector bites against a rigged fixture.
- Lives where the root vitest config actually globs it (`packages/*`) — a root-level test would never run and would be its own vacuous law.
- The PR shows it red, listing every currently-broken citation from an in-scope file.

## Corrected 2026-08-24 — the exclusion is the whole PRD tree

This DoD previously excluded `docs/prds/done/` only, which came from prd-43 ruling 1 contradicting prd-43's own non-goal ("the citation law excludes them by directory, deliberately", naming `docs/prds/`). Review of #25 caught it and `c143b72` resolved it toward the wider exclusion.

Built to the old wording, this law's first run would have gone red on the document that commissioned it. A PRD's Evidence section names dead paths **on purpose** — that is what the evidence is — and prd-43's own names four, all verified absent:

```
packages/web/src/fleet/buildFleet.ts              MISSING
packages/web/src/panels/spend/format.ts           MISSING
docs/research/2026-08-04-semantic-judge-spike.md  MISSING
docs/decisions                                    MISSING
```

The four PRDs beside it cite moved paths of their own. A law that must be suppressed to pass teaches its readers to suppress it, which prd-43's rejected-alternatives section already says in as many words.

The honesty check is **unaffected and still bites**: `docs/prds/` holds those four paths, so the exclusion cannot go stale silently. Do not narrow this back to `done/`.

gabriel-canaan:
Closes #16

prd-43's **keystone**. It lands the law that every later wave consumes: a path
cited from a document or a code comment must exist. Zero-claimant — it writes the
law and fixes none of what the law finds.

## Why this ships alone rather than bundled

`AGENTS.md` bundles by (milestone, wave), and prd-43 wave 1 now has two open
issues — this one and #186, filed from this PR's own verification. They are a
**stack, not a bundle**: #186 edits this same file and is blocked on this
landing, and a stack is the thing bundling exists to avoid. Waves 2–5 all consume
this law, so holding the keystone to bundle it with its own follow-up would stall
the PRD to save one queue toll.

## The green/red ruling, and why it is green

The issue left one thing open: the DoD says the law should show red, but a law
that lands red on `main` is red CI for everyone and the landing gate blocks on
the suite. The lane chose **(a) — commit green with an explicit allowlist that
wave 2 empties entry by entry** — and made the allowlist honest rather than a
place to hide things. Every one of the 16 entries is asserted to still be cited
**and** to still fail today; the moment a citation resolves, the test says
`"now resolves — remove this allowlist entry"`. That is what makes it emptyable
instead of write-only.

## Commits

|---|---|
| `174afc0` | the law exists: 468 lines, an enumerated input-class table, 16 honesty-checked allowlist entries |
| `92180d4` | the docs sweep pin reads the sweep it pins |
| `c14eb1b` | `:NNN`, `:78-85` ranges and `#anchor` reach the existence check |
| `9e384c8` | `existsSync` runs before `isIgnored`, so the law stops forking git per citation |
| `29eeeea` | a compound line-and-anchor suffix strips in either order |
| `1451c6f` | the stripper covers the whole class, and the numbers beside it are measured |

The five `(review of #16)` commits are repairs from verification, not lane work.

## What is the sibling case?

**Answered five times, because each answer produced the next one — that is the
finding worth reading here.**

The law is a detector over text, and this repo's most expensive recurring defect
is a hand-typed pattern covering the form its author tested while staying silent
on the sibling nobody tried. This branch walked straight into it:

| round | what was fixed | what it opened |
|---|---|---|
| `c14eb1b` | `:` and `#` admitted to the character class | compound `x.md:1#anchor` now matched and mis-stripped |
| `29eeeea` | run-alternation, order-independent | `file:line:col` and `#v1.2` still unstripped |
| `1451c6f` | `.` in the anchor class | — |

`c14eb1b` is the sharpest case: before it, compound spans did not match at all and
were silently skipped. After it they matched and were mis-stripped, so a **valid
citation was reported as a violation**. The repair turned a false negative into a
false positive — louder, and still wrong.

The fix is now an alternation over a RUN of suffix groups rather than a sequence
of anchored replaces, which is the class instead of an enumeration of it. A third
sequential replace would have closed the reported spelling and left its mirror,
which is how the defect got there in the first place.

Other siblings, verified rather than assumed: **line ranges** (7 real instances —
widening the class without a range arm would have turned them into fresh
violations), and the mirrored order of every compound form.

None of these. All EXECUTED, each re-check using a spelling the original finding
did **not**, because re-running the probe that found a defect proves that input is
caught and says nothing about its neighbours:

    sweep glob -> docs/**/*.md                     1 red   (the spelling that defeated the old pin)
    sweep glob -> docs/*/*.md                      1 red   (never probed before)
    CITATION_RE class reverted                     1 red
    range arm removed                              2 red   (the test AND the main sweep)
    stripper reverted to two sequential replaces   1 red
    anchor class reverted to exclude `.`           1 red
    isIgnored guard deleted                        2 red
    citationExists forced true / forced false      7 / 6 red
    EXCLUDED_DIRS emptied                          1 red
    self-exclusion reverted                        2 red
    hoist removed (`9e384c8`)                     14 green — a pure reorder, exactly as claimed

End-to-end, the control that matters — the same break in two spellings, before
`c14eb1b`:

    `packages/core/src/definitely-not-real.ts`      1 failed, named exactly
    `packages/core/src/definitely-not-real.ts:42`   14 passed, invisible

## Fence widening

**None.** Every commit touches `packages/server/src/doc-citation-law.test.ts` and
nothing else; `git diff --name-only origin/main...HEAD` is one file.

## Verification

Two independent seats (Opus `--effort xhigh`, Codex `gpt-5.6-terra` `xhigh`), each
in its own disposable worktree with `main` merged in, plus orchestrator
reconciliation against the code — every finding re-executed before it was acted
on. Then, because the repairs had been read only by the person who requested them,
a **second** pass of both seats against the amended head.

That second pass is the one that paid: both seats found defects **in the repairs,
not in the lane's work** — the compound-suffix regression above, and two numbers
in a comment that were swapped. The timing pair had been produced by sorting two
durations and attaching test names by assumption; re-measured by name,
`exclusion 3156 ms -> ~380 ms` and `main sweep 2466 ms -> ~145 ms`. `9e384c8`'s
body carries the uncorrected pair and `1451c6f` is the correction of record.

One reported finding did **not** survive verification and is filed as a correction
rather than a defect: a "real dead path" in an unswept `.mjs` file.
`packages/web/src/scene/parity/capture.mjs` names a deleted `paint.ts`
deliberately and resolves it out of git history, because `8686f24` (#578) replaced
the 2D painter and the parity harness compares against the pre-deletion version.
Extending the sweep there naively would false-positive on working code — which
turns that scope question into a ruling with evidence, and is how #186 files it.

    VITEST_MAX_WORKERS=6 npm test     0   ->  387 files, 6944 passed

## Follow-up

**#186** (`prd43 w1: the citation law is honest about what it does not sweep`)
carries the nine deferred findings: the inert fenced-code stripper, `//` inside a
string literal, `**`-glob truncation, gitignore-as-existence-exemption, the
index-only sweep, the vacuous-on-empty exclusion test, the identity control that
cannot distinguish derived from hardcoded, the misplaced coupling comment, and the
`.tsx`/`.mjs` scope ruling. One issue rather than nine, because they share this
fence and separate filings would be a `fence-lint` overlap scheduled in advance.


ciaran-slow (approved):
Approving. Read commit by commit against #16's DoD, in a clean clone at `420930a` with
`origin/main` (`835ae52`) as the merge base. One fix pushed, `420930a`, mechanical.

**The flake `9e384c8` was landed on, and could not reproduce, is real. I reproduced it.**

That commit's body is unusually honest: *"I could not reproduce that timeout — 4 of 4
full-suite runs green here… this lands on the strength of the measured margin and the
gate's concurrency, NOT on a flake I could demonstrate."* Landing a commit on an
unreproduced report is the thing most likely to be waved through here, so it is the
thing I probed hardest. **EXECUTED**, `--maxWorkers=5`, four concurrent runs per round,
two rounds each, same machine, same load:

```
HEAD (hoisted)     8/8 runs   Tests 15 passed (15)
hoist reverted     8/8 runs   Tests  2 failed | 13 passed (15)
                              Error: Test timed out in 5000ms.   (x2 per run)
```

8/8, not 2/3. The two tests that time out are exactly the two the commit names. The
reporting seat was right, the author's caution was warranted, and the commit is load
bearing rather than precautionary.

**The timing correction in `1451c6f` holds, by the method it claims.** Re-measured per
test BY NAME rather than by sorting durations — the exact error being corrected:

```
every excluded directory … trips the detector    1272 ms
every in-scope citation exists                    293 ms
```

Exclusion is the slower of the two, ~4x, as the corrected comment says and the
superseded one had backwards. My absolute numbers differ from the comment's
`3156 -> ~380` / `2466 -> ~145` (different machine, unloaded) — the direction and ratio
are what the comment reasons from, and both hold.

**The run-alternation really is the class, not an enumeration of it.** The strongest
claim in the PR body is that `1451c6f` stopped closing spellings one at a time. Probed
with four spellings **none** of the commits used:

```
docs/architecture.md#a#b          -> docs/architecture.md
docs/architecture.md:1-2:3-4      -> docs/architecture.md
docs/architecture.md#L42          -> docs/architecture.md
docs/architecture.md:1#a:2#b      -> docs/architecture.md
```

All four strip. Re-running the probes that found the defects would have proved nothing;
these are the neighbours.

**The zero-instance claims are true.** `git grep` over `docs/*.md` + `packages/*.ts`
returns 0 for compound `:N#anchor`, 0 for `file:line:col`, 0 for trailing bare `:`,
0 for trailing bare `#`. Nothing is oversold.

**Gates, all EXECUTED here at `420930a`, node 22.23.2:**

```
VITEST_MAX_WORKERS=6 npm test     0   ->  387 files, 6940 passed, 3 expected fail, 6 skipped
npx vitest run doc-citation-law   0   ->  15 passed
```

**Fence:** `git diff --name-only origin/main...HEAD` is one file, as claimed.

**Against the DoD:** every bullet in #16 is met, including the corrected 2026-08-24
wording (whole `docs/prds/` tree excluded as a citing source, honesty-checked). The
green/red ruling was posted on the issue before the code was written, which is the
order this repo asks for.

## The one fix — `420930a`, mechanical

`9e384c8`'s hoist added `existsSync` at the top of `citationExists` and left the
original twelve lines below, on the same unmodified `stripped`, after two branches that
both return unconditionally. The condition can never be true.

Proved by unreachability rather than by deletion, because deleting it and seeing 15/15
would only show it is *removable*. The statement is on the control path to the bare-ADR
branch, so it does execute — it is the condition that cannot hold:

```
if (existsSync(...)) throw new Error("DEAD BRANCH REACHED")   -> 15 passed
```

A throw that never fires is the assertion. Sibling checked by reading every call in the
function: of the four remaining `existsSync`/`readdirSync` calls, one is the live
hoisted check and three test `target` or `dir` inside their own branches. No second
leftover.

## Deferred, not blocking

One new finding, posted to #186 as item 10 rather than fixed here: a bare trailing `:`
or `#` is extracted but not stripped, so `` `docs/architecture.md:` `` would report a
real file as broken. Zero live instances; it is the fourth spelling in the family
`29eeeea` and `1451c6f` closed. Kept off this PR because it changes what
`stripCitationSuffix` accepts, and this PR's repairs were scoped to the must-fix set —
widening them here is the thing #186 exists to prevent.

I re-derived #186's other nine items independently while reading and found the same
set. That issue is an accurate remainder, not a parking lot.

## What I did NOT run

- **`scripts/gate.sh`** — that is the operator's landing step, not a reviewer's, per
  `AGENTS.md`. This repo has no `scripts/land.sh`.
- **CI on `420930a`** — all 7 jobs were still pending when I posted. The previous head
  `1451c6f` was green on 5 of 7 with both ubuntu `build-test-boot` legs unfinished, so
  **no ubuntu test leg has been observed green on this branch yet.** Confirm both
  before merging; `Build` runs before `Test`, so a red leg silently skips typecheck,
  lint, packaging and boot smoke.
- **Windows** — not in `ci.yml`'s grid at all. This law shells out to `git` and joins
  paths with `path.join`, so it is exactly the shape that grid does not cover.

ciaran-slow:
**CI settled green on `420930a`** — both jobs, full grid, all 7 legs:

```
build-test-boot (macos-latest, current)   pass  4m16s
build-test-boot (ubuntu-latest, current)  pass  7m24s
build-test-boot (ubuntu-latest, min)      pass  9m20s
pack-smoke      (macos-latest, current)   pass  1m00s
pack-smoke      (macos-latest, min)       pass  1m09s
pack-smoke      (ubuntu-latest, current)  pass  55s
pack-smoke      (ubuntu-latest, min)      pass  1m03s
```

This resolves the caveat in my review. The two ubuntu `build-test-boot` legs are the
ones that had never been observed finishing on this branch — they ran for 7m24s and
9m20s, so they executed rather than short-circuiting, which is the distinction
`AGENTS.md` asks to check rather than reading the green bar.

Approved and ready. Not merging — landing on this repo is the operator's step.

PR #191, opened 2026-09-01
prd46 w5: the honesty law convicts the class, and the registry records its hold

### LINKED ISSUE #72: prd46 w5: the coupling registry records the law's collection-time hold on gate.sh

### LINKED ISSUE #179: prd46 w5: the law's own proofs and predicate cover the class, not one spelling
## Fence (may touch ONLY)

## Blocked by
**PR #155 (prd46 w2+w4) landing.** It holds this file as a live fence, and both
defects below were found by its verification pass. The line citations here must
be re-derived after it lands rather than trusted — #155 moves anchors in this
file and in `scripts/gate.sh`.

Sibling lanes own other surfaces: `scripts/gate.sh` belongs to #73 (w3), and
`.swarm/coupling.txt` to #72 (w5). Do NOT touch either.

## Why one issue and not two

Both defects live in the same file, so they are one fence. Filed separately they
would be a guaranteed fence-lint overlap — a rebase conflict scheduled in
advance — which is exactly what that check hard-fails on.

## The defects

Both were found by PR #155's verification and deliberately not fixed there: each
would have widened scope beyond what that PR verified. #155's body lists them as
follow-ups 3 and 4 and says they are worth filing; this is that filing.

### 1. The structural predicate is blind to four producer spellings

`gate-honesty-law.test.ts:437` — ruling 1's predicate is *"every `VAR=$(...)`
line whose exit status is checked neither on the same line nor on the next."*
It only recognises the bare `VAR=$(` spelling. It does not see:

| spelling | seen? |
|---|---|
| `VAR="$(cmd)"` (quoted) | no |
| `export VAR=$(cmd)` | no |
| ``VAR=`cmd` `` (backticks) | no |
| multi-line `VAR=$(` … `)` | no |

**Pre-existing, not a regression** — #155 introduced none of these. But the law's
whole claim is that it convicts *the class*, and it currently convicts one
spelling of it. A producer written any of these four ways is unchecked and the
law stays green, which is the failure mode this law exists to abolish.

This is the same shape prd-46 is named for: **the class, not the enumeration.**
`#174` (prd42 w11) is the same lesson on a different grammar — enumerate the
productions as a table and give each a verdict, rather than widening the
alternation one spelling at a time.

### 2. A proof degrades to vacuity under root instead of skipping

One old-form `lanes.json` proof silently passes when running as root, while its
sibling one screen away openly declares itself skipped
(`it.skipIf(RUNNING_AS_ROOT)(...)`, `:1812`, the standard #74 landed).

A third instance of exactly this class was found and fixed during #155's review
(review of #155, the empty-diff fence proof): a fixture labelled EXECUTED
that could not fail on the platform it ran on, because `runFragment` invoked
PATH's bash rather than `/bin/bash`. It now runs `/bin/bash` and skips out loud
above 4.4. **That makes this the second known instance of the class in this file
and the third overall, so the fix should close the class rather than the
instance.**

## Definition of done

- The predicate recognises the quoted, `export`-prefixed, backtick and
  multi-line spellings, or the tolerance table states which are deliberately out
  of scope and why. Whichever is chosen, the comment matches what the code does.
- **The productions are enumerated as a table with a verdict per row**, in the
  file, so the next reader can see which spellings were considered rather than
  inferring it from the regex. A spelling nobody listed is a spelling nobody
  reviewed.
- EXECUTED both directions per spelling: a producer written that way with an
  unchecked status is convicted, and one with a checked status is not.
- The pinned counts (`15` total / `1` unchecked / `10` same-line / `4`
  next-line at #155's head) are re-derived after the predicate widens, not
  retyped — a widened predicate almost certainly moves them.
- The root-vacuous `lanes.json` proof either holds as root or skips and says so,
  matching `:1812`.
- **No proof in this file passes on a platform where it cannot fail.** A sweep,
  not a fix of the two known instances: every `EXECUTED`-labelled fixture is
  checked for a platform or privilege condition that makes it vacuous, and each
  either holds or skips with the reason in its title.
- The law is run 4x concurrently as part of the proof; this file raced itself
  once before.

`scripts/gate.sh` IS the landing tool: it merges to local `main` and pushes to
`origin`. Never run it to check any of this. Every probe uses a block extracted
into a scratch repo, which is the standard this file already sets throughout.

## Provenance

PR #155's verification pass (`prd46 w2+w4`), follow-ups 3 and 4, plus a third
instance of defect 2 found while reviewing that PR and fixed there.

gabriel-canaan:
Closes #179
Closes #72

prd-46 wave 5. Both issues are the same wave with disjoint fences, so they pay
the queue toll once.

| commit | fence |
|---|---|
| `f52b74e` `test(gate)` (#179) | `packages/server/src/gate-honesty-law.test.ts` |
| `82f4e75` `fix(gate)` (review of #179) | same |
| `06aea73` `docs(coupling)` (#72) | `.swarm/coupling.txt` |

## #179 — the law convicts the CLASS of assignment, not one spelling

Ruling 1's predicate recognised only bare `VAR=$(`. It replaces that regex with a
scanner — a keyword-prefix token loop, `$()`/backtick/double-quote skippers, a
heredoc-body map — rather than widening an alternation one spelling at a time.
That instrument change was the point: this file had recorded the same
one-spelling-at-a-time defect six times.

## #72 — the coupling registry records a collection-time hold

`gate-honesty-law.test.ts` anchors assertions on exact line TEXT in
`scripts/gate.sh`, so rewording an anchored line fails during **collection** and
takes the whole file out of service — not one red test. The entry cites **anchor
text, not line numbers**: the same anchors have had three different sets of line
numbers within this one milestone.

## What is the sibling case?

**The first round of #179's own fix was the sibling case, and it made the law
weaker than before.** Verification found it and it is the reason `82f4e75`
exists.

For all five keyword prefixes bash reports the **keyword's** exit status, not the
substitution's:

    V=$(false) || echo GUARD_RAN          -> GUARD_RAN
    export V=$(false) || echo GUARD_RAN   -> (nothing)
    next-line RC=$?:  bare -> 1           export -> 0

So `export V=$(cmd) || fail "..."` looks like a guard, reads correct to a human,
and **cannot fire**. Round 1 credited it as checked and shipped three fixtures in
`HONEST_LINES` certifying that shape as honest. Before #179 such a line was
merely invisible; after round 1 it was *seen and excused* — strictly worse.
`82f4e75` forces keyword-prefixed producers unconditionally into the unchecked
set via `hasKeywordPrefix`, moves those three fixtures to `RIGGED_LINES`, and
records the bash semantics as the reason.

Four more from the same pass: arithmetic `$((…))` convicted as a command
substitution (no command runs); heredoc bodies scanned as assignments; and
`declare -r`, `export readonly`, `V+=$(cmd)` invisible **and** unlisted — the
table's own contract being that an unlisted form is an unreviewed one.

Re-checked with spellings the findings did **not** use, because re-running the
probe that found a defect proves only that one input is caught:

    typeset V=$(cmd) || fail            unchecked=1   guard correctly NOT credited
    declare -x -r V=$(cmd) || fail      unchecked=1   two flags, still seen
    readonly V=$(cmd) / RC=$?           unchecked=1
    CONTROL  bare V=$(cmd) || fail      unchecked=0   a REAL guard still credited
    CONTROL  bare V=$(cmd)              unchecked=1
    COUNT=$((COUNT+1)) / $(( X * 2 ))   producers=0   false positive gone
    heredoc, unquoted <<EOF             producers=0
    V+="$(cmd)"                         producers=1   now recognised

The two CONTROLs are the load-bearing rows: they show the fix did not simply mark
everything unchecked.

Both seats independently recomputed the pinned counts from the real `gate.sh` and
agreed with the file: **17 total / 1 unchecked / 11 same-line / 5 next-line**.
Mutation-wise, disabling each new scanner branch reddened 3, 2 and 5 **named**
tests respectively.

For #72, both seats reproduced the coupling in isolated clones: rewording
`fail()` produced `expected exactly one line … found 0` and `Tests: no tests`.
The one finding against it — that "anchors *every* assertion" overstated the
scope — is fixed in `06aea73`; the collection-time claim itself is unchanged.

## Fence widening

**None.** `git diff --name-only origin/main...HEAD` is exactly the two fenced
files.

## Verification

Full pipeline per issue-pair: two independent seats (Opus `--effort xhigh`, Codex
`gpt-5.6-terra` `xhigh`) on the assembled tree with `main` merged, each
enumerating the shell-assignment grammar as a table with a probe per row.
**Both returned not-merge-ready**; the findings above are theirs, re-executed by
the orchestrator before being acted on, then repaired by the lanes and re-checked
by execution.

**Declared gap:** the repairs themselves have not been through an independent
fix re-review pass. On the sibling PR (#187) that pass found a real defect in a
repair, so this is a known, stated risk rather than an oversight — the repairs
here were verified by orchestrator execution only, with the fresh-spelling table
above as the evidence.

Gates on this branch, based on `f95f9d3`:

    VITEST_MAX_WORKERS=6 npm test     0   ->  387 files, 6987 passed


ciaran-slow:
## Reviewed commit-by-commit, on a clean clone at `06aea73`, node 22.23.2

**Verified in the PR's favour first.** Nearly everything this PR claims, holds:

| claim | how I checked | result |
|---|---|---|
| pinned counts `17 / 1 / 11 / 5` | ran the predicate against the real `scripts/gate.sh` | **17 producers confirmed** |
| fence: exactly the two declared files | `git diff --name-only origin/main...HEAD` | **confirmed, no widening** |
| keyword-prefix blocker (`82f4e75`) | `export V=$(cmd) \|\| fail` → `unchecked=1`; bare `V=$(cmd) \|\| fail` → `unchecked=0` | **both correct — the CONTROL is the load-bearing half and it passes** |
| row 19 arithmetic exclusion | `V=$((1+2))` → `producers=0` | **confirmed** |
| row 21 heredoc bodies | quoted + unquoted delimiter, scan resumes after terminator | **confirmed** |
| multi-line producer (row 7) | checked → `unchecked=0`; unchecked → `unchecked=1` | **confirmed** |
| #72's anchors exist in `gate.sh` | grepped each cited anchor | **all present**; `"timing pass matches ZERO files"` at `gate.sh:264`, ratchet rise-is-reported-never-failed at `:267-270` and `:360` |
| #72's entry is actually parsed | `scripts/fence-lint.sh 179 72` | **PASSED**, and the new entry shows as `ok` — owned by #179's fence |
| gates | `typecheck` 0, `lint` 0, `VITEST_MAX_WORKERS=6 npm test` → 387 files, 6987 passed | **reproduced green** |
| CI actually ran | `gh pr checks 191` | **all 7 expected jobs executed** (3 `build-test-boot` legs + 4 `pack-smoke`); macOS×min-node excluded by design |

The declared gap in the PR body — that the repairs had no independent fix re-review pass — is what this review is. It found one thing.

---

## Finding — `EXECUTED` — row 21's opener scan fires on two `<<` that open nothing

Fixed in **`6103a28`**, pushed to this branch.

`findHeredocStart` decides a line opens a heredoc by matching any `<<` whose next word looks like a delimiter. Two spellings reach that match without opening a heredoc:

```
grep -q x <<< foo          # `<<<` here-string — operand is stdin, not a body
if (( a << b )); then      # `<<` inside `((...))` — the LEFT-SHIFT operator
```

Each is read as an opener whose terminator never arrives, so `computeHeredocBodyLines` marks **every remaining line** as heredoc body and `findAllProducers` skips all of them.

**EXECUTED, against the real `scripts/gate.sh`, predicate as shipped** — one inserted line above the first producer:

```
real scripts/gate.sh                    producers=17
+ 'grep -q x <<< foo'                   producers=0
+ 'if (( a << b )); then :; fi'         producers=0
CONTROL, no intruder                    producers=17
```

**This is the sibling case, and row 21 names it itself.** The row already identifies this exact direction as the dangerous one — *"it would hide every real producer between the false opener and wherever a same-named terminator line next happens to occur"* — and its CONTROL proves a `<<EOF` inside a **quoted message** is not mistaken for an opener. These two are the structurally identical sibling: the `<<` is real, unquoted text and still opens nothing.

**Bounding it honestly: this is substantive, not a blocker.** The law does not silently pass — the pinned-count test asserts 17, so a count of 0 reddens it. What it reddens *with* is a count pointing nowhere near the offending line, and a mid-file here-string producing a 17→9 drop looks exactly like a pin that needs re-deriving. Re-deriving it would blind the predicate below that line for real — the same *seen-and-excused is worse than invisible* shape `82f4e75` fixed for keyword prefixes. `gate.sh` has no `<<` of any kind today, so nothing is live.

One trap worth flagging, since it defeats the obvious fix: refusing to match at the first `<` of `<<<` does **not** work — the loop retries at `i + 1`, finds the trailing `<<`, and reads the operand as a delimiter anyway. All three characters must be skipped. And `(( a << 2 ))` was safe only by accident: the delimiter regex requires an identifier and `2` is not one, so `(( a << b ))` fires while `(( a << 2 ))` does not.

**The fix** adds a three-character skip for `<<<` and a paren-depth skip for an arithmetic `((...))` span, plus production-table **row 22** (the file's own contract: a form nobody listed is a form nobody reviewed) and two fixtures.

**Mutations, both EXECUTED and reverted before commit:**

```
1. drop the `<<<` three-character skip     -> 2 failed | 180 skipped
2. drop the arithmetic ((...)) span skip   -> 2 failed | 180 skipped
```

Both new tests carry CONTROLs so the fix cannot pass by disabling heredoc detection wholesale: a real `<<EOF` still hides its body, and a real heredoc on a line that *also* carries an arithmetic span is still found.

Re-verified after the fix: `typecheck` 0, `lint` 0, law file **182 passed** (was 180), full suite **387 files / 6985 passed / 0 failures**, `fence-lint.sh 179 72` PASSED. Fence unchanged — `gate-honesty-law.test.ts` is #179's own.

---

**The third `chmod 0o444` site (`:2540`) is arguably still root-vacuous.** The commit body argues it needs no `skipIf` because the assertion "never inspects the file's final content, only that the OLD script's `catch {}` prints pruned either way." That defends *the assertion stays true*, which is not quite *the proof stays meaningful*: under root the chmod does not hold, the file is writable, and the case degenerates into a duplicate of the malformed-JSON case directly above it in the same test — it stops exercising the unwritable dimension it exists for.

Against that: you investigated it explicitly rather than assuming, ran it under `unshare -r`, and recorded the reasoning in the file, which is what #179's DoD asked for. Reasonable people differ on whether "passes for a different reason" counts as "passes where it cannot fail". Flagging it, not blocking on it — entirely your call whether it wants an `it.skipIf(RUNNING_AS_ROOT)` for symmetry with `:2567` and `:2729`.

---

## What I did NOT run

- **`scripts/gate.sh`** — never run by a reviewer; it *is* the landing.
- **The 4x-concurrency pass and the serial timing pass.** I reproduced the single-run suite only. Your `f95f9d3` 4x figures stand unchallenged by me rather than confirmed.
- **The Windows leg** — does not exist in `ci.yml` for this grid.

## Verdict

**Approve, with `6103a28` on the branch.** Both issues meet their DoD; the one defect I found is fixed with a mutation proof and a table row. I am not merging: per this repo's practice the merge is a human act, and GitHub would not let me formally approve a branch I have now committed to — so treat this as a reviewed-and-fixed handoff rather than as two approvals.

PR #192, opened 2026-09-01
prd42 w11: the barrel law derives names for the whole fail-open family

### LINKED ISSUE #185: prd42 w11: the barrel law's grammar covers the fail-open family, not one form at a time
## Fence (may touch ONLY)

## Blocked by
**PR #183 (prd42 w11) landing.** It holds this file as a live fence and changes
it substantially. Every line citation below is against #183's head and must be
re-derived after it lands rather than trusted.

**Model:** sonnet. **Wave:** 11 — the SAME wave as #174, deliberately, rather
than a new one. This is #174's own remainder in #174's own file; a w12 would
have been a wave invented for one follow-up, and the tracker would have claimed
a wave `docs/prds/prd-42-one-path-one-spelling.md` never declares. EXECUTED:
filed as w12 first, and `scripts/dev/prd-reconcile.sh 42` immediately reported
`UNDECLARED WAVE`; retitled to w11 and it reconciles clean again.

`scripts/fence-lint.sh 174 185` will report an OVERLAP on this file. It is
**historical** once PR #183 lands and #174 closes — fence-lint reads fences, not
issue state, the same way `fence-lint 47 72` still reports a closed issue's
claim on `.swarm/coupling.txt`. Do not dispatch this alongside a reopened #174.

## The defect

`rotateModuleExportNames` derives the names a barrel must republish. Four legal
export forms derive NOTHING, so the completeness guard passes on a barrel that
omits them — it fails OPEN. All were found by #183's verify pipeline, all
EXECUTED against the real helper, and all deliberately NOT fixed there: they
would have widened a wave reviewed for something else.

**None occurs in `rotate.ts` today** — an EXECUTED sweep found zero instances of
each in production recorder source. They are filed as a family because the
enumeration is the point: this file has now recorded the same shape SIX times
(`#154`'s enum/namespace/abstract, `#154`'s `export type { X }`, and the four
`#174` closed), and each round found it one spelling at a time.

| production | derives | direction |
|---|---|---|
| `export const a = {\n  x: 1,\n}, b = 2` (multi-line declarator list) | `["a"]` | **fails OPEN** — `b` is a real binding, silently dropped |
| `export const { a, ...rest } = obj` | `["a"]` | **fails OPEN** — `rest` is a real binding |
| `export const [a, ...rest] = arr` | `["a"]` | **fails OPEN** — same |
| `export * as ns from './x.js'` (module side) | `[]` | **fails OPEN** — `ns` is a real export name |
| `export const a: Map<string, number> = …` | `["a","number"]` | fails closed — noise |
| `export default function*gen() {}` (no space) | `[]` from `defaultExportLocalNames` | fails closed — spurious gap |

The multi-line row is the most reachable: `[^\n;]*` stops at the first newline,
and wrapping a declarator list at `lineWidth: 100` is ordinary formatting, not a
contrived input. It was mislabelled "fails closed, low" by one review seat and
corrected by another — the direction matters, because a dropped name is silent
and a spurious one is loud.

The rest-binding rows have a second problem: nested patterns are a DOCUMENTED
exclusion with a CONTROL pinning them, and rest elements are not — the skip is
silent, which is the one undocumented exclusion in the function.

## Done when

- Every production above has a row in the enumerated grammar table **before**
  the derivation changes, per that table's own contract, with a verdict.
- Each fail-OPEN form derives its real binding(s), or the doc says plainly why
  it is excluded — a silent exclusion is what this issue exists to end.
- `export * as ns from` gets a RULING rather than silence: a namespace re-export
  changes HOW a caller reaches the name, and whether that satisfies completeness
  is a decision the file currently declines to make. Decide it and say so.
- Mutation-certified, each new arm reddening a NAMED row.
- **Assertions that pin an absence use exact equality, not `toContain`.** #174
  shipped two `toContain` rows for a defect whose symptom was an EXTRA name;
  reverting the fix left the file green at 52/52. Use `toEqual`.
- The fail-closed rows (generic-type noise, `function*gen`) may be left as
  documented residuals — but say which, rather than going quiet.

## Proving it

Run the real helper, not a retyped copy. `~/.claude/skills/verify/scripts/certify-mutation.sh`
takes the original line from git and verifies its own restore; use it rather
than hand-rolling the revert/run/restore loop.

## Provenance

PR #183's verify: two independent seats, two cross-reviews, a reconcile, and
four fix re-review rounds. Every row above is EXECUTED with a control. Recorded
in #183's body as deliberately out of scope.

gabriel-canaan:
Closes #185

prd-42 wave 11. Ships alone: #185 is the only open prd-42 issue, so there is
nothing in its wave to bundle with.

It stays in **wave 11** rather than opening a w12 because it is #174's own
remainder in #174's own file. Filed as w12 first, `scripts/dev/prd-reconcile.sh 42`
immediately reported `UNDECLARED WAVE` — the PRD declares no w12.

## What becomes true

`rotateModuleExportNames` derives the names a barrel must republish. **Four legal
export forms derived nothing**, so the completeness guard failed OPEN — it passed
on a barrel omitting a real binding. This adds arms for multi-line declarator
lists, object and array rest elements, and `export * as ns from`, and takes a
ruling on the last of those rather than leaving it silent.

None of the four occurs in `rotate.ts` today. They are filed and fixed as a
family because the enumeration is the point: this file has recorded the same
one-spelling-at-a-time defect **six times**, and the table is cheaper than a
seventh round.

## What is the sibling case?

**The first round fixed the multi-line VALUE and left the wrapped DECLARATOR
LIST — the same production.** Verification measured both against the shipped
helper:

    export const a = {\n  x: 1,\n}, b = 2   -> ["a","b"]   fixed by round 1
    export const a = 1,\n  b = 2            -> ["a"]       b silently dropped

The bracket-depth fix never touched the second, because that wrap has no bracket
to track. Both are now pinned with exact equality.

Three more fail-OPEN forms were found and are now **declared with a verdict and a
reason each**, with a CONTROL test pinning all three so an accidental change
reddens:

- **nested destructuring** (`export const { outer: { nested } } = value`) —
  deliberately left; reading only the outer key would mint a name nobody can
  import, and reading it correctly needs a real parser rather than a comma split.
- **module-side `export * from './helpers.js'`** — deliberately left, and
  asymmetric with its `export * as ns from` sibling **on purpose**: `as ns` names
  something locally whether or not the target resolves, while a bare
  `export * from` has no local name at all and deriving its names means resolving
  and reading an arbitrary file.
- **a barrel string literal containing re-export text** — fails OPEN on the
  barrel side, the mirror of the already-declared module-side residual that fails
  CLOSED. Same root cause (neither side is string-literal-aware), opposite
  consequence.

That is the DoD's own alternative: derive the binding, **or** say plainly why not.
A silent exclusion is what the issue exists to end.

Re-checked with spellings the findings did not use:

    REPORTED  export const a = 1,\n  b = 2        -> ["a","b"]     fixed
    FRESH     three wrapped declarators           -> ["a","b","c"] the class, not the instance
    FRESH     export default function*tight(){}   -> ["tight"]     fail-open closed
    CONTROL   export const a = 1, b = 2           -> ["a","b"]
    CONTROL   export function realDoor() {}       -> ["realDoor"]

Each new arm reddens a **named** row when reverted: the multi-line walker on
missing `b`; the rest arm on missing `rest` including the completeness gap; the
namespace-export arm on missing `ns`; the barrel ruling on an incorrectly
satisfied `realDoor`.

Absence assertions use `toEqual`, not `toContain` — #174 shipped two `toContain`
rows for a defect whose symptom was an EXTRA name, and reverting its fix left the
file green at 52/52.

## A premise in the issue body that does NOT hold

The issue calls the wrapped declarator list *"ordinary formatting, not a
contrived input"* at `lineWidth: 100`. **That is false.** `biome.json` sets
`"formatter": { "enabled": false }` — only the linter runs, and that `lineWidth`
sits in a dead block. Nothing in this repo emits that wrap automatically.

The defect is real and worth closing because the form is legal and hand-writable.
The reachability argument is not, and it is recorded here so it does not get
copied forward.

## Fence widening

**None.** `git diff --name-only origin/main...HEAD` is one file.

## Verification

Two independent seats (Opus `--effort xhigh`, Codex `gpt-5.6-terra` `xhigh`) on
the assembled tree with `main` merged, each enumerating the TypeScript export
grammar as a table with a probe per row. **Both returned not-merge-ready**; every
finding above is theirs, re-executed by the orchestrator before being acted on.

**Declared gap:** the repairs have not been through an independent fix re-review
pass. On the sibling PR (#187) that pass found a real defect in a repair, so this
is a stated risk rather than an oversight — the repairs here were verified by
orchestrator execution only, with the fresh-spelling table above as evidence.

**Two forms remain undeclared**, found in that re-check and worth a table row each
rather than a code change (neither is reachable from any formatter, since none is
enabled):

    export const a = 1\n  , b = 2       -> ["a"]   comma-at-BOL
    export type * from './helper.js'    -> []      no mention in the file

Gates on this branch, based on `f95f9d3`:

    VITEST_MAX_WORKERS=6 npm test     0   ->  387 files, 6954 passed


ciaran-slow (commented):
Reviewed commit-by-commit on a clean clone at `b1500ea`, Node 22.23.2. **Everything the PR body claims, I re-ran and it holds.** One fail-OPEN sibling of the arm this PR adds was still open; I fixed it in `e2d392d` rather than hand it back.

## What verified in your favour — EXECUTED, all of it

**Your mutation claims are true, and understated.** I re-derived them rather than trusting the commit message. All four arms redden a *named* row:

| mutation | result |
|---|---|
| `\n`-at-depth-0 back to unconditional terminator | **3 failed** — the enumerated-forms table, the named "wrapped at the COMMA … round 2 (#185)" row, and its end-to-end |
| `defaultExportLocalNames` back to `function\s*\*?\s+` | **1 failed** — the named "TIGHT generator default … round 2 (#185)" row |
| rest-element derivation removed | **3 failed** — the table, the named rest row, and its end-to-end |
| the new `export * as ns from` arm removed | **3 failed** — the table, the named namespace row, and its end-to-end |

**The amplification risk in "change the instrument" does not materialise.** Walking raw source instead of a pre-cut line means an unbalanced bracket can drive `depth` negative and never recover, which could in principle swallow the rest of the file. It does not, because `matchAll` restarts the walk per declaration. EXECUTED against the real helper:

```
export const msg = 'oops )' / 'oops {' / 'ok'   ->  ["keep","msg","later"]  (identical in all three)
```

**The new comma-continuation rule does not mis-fire.** The case I most expected to break it is a depth-0 comma inside a generic — `<>` are not tracked as depth — wrapped right after the comma. `DECLARATOR_BINDING_RE`'s anchor holds:

```
export const m: Map<string,\n  number> = new Map()   ->  ["m"]        no phantom `number`
export const a = {\n x: 1,\n}, b = 2\nexport const c = 3  ->  ["a","b","c"]   no bleed
export const a = 1,\n  b = 2,\nexport const c = 3        ->  ["a","b","c"]   no phantom `export`
```

**Your CORRECTION to the issue's own premise is right**, and worth keeping: `biome.json:12` is `"formatter": { "enabled": false }` with `lineWidth: 100` sitting in that dead block. The reachability argument in #185's body is false; the defect is real anyway.

**Gates, my run, Node 22.23.2 on `b1500ea`:** `typecheck` 0 · `lint` 0 · `VITEST_MAX_WORKERS=6 npm test` 0 → 387 files, 6950 passed. File alone: 64/64.

**CI:** both jobs, all 7 legs green, and I confirmed the steps actually executed rather than reading the bar — the macOS leg ran Build → Test → Typecheck → Lint → Packaging guard → Boot smoke, none skipped.

**Fence honoured:** `git diff --name-only origin/main...HEAD` is exactly `packages/server/src/recorder/namespace-law.test.ts`.

## Finding — fixed in `e2d392d`, needs your confirmation

**The arm you added has a type sibling that was still fail-OPEN.** EXECUTED on your head, via the real helper:

```
export * as ns from './helpers.js'        ->  ["ns"]
export type * as tns from './helpers.js'  ->  []      <- fails OPEN
```

`tns` is a real local export name either way, and `rotateModuleExportNames` already derives type-only names (`export type X`, `export interface X`) into the same set — so a barrel omitting it read as complete. It was the one member of the family carrying neither an arm nor a verdict: not a row in the table, and not in the PR body's undeclared list.

This is the shape the file has now recorded seven times, and it is the sibling of *this PR's own new arm* — the value spelling shipped, the type spelling one keyword away did not. That is why I fixed it here rather than filing round eight.

The widening is `\bexport(?:\s*|\s+type\s+)\*`, not the shorter `\bexport\s*(?:type\s+)?\*`. The first branch preserves your value form's zero-whitespace tolerance byte for byte; the second requires real whitespace before `type`, so it cannot newly match an identifier that merely starts with the keyword. Pinned by its own CONTROL (`exporttype * as sneaky from` → `[]`).

Mutation-certified both ways: reverting the widening → 3 failed (table + both named rows); reverting to the shorter spelling → 1 failed, on the `exporttype` CONTROL.

**Also: the two forms your body names as "undeclared" are now declared in the file.** A PR body is not the doc, and the DoD is explicit — *"the doc says plainly why it is excluded — a silent exclusion is what this issue exists to end."* Both now carry a verdict and a CONTROL row:

- `export const a = 1\n  , b = 2` → `["a"]` — comma-at-BOL, fails OPEN. **Declared, not fixed**, on your own stopping rule: that production has been defeated twice and rebuilt once, and a "next non-whitespace is a comma" exception would be a third patch on it by the same method. What closes it is the real-continuation awareness your `export const c =\n  1` residual already names — one tool, both shapes. Your call if you disagree.
- `export type * from './helpers.js'` → `[]` — the type spelling of the bare-star residual, benign for the same reason (no local name). Recorded because its `as ns` sibling is now derived in *both* spellings, and a reader shouldn't have to guess which halves of that grid are covered.

**After `e2d392d`:** file 66/66 · `typecheck` 0 · `lint` 0 · full suite 0 → 387 files, 6952 passed.

## What I did NOT run

- **Nothing was run against `main` merged in.** `origin/main` has moved 9 commits past your base (`f95f9d3` → `ebc7681`). None of them touch this file, but my suite runs are pre-merge; `land.sh` merges main in and re-runs before it will merge.
- CI has not yet run on `e2d392d`.
- I did not run `scripts/gate.sh` — that is the operator's step.

## Where this sits

Not approving: the fix in `e2d392d` changes derivation behaviour, and approving my own substantive fix is the thing the runbook says not to do. **Over to you** — confirm `e2d392d` (or push back on the comma-at-BOL call, which is the one judgement in it), and it's ready for the operator to land.

gabriel-canaan:
## `e2d392d` confirmed — re-run here, including two spellings you did not use

Checked out `origin/prd42-w11-185` at `e2d392d` in a clean detached worktree and
drove the real `rotateModuleExportNames`, rather than reading the diff:

    ["ns"]       export * as ns from './h.js'              value form, unchanged
    ["tns"]      export type * as tns from './h.js'        the fix
    ["spaced"]   export  type  *  as  spaced from './h.js' FRESH — extra whitespace
    ["dq"]       export type * as dq from "./h.js"         FRESH — double quotes
    []           exporttype * as sneaky from './h.js'      your CONTROL, holds
    []           export type * from './h.js'               bare type-star, declared

The two fresh rows are the point: re-running your own probe would only have shown
that one input is caught. The widening generalises across whitespace and quote
style, and `exporttype` still returns `[]`, so the `\s+type\s+` branch is doing
what the shorter spelling would not have.

**Your reading of the defect is right and I had it wrong.** I listed
`export type * from` in the PR body's undeclared set as a bare-star sibling. It is
that, but the one that mattered was `export type * as tns from` — the type
spelling of *this PR's own new arm*, which I did not probe at all. `tns` is a real
local name and the helper already derives type-only names into the same set, so a
barrel omitting it read as complete.

## On the comma-at-BOL judgement — I agree, and would not want it fixed here

Declaring it is the correct call, and it is the stopping rule applied honestly
rather than as an excuse. That production has been defeated twice and rebuilt
once; a "next non-whitespace is a comma" exception would be a third patch on the
same instrument by the same method, which is the pattern this file has now
recorded seven times. Your framing — that real-continuation awareness closes it
together with the `export const c =\n  1` residual, one tool for both shapes — is
the better boundary, and it belongs to whoever builds that tool.

## And you are right that a PR body is not the doc

Moving those two into the file with verdicts and CONTROL rows is the DoD being
honoured, not a nicety. I had recorded them where a future reader of the *file*
would never see them, which is the exact failure the "a form nobody listed is a
form nobody reviewed" contract exists to prevent.

## What I did not run

I did not re-run the full suite on `e2d392d`, and I did not merge `main` in —
`origin/main` is now 9 commits past this base (`f95f9d3` → `ebc7681`), none of
them touching this file. Your pre-merge figures stand as the last full run;
`land.sh` merges main and re-runs before it will merge.

No objection to anything in `e2d392d`. Over to the operator.

Issue #190, opened 2026-09-01
scene: the variance question is re-asked if the fleet outgrows the measured regime


ciaran-slow:
prd-47 ruling 4 was answered NO-GO on 2026-09-01 and the pooling was not built.
That verdict is **conditional on the scale the instrument renders**, and this
issue carries the condition so it is a tripwire rather than a memory.

> **This body was rewritten on 2026-09-12 (prd-49 ruling 3).** It had carried the
> original pre-composition table and a threshold of *"~60 % of budget on the
> reference box"* — both superseded, twice, by amendments recorded in the paper and
> in comments below it. An issue whose whole job is to carry a condition must state
> the live one, so the superseded numbers are gone from the trigger and kept below
> as history.

## The trigger, as it now stands

**A selectable stream source, or a real session, renders more than ~90 threads.**

That is the whole trigger. It is a fact about the product — the scale the
instrument renders — not a reading of anybody's box.

- **The shipped half is a law, not a memory.** `packages/web/src/scene/tripwire-law.test.ts`
  asserts that no source the operator can select composes more than 90 threads, and
  fails by name at prd-49 ruling 3 and this issue. Nobody has to be watching: when
  someone ships a bigger fixture, or the frame loop begins composing more than one
  colony, the suite says so.
- **The live half is a human observation, and says so.** The frame loop composes
  exactly one colony (`layoutWorld([{ id: LOCAL_COLONY, … }])`), so a session's
  threads are one repo's lanes and the count is on the fleet table in front of the
  operator. Nothing about a stranger's repo is assertable from here.

**Where it stands today:** not tripped. The largest selectable fixture is 20 lanes,
and the frame loop composes one colony.

## What to do when it trips — and what NOT to do

**Do not reach for pooling first.** The measurement already says it would be the
wrong instrument in that regime: at 180 threads the **median** frame is over budget,
not just the occasional one, and a free-list does not fix a median.

prd-49 ruling 2 binds, in order:

1. Re-run the forced-collection cells — still in `packages/web/src/scene/perf.test.ts`:
   `npx vitest run packages/web/src/scene/perf.test.ts --maxWorkers=1 --disable-console-intercept`
2. Re-record a Chrome trace at the new scale, against a **production build** — the
   2026-09-01 trace was taken on the vite dev server, which allocates more than
   production and therefore over-stated GC.
3. Only if the *tail* (not the median) is then the binding constraint, and GC is in
[...]

PR #194, opened 2026-09-01
prd47 w3: the fleet's beat, the hover path, and ruling 4's verdict

### LINKED ISSUE #158: prd47 w3: the fleet is rebuilt on its beat
## Fence (may touch ONLY)
- `packages/web/src/fleet/FleetContext.tsx`
- `packages/web/src/fleet/FleetContext.test.tsx`
- `packages/web/src/panels/ledger/index.tsx`
- `packages/web/src/panels/ledger/index.test.tsx`

## Blocked by
**#157** (wave 2) — wave ordering. Nothing within wave 3 depends on this, and nothing here depends on the other two wave-3 issues — the three fences are pairwise disjoint.

Sibling lanes own other surfaces. scene/view/ and scene/SceneView.tsx belong to the hover issue in this same wave; scene/perf.test.ts belongs to the variance issue in this same wave. Do NOT touch either.

Two coupling points checked by hand, because fence-lint.sh's WARN branch cannot reach them and neither is in this fence. packages/web/src/App.test.tsx mocks eight lazily-loaded panels and is forced only by **adding, renaming or moving** a lazy panel — this issue does none of those. packages/web/src/panels/ledger/no-panel-refolds.test.ts greps every panel's source for a `reduceAll` import — selector work adds none. If either turns out to redden, stop and record a widening on the PR **before** the change.

## Definition of done

- **The fleet is rebuilt on its stated beat** (PRD success 4). An SSE event that changes nothing `buildFleet` reads must not re-render a FleetContext consumer; the 1 Hz tick remains the only scheduled rebuild.
- **The defect is the memo key, not the design.** `fleet/FleetContext.tsx:74-77` memoises on `[state.session, clock, manifest]` under a comment that already states the intent — *"the fleet is rebuilt on a beat rather than per event"* — while `state.session` is a fresh reference from `reduce()` per event, so the beat is fiction and five consumer subtrees re-render. Ruling 3 makes two existing comments true rather than adding architecture.
- **The ledger's selectors key on the slices they read**, not on whole-`session` (`panels/ledger/index.tsx:46-58`).
- **`FLEET_TICK_MS` and the clock already in the memo key are reused, not replaced** — the PRD lists them under "what already exists (do not rebuild)".
- **A counting law, not a wall clock** (PRD success 6): assert render generations per event, not milliseconds.
- **The context-selector refactor is explicitly NOT in scope.** Ruling 3 refuses it by name — 25 `useStream()` call sites, spike first, it re-plumbs half the app — and prd-47 open question 3 leaves its home undecided. A change that touches those call sites is out of scope and belongs on the open question.

### LINKED ISSUE #159: prd47 w3: the hover path reads no layout and the cursor costs no render
## Fence (may touch ONLY)
- `packages/web/src/scene/view/hitTest.ts`
- `packages/web/src/scene/view/hitTest.test.ts` (new file — this module has no test today)

## Blocked by
**#157** (wave 2) — it holds useFrameLoop.ts, and so does this. Wave ordering already sequences them; do not start before wave 2 is on `main`. Nothing within wave 3 depends on this, and no other wave-3 issue claims any path above.

Sibling lanes own other surfaces. fleet/FleetContext.tsx and panels/ledger/ belong to the beat issue in this same wave; scene/perf.test.ts belongs to the variance issue in this same wave. Do NOT touch either.

Path note, checked against `main` rather than copied: the PRD's Sequencing writes this lane as view/hitTest.ts + SceneView.tsx, but SceneView.tsx lives at packages/web/src/scene/SceneView.tsx, not under view/. And `setPanning` is defined and used entirely in useFrameLoop.ts (`:149`, `:398`, `:409`) — SceneView.tsx only consumes `panning` for the cursor — which is why this fence reaches the hook.

## Definition of done

- **The hover path reads no layout.** `pickAt` opens with `canvas.getBoundingClientRect()` per mousemove (`view/hitTest.ts:47`), a forced-layout risk on every hover. The ResizeObserver at `useFrameLoop.ts:236-242` already has the measured rect; publish the two fields it holds and read those instead.
- **The rect stays correct on scroll.** The audit names this as the trap in the obvious fix — a cached rect that never refreshes is wrong the moment the page scrolls. A law covers it.
- **The cursor costs no render.** `setPanning` routes a cursor change through React state on the first move of every gesture. The gesture bracket itself is listed under "what already exists (do not rebuild)" and must survive — ruling 1's repaint path depends on it.
- **A counting law, not a wall clock** (PRD success 6). The per-event handler is already 15–21 ns and hit-test 0.9–2.2 µs, so a timing here measures the box; assert layout reads and render counts.
- **SceneView.test.tsx carries a `// @gate-timing` marker and it must survive this change.** scripts/gate.sh derives its serial timing set from that marker and ratchets .swarm/timing-count; un-marking, renaming or merging away one timing test breaks the operator's landing tool twice over, and only a human can clear the ratchet. Do not add the marker to any new file either.
- **Quadtree hit-testing is a rejected alternative**, refuted in the audit with numbers: the linear scan is 0.9–2.2 µs and is skipped during drag anyway by d3's capture-phase `stopImmediatePropagation`.

### LINKED ISSUE #160: prd47 w3: the variance is attributed and the tail is printed

ciaran-slow:
Wave 3 of prd-47, three issues that share no file between them — the fleet's beat, the hover path, and ruling 4's spike. Three commits, one per issue, each standing alone.

Closes #158
Closes #159
Closes #160

Every lane was built and verified by a **different model** than built it (`Co-Authored-By` trailers): #158 and #159 built on sonnet and verified on opus; #160 built on opus and verified on sonnet. Verify reports are on each issue.

---

## `perf(fleet)` — the fleet is rebuilt on its beat, not on reference churn (#158)

`FleetContext` memoised on `[state.session, clock, manifest]`, but `core`'s `reduce` spreads the whole state object on **every** event to bump `eventCount` (`withEnvelope`, `reduce.ts:154`). So `state.session` was a fresh reference after an event that changed nothing, and five consumer subtrees re-rendered for it. The memo now keys on the eleven slices `buildFleet` actually reads; the ledger's four memos key on the slices each selector reads.

**What is the sibling case?** `FleetContext`'s memo and `LedgerPanel`'s four memos are the same defect in two places — hence one commit covering both. The third sibling, `LedgerPanel` re-rendering per event because it calls `useStream()`, is ruling 3's *named refusal* (25 call sites, spike first) and is deliberately untouched.

**What mutation would this test survive?** Reverting the key to `[session, clock, manifest]` fails 3 tests; re-keying the ledger's exemplars to `[session]` fails 2 — both executed at verify, including the positive controls, which fail under the mutation too and so cannot pass on a component that computes nothing. A partition law against `initialSessionState()` fails the moment `core` grows a slice that joins neither the input list nor the declared non-inputs.

**Deviation from the plan.** The plan's prose for one ledger case claimed `tool.activity` should not move the spend rows. It was wrong: `BranchSpend.toolCounts` reads `telemetry.tools`, so `tool.activity` genuinely is spend input. The builder proved it with a scratch test, traced it to `groupSpendBy`, and corrected the test's assertion rather than swapping it in silently. No source change resulted — `telemetry` already covers `tools`.

**One deliberate consequence, recorded rather than assumed.** `eventCount` is read by `buildFleet` but is *not* a key, because it changes on every event and keying on it would leave the ruling a no-op. Its only reader is StatusBar's event counter (`app/StatusBar.tsx:277`), which is therefore beat-refreshed — bounded by `FLEET_TICK_MS`, and only for events that move nothing else.

## `perf(scene)` — the hover path reads no layout and the cursor costs no render (#159)

`pickAt` opened with `canvas.getBoundingClientRect()` on every mousemove. The frame loop's ResizeObserver had already measured the same box, so it now publishes the origin and `pickAt` is pure. Separately, `panning` React state existed solely to set a CSS cursor, committing a render at both ends of every drag; the gesture bracket now writes `canvas.style.cursor` directly.

**What is the sibling case?** `pickAt` is called from `onClick` as well as `onMouseMove` — the click path had the identical forced read and is fixed by the same signature change, not just the hover path that names the issue. Second sibling, checked at verify: the two new `window` listeners are registered once and removed in **both** of the mount effect's cleanup paths (2 registrations, 4 removals); every `return` in the effect was enumerated to confirm no third exit leaks them.

**What mutation would this test survive?** Three, all executed at verify: removing the scroll listener fails 2 tests; reintroducing `getBoundingClientRect()` on the hover handler fails 2; and making the gesture bracket commit a React render again — which reproduces the *original defect* rather than merely deleting the fix — fails the cursor law. That last one establishes the `cursorOf`-call-count probe is a real render counter rather than a correlate.

**Fence widening**, recorded on the issue before the change: `packages/web/src/scene/view/input.ts` joined the fence so `cursorOf` could lose its now-unreachable `panning` parameter. One five-line function, one call site, no test file of its own. The alternative — `cursorOf(false, …)` forever — ships a dead branch.

The `@gate-timing` marker count is **3 on `main` and 3 on this branch**, so no timing test fell out of the gate's serial pass, and the marker is correctly absent from the new `hitTest.test.ts`. `useFrameLoop.test.ts` is byte-for-byte unmodified, so its `{ builds: 0, repaints: 4 }` drag law passed unedited.

## `test(scene)` — the variance is attributed and the tail is printed (#160)

Ruling 4 is measurement, and the answer is **NO-GO: the variance is not GC**. Two independent methods, both in `research/2026-08-28-variance-attribution.md`:

- **Ruling 4's own falsifier was run and not met.** The 60x3 cells under forced out-of-band collection: `plain 22.633 ms median / 1.25x tail` against `forced 7.821 ms / 1.66x tail`. The tail did not collapse — it rose. A throwaway A/B/C probe separated the confound (the forced sample always ran on a freshly collected heap): the effect tracks accumulated allocation pressure, not the act of collecting.
- **GC is not on the rAF timeline.** A Chrome trace of a real drag on the 20-lane fixture, 13.4 s, 764 frames: GC is **0.74%** of main-thread animation-frame time and **0 of 764** frames missed the 16.67 ms budget.

**What is the sibling case?** `:1517` was the model-floor line missing its `worst`; the structurally identical sibling one screen up (`:1440`) has printed one since #579 — verified against `git show main:` rather than taken from the plan. The two lines that print medians only (`:489` by-stage, `:746` by-builder) are deliberately left alone: a "worst" there would be a different frame's tail in every column.

**What mutation would this test survive?** The two count laws — matching marks between series, and both series being full length — exist to stop a forged collapse, since `Math.max(...[])` is `-Infinity` and `median([])` is `0`, so an empty forced series would report a `0x` tail and read as total collapse. At verify both were mutation-tested (a dropped sample, and a forged cheaper fleet); both caught it. The `collect === null` branch was forced via a probe copy and confirmed to print `NOT MEASURED` and exit before any assertion, so CI — which exposes no gc — cannot mistake it for a measurement that ran.

**Consequence, and it is binding:** `prd47 w4` is **not groomed**, the cross-PRD note for `scene/marks/` is not owed, and ruling 4 closes as answered. The decision is conditional and instrumented rather than final — the forced-collection cells stay in the file so re-asking costs one command, and **#190** (`later`/low) carries the tripwire for a fleet that outgrows the regime the browser could test, including an explicit instruction *not* to reach for pooling first if it trips, because at 180 threads the model floor's **median** is already 106.7% of budget and a free-list does not fix a median.

---

## Gates, on the assembled wave

Each lane gated alone; this is the first run where all three coexist, and on current `main` (base `f95f9d3`, which moved after the lanes branched).

| gate | result |
| --- | --- |
| `VITEST_MAX_WORKERS=6 npm test` | **388 files, 6973 passed**, 3 expected-fail, 6 skipped |
| `npm run lint` | pass |
| `npm run build` | pass |

No cross-lane breakage. No server or CLI surface is touched, so the boot smoke does not apply. Cherry-picks applied clean, which is itself the fence check: nothing in this wave shares a file.


gabriel-canaan (approved):
## Review — approve

Read commit-by-commit at `2d6e059`, in a detached review worktree with `origin/main` already an ancestor (the branch carries `f95f9d3`, so nothing was exercised against a tree it will not land on). Every finding below is labelled `EXECUTED` or `REASONED`. **Nothing here asks for a change** — the checks that could have blocked all came back in the author's favour, so I have led with those, as the evidence rather than the opinion.

### What I re-derived, and it held

**1. `FLEET_INPUT_SLICES` is not under-inclusive.** `EXECUTED.` This was the one defect that would have mattered — an under-declared slice makes the fleet render *stale*, and the partition law in `FleetContext.test.tsx` proves set membership, not read-set truth: it would pass just as green if `judge` were secretly an input. So I measured the read set directly, wrapping `SessionState` in an access-recording `Proxy` and running `buildFleet` over the `fleet20Spec` fixture:

```
TOUCHED           : agents branches collectors commits eventCount mainBranch panes session telemetry traces worktrees
READ-NOT-DECLARED : eventCount
DECLARED-NOT-READ : errors
```

`eventCount` is the documented, deliberate exclusion. `errors` is declared but was not reached on that fixture — it is a genuine input, read at `packages/core/src/fleet/ladder.ts:79` (`latestErrorFor(state.errors, …)`) on the collector-error path the fixture never enters, so declaring it is correct and conservative. **None** of the six declared non-inputs (`checkpoints`, `forks`, `judge`, `refusals`, `firstEventTs`, `lastEventTs`) is touched at all. The list is exactly right.

**2. All four ledger memo keys are exact or conservative.** `EXECUTED.` Same proxy technique, since these are hand-written narrow lists and therefore the riskier half of the commit:

| memo | actually reads | declared key | |
|---|---|---|---|
| `selectSpendByBranch` | `branches telemetry worktrees` | same three | exact |
| `heaviestLlmRequestSpanByLane` | `traces` | `traces` | exact |
| `selectThreadRowsForBranch` | `telemetry` | `+ branches worktrees` | conservative |
| `usageEventsByBranch` | — | `session.telemetry.usage` | keyed on the exact value passed |

No under-inclusion anywhere, so no stale-render path.

**3. The two new `window` listeners cannot leak.** `EXECUTED.` I enumerated every `return` in the mount effect (`useFrameLoop.ts:214`–`801`) by indentation rather than trusting the count: exactly two of them exit the effect — `:769` (the `pinned` branch) and `:785` (the main path) — and both remove both listeners. Every other `return` in that range sits inside a nested function. 2 registrations, 4 removals, as claimed.

**4. The three mutations reproduce, one of them harder than reported.** `EXECUTED.` I broke each thing myself rather than taking the verify report's word:

|---|---|---|
| drop the `scroll` listener | 2 fail | **2 fail** — the scroll-trap case and the repetition case |
| reintroduce `getBoundingClientRect()` on the hover path | 2 fail | **3 fail** — the counting law goes red too |
| revert the memo key to `[session, clock, manifest]` | 3 fail | **3 fail**, the positive control among them |

That last one is the point worth stating: the positive control fails under the mutation as well, so the falsifier pair cannot pass on a hook that rebuilds nothing.

**5. Ruling 4's verdict reproduces on different hardware.** `EXECUTED.` Ran the cell serially and alone on my own box:

```
model floor 60x3 plain:  26.603 ms median · 41.776 ms worst · 1.57x tail
model floor 60x3 forced:  9.832 ms median · 15.248 ms worst · 1.55x tail
```

Different absolute times from the note's (22.633 / 7.821), and the tail moves the other way here — 1.57 → 1.55 rather than 1.25 → 1.66. **Either reading is a NO-GO:** the tail does not collapse toward 1, which is the falsifier as ruling 4 stated it. The verdict is independently confirmed, and the note's "the ratios and the ordering are the durable figures, not the raw times" is exactly the right caveat to have written.

`research/2026-08-28-variance-attribution.md` is the strongest part of this PR. The confound is named and tested rather than waved at (Q1a's A/B/C ordering, C sitting near B not A), and *What this does not establish* gives each limitation a direction of bias — including the one that runs against the conclusion and is reported anyway. The 180-thread median already being 106.7% of budget is the argument that makes the NO-GO robust rather than merely measured.

**6. The mechanism claims check out.** `EXECUTED` where noted.
- Host/canvas box equivalence, which the whole origin-publishing fix rests on: host is `relative h-full w-full overflow-hidden` — no border, no padding — and the canvas is `absolute inset-0 h-full w-full`. One box, two elements, as the comment says. `REASONED` from the markup.
- `@gate-timing`: 9 marked files on `f95f9d3` and 9 on this branch, byte-identical sets; the three under `packages/web/src/scene/` are untouched and the marker is correctly absent from the new `hitTest.test.ts`. `EXECUTED.`
- `pickAt` and `cursorOf` have no remaining caller on the old signatures, and `FrameLoopResult.panning` has no consumer left. `EXECUTED` (grep + typecheck).
- `#190` is open and carries the tripwire; the ruling-4 verdict is recorded as a comment on `#160`, which its DoD required before wave 4 could be said not to exist. `EXECUTED.`

**7. Fences and hygiene.** `EXECUTED.` Each commit touches exactly its fence — 4 / 6 / 2 files, pairwise disjoint, so the "cherry-picks applied clean" claim is structurally true rather than lucky. The one widening (`scene/view/input.ts`) was recorded on `#159` at `20:34:27Z`, before the build report at `21:21:48Z`, with a rejected alternative stated. No changed file — the new research note included, which is the file most exposed to it — names a home path, a hostname or a real user.

### Gates I ran, and what I did not run

On the post-merge tree, at `2d6e059`, node 22.22.2:

| gate | result |
|---|---|
| `npm install` | pass |
| `npm run lint` | pass — 859 files |
| `VITEST_MAX_WORKERS=6 npm test` | **388 files, 6977 passed**, 3 expected-fail, 2 skipped |

`gh pr checks 194` — all **seven** jobs executed and passed, both `build-test-boot` and `pack-smoke` across the grid, so no leg's later steps were silently skipped behind an early failure.

I did **not** run `npm run build` or the 4x-concurrency load batches; CI's build legs cover the former and landing covers the latter.

One number to reconcile, and it favours the author: the body reports **6973 passed / 6 skipped**, I measured **6977 passed / 2 skipped** on the same 388 files. Four tests that skipped on your box passed on mine — environment-gated, and the ruling-4 gc cell is precisely that shape. Nothing structural; worth knowing only if someone later diffs the two.

Neither is a defect and neither blocks. Both are about a test being narrower than its comment, on code I verified is correct.

1. **The scroll-trap test does not exercise `capture: true`.** `EXECUTED.` It dispatches `new Event('scroll')` on `window` directly, which a non-capture listener receives just as well — so the flag could be dropped and the law would stay green. The flag is genuinely needed and genuinely right (a `scroll` on a scrolling ancestor does not bubble, and only the capture phase sees it at `window`), which is why this is a note and not a finding. A case that scrolls an actual ancestor would close it.

2. **The canvas can still move without either event.** `REASONED.` `ResizeObserver` fires on host *size*, the listener catches scroll and window resize — but an ancestor transform moves the box past all three, where the old per-mousemove rect was correct by construction. I checked whether that is reachable: nothing under `packages/web/src/app/` moves the scene by transform (the only `animate-*` uses are opacity, on unrelated elements). Unreachable today; recording the narrowing rather than asking for anything.

3. `plainMarks` / `forcedMarks` are reassigned each round, so `expect(forcedMarks).toBe(plainMarks)` compares the last round only. Sound for the forgery it exists to catch (a cheaper fleet on one series), just not per-round.

---

Approving. Three commits, three issues, three disjoint fences, one recorded widening, and the two questions `AGENTS.md` asks are answered substantively in the body rather than ceremonially — the `#160` note in particular is the shape this repo should be citing as the example. I am not merging: landing is the operator's step.

Issue #167, opened 2026-08-28
prd48 w2: an ack-after-durable-journal shipper loses no batch when the server dies


hchristinasmith:
## Fence (may touch ONLY)
- `docs/research/2026-08-29-shared-record-s2-ack-after-journal.md`

## Blocked by
#165.
Sibling lanes hold other surfaces: `...-s2-concurrent-actors.md`, `...-s3-windows.md` and `...-s5-pending-invite.md` belong to other lanes in this wave — if the work needs them, stop and surface it rather than absorbing it. The landed S2 note is read-only history; do not edit it.
**Model:** sonnet. **Wave:** 2.

## Definition of done
Success 2 is QUALIFIED, not met: the accept-fast 202 lost one batch on server death. This is the named re-run that discharges it.

- The shipper acks only after a durable journal write, and the design is stated before the chaos run, not fitted to it.
- Chaos row (c) is re-run under the same shape that lost a batch before — server killed mid-flight — and reported as a count of lost batches, which must be zero.
- The other chaos rows (kill mid-batch, offline, replay, cold backfill of a 63k-event log) are re-run against the new ack discipline, so the fix is not shown to cost losslessness elsewhere.
- Chain digests reported end to end for every row, identical local and server-side.
- Falsifier stated and its verdict recorded: if ack-after-durable-journal still loses a batch, the accept-fast shape is not repairable by ordering alone and the build PRD needs a different ingest contract.

hchristinasmith:
Closed by PR #199, squash-merged as `2abc789`.

**Merged into `prd48`, not into `main`.** prd-48 lands on `main` as one integration PR when the whole PRD is done — the operator's explicit decision — so this work is not yet on the default branch.

Reviewed by the dispatching session before merge. One finding, fixed in `8d08f2c` before the merge: the headline led with 75 kills, where the window that reproduces S2's defect fired twice and one of those had fsync deliberately disabled — so the full design was exercised at that window exactly once. Rewritten so the closed-window count carries the headline and the falsifier table.

Success 2 moves from QUALIFIED to met **for the window tested**, at n=1 with the full design. #205 is filed to raise that n and to fault-test the fold worker's own commit-then-cursor ordering, which this sweep never aimed at.

[...]

PR #198, opened 2026-09-01
docs(research): eight concurrent shippers into one server lose and reorder nothing (#166)

### LINKED ISSUE #166: prd48 w2: two shippers writing at once lose and reorder nothing
## Fence (may touch ONLY)
- `docs/research/2026-08-29-shared-record-s2-concurrent-actors.md`

## Blocked by
#165. The corpus shape decides the batch sizes worth testing.
Sibling lanes hold other surfaces: `...-s2-ack-after-journal.md`, `...-s3-windows.md` and `...-s5-pending-invite.md` belong to other lanes in this wave — if the work needs them, stop and surface it rather than absorbing it. The landed S2 note (`docs/research/2026-08-28-shared-record-s2-shipper.md`) is read-only history; do not edit it.

## Definition of done
The executed S2 spike ran one shipper. The gap it named is more than one actor shipping at once.

- Two or more shippers write concurrently into one server for a sustained run, with the actor count and batch rate stated.
- Zero duplicates and zero reorders after dedup, reported as counts, not as "passed".
- Per-actor append order holds for every actor independently — the ordering rule the record format already pre-rules.
- The server-side re-export's `verifyRecord` chain closes to each local export's digest, per Success 2.
- Throwaway lives outside the tree and is named in the note so a reader can rerun it.
- Falsifier stated and its verdict recorded: any interleaving that produces one duplicate, one reorder, or one chain that does not close means concurrent shipping needs a design answer before the build PRD, not after.

hchristinasmith:
Closes #166

## What this is
Extends the single-shipper S2 spike (`docs/research/2026-08-28-shared-record-s2-shipper.md`, read-only, not edited) to N=8 concurrent shipper processes into one server/queue/drain loop. Six runs: concurrent full-ledger backfill (= falsifier row (f)), a 10-minute sustained live-tail, and four more falsifier rows (staggered start, one throttled actor, kill-mid-flight, mixed backfill+live-tail regime).

**Falsifier verdict: PASS, stated to the precision the evidence supports.** This design is lossless (zero gaps, all 48 chain digests close) under five deliberately adversarial interleavings plus one real shared-box outage, across 902,640 rows and 48 actor-checks. Per-actor ordering is guaranteed by this design's construction (one shipper in flight at a time, one FIFO drain loop) rather than demonstrated by these runs — the one actor-check with any genuine path to a reorder (actor-5, `kill-i`) held. Concurrent shipping into this single-queue/single-drain-loop design does not need a design answer before the build PRD, on this evidence; a design with concurrent Postgres writers would have to re-earn the ordering guarantee this one gets for free, and this spike says nothing about whether it would.

Throwaway harness lives outside this tree at `~/rhizomorph-spikes/concurrent-actors/`; full rerun recipe is in the note's Reproduction section.

## What is the sibling case?
S2 found event ids repeat *within one actor's own log* (the `evt-NNNNNN` collector-cycle counter). This spike asked the sibling question — do ids also collide *across different actors'* logs — and the answer is yes, heavily: 68–100% pairwise overlap across all 28 real-actor pairs checked, even between two actors that are each internally unique. That blocks id-based per-actor attribution from a merged stream on real data; the `mergeRecords`-based order check only ran cleanly on synthetic globally-unique-id data (4/4 pairs), per the plan's own pre-authorized fallback. The raw per-actor `events_n` order check (0 inversions, 0 gaps, every actor, every row) is the load-bearing per-actor-order evidence instead.

Which assertions could actually have failed, and which could not, is now stated in the note itself (see "Results," "Check 1"): `events_n`'s primary key `(project, actor_instance, n)` with `ON CONFLICT ... DO NOTHING` makes the dup check unable to return anything but zero, in any run — a mutation that broke duplicate handling entirely would not be caught by this check, because the schema itself prevents the row from ever landing twice. The reorder check (`lag(n) OVER (PARTITION BY actor_instance ORDER BY serial)`) is guaranteed to read zero for any actor never killed mid-flight, because each shipper awaits its own POST before building the next batch and the server drains with one FIFO loop — a mutation that broke per-actor ordering under a design with actual write concurrency would not be caught here either, since this design cannot produce concurrent writes to catch it with. The one check that would catch a real mutation across the whole matrix: the gap check (a batch that never lands, or lands and is never retried, shows up as a hole in `n`) and the chain-digest closure (any dropped, duplicated-in-content, or reordered line changes the hash). Those two are what this spike's PASS verdict is actually built on, and the note now says so in the headline, not only in a caveat.

## Findings from PR review — both verified and fixed
The conductor's review of the first commit found two issues, both confirmed before fixing:
- **The outage's cause was wrong.** The note first attributed the mid-run server death to sandbox process supervision. It was actually a sibling lane's (#167) accidental process kill on this exact port and build area during a shared-box cleanup, per that lane's own note — confirmed by matching port 5561 and the `~/rhizomorph-spikes/concurrent-actors/` build area. Corrected, citing the sibling issue number and its title (no path — that file does not exist on this branch).
- **The headline overstated what the dup/order checks demonstrated** (see "What mutation would this test survive?" above) — this is the repo's named "test that cannot fail for the reason it claims" shape. Rewritten throughout: headline, results table, falsifier verdict, and "what this note did not test." Nothing was re-run or weakened — the gap and chain-digest results, which do carry the finding, are unchanged.

Both corrections are also recorded in the retro as their own items, including the process gap each one surfaces (checking for a sibling-lane cause before naming an environmental one; asking whether a check can fail at all before headlining its result).

## Fence widening — recorded here, before merge
The declared fence was one file: `docs/research/2026-08-29-shared-record-s2-concurrent-actors.md`. This PR also adds `docs/research/2026-08-29-shared-record-s2-concurrent-actors-retro.md` — the retro file gear-3's pipeline requires (CLAUDE.md feature-pipeline step 9), not anticipated in the issue's own fence. No other file touched; negative-fenced files (`...-s2-ack-after-journal.md`, `...-s3-windows.md`, `...-s5-pending-invite.md`) not needed, not touched.

## Recording gap — reported, not papered over
`baseline finish` could not complete: the reviewer agent's subagent transcript ends with no terminal `stop_reason` (the #462-shaped gap CLAUDE.md already names), so `--from-transcript` refuses to measure it, and only a combined token count (no in/out split) survives in the task notification — not enough to state `--tokens-in`/`--tokens-out` without guessing. Per the house rule against hand-stating telemetry into a write-once row, those numbers were not fabricated. The Baseline run `78e9c402-afc4-47ae-a2e6-1797c2a64935` stays `running`; the retro was not POSTed to Baseline (the route requires `finish` first), but the retro file is committed in this PR per the repo's convention.

## Gate
`npm run typecheck`, `npm run lint`, `VITEST_MAX_WORKERS=6 npm test` all green (384 test files, 6792 passed, 3 expected-fail, 6 skipped) — the diff is two new markdown files, so nothing else could have regressed. Re-run clean after the review-fix commit too.

## Gear
Stayed at 3 (build), as issued. No new exported repo symbols, no migration, no file outside the (widened) fence.


hchristinasmith:
## Review record — posted after the fact

This PR was reviewed before merge by the dispatching session (a different session from the lane that authored it), which read the diff commit by commit and the throwaway harness. That review happened in a terminal and left no trace here, which is the gap this comment closes. Posting it retroactively so the record survives the session.

**Findings, both fixed in `6ad38f7` before merge:**

1. **[EXECUTED] The outage's stated cause was wrong.** The note attributed the mid-run server death to sandbox process supervision. The sibling #167 lane recorded killing a process on port 5561 in `~/rhizomorph-spikes/concurrent-actors/` during a shared-box cleanup — this spike's own server, by both port and build area. A committed research note stating a false cause is read later as fact.

2. **[EXECUTED] The dup and reorder checks cannot fail.** `events_n` declares `PRIMARY KEY (project, actor_instance, n)` and inserts with `ON CONFLICT ... DO NOTHING`, so `count(*) - count(DISTINCT (actor_instance, n))` is zero by constraint. The shipper awaits each POST before forming the next batch and the server drains with one FIFO loop, so `serial` is monotonic in `n` by construction for any actor never killed — 47 of 48 actor-checks. This is the repo's named "test that cannot fail for the reason it claims" shape.

**Verified and left alone:** fences held; no real machine paths or home directories; the gap check and chain-digest closure are real, can fail, and are what the verdict rests on.

**Follow-up filed:** #204 — cross-actor id collision blocks attribution from a merged stream.

Merged into `prd48`, not `main`.

PR #199, opened 2026-09-01
docs(research): an ack-after-durable-journal shipper loses no batch when the server dies (#167)

### LINKED ISSUE #167: prd48 w2: an ack-after-durable-journal shipper loses no batch when the server dies

hchristinasmith:
Closes #167

## Baseline recording — one seat red, cause outside this diff

`baseline finish` cannot complete: the `reviewer` seat's subagent transcript stopped with no terminal `stop_reason` (the #462 shape named in `CLAUDE.md`, confirmed by checking the transcript file directly — it has been unmodified for 8+ minutes, so this is not a race, and its content is complete: a full final report exists, only the technical stop marker is missing). Per this repo's own rule, I have **not** hand-stated the seat's model/duration/tokens/tool-calls to force a green record — that estimate would land in a write-once row that can never be corrected. Evidence the review actually happened: its full findings are quoted below. `baseline finish` / the retro POST will need to be run once this is resolved, by whoever picks this back up.

## Headline result

Falsifier: **PASS, and read the window count, not the total.** The sweep fired 75 real process kills of the ingest server total (far more than the issue's 5–10 ask — a sweep-driver limitation meant every port-clearing ambiguity produced a real kill, all genuine and evidenced), spanning all four fault windows the design names. But the window that actually reproduces S2's defect — server dies after the 202 was fully flushed — fired only **twice** within that 75, and one of those two was the `RZ_SKIP_FSYNC=1` ablation with fsync deliberately disabled. **At full design, the precise S2 reproduction target was exercised exactly once: zero batches lost.** Across every window, plus 20 shipper-side kills, a 5,000-line rewind, and a total garbage-cursor loss: **zero lost batches, zero gaps, zero duplicate rows, identical chain digests local vs. server on every row.** The specific S2 defect (`events_n` count landing 500 short of the shipper's cursor) did not recur anywhere. What would strengthen this most: more trials aimed specifically at the closed window with fsync enabled — not re-run here, noted as a cheap follow-up since the fault-injection hooks already exist. Full detail, the stated design (written before any chaos re-run — verified by the reviewer to carry zero post-run numbers), and the falsifier table are in the new note.

A genuine finding beyond the ask: an `fsync`-skip ablation shows a `kill -9` sweep (this one, and S2's) proves the *ordering* claim but cannot prove `fsyncSync`'s actual host-crash durability — that needs a different kind of test neither spike attempted. Flagged as the highest-priority open gap for the real-VPS work.

## What is the sibling case?

The reviewer found one, and the note now names it in "What this did not test": the falsifier aims four fault points at the HTTP request handler's write→fsync→ack ordering, but **none at the fold worker's own commit-then-advance-cursor ordering** — the structural sibling of the exact bug this issue exists to fix, one layer down. The code is correct (commit precedes cursor-persist) and no fault point proved it wrong, but nothing tested it either. A follow-up fault point (`f1`, between commit and cursor persist) is named as the next step, along with a smaller asymmetry: the fold cursor's own tmp-then-rename write carries no `fsync`, unlike the journal's.

If the fold worker's cursor were advanced *before* the Postgres commit instead of after (an ordering bug one line different from the correct code), this sweep would very likely **not** have caught it — see the sibling-case gap above. The falsifier is structurally capable of catching it (an acked-but-unfolded batch would show as a gap), but no kill in the 75 was aimed at that narrow window; catching it would have been coincidence, not design.

## Fence

Touched exactly the two files the issue's fence and the retro convention call for:
- `docs/research/2026-08-29-shared-record-s2-ack-after-journal.md` (the declared fence)
- `docs/research/2026-08-29-shared-record-s2-ack-after-journal-retro.md` (the retro file, following the same-wave precedent set by issue #166's own note-plus-retro shape — recorded here as the fence widening, since the issue's fence named only the note)

No other file changed. `npm run typecheck`, `npm run lint`, and `VITEST_MAX_WORKERS=6 npm test` all pass (384 files, 6792 tests) — expected, since no application code changed.

## Gear

Ran at gear 3 as scoped — no shift.

## Review round 2 (conductor)

Fixed: the headline, row (c) results, and falsifier table overclaimed 75 as the test count for the window that matters — corrected to state the closed-window count (2, 1 at full design) explicitly, everywhere the total appeared. Added a one-line cross-reference in mistake 4 to issue #166 (whose own note recorded, from its side, the outage this lane's accidental kill caused) — by issue number only, no path cited, since that file doesn't exist on this branch. Re-ran the gate clean. Not merged.


hchristinasmith:
## Review record — posted after the fact

This PR was reviewed before merge by the dispatching session (a different session from the lane that authored it), which read the diff commit by commit. That review happened in a terminal and left no trace here; posting it retroactively so the record survives the session.

**Finding, fixed in `8d08f2c` before merge:**

**[EXECUTED, from the note's own table] The headline rested on n=1, not n=75.** The sweep fired 75 kills, but the window that reproduces S2's defect (`closed` — after the 202 flushed) fired **twice**, and one of those two was the `RZ_SKIP_FSYNC=1` ablation with fsync deliberately disabled. So the precise defect window, with the full design enabled, was exercised **once**. 61 of the 75 landed in `c2`, a window that was never unsafe even under S2's old design, since no ack had been sent. Everything needed to see this was already in the note; the headline and PR body did not carry it.

**Verified and left alone:** fences held; no real machine paths or home directories; the design is stated before the chaos run as the issue demanded; the fsync ablation finding — that a `kill -9` sweep proves the ordering claim but cannot prove `fsyncSync`'s host-crash durability — is the strongest result in either of this wave's notes.

**Better than what the review asked for:** the lane's own pass then found a real sibling case (no fault point aims at the fold worker's commit-then-advance-cursor ordering) and a real mutation answer (a reversed ordering there would probably not have been caught).

**Follow-up filed:** #205 — fault-test that fold-worker ordering, and raise the closed-window n.

Merged into `prd48`, not `main`.

PR #200, opened 2026-09-01
prd43 w2: the cited paths exist, and the allowlist shrinks 16 to 3

### LINKED ISSUE #17: prd43 w2: architecture.md points at the paths that exist
## Fence (may touch ONLY)
- `docs/architecture.md`

## Fence widened at grooming, 2026-09-01 — recorded before the change

The law file is added, and this issue holds it **on behalf of the whole of wave
2**. The reason is a coupling the PRD's sequencing did not anticipate.

Wave 2 is declared "parallel, fenced apart", and its three issues do own three
disjoint document sets. But the wave-1 law carries an allowlist of
currently-broken citations, and its honesty test asserts every allowlisted
citation **still fails today**. So the moment a wave-2 issue repoints a citation,
that entry stops failing and the law goes red until the entry is deleted.
EXECUTED on origin/main before this grooming: creating one allowlisted target
produced

    packages/web/src/fleet/buildFleet.ts now resolves — remove this allowlist
    entry, docs/architecture.md is fixed

13 of the 16 entries belong to #17 (5), #20 (4) and #19 (4). Every wave-2 lane
therefore has to edit one shared array to leave its own suite green, which is why
the three ship as ONE lane rather than three: three lanes in one array is a
rebase conflict scheduled in advance, and the fence linter hard-fails on exactly that.

The claim is recorded on this issue alone rather than on all three, so linting
the wave does not report a phantom overlap between siblings that are not actually
concurrent.

## Blocked by
**Nothing — prd43 w1 landed on origin/main as `ed041c8`.** Both #16 (the law) and
#186 (its honesty follow-up) are in.

The issue previously read "the law lands red first". That is superseded: the law
landed GREEN with the currently-broken citations as an explicit, honesty-checked
allowlist, which is what makes this wave's work "delete your entry" rather than
"fix a red build".

Sibling issues #19 and #20 own `docs/design-notes/` and the four judge source
files; this lane covers all three, so those are not held by anyone else.

## Definition of done
- `:111`, `:227` and `:346` repointed from `packages/web/src/fleet/buildFleet.ts` to `packages/core/src/fleet/buildFleet.ts` (moved by `#246`).
- The two legacy fence-module citations at `:314` and `:330` are repointed to where those files now live.
- `:2070`'s `packages/web/src/panels/spend/format.ts` repointed or removed.
- prd43 w1's law goes green for this file.

### LINKED ISSUE #19: prd43 w2: the judge's rationale is reachable
## Fence (may touch ONLY)
- `packages/core/src/events/judge.ts`
- `packages/server/src/collectors/judge/collector.ts`
- `packages/server/src/judge/mergetree.ts`
- `packages/server/src/judge/symbols.ts`
- `docs/research/2026-08-04-semantic-judge-spike.md`

## Blocked by
Blocked by prd43 w1.

Sibling lanes own `docs/architecture.md`, `docs/roadmap.md` and `docs/design-notes/`.

## Definition of done
- All four files cite `docs/research/2026-08-04-semantic-judge-spike.md`, deleted by `756e1bf` — whose sweep covered documents citing documents, not code citing documents.
- Either restore the note from `756e1bf^` (cheaper, keeps the citation resolving) or inline the load-bearing rationale into the four files and drop the citation. One or the other, not a mix.
- prd43 w1's law goes green for these four files.

### LINKED ISSUE #20: prd43 w2: the design-notes link to where their targets live
## Fence (may touch ONLY)
- `docs/design-notes/`

## Blocked by
Blocked by prd43 w1.

Sibling lanes own `docs/architecture.md`, `docs/roadmap.md` and the judge sources.

## Definition of done
- Four links to a non-existent `docs/decisions/` are repointed to `docs/design-notes/`, where the targets actually live (`root-depth-tissue-vibrancy.md`, `node-persist-lane.md`, `node-apical-tuft-glow.md`).
- prd43 w1's law goes green for this directory.

gabriel-canaan:
Closes #17
Closes #19
Closes #20

prd-43 wave 2. Three issues, three commits, one PR — they share a surface, so
they cannot be three lanes. See "Fence widening" below.

## What becomes true

Wave 1 landed a law asserting every path cited from a document or a code comment
exists, carrying an **allowlist** of the citations that were already broken. Its
honesty test asserts each allowlisted citation **still fails today**, so fixing
one forces its entry out.

This wave fixes them. **The allowlist goes 16 → 3.**

| commit | issue | what |
|---|---|---|
| `6469800` | #17 | `docs/architecture.md` cites where `buildFleet`, `fences` and the spend panel actually live |
| `42e631c` | #19 | the judge's rationale note is restored, not just cited |
| `5c884c0` | #20 | the design-notes link to where their own targets live |

`#19` took the restore branch of its Done-when rather than the inline branch —
the deleted note comes back byte-identically from `756e1bf^` (218 lines,
sha256-matched), so the four judge source files are **untouched**. Its issue
allows either, and is explicit that it must not be a mix.

## Fence widening — recorded on #17 at grooming, before the change

`packages/server/src/doc-citation-law.test.ts` was added to #17's fence, and #17
holds it **on behalf of the whole wave**.

The PRD declares wave 2 "parallel, fenced apart", and its three document sets
genuinely are disjoint. What the sequencing did not anticipate is the allowlist:
repointing a citation makes its entry stop failing, and the law goes red until
the entry is deleted. EXECUTED on `origin/main` before grooming — creating one
allowlisted target produced

    packages/web/src/fleet/buildFleet.ts now resolves — remove this allowlist
    entry, docs/architecture.md is fixed

13 of the 16 entries belong to these three issues. Three lanes editing one shared
array is a rebase conflict scheduled in advance, so they ship as one lane. The
claim is recorded on #17 alone so linting the wave does not report a phantom
overlap between siblings that are not concurrent.

## What is the sibling case?

**A verb, not a citation — and it was missed by the commit that fixed its
siblings.** `6469800` repointed three `buildFleet` call sites and correctly
changed `now computes` → `computed` where the spend panel is described. Four
lines later, in the same predicate, `and renders` survived. The sentence asserted
that `SpendPanel` was *"deleted outright by `b1cb3d5`"* **and** that it renders
something today. `selectCostOverhead` and `formatCostOverhead` return zero hits
across `packages/` — they died with the panel.

That one matters more than an ordinary tense slip, because the two spend
citations in that paragraph were correctly **un-backticked** (no successor file
exists) — which means the law can never look at that paragraph again. Nothing
would have resurfaced it. Fixed, along with the stale `see docs/telemetry.md`
clause inside the string a dead panel used to render.

The wave's other sibling axis is the allowlist itself, in both directions:
deleting an entry whose citation still fails silently narrows the law, and
keeping one that now resolves reddens it. Both were checked entry by entry.

None of these — all EXECUTED by the verification pass, independently re-derived
rather than taken from a commit body:

- **Over-deletion, all 13 entries individually.** Wave-1's full 16-entry
  allowlist was reinstated into the post-wave tree and each entry classified by
  *why* it left, using the law's own `extractCitations`/`citationExists` — which
  distinguishes "no longer cites" (citation deleted or hidden: the dangerous
  case) from "now resolves" (the target really exists). Result: 4 judge entries
  `RESOLVES-AT-SAME-PATH`, 7 repointed with each new path confirmed extracted
  from the *stripped* text and resolving, 2 spend entries genuinely un-cited with
  no successor file. **No entry was removed while its citation was still
  broken.** CONTROL: the same probe reported the 3 survivors as
  `STILL-BROKEN+CITED`, so the classifier can tell the states apart.
- **Under-deletion.** All 3 remaining entries still fail today.
- **Did the law get weaker?** A fresh dead citation appended to each of the four
  in-scope touched documents was convicted, including one inserted inside the
  rewritten spend paragraph. Deleting the restored note reddens all four judge
  citations, so `#19`'s restore is load-bearing rather than vacuous.
- **Repointed to the wrong file?** Checked against history, not the commit body:
  `69249f0` (#246) is the real `packages/{web → core}/src/fleet/buildFleet.ts`
  move, `fences.ts`/`fences.test.ts` are pure renames (0 line changes), and
  `b1cb3d5` deleted `packages/web/src/panels/spend/` outright (4 files, 741
  deletions) — which *corrects* wave-1's allowlist reason, that had wrongly said
  the panel moved to `lane-page/`.
- **Races.** The law has raced itself twice before. 5 rounds × 4 concurrent runs
  = 20/20 green; full suite 4× at `--maxWorkers=5` = 4/4 green.

    VITEST_MAX_WORKERS=6 npm test     0   ->  401 files, 7128 passed
    doc-citation-law 4x concurrent        ->  4/4 green

## Verification

Full pipeline: two independent seats (Opus `--effort xhigh`, Codex
`gpt-5.6-terra` `xhigh`) on the assembled tree with `main` merged, each in its
own disposable worktree. They **split** — one merge-ready-after-one-fix, one
not-merge-ready — and the split was resolved by execution rather than by vote.

Two findings were fixed: the verb above, and a false historical claim in #20's
commit body (it said `docs/decisions/` "has never existed in this repo"; `eca1b33`
renamed it to `design-notes` and six commits touch it). That message now records
the correction rather than quietly replacing it.

**One seat finding was overruled, and it is worth stating rather than burying.**
The restored note cites `docs/prd11.md` and `docs/prd3.md` at `:217`, both
absent, and `docs/research/` is in `EXCLUDED_DIRS` so the law structurally cannot
see them. One seat called that must-fix. It is not: the law's own comment says
those directories are excluded as citing sources precisely because *"a dated
artefact's own citations are a record of the tree at a commit, not a live
claim"*, and a 2026-08-04 "sources read this session" footer is exactly that. It
is recorded here because 218 lines entered the tree unchecked and that should be
a decision, not an oversight.

The repairs themselves are three lines of prose in one document — §12's cost
guard for an inert fix, declared rather than skipped. They did not touch the law
file, so the substantive change (the allowlist) is exactly what both seats
reviewed.

## Follow-ups, out of this wave's fence

Filed rather than widened into this PR:

- `docs/architecture.md:199` and `:469` attribute `detectFrozen`/`detectWaiting`
  to `buildFleet.ts`; they live in `diagnose.ts`. Pre-existing and outside #17's
  Done-when, which is about paths rather than claims — but the repoint turns it
  from a dead path someone would eventually look at into a **live** path nothing
  will ever flag.
- `docs/adr/0007-one-derived-fleet-object.md:66` still says `buildFleet` is in
  `packages/web/src/fleet/` and cites #246 as pending. That path still resolves
  (the directory exists, holding other files), so the law will never flag it. The
  ADR log is append-only, so the fix is a superseding record, not an edit.
- `packages/server/src/doc-citation-law.test.ts:587`'s exclusion scan lacks the
  ENOENT guard its sibling got in #186 — the outstanding half of that pair.


ciaran-slow (approved):
## Approved — one fix pushed as `2dca1bb`

Reviewed commit-by-commit in a clean clone at `7d42a81`, node 22.23.2. Every
claim below is **EXECUTED** unless labelled otherwise.

The PR body's central claims are true, and I re-derived them rather than
reading them:

- **The allowlist really goes 16 → 3**, and the 13 removed map exactly onto the
  three issues (5 architecture.md / 4 judge / 4 design-notes). The 3 survivors
  are the adr-0029, `placeholder.ts` and `docs/prd2.md` entries, owned by no
  wave-2 issue.
- **Every repointed target exists**: `packages/core/src/fleet/{buildFleet,fences,fences.test}.ts`,
  the restored research note, and all three design-notes targets.
- **The restore is byte-identical.** `sha256(docs/research/2026-08-04-semantic-judge-spike.md)`
  == `sha256(git show 756e1bf^:…)` = `1342a8ea…fdb9fc`, 218 lines. The four
  judge source files are genuinely untouched.
- **No over-deletion.** Three mutations, each against this tree, all convicted
  (exit 1), control green (exit 0):

  | mutation | result |
  |---|---|
  | revert architecture.md's `buildFleet.ts` citation | caught |
  | delete the restored note | caught — **all four** judge citations redden, so #19's restore is load-bearing, not vacuous |
  | revert one `docs/design-notes/` link | caught |
  | a fresh backticked dead path injected into the rewritten spend paragraph | caught |

- **History claims check out.** `b1cb3d5` deleted `packages/web/src/panels/spend/`
  outright (4 files, 741 deletions) — so the body is right that wave-1's
  allowlist reason ("moved to `lane-page/`") was wrong. `69249f0` (#246) is the
  real `web → core` fleet move. `eca1b33` did rename `decisions` → `design-notes`,
  and `docs/decisions/` *did* exist (6 commits touch it) — #20's commit-body
  correction was the right call, not a cosmetic one.
- **The overruled seat finding was correctly overruled.** The note does cite an
  absent `docs/prd11.md`/`docs/prd3.md` at `:217`, and `docs/research/` is in
  `EXCLUDED_DIRS` (`doc-citation-law.test.ts:266`), so the law structurally
  cannot see them. That exclusion is by *citing-file* directory and the law's
  own comment gives the reason — a dated artefact's citations record the tree at
  a commit. A 2026-08-04 spike note is exactly that. Agreed.
- **The 218 restored lines carry no personal paths** — no `/Users/`, `/home/`,
  `C:\Users`, hostname or `.local` state path. Worth stating, since restoring a
  capture-era document from history is precisely how one gets in.

Gates I ran on the merged tree:

    npm run typecheck              0
    npm run lint                   0
    VITEST_MAX_WORKERS=6 npm test  0   ->  401 files, 7124 passed, 3 expected fail, 6 skipped

`gh pr checks 200`: all **7** expected jobs actually executed with real
durations (58s–9m35s) — `build-test-boot` × 3 legs and `pack-smoke` × 4. Not a
green bar over skipped steps.

### The one fix — `2dca1bb`, mechanical

`docs/architecture.md:469` says `parked`'s three consequences are *"all in
`packages/core/src/fleet/buildFleet.ts`"*. They are not:

    detectFrozen        packages/core/src/fleet/diagnose.ts:110
    detectWaiting       packages/core/src/fleet/diagnose.ts:154
    buildLadder's skip  packages/core/src/fleet/ladder.ts:29
    the dimmed PARKED   packages/web/src/panels/fleet/index.tsx:358

The sentence contradicted itself already, citing the fleet table's own `.tsx`
path four lines after "all in buildFleet.ts".

This is the repoint's **own sibling case**, which is why I fixed it here rather
than leaving it filed. Before `6469800` the line carried a dead path, so the law
convicted it and someone would eventually have read the predicate around it.
After the repoint the path is live, the law is satisfied, and the false claim
beside it is exactly the shape nothing will ever resurface — the same trap the
commit body already names for the un-backticked spend paragraph. The follow-up
note in the body reads `:199` and `:469` as one item; `:199` is actually fine
(it says buildFleet *uses* `parked`, which it does, at `buildFleet.ts:349`) —
`:469` is the one that asserts location.

Fixed by naming the owning module rather than deleting the clause, which turns
the correction into two more law-checked citations instead of one fewer.
Mutation-proven: `diagnose.ts` → `diagnoze.ts` and `ladder.ts` → `laddr.ts` are
both convicted (exit 1, bad path named); unmutated tree 26/26. The pre-fix text
could not be mutated this way at all — that was the defect.

**The un-backticked spend paragraph is a real, correctly-declared blind spot.**
`CITATION_RE` requires backticks, so `packages/web/src/panels/spend/` as bare
prose (`architecture.md:2333`) is now invisible to the law forever. That is a
legitimate reading of #17's "repointed or **removed**" and I am not blocking on
it — but it is the one place this PR trades law coverage for prose, and the
commit body is right to flag that nothing will resurface that paragraph.

### Not run

`scripts/gate.sh` (operator's step, not a reviewer's) and the concurrency probe
— I did not re-measure the body's 20/20 law-race claim, so that one stands as
the author's evidence, not mine.

Fence discipline is clean: #17 holds the law file on behalf of the wave,
recorded at grooming before the change, and no commit strays outside its
declared paths.

Issue #66, opened 2026-08-24
prd43 w7: the docs cite a tracker that no longer exists


gabriel-canaan:
## The defect

The documentation corpus cites issue numbers from a previous incarnation of the tracker. **None of them resolve.** The highest issue that has ever existed in this repository is `#64`; the docs cite numbers into the `#600`s as though they were live links.

Measured 2026-08-25 on `b66fd6d`:

- **216 distinct `#NNN` references** (3–4 digits) across **111 markdown files** under `docs/` plus `AGENTS.md` and `CLAUDE.md`.
- Every one checked resolves to nothing: `gh issue view` fails for `#217`, `#228`, `#243`, `#299`, `#401`, `#649`.

## Why it matters more than a broken link usually does

**`AGENTS.md` is affected, and it is the file that tells every agent and contributor how the repo works.** Its citations are `#209 #275 #281 #362 #396 #399 #408 #649` — which means:

- The **working agreement** cites `#399` as its own origin ("Adopted 2026-08-11 (#399), from measurement over the 45 PRs merged 08-09 → 08-11"). A reader cannot check the measurement.
- The **landing section** is built around the `#408` story — a lane that ran `gate.sh` and came within a stale checkout of pushing unreviewed work. That story is the entire reason the section is worded as it is, and it is unverifiable.
- The **queue evidence** cites `#275` (98 lines, waited 62 h) and `#396` (1,235 lines, merged in 0.5 h) as the two data points proving the toll is per-PR rather than per-line.
- `#362` is cited as the live reason `issues.sh show` is broken.

`AGENTS.md`'s own closing rule is *"The repo's own history is why findings are checkable later."* Right now none of its findings are checkable, which undercuts the authority of the file that governs everything else.

`docs/prds/prd-42-one-path-one-spelling.md` is a worked example: its Status line stands on `#217`/`#228`/`#299`/`#401` and its Evidence cites `#243` as the territory a residual should be folded into. `#47` found that `#243` does not exist, which is why it was filed as its own issue rather than commented onto the phantom.

## What this is NOT

Not a request to delete the references. They are real provenance — the work happened, in a tracker this repo no longer has. Erasing them would trade a broken link for a missing fact, which is worse.

## Definition of done

The decision is what those references should become, and it is a decision rather than a sweep. One shape, chosen deliberately and applied consistently:

- a marker distinguishing prior-tracker numbers from live ones (e.g. `#399 (prior tracker)`), so a reader knows not to click; or
[...]

Issue #63, opened 2026-08-24
prd43 w4: a remedy names something a reader can reach


KelliherL:
## Fence (may touch ONLY)
- `packages/core/src/fleet/gaps.ts`
- `packages/core/src/fleet/remedy-reachability-law.test.ts`
- `packages/core/src/fleet/buildFleet.test.ts`
- `docs/user-guide/watching.md`

## Fence corrected at grooming, 2026-09-02 — the groomer's error, recorded before the audit finds it

The fence first written here named packages/core/src/selectors/gaps.ts and a
sibling law beside it. **That directory has never existed.** The module is, and
always has been, packages/core/src/fleet/gaps.ts — EXECUTED against origin/main:
`git cat-file -e` fails on the selectors path and succeeds on the fleet one, and
`git log --diff-filter=D` shows it was never deleted from selectors either. The
path was lifted from this issue's own prose at filing and written into a fence
without checking that it resolved.

Nothing in the grooming chain catches that. The body preflight asks whether a
body claims the paths it means to; the fence linter asks whether two bodies claim
the same path. **Neither asks whether a claimed path exists**, so a fence naming
a file that is not there lints clean, dispatches, and is discovered only when the
lane either stops or — as here — quietly does correct work outside its declared
boundary, which the landing gate's fence audit would then reject.

buildFleet.test.ts is added for the second half of the same mistake: it encodes
the OLD remedy string, so changing the remedy necessarily changes it. An issue
that changes behaviour must fence the tests that encode the old behaviour, and
this one did not.

## Fence widened by the operator, 2026-09-02 — recorded before the change

`docs/user-guide/watching.md` was added while assembling the prd43-w4 bundle.
It reproduces this issue's defect **verbatim**, one file over: its
"Exact strings you may see" list quoted the old remedy, `run: dispatch.sh`,
so the fix as built removed the unreachable script from the code and left it
standing in the user guide. Full rationale, and why no law is added for it
here, in [this comment](https://github.com/launchpad-26/rhizomorph/issues/63#issuecomment-5504669971).

## Blocked by
Nothing.

Sibling lanes hold other surfaces this wave: #23 owns README.md, SECURITY.md,
mutation-guard.ts, route-class-law.test.ts and the CI workflow. Do not touch any
of them.

**Model:** sonnet. **Wave:** 4.

## Groomed 2026-09-02 — why wave 4, and why this is ruling 1's declared blind spot

[...]

Issue #62, opened 2026-08-24
prd17 w1: every event reaches the reducer through the upcast chokepoint


KelliherL:
## Fence (may touch ONLY)
- `packages/core/src/events/upcast.ts` (new file)
- `packages/core/src/reduce.ts`
- `packages/core/src/upcast-chokepoint-law.test.ts` (new file)

## Blocked by
Nothing. Additive and zero-claimant: prd-17 ruling 3's item 3 is the home every later migration flows through, so it lands before anything that would flow through it.

Sibling lanes hold other surfaces, and none of these may be touched here. #219 (prd17 w2, re-waved behind this issue on 2026-09-02) owns the rest of `packages/core/src/events/` — every existing file in it, including `index.ts` and its `parseEvent` / `parseEventLenient`; this issue adds exactly one new file there and edits nothing else in the directory. #210 owns `packages/core/src/eras/` and is in progress — the era corpus is the identity *proof* here and is read, never edited. #232 owns `docs/architecture.md`. #204 owns `packages/core/src/record/`. No document is edited by this issue at all, and that is the point (see Why).

**Model:** sonnet. **Wave:** 1.

## Why

prd-17 ruling 3, item 3: *"An identity `upcast()` chokepoint reserved now between parse and reduce, so the day a migration is needed it has a home that every event already flows through."* ADR-0011 chose it as option E and rejected D (retrofit when first needed) with the argument that retrofitting means finding folds *"in the live stream, in replay, in the record reader, in the era corpus, and in a hundred tests."*

**It was never built.** On `main` today `packages/core/src/events/upcast.ts` does not exist and no `upcast` function exists anywhere in `packages/*/src`. The word appears twice, both as comments treating it as already reserved: `packages/core/src/state.ts:52` (*"no upcast needed, additive events refold cleanly"*) and `packages/server/src/log/retention.ts:40`.

**Three documents say it landed, and none is checked.** `docs/architecture.md:2147` stamps item 3 *"(landed)"* and describes the file as *"today, the identity function ... the one function both the live path and the replay path bottom out in."* prd-17's own Outcome header reads *"ruling 3 landed, all five laws."* ADR-0011 goes furthest: *"D was rejected in `upcast.ts`'s own comment"*, and then **blockquotes that comment**. The citation law cannot see any of this — it sweeps backticked `packages/`-rooted paths, and every one of these writes the file as bare `upcast.ts` — which is how a phantom survived three separate documents.

[...]

Issue #39, opened 2026-08-24
theme: the greyscale-survival law covers the status family


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/theme/theme.css`
- `packages/web/src/theme/category.test.ts`
- `packages/web/src/scene/palette.ts`
- `packages/web/src/scene/palette.test.ts`
- `packages/web/src/theme/oklch.ts` — *widened at plan time, 2026-09-07:* the dichromacy simulation and `deltaE` are shared colour arithmetic and belong beside `oklch()`/`hueGap()`, not in a test file
- `docs/design-notes/palette-paper-broken-ink.md` — *widened at plan time:* new; the Definition of done below requires it
- `docs/design/charter.md` — *widened at plan time:* law 9's row in the §2 register says "no automated greyscale test exists", which this issue makes false; one cell moves

## Blocked by
Nothing.

**The changelog is deliberately not claimed by this issue** (2026-09-07): #216 (prd20 w1, In progress) already claims CHANGELOG.md, and the wave lint reports the overlap. The Unreleased / Fixed entry is drafted verbatim in the plan comment (§7) and in the build report, and is added when the wave PR is assembled or once #216 lands, whichever comes first. The builder does not edit the changelog.

Outside any PRD — a theme law, not a cost that scales, so it is deliberately not part of prd-44 (recorded in that PRD's Non-goals). **Do NOT touch `packages/web/src/theme/tokens.test.ts`, `theme/legibility.test.ts` or `theme/kind.test.tsx`** — all three are coupling points holding exact counts and snippet anchors across the package (`.swarm/coupling.txt`), and none of them is this lane's to reconcile.

**Model:** sonnet.

## Definition of done
- In the **light** theme, `--color-broken` and `--color-working` are separable by a red-green colourblind reader and in greyscale. Today they measure **deltaE 1.4 under deuteranopia** and **1.25:1 in greyscale** — near-identical on both axes.
- Cap 5's greyscale-survival rule covers the **status** family, not only the category family. Today `category.test.ts` holds the four category tints to it (`:199`, `:207`) and reaches the status colours only as `STATUS_HUES` (`:154`) — the reference the categories are measured *against*, never subjects themselves. The law bites for status too, proven with a rigged pair that trips it.
- The **dark** theme already passes both axes (deltaE 10.9, 1.88:1) and must stay passing — this is a light-theme change.
[...]

PR #208, opened 2026-09-02
fix(issues): orphans finds an issue with no milestone, which AGENTS.md always said it did

ciaran-slow:
`AGENTS.md` has said, since the runbook was written:

> Every issue must carry a milestone. `scripts/dev/issues.sh orphans` finds the ones that do not.

`cmd_orphans` never read `milestone`. It compared open issues against **board membership** only — which the same file documents correctly twelve lines earlier as *"open issues missing from the board"*. So the runbook contradicted itself, and the half a reader acts on was the wrong half.

The failure mode is worse than a dead pointer: running the named check prints `all open issues are on the board`, a **success message for a different question**. Nothing enforced the rule, and nothing said so.

Four open issues had drifted milestone-less by 2026-09-02 — `#39`, `#62`, `#63`, `#66`. Three were placed under `prd43` while tidying the board; `#39` is still unplaced and is what the new arm reports today:

```
$ scripts/dev/issues.sh orphans
all open issues are on the board
no milestone:
  #39 theme: the greyscale-survival law covers the status family
```

No `Closes` line: this is non-PRD work with no issue, filed as its own change at the operator's direction rather than under a milestone.

## What is the sibling case?

**Two findings must stay two findings.** An issue can be off the board *and* milestone-less, and a merged `N orphans` count would hide which rule it broke. Each clean line names only what it checked, so neither can be read as covering the other. The board-clean/milestone-dirty case is asserted precisely because the old command printed only the first line and a reader would have taken it for the whole answer.

**"Not read" is a third state, distinct from "read, and null".** An older `gh`, or a caller that trims the `--json` list, returns rows with no `milestone` key. Collapsing that into null would report *every* open issue as drifted the day `gh` changes — a check announcing a verdict it did not earn, which is prd-45's whole subject. It prints `milestone not read` and claims nothing.

**The sibling command was checked and left alone.** `prd-reconcile.sh` sits beside this one and is also a finder rather than a gate; nothing here changes that, and exit status is untouched (findings still exit 0; only a truncated read exits non-zero). Making either fail a build is a CI decision that belongs in its own issue.

**No second API call.** The field rides along on the `gh issue list` call already being made. `#581`'s whole subject is how many times this script reads a remote, so adding a call would have been the wrong fix even for a right check.

Three run. **The second is why this is not the first draft.**

| # | Mutation | Result |
|---|---|---|
| 1 | Delete the milestone reporting block | **6 of 7 fail** |
| 2 | Collapse not-read into read-and-null | **SILENT — 41/41 green** |
| 3 | Hardcode the field-read decision to `True` | **1 fails** |

Mutation 1's one survivor asserts the *off-board* line, which prints either way; its paired partner (*"...and the milestone finding is not swallowed by it"*) does fail, so the pair is sound.

**Mutation 2 caught a dead guard in my own first draft.** I had written `"milestone" in i and i.get("milestone") is None` on the comprehension — a guard that could never fire, because the branch above it returns first. A check that cannot fail, inside a commit whose entire subject is a check that could not fail. It is deleted rather than tested: one line now decides whether the field was read, and every later branch defers to it. Mutation 3 exists to hold that line, and it bites.

The mock learned two knobs: `MOCK_NO_MILESTONE` (issue numbers whose milestone is null) and `MOCK_DROP_MILESTONE` (the key omitted entirely). Default is "every issue carries one", so a test that does not opt in cannot accidentally depend on a drifted issue existing.

## Fence

`AGENTS.md`, `scripts/dev/issues.sh`, `scripts/dev/issues.test.sh`. **No widening** — the runbook paragraph is the claim being made true, so it is in scope by construction.

## EXECUTED

- `scripts/dev/issues.test.sh` — **41 passed, 0 failed** (34 before)
- the three mutations above
- `scripts/dev/issues.sh orphans` against the live tracker — returns `#39`
- `scripts/dev/coupling.test.sh` — 11 passed
- `doc-citation-law` + `no-personal-paths-law` + `runbook-delivery-law` — 47 passed
- `npm run lint` — clean, 872 files

Node v22.23.2.


gabriel-canaan (approved):
Reviewed commit-by-commit in a detached worktree with `origin/main` merged in (the branch was 5 behind; main touched only `docs/`, no fence overlap). Everything below is labelled **EXECUTED** or **REASONED**. Node v22.22.2.

**The live behaviour is exactly what the body claims.** EXECUTED — `scripts/dev/issues.sh orphans` against the tracker:

```
all open issues are on the board
no milestone:
  #39 theme: the greyscale-survival law covers the status family
```

**The `is None` test is live, not decorative.** EXECUTED — I checked the shape `gh` actually emits before trusting the predicate, because a check that can never fire is the failure mode this commit is about. `gh 2.97.0`, `--json number,milestone`, 18 open rows: 17 objects, **1 literal `null`**, `milestone` key present on every row. So `i.get("milestone") is None` fires for exactly one issue, and `field_read` is True on the real path.

**All three of your mutations reproduce.** EXECUTED, independently, each restored after:

|---|---|---|
| delete the milestone reporting block | 6 of 7 fail | **6 of 7 fail** — 35 passed, 6 failed |
| collapse not-read into read-and-null | SILENT, 41/41 | **1 fails** — 40 passed, 1 failed |
| hardcode `field_read = True` | 1 fails | **1 fails** — 40 passed, 1 failed |

Mutation 1's survivor is `"board-clean and milestone-dirty are both reported"`, which asserts the off-board line — exactly the one you named, and its paired partner does go red. The pair is sound.

The second row is a **correction in your favour**: the table reads as results against the shipped code, and against the shipped code that mutation is *not* silent — `ms_absent` catches it. The prose under the table says it was the first draft; the table doesn't. Worth a one-word edit to the body, since it undersells the tests you wrote.

**Gates.** EXECUTED — full suite against branch + merged `origin/main`: **401 files, 7128 passed | 3 expected fail | 2 skipped**. `typecheck` clean. `lint` clean, 872 files. `issues.test.sh` 41 passed / 0 failed. `doc-citation-law` + `no-personal-paths-law` + `runbook-delivery-law` 47 passed. `gh pr checks 208`: all **7** jobs (3 × `build-test-boot`, 4 × `pack-smoke`) executed and passed — checked against the two-job list in `AGENTS.md`, not the green bar.

**The sibling claim holds.** EXECUTED — grepped every `orphans` mention repo-wide. `AGENTS.md:77`, `AGENTS.md:133` and `issues.sh:22` are all updated; nothing else describes the command as board-only. `prd-reconcile.sh` reads *by* milestone rather than *for* one, so it is not the same check wearing a different hat.

## One finding — fixed in `c08d45d`

**A fourth output state, unheld.** EXECUTED (mutation).

`cmd_orphans` now prints four milestone verdicts. Three have assertions. The fourth does not:

```python
if not issues:
    print("no open issues — nothing to check for a milestone")
```

Deleting those two lines was a **SILENT mutation — 41/41 green**. And the branch ordering makes the deletion worse than a gap: `field_read = bool(issues) and all(...)`, so an empty list makes it False, and the empty case falls through to the *next* arm:

```
$ MOCK_OPEN_ISSUES=0 issues.sh orphans     # with the empty-list line deleted
all open issues are on the board
milestone not read — this gh did not return the field, so nothing is claimed about it
```

`gh` returned the field perfectly well. There was nothing in it. That is a verdict nothing observed — one branch below the one this commit exists to close.

One assertion, using the `MOCK_OPEN_ISSUES` knob the mock already had:

```sh
fresh ms_none
out=$(MOCK_OPEN_ISSUES=0 "$SCRIPT" orphans 2>&1)
has "orphans: an empty issue list claims nothing about the field" "no open issues" "$out"
```

EXECUTED: 42 passed / 0 failed. Replaying the mutation against it: **41 passed, 1 failed**, and the one that fails is this one. Also corrected `"These four assertions"` in the comment above — there were seven.

REASONED. `bool(issues) and` in `field_read` cannot change any output: `if not issues:` returns first in every case where it would matter. It is the same species as the `"milestone" in i` guard your commit message describes deleting from the first draft — a term whose justification is real but whose branch is already covered. I left it: defence-in-depth is fine, and the new assertion now pins the output it protects. Noting it only because your own comment two lines up argues for deleting exactly this shape.

REASONED. The body has no `Closes` line and says why. `AGENTS.md`'s PR rule assumes one; you named the deviation rather than leaving it silent, which is the right handling, but the milestone rule this commit makes enforceable would have covered this change too if it had an issue.

---

Approving. Not merging — landing is the operator's step in this repo.

PR #223, opened 2026-09-02
prd43 w3: the install identity has one source, and a screenshot is bound to its manifest

### LINKED ISSUE #21: prd43 w3: the install identity has one source, and the CLI surface is one list
## Fence (may touch ONLY)
- `package.json`
- `README.md`
- `docs/user-guide/getting-started.md`
- `docs/demo.md`
- `packages/app/src/host/manifest-law.test.ts`

## Blocked by
Nothing.

**This is the ONLY issue in wave 3 that may claim `README.md`** — that file is a recorded coupling point in `.swarm/coupling.txt`, anchored from the install identity, the support matrix, the route counts, the CLI surface and its own grep recipes at once. The support matrix, the route counts and the recipes are owned by prd43 w4; the lab-session location is owned by prd43 w5.

Sibling lane owns the screenshot law and `scripts/dev/visit.mjs`.

## Definition of done
- `manifest-law.test.ts` is widened to the ROOT `package.json`'s `homepage`, `repository.url` and `bugs.url` — the app manifest's identical field was already guarded and its published sibling was not.
- Every other tracked mention of the repository URL is derived from that one source rather than corrected in place.
- All 8 live `KelliherL/rhizomorph` sites resolve to `launchpad-26/rhizomorph`. (`manifest-law.test.ts`'s own comment mentioning the old value is history, not a defect — leave it.)
- `rhizomorph export-otlp` is documented in the README's CLI section; `CHANGELOG:14` makes that surface the breaking-change contract.
- A law asserts the subcommands registered in `cli/index.ts` and the README's list are the same set, both directions.

### LINKED ISSUE #22: prd43 w3: a tracked screenshot is bound to the manifest that vouches for it
## Fence (may touch ONLY)
- `packages/server/src/no-personal-paths-law.test.ts`
- `scripts/dev/visit.mjs`
- `docs/screenshots/`

## Blocked by
Nothing.

Sibling lane owns `README.md`, `package.json` and the CLI-surface law. Do NOT touch `README.md` in this lane.

## Definition of done
- The blanket `.png` entry in `BINARY_EXTENSIONS` is replaced by a provenance check — the exemption is the gap, not the law.
- Every tracked PNG carries a sidecar manifest recording the synthetic root it was captured against **and the SHA-256 of the image bytes it describes**. The law recomputes that digest and fails on a mismatch, so a manifest cannot be paired with a different image.
- `scripts/dev/visit.mjs` gains a capture mode that substitutes the synthetic root **before** the screenshot and emits the sidecar itself, and it is the **only** sanctioned way to produce a tracked PNG. A hand-made capture cannot produce a valid pair — provenance is enforced by the tool that holds the root, not by a human remembering to declare one.
- A fixture PNG with no manifest makes the law red, **and so does a fixture PNG whose manifest digest names different bytes** — the detector is shown to bite on both halves.
- All 16 currently-tracked images still pass.

## Widened 2026-08-24 — the manifest must bind to the bytes

Retitled and widened from *"a tracked screenshot names the root it was captured against"*, per prd-43 ruling 4 as corrected by `94b2950`.

Naming the root is not enough, and review of #25 is where that surfaced. A sidecar declaring `/Users/operator` sits happily beside a PNG rendered against a real home directory: the law reads the sidecar, goes green, and the leak ships — because **no grep can read a path drawn into pixels**. Unbound, the manifest is an assertion the law believes rather than evidence the law checks.

The digest closes the pairing. It does not prove the pixels are clean, which is why the second half matters just as much: if `visit.mjs` is the only thing that can mint a valid pair, and it substitutes the root before the shutter, then a valid pair is itself the provenance claim. Both halves or neither — the digest alone would be a law that is true until the first time it matters, which is the shape ruling 4 exists to avoid.

---

## Correction, 2026-09-02 — the provenance half was not achievable, and the law no longer claims it

The section above says *"if `visit.mjs` is the only thing that can mint a valid
pair, and it substitutes the root before the shutter, then a valid pair is itself
the provenance claim"*, and *"both halves or neither"*. **The first half is
false, and it was false as written.**

A SHA-256 is a public, unkeyed function. Anyone holding the PNG bytes computes
the same digest `visit.mjs` would. EXECUTED by **two independent verification
seats and the orchestrator**, three times independently: a real 1×1 PNG carrying
a real contributor's home directory in a `tEXt` chunk, paired with its own
correctly-computed digest and a `syntheticRoot` of `/repo/rhizomorph`, passes
every check the law makes. `file` confirms it is a genuine PNG. Nothing in the
pair distinguishes it from one the capture tool minted.

The evidence was already in the tree before review: `visit.mjs` minted
`/tmp/rhizomorph-synthetic-repo` while all ten committed manifests declared
`/repo/rhizomorph`, so the ten tracked pairs were themselves hand-written. That
mismatch is fixed — both are now `/repo/rhizomorph` — but it is recorded here
because it is what proved the sole-minting property had never held in practice.

### What the law proves, and what it does not

**Proves:** a tracked screenshot cannot ship with zero manifest, and a manifest
cannot be silently re-paired with different image bytes — the digest is
recomputed from the actual PNG bytes, not read off the sidecar's own word. That
closes the failure this issue was FILED for: a stale manifest surviving a
re-capture, or a screenshot added with no manifest at all.

**Does not prove:** that the pixels are clean, or that `visit.mjs --capture`
produced the pair.

The gap is now **pinned by a test** rather than left implicit — a hand-forged
pair is asserted to pass, so anyone who later closes the hole must redden that
test and update this record deliberately.

### Closing it for real is future work behind an ADR

It needs an attestation only the capture tool can make: asymmetric signing, with
`visit.mjs` holding a private key that never enters this repository and the law
verifying against a public key that does. This repo has no key-custody story —
no secret store, no rotation policy, no ADR naming who holds the key or how a
fork or a new contributor's machine obtains one. Building the signing half
without that story would hand this law a checked-in secret in every way that
matters except the name, so it was not attempted.

**"Both halves or neither" is withdrawn as a gate on this issue.** It was the
right instinct — a binding that proves less than it claims is exactly the shape
ruling 4 exists to avoid — but the correct response to an unachievable half is
to stop claiming it, not to block the achievable half behind it. The section
above is left standing rather than edited, because a reader hitting the original
reasoning and then this correction can see which is current; deleting it would
hide that the claim was ever made.

### LINKED ISSUE #202: prd43 w3: ADR 0007 records that #246 lifted its packages/web constraint
## Fence (may touch ONLY)
- `docs/adr/`

## Blocked by
Nothing. `#246` landed long ago; this is the record catching up.

## The defect

`docs/adr/0007-one-derived-fleet-object.md:66` says `buildFleet` is in
`packages/web/src/fleet/`, "so nothing outside the browser can reach it", and
tracks that as issue `#246`.

`#246` landed: `69249f0` moved `buildFleet.ts` from `packages/web/src/fleet/` to
`packages/core/src/fleet/`. The ADR's stated constraint is no longer true, and
the issue it defers to is closed.

## Why the law will never catch it

EXECUTED: `packages/web/src/fleet/` **still exists** — it holds `FleetContext.tsx`
and siblings — so the cited path resolves and prd-43's citation law is satisfied.
A path that resolves while naming the wrong thing is invisible to a law that
checks existence.

This is the same class as the architecture.md finding filed alongside it, in a
file no wave-2 fence covered.

## The ADR log is append-only — do not edit the record

`docs/adr/README.md` states the log is append-only: a changed mind gets a NEW
record that supersedes the old. So the fix is **not** an edit to `0007`'s prose.
Either:

- a superseding ADR recording that the constraint was lifted when `#246` moved
  the module to `core`, and what that changed about who can reach it; or
- a "resolved by #246" note in whatever form the log's own convention allows for
  annotating a superseded record.

Read `docs/adr/README.md` first and follow what it actually says, rather than
either option above.

## Done when

- A reader of `0007` can tell that its `packages/web/` constraint no longer
  holds, without having to check the tree.
- The append-only rule is honoured — the original record is not rewritten.
- Whatever form is chosen, `#246` is named as what changed it.

## Provenance

PR #200's verification, recorded there as out of fence.

gabriel-canaan:
Closes #21
Closes #22
Closes #202

prd-43 wave 3. Three issues, one PR — the wave is the unit of queueing. Fences
are disjoint, so these ran as three parallel lanes and assemble here.

| commit | issue |
|---|---|
| `fb53fc7` | #21 — the install identity resolves to one manifest, and `export-otlp` joins the checked CLI list |
| `87ee2d9` `cbc3184` `f42135c` | review rounds on #21 |
| `305de07` | #22 — a tracked screenshot cannot ship unmanifested or be re-paired |
| `5e035f8` | #202 — ADR 0007's layering amendment names the commit that did it |

## #22 retracts a security claim rather than pretending to hold it

This is the substantive change in the PR and it is worth reading before the code.

Ruling 4, as widened by `94b2950`, said a valid manifest/PNG pair **is** the
provenance claim — because `visit.mjs` would be the only thing able to mint one.
**That was false as written, and three parties proved it independently**: two
verification seats and the orchestrator each hand-forged a passing pair in about
a minute. A SHA-256 is a public unkeyed function; anyone holding the bytes
computes the digest the tool would.

The evidence was already in the tree: `visit.mjs` minted
`/tmp/rhizomorph-synthetic-repo` while all ten committed manifests declared
`/repo/rhizomorph`, so the ten tracked pairs had themselves been hand-written.
That mismatch is fixed; it is recorded because it is what proved the sole-minting
property had never held in practice.

**What the law now proves:** a tracked screenshot cannot ship with zero manifest,
and a manifest cannot be silently re-paired with different bytes — the digest is
recomputed from the actual image, not read off the sidecar's own word. That
closes the failure #22 was filed for.

**What it does not prove, and now says so:** that the pixels are clean, or that
`visit.mjs --capture` produced the pair. **The gap is pinned by a test** that
asserts a forged pair passes, so closing it later requires reddening that test
and updating the record deliberately rather than by accident.

Closing it for real needs an attestation only the capture tool can make —
asymmetric signing, with a private key that never enters this repository. This
repo has no key-custody story: no secret store, no rotation policy, no ADR naming
who holds the key or how a fork obtains one. Building the signing half without
that would hand the law a checked-in secret in every way but the name, so it is
future work behind an ADR. #22's issue body carries the same correction, with
"both halves or neither" explicitly withdrawn.

## What is the sibling case?

**Four rounds of it, each closing the spelling in front of it — that is the
finding, more than any individual defect:**

| round | closed | what the next round found |
|---|---|---|
| 1 | substring → exact assertions | dot-suffix `rhizomorph.archive.git`, wrong host `mirror.github.com`, only the first clone line read |
| 2 | whole-URL parse, `matchAll` | a **false positive**: `git clone <url> <dir>` now rejected |
| 3 | optional target directory | four more clone spellings; and the **case** axis on PNG |
| 4 | tokenized parsing; case-insensitive classifier | — |

Two axes turned out to matter, and both were closed only when the lane stopped
patching spellings and enumerated:

- **#21's clone law** stopped matching lines and started **tokenizing** them. A
  pattern that skips what it cannot parse treats an unparseable line as absent
  rather than suspicious — the wrong default for a law whose job is to catch a
  wrong clone URL.
- **#22's classifier** was case-sensitive, then case-insensitive-but-`.png`-only.
  The blocker moved one axis over: `.jpg`/`.webp`/`.gif` bypassed it exactly as
  `.PNG` had. Its own retraction reasoning transfers verbatim — *a real leaked
  screenshot carries its path in pixels* — which is as true of a JPEG.

None of these. All EXECUTED **on the assembled tree**, not taken from the lanes'
own runs — cherry-picking three lanes onto a base that moved three times is
exactly where a fix quietly stops biting. Spellings marked ★ appear in no
finding; they were invented for the re-check and fell out for free, which is the
signal the classes are closed rather than the reports.

**#21 — attacker's repository appended to the real `docs/demo.md`:**

    git clone <wrong> # alternate source        1 failed
    git clone --depth 1 <wrong> rhizomorph      1 failed
    ␣git clone <wrong>   (one space indent)     1 failed
    $ git clone <wrong>  (prompt prefix)        1 failed
    ★ TAB-indented git clone <wrong>            1 failed
    ★ git clone -q <wrong>                      1 failed
    git clone --depth=1 <CORRECT>              28 passed   ← the round-2 false positive, gone
    CONTROL untouched                          28 passed

**#21 — `repository.url`:** `rhizomorph.archive.git` → 8 failed;
`mirror.github.com/...` → 8 failed; correct → 19 passed. One seat verified the
direction that matters: **no spelling makes a wrong URL parse to the correct
`launchpad-26/rhizomorph`**, and both describe blocks pin that literal, so a
mis-parse cannot produce a false pass.

**#22 — unmanifested files under `docs/screenshots/`:**

    leak.png                                    1 failed   (baseline)
    UPPER.PNG / .PnG / .Png                     1 failed   (case axis)
    leak.jpg / .webp / .gif                     2 failed   (format axis)
    ★ fmt.JPEG                                  2 failed
    ★ fmt.png.bak                               1 failed
    identical bytes, correctly manifested      32 passed   ← not over-tightened

Other probes: a PNG in a **subdirectory** with no sidecar → red; a manifest
re-paired to another screenshot's real digest → red; a manifest with no `sha256`
key, or invalid JSON → red; `NON_SCREENSHOT_BINARY_ASSETS` edited in one copy
only → red. The pin list cannot exempt anything, since `isTrackedScreenshot` is a
`startsWith`. All 16 tracked PNGs scanned byte-wise for home paths, slug forms,
`DESKTOP-*`/`LAPTOP-*` and `tEXt`/`iTXt`/`zTXt`/`eXIf` chunks: **zero hits, no
metadata chunks at all**, with a control proving the scanner works.

## Fence widening

**None.** `git diff --name-only origin/main...HEAD` is the 19 files the three
issues declare between them. `README.md` is a `.swarm/coupling.txt` coupling
point and was touched only by #21, which is the one issue in this wave permitted
to claim it.

## Verification

Full pipeline plus **two fix re-review rounds**, each with two independent seats
(Opus `--effort xhigh`, Codex `gpt-5.6-terra` `xhigh`) in disposable worktrees.
The seats split on the verdict in both re-reviews, and both splits were resolved
by execution rather than by vote:

- Re-review 1: one seat called the uppercase-`.PNG` gap mitigated, because the
  text sweep catches the bytes. That reasoning fails on the case that matters —
  a probe with a path in a `tEXt` chunk is caught, a **pixel-drawn** leak is not.
  A clean uppercase PNG passed 22/22.
- Re-review 2: one seat called the clone-parser gaps non-blocking as
  pre-existing. They are pre-existing, but the commit subject and test name both
  said *every*, so the claim was false either way.

    VITEST_MAX_WORKERS=6 npm test     0   ->  402 files, 7175 passed
    screenshot law 4x concurrent          ->  4/4 green

## Known and deliberately out of scope

- **`scripts/gate.sh:218`'s NUL-byte exclusion is case-sensitive** while the law
  is now case-insensitive, so a correctly-manifested `.PNG` would fail the
  landing gate on its ~4,005 NUL bytes. It fails **closed** — nothing unsafe
  lands — and `gate.sh` is outside every wave-3 fence, so no lane touched it.
  Being filed separately.
- `/repo/rhizomorph` is written literally in three files with nothing binding
  them (`packages/core/src/fixtures.ts:21`, `scripts/dev/visit.mjs:105`, and the
  law). Drift would make future captures red while the ten backfilled stay green
  — loud, not silent.
- `CLONE_SITES` is a hand-pinned list of three files with no test proving it
  complete. A repo-wide discovery sweep would close it.


ciaran-slow (approved):
## Approving. What verified, and the one thing that did not.

Reviewed commit-by-commit in a disposable clone at `d666173` (the branch is level with `origin/main`, so no rebase drift). Node **v22.23.2**.

### What verified in the PR's favour — all EXECUTED

I re-derived the PR body's claims rather than reading them, and **the laws bite where they say they do**. Eight mutations, each applied to the assembled tree and reverted:

| # | mutation | result |
|---|---|---|
| A | unmanifested `docs/screenshots/leak.png` | **1 failed** |
| B | unmanifested `LEAK.PNG` (case axis) | **1 failed** |
| C | unmanifested `leak.jpg` (format axis) | **2 failed** |
| D | `drawer.png.manifest.json` sha256 → `0`×64 (re-pairing) | **1 failed** |
| E | wrong clone URL in `README.md` | **1 failed** |
| F | wrong `repository.url` in root `package.json` | **6 failed** |
| G | wrong URL behind `--depth 1` (the round-3 finding) | **1 failed** |
| H | `README` drops `export-otlp` | **2 failed** |

A–D reproduce the PR body's own failure counts exactly (1/1/2/1). **G is the one worth calling out**: the defect three review rounds were chasing is genuinely closed, not just described as closed.

I also invented seven clone spellings that appear in no finding — `-b main <evil>`, tab-indent, `--config x=y`, `--` end-of-options, backslash continuation, CRLF. **Six of the seven reddened**, which is the signal the class is closed rather than the reports.

**#202 verified factually**: `git show 69249f0` is the `git mv` for #246; `packages/web/src/fleet/` still exists holding `FleetContext.tsx`; `buildFleet.ts` is in `packages/core/`. The amendment appends and does not rewrite, honouring `docs/adr/README.md`'s append-only rule.

**#22's retraction is the right call and is honestly recorded.** A SHA-256 is unkeyed; the sole-minting property never held. `packages/server/src/no-personal-paths-law.test.ts:716` pins the gap by asserting a forged pair *passes*, so closing it later requires reddening that test deliberately. That is the correct way to record a limitation.

Two structural things I checked because they are where this shape usually leaks:
- `SANCTIONED_SYNTHETIC_ROOT` is **imported** (`FIXTURE_REPO_PATH`), not retyped, and `visit.mjs`'s literal is bound to it by test at `:675`.
- The text sweep excludes binary extensions repo-wide, but `:580` forces every tracked image into either "manifested screenshot" or the pinned allowlist — **no third silent category**. That is the sibling case closed properly.

Gates I ran myself on the assembled tree: `typecheck` **0**, `lint` **0**, `VITEST_MAX_WORKERS=6 npm test` **0** → **402 files, 7171 passed**.

---

### The one finding — fixed in `711346c`, not handed back

**Whitespace between `git` and `clone` hid a wrong URL, at both layers.** EXECUTED.

Rounds 2 and 3 normalized whitespace *before* `git` — indentation, tab, a `$ ` prompt — and left the gap *between* `git` and `clone`. The shell collapses it, so `git  clone <url>` is the same invocation; `git  clone --depth 1 /nonexistent` reaches `fatal: repository ... does not exist`, i.e. git parsed the subcommand fine.

Both halves missed it, which is why it was **silent rather than loud**:

- `cloneLinesIn` matched `startsWith('git clone ')` → the line was not recognized as a clone invocation at all and was skipped. That is exactly the *"treats an unparseable line as absent rather than suspicious"* default the round-3 comment directly above it argues against.
- `CLONE_INSTRUCTION_RE` carried the same single space → a **new** doc whose only clone instruction was written that way was never discovered, `CLONE_SITES` still pinned to exactly three, and every check below passed vacuously.

Measured before the fix:

```
git  clone <wrong-repo>  appended to docs/demo.md   ->  28 passed   (invisible)
same line in a new docs/*.md file                   ->  28 passed   (invisible)
single-space equivalents                            ->  1 failed / 2 failed
```

After: **1 failed / 2 failed**, matching single-space behaviour.

**What mutation would the new test survive?** None — reverting the two regexes to their single-space form while keeping the test reddens exactly that test (`1 failed | 28 passed`), so it cannot pass for a reason other than the one it claims.

Gates with the fix, on the committed tree: `typecheck` **0**, `lint` **0**, suite **0** → **402 files, 7172 passed**. CI: all **7** jobs pass, and I confirmed the steps genuinely executed (Build / Test / Typecheck / Lint / Packaging guard / Boot smoke all `success`, none skipped) rather than trusting the green bar.

---

1. **Two "known and out of scope" bullets understate what you shipped.** The body says `CLONE_SITES` is "hand-pinned … with no test proving it complete" — but `discoverCloneSites()` sweeps the tracked tree and `:259` pins the result. It also says `/repo/rhizomorph` sits in three files "with nothing binding them" — the law imports `FIXTURE_REPO_PATH` and `:675` binds `visit.mjs` to it. Both claim *less* than the code does, which is the safe direction; worth correcting only so the record is accurate.
2. **Suite count.** The body reports 7175 passed; I measured **7171** on your tree (`7171 passed | 3 expected fail | 6 skipped`). Probably a different tree state — flagging it only because the numbers are load-bearing in this repo.
3. **`gate.sh:218`** — your caveat is accurate: the exclusion list is lowercase-only (`*.png|*.jpg|…`) while the law is now case-insensitive. Confirmed unreachable today — zero uppercase-extension images are tracked. Fails closed. Correctly out of fence.
4. **`.nvmrc` pins `22.22.2` exactly**, which `nvm use` cannot resolve unless that precise patch is installed; I used `v22.23.2`, which satisfies the `^22.22.2` engines range. Pre-existing, untouched by this PR.

### What I did NOT run

- **`scripts/gate.sh`** — that is the operator's step and running it *is* the landing.
- **The load gate** (4× concurrent batches) and the serial timing pass. You report 4/4 green on the screenshot law; I did not re-measure it.
- **Windows** — no coverage exists in `ci.yml` for it.

I have **not merged this** — merges here are manual, and my commit `711346c` should have your eyes before it lands.

PR #209, opened 2026-09-02
docs(prd): prd-42, prd-41 and prd-29 are shipped, with the drift each was hiding

ciaran-slow:
Three PRDs whose milestones had zero open issues while the documents still read `BLESSED`. Each retires to `done/` with the closeout form prd-44/45/46 established: rulings as they landed, success criteria assessed with executed evidence, open questions answered, **what the plan got wrong**, and residuals with owners named.

| PRD | Issues | Waves | Criteria |
|---|---|---|---|
| `prd-42` — one path, one spelling | 28 | 11 | 4/4 met |
| `prd-41` — the laboratory is confined in fact | 7 | 3 + 2 late defects | 5/5 met |
| `prd-29` — the identity seam | 4 | 3 | 5/5 met |

**No `Closes` line, deliberately.** Every issue in all three milestones is already closed; this is the record catching up with the tracker, not work closing an issue. The milestones (`prd29`, `prd41`, `prd42`) were closed on GitHub alongside this.

**Independent of #207.** That PR closes out prd-45/46/47. No file overlap, and nothing here cites a prd-45/46/47 path — only prose references like *"prd-45 ruling 4"*, which survive either merge order. Either PR can land first.

## The one that is not just record-keeping

`SECURITY.md` carried a **live, false disclosure**. Its *"One live gap in that confinement"* paragraph told readers to treat `lab fork` on an untrusted checkpoint as running that repo's install hooks — *"because that is what it does"*. That stopped being true on 2026-08-26 when `#7` landed `--ignore-scripts` and the namespace law started walking the installing path.

prd-43 ruling 1's citation law checks that a cited **path resolves**; nothing checks that the **sentence is still true**. So a security document warning about a closed hole sat there for a week.

Rewritten to record the gap as closed rather than deleting it — the file explicitly invites reports of that class — and keeping the distinction a reader needs: `install` still defaults to true, so a restore still installs dependencies. What it no longer does is execute the checkpointed tree's lifecycle scripts.

## What is the sibling case?

**The stated mechanism vs. the ruling — found three times, independently, all predating prd-47 writing it down.**

[...]

Issue #221, opened 2026-09-02
prd30 w2: the one-card law says the whole sentence it was written to say


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/disclosure/`
- `packages/web/src/tide/`
- `packages/web/src/app/title-residue-law.test.ts`

## Blocked by
Wave 1 (#220) — **landed** on `main` as `8ee1b498`. It retired the third of the
three idioms, which is what unblocks this one.

**#39** owns `packages/web/src/scene/palette.ts` and `packages/web/src/theme/theme.css`.
This issue claims neither.

**Model:** sonnet. **Wave:** 2.

## Regroomed 2026-09-08, before any plan was written

This issue's body was wrong in three places and would have sent a cold builder
at work that cannot be done. The original text asked to *re-seat* `MarkHoverCard`
and the loupe read-out onto the shared disclosure card, preserving their content,
and to empty an allowlist. All three premises fail against the code — see
**What was wrong** below, and the timeline comment for the measurements. The
issue's *purpose* is unchanged: prd-30 S1's acceptance says **"a law test asserts
no other directory renders card chrome"**, and today's law is narrower than that
sentence. Paying that is still what closes prd-30.

## Why

`packages/web/src/disclosure/one-card-law.test.ts:14-26` says what it is not yet
saying, and why:

> A law written to the full sentence would fail on landing, so wave 1 would have
> to disable it — and a disabled law is worth less than a narrow one. So the law
> lands on **this card's own markers** ... When wave 3 retires the three idioms,
> the same test widens to the full sentence with the allowlist empty.

`cardChromeOutsideDisclosure` matches one string, `data-disclosure-card`. It
catches a surface that **copies this card**. It says nothing about a surface that
grows *a different* card — which is the thing prd-30 exists to prevent, and the
thing the sentence in S1 actually asks for.

**The three idioms are named in prd-30 S1:159** — *"It replaces `MarkHoverCard`,
the loupe read-out and every `title=`"*. #220 retired the third. What is left is
to make the law say its whole sentence.

## What was wrong with this issue's previous body

Recorded rather than deleted, because the next reader deserves to know the
premise was checked rather than assumed.

1. **"Re-seat `MarkHoverCard` onto the shared card, content preserved" — not
   possible, and not desirable.** `tide/ChapterMarks.tsx:399` is an interactive
   **menu**: `group.members.map(...)` renders one `<button>` per member, each
[...]

Issue #220, opened 2026-09-02
prd30 w1: no surface explains itself with a native title attribute


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/panels/`
- `packages/web/src/drawer/`
- `packages/web/src/recordings/`
- `packages/web/src/lane-page/`
- `packages/web/src/app/`
- `packages/web/src/why/`
- `packages/web/src/trace/`
- `packages/web/src/interaction/`
- `packages/web/src/theme/tokens.test.ts`
- `packages/web/src/theme/legibility.test.ts`
- `packages/web/src/theme/kind.test.tsx`
- `packages/web/src/App.test.tsx`

## Blocked by
**#39** owns `packages/web/src/theme/theme.css`, `packages/web/src/theme/category.test.ts`, `packages/web/src/scene/palette.ts` and `packages/web/src/scene/palette.test.ts`. This issue claims neither `theme/theme.css` nor anything under `scene/` — **`packages/web/src/scene/` is deliberately excluded from the sweep** and its three tooltips move in a later pass, after #39 lands.

**Narrowed 2026-09-07, before any code moved.** Three directories that were claimed at groom
time are now **out of scope** for this sweep and owned by live lanes: `packages/web/src/replay/`,
`packages/web/src/lab/` and `packages/web/src/concierge/` are out of scope here.
The fence lint found each overlapping a live lane, and the overlap is not a granularity
artefact — the law tests at `replay/mutating-calls-law.test.ts:722-732` and
`concierge/explicit-invocation-law.test.ts:245-254` `readFileSync` the very surfaces this sweep
would edit (`RotateButton.tsx`, `InstrumentButton.tsx`), and #216 (live on `origin/prd20`, pushed
5 h before this edit) owns both law files. #235 owns `lab/no-live-fleet-law.test.ts` and
`lab/launch/explicit-invocation-law.test.ts` — out of scope here too.

The **11 tooltips** this defers are the recorded residue the Definition of done already provides
for, and this line is the record: 8 in `replay/` (`index.tsx`, `RotateButton.tsx`, `Banner.tsx`),
2 in `lab/` (`compare/ComparisonSurface.tsx`), 1 in `concierge/` (`InstrumentButton.tsx`) — all
owned by other lanes for now. **77 of the 88** in the claimed directories convert here. Whoever
picks up the residue should re-run the lint first: once #216 and #235 have landed, the three
directories are free and it is one small commit.

[...]

Issue #216, opened 2026-09-02
prd20 w1: the wizard switches the watched repo through the retarget route


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/connect/`

**Widened by the plan (2026-09-04).** The client for a sixth mutating call cannot live in connect/ (its ruling-7 law forbids a verb or a request init there), and a new call site moves four exact enumerations and two documents. Listed here so fence-lint sees the real footprint before dispatch; recorded again on the PR before the change, as AGENTS.md requires:
- `packages/web/src/concierge/retarget.ts` (new)
- `packages/web/src/concierge/retarget.test.ts` (new)
- `packages/web/src/concierge/explicit-invocation-law.test.ts`
- `packages/web/src/replay/mutating-calls-law.test.ts`
- `packages/web/package.json`
- `packages/contract/src/contract-coverage-law.test.ts`
- `packages/contract/src/retarget.contract.test.ts` (new)
- `packages/server/src/api/route-class-law.test.ts` (the outbound call-site table only)
- `README.md`
- `SECURITY.md`
- `CHANGELOG.md`

## Blocked by
**Corrected 2026-09-04 — the paragraph below was true of the original fence and is false of the real one.** The new client module is a new outbound call site, and the README's call-site list is pinned by the outbound-fetch table in `packages/server/src/api/route-class-law.test.ts` — which is **#232**'s fence (prd43 w6, In progress). `scripts/fence-lint.sh 216 232` reports the OVERLAP (executed 2026-09-04, after the widening above). Dispatch after the prd43 w6 bundle merges to `main`; #213 waits on the same lane. #234 (prd43 w6, Backlog) will edit the same table and README paragraph — re-sequence behind it too if it goes In progress first. The original text is kept, not deleted.

Nothing. No open issue claims `packages/web/src/connect/`.

Sibling lanes: #21, #23 and #24 own `README.md`; #23 also owns `packages/server/src/api/route-class-law.test.ts` and `.github/workflows/ci.yml`. This issue adds no route and changes no route class — `POST /api/retarget` already exists, gated and tested — so it touches none of them.

**Model:** sonnet. **Wave:** 1.

## Why

prd-20's 2026-08-24 amendment cut this PRD to two outcomes. **One of them is already true** (see below); this issue is the other, and it is the whole remaining build.

The wizard tells the operator the capability does not exist. `packages/web/src/connect/wizard.tsx:753`:

[...]

PR #224, opened 2026-09-02
docs: the runbook and the retained PRDs say what is true on main

### LINKED ISSUE #215: AGENTS.md's CI claims match what the workflow actually runs
## Fence (may touch ONLY)
- `AGENTS.md`
- `docs/prds/done/prd-24-the-seam-that-lies.md`
- `packages/server/src/runbook-delivery-law.test.ts`

## Blocked by
Nothing. No open issue claims any of these three.

Sibling lanes: #23 owns `README.md` and `.github/workflows/ci.yml`; this issue corrects prose *about* that workflow and must not edit the workflow itself. #201 owns `docs/architecture.md`; #202 owns `docs/adr/`; #203 owns `packages/server/src/doc-citation-law.test.ts`.

**Model:** sonnet. **Wave:** none — this is deliberately unwaved, like #62, #63 and #66. It shares no path with prd43 waves 3, 4 or 5 and must not queue behind them: `AGENTS.md` is loaded into every agent session, so every lane dispatched before this lands inherits the stale claim it corrects.

## The defect

**`AGENTS.md` states a CI behaviour the workflow no longer has, and it is the file that tells every agent and contributor how this repo works.**

Its CI section says:

> `Build` runs **before** `Test` (`ci.yml:60`), and a failing step skips every later step on its leg: if `Test` fails, typecheck, lint, packaging and boot smoke silently do not run. A red leg is therefore worth more than one failing test

Measured on `main` today, that is false. `.github/workflows/ci.yml`:

- `Typecheck` — `if: "!cancelled()"`
- `Lint` — `if: "!cancelled()"`
- `Packaging guard` — `if: "!cancelled() && steps.build.outcome == 'success'"`
- `Boot smoke` — `if: "!cancelled() && steps.build.outcome == 'success'"`

(Deliberately cited by step name, not line number — see the two rotted citations below.)

All four run after a red `Test`, provided `Build` was green. The workflow even carries a comment above the packaging guard explaining why `always()` was rejected in favour of gating on `Build`'s own outcome: the packaging guard passes vacuously over an empty `dist/`, and the boot smoke's bin falls back to TS source when `dist` is absent, so `always()` would report both green having verified nothing built. That is a better answer than the one `AGENTS.md` describes as missing.

## Why this one matters more than an ordinary stale line

`AGENTS.md` is the shared runbook and is loaded into every agent session through `CLAUDE.md`'s import. A contributor reading this line draws two wrong conclusions:

1. That a red `Test` leg means four unrun gates — so they re-run CI or discount the leg, when the evidence is already there.
2. That adding `if: always()` is outstanding work. It is not; it was considered and deliberately improved on.

**prd-24's closing amendment carries the same stale claim** and books it as future work: residual 2, "CI stops producing evidence after a red Test — no `if: always()` on Typecheck, Lint, the packaging guard or the boot smoke ... **Described for the next groom as one small issue**". That residual is discharged by the tree and must be recorded as discharged, or the next groom mints an issue for work that is done.

**prd-25 ruling 1 leans on the same premise** to argue the Windows leg belongs on `pack-smoke`. Its conclusion survives on other grounds and #211 already records why; this issue does not need to touch prd-25.

## Definition of done

- `AGENTS.md`'s CI paragraph states what the workflow actually does: the post-`Test` steps run under `!cancelled()`, the two build-dependent ones additionally require `Build` to have succeeded, and the reason `always()` was rejected.
- **Two rotted line-number citations in the same section are corrected.** Both re-verified on `main` at `26c48c7`:
  - `` `.github/workflows/ci.yml:166` `` for the `pack-smoke` job — the job is at **line 197**. Wrong by 31 lines.
  - `` `ci.yml:60` `` for `Build` — this one is *currently* right, and that is the trap: it was wrong before #209 merged and drifted back into correctness by accident. Do not "verify" it and move on.

  Cite the step and job by **name**, not by number. AGENTS.md's own guidance on `scripts/gate.sh` already says why — those anchors "have carried three different sets of line numbers within one milestone" — and `.swarm/coupling.txt` repeats it: *a registry entry pinned to a number for a file that moves is a registry entry that rots by construction*. This section is the same shape and has not had the same lesson applied.
- prd-24's closing amendment marks residual 2 **discharged**, naming the gating that discharges it, rather than leaving it described-for-the-next-groom. Ruling numbers stay citable and are not renumbered.
- While in that amendment: residual 1 is **also discharged** and gets the same treatment. It claims `web/src/recordings/no-live-fleet-law.test.ts:31` is "the last flat walk of the three ruling 3 named" — that file is recursive today (`walkSourceFiles`, plus a `realSourceFileNames()` recursive `readdirSync` computing the floor, plus a committed nested fixture under `recordings/recursion-fixture/`). Verify the two walks it calls "only vacuously correct" — `interaction/no-model-call-law.test.ts` and `connect/index.test.tsx` — before writing either of them off, and say which state you found.
- `packages/server/src/runbook-delivery-law.test.ts` gains a check that fails if `AGENTS.md` claims a CI step is skipped that the workflow actually runs — parsed from `ci.yml`, not from a copy of the sentence. Without it this line simply rots again the next time a gate changes.

## Sibling case

Already found by looking, and all in the same ~25-line section — do not stop at the behavioural claim:

- **The `pack-smoke` line number is wrong** (166 vs 197), as above.
- **`#281` is a dead tracker reference.** The CI paragraph says a built clone "could not boot on Windows at all until `#281`". That issue does not resolve — it is from the previous tracker, which is #66's whole finding, and `AGENTS.md` is the file #66 singles out as worst affected. Say what fixed it (`pathToFileURL` in `packages/server/bin/rhizomorph.mjs`) rather than citing a number no one can open. Coordinate with #66 if it is in flight; otherwise fix it here, since this issue is already in the file.
- **The `windows-latest` gap paragraph is now owned.** It correctly says the leg is absent from the test/typecheck/lint/boot grid and present only on `desktop.yml`'s installer-packaging leg — verified, `desktop.yml:40`. That gap is now #211's, and the paragraph should name the live issue instead of leaving it as an unowned observation.

Three defects of three different kinds in one paragraph — a false behavioural claim, a rotted line number, and a dead citation. Fixing only the one this issue is titled after is the exact "fix that handles the case its author considered" shape AGENTS.md itself warns about.

Remove `if: "!cancelled()"` from the `Lint` step in `ci.yml`. The new runbook law must go red. A law that only greps `AGENTS.md` for the word "skips" survives that mutation and proves nothing.

### LINKED ISSUE #222: the retained PRDs state what is true on main, and the coupling map names two registration seams
## Fence (may touch ONLY)
- `docs/prds/prd-25-the-third-platform.md`
- `docs/prds/prd-20-the-concierge.md`
- `docs/prds/prd-30-the-open-hand.md`
- `.swarm/coupling.txt`

## Blocked by
Nothing. No open issue claims any of these.

Sibling lanes: #215 corrects `AGENTS.md` and `docs/prds/done/prd-24-the-seam-that-lies.md`; this issue must not enter either. #201 owns `docs/architecture.md`; #202 owns `docs/adr/`.

**Model:** sonnet. **Wave:** none — this is deliberately unwaved, like #62, #63 and #66. It shares no path with prd43 waves 3, 4 or 5 and must not queue behind them: `AGENTS.md` is loaded into every agent session, so every lane dispatched before this lands inherits the stale claim it corrects.

## The defect

Three retained PRDs describe remaining work that the tree has already done. They were reconciled on 2026-08-22 and amended on 2026-08-24; `main` has moved past them, and grooming them on 2026-09-02 turned up four specific contradictions. Left as-is, the next groom mints lanes for finished work — which is precisely what the 2026-08-22 reconciliation was written to prevent.

Each item below was verified against `main` during that groom.

### prd-25 — three claims, all overtaken

1. **Ruling 4's node-floor call is discharged.** It asks for the README's "Node >= 22" to change so a first `EBADENGINE` reads as a stated requirement. `README.md:352` already reads **"Node >= 22.22.2"**, names `engines` as the source of truth, and points at CONTRIBUTING.md for the Node 20 behaviour.
2. **Ruling 5's matrix rows are discharged.** It asks for Windows to enter as an honest partial and for the macOS row's "treat it as untested" to be corrected. `README.md:347-350` already carries four rows; macOS reads "CI-verified on every push" with its legs named, and Windows reads as an honest partial citing the fixed boot defect and prd-25 itself.
3. **The Evidence bullet for the boot defect is history.** The amendment already says so; the PRD body still states it in the present tense. `packages/server/bin/rhizomorph.mjs:52` uses `pathToFileURL(distEntry).href`.
4. **Ruling 1's mechanical premise is stale.** It argues the leg must start on `pack-smoke` because "a failed `Test` step **skips every later step on that leg**". That has not been true since the post-`Test` steps were gated on `!cancelled()` — see #215, which corrects the same claim in `AGENTS.md`. The ruling's *conclusion* stands on other grounds and #211 records which; the ruling should say so rather than resting on a premise the repo has retired.

What is genuinely left of prd-25 is #210, #211 and #212 — nothing else.

### prd-20 — ruling 7 is satisfied by the tree

The 2026-08-24 amendment cut this PRD to two outcomes. The second, ruling 7's narrowed no-tmux launch promise, already ships: `packages/web/src/connect/wizard.tsx:908-913` renders the detached launch as its own answer — no tmux window, nothing attached, the harness may exit on its own, run it yourself in a terminal. That is ruling 7 in behaviour. Only the first outcome remains, and it is #216.

### prd-30 — "MarkHoverCard remains" is the wrong word

The Outcome header says `MarkHoverCard`, the loupe read-out and the `title=` sweep "remain", which reads as *absent*. `MarkHoverCard` exists — `packages/web/src/tide/ChapterMarks.tsx` — and so does the loupe read-out (`tide/Loupe.tsx`, `tide/TideDock.tsx`). What is owed is a **re-seat** onto the shared card so `disclosure/one-card-law.test.ts` can widen to its full sentence, which is what that law's own doc comment says. #220 and #221 hold it.

## Also: two registration seams are missing from the coupling map

Both were found the hard way while fencing this groom, and both are the shape `.swarm/coupling.txt` exists to record — a file a change *elsewhere* forces an edit to:

- **`packages/server/src/api/route-class-law.test.ts`** — line 99 pins `expect(ROUTE_CLASSES.length).toBe(25)`, deliberately asserted independently of the array so the walk cannot pass vacuously. Any new route anywhere reddens it, from a directory the lane never entered. It is why #213 is blocked on #23.
- **`packages/server/src/server/collector-loader.ts`** — the six collectors are registered here by static import, because a variable dynamic import cannot be bundled. A new collector anywhere forces an edit to this one file. Note the *sibling* law, `server/collector-wrap-boundary.test.ts`, is deliberately **not** an entry: it derives its set by walking the collectors directory and its own test proves a sixth collector is picked up automatically. Recording it would teach the next lane to edit a file that correctly needs no editing.

Each entry carries its reason, per the file's own format — an entry with no reason is worse than no entry.

## Definition of done

- The three PRDs' Outcome headers and the named rulings state what is true on `main`, with the superseding evidence cited by path.
- **Nothing is renumbered and no ruling is deleted.** A ruling that has been overtaken is marked discharged with what discharged it, the way prd-24's closing amendment handles its residuals. The repo's history is why findings are checkable later.
- Where a ruling's premise is stale but its conclusion holds (prd-25 ruling 1), both facts are stated. Silently repairing the premise hides that the repo learned something.
- `.swarm/coupling.txt` gains the two entries above, each with its reason, and does **not** gain `collector-wrap-boundary.test.ts`.
- No PRD in `docs/prds/done/` is touched — #215 owns the one that needs it.

## Sibling case

Four contradictions were found in three PRDs during one groom, all of the same shape: an Outcome header written at reconciliation time describing a tree that has since moved. The other retained PRDs — prd-14, prd-17, prd-27, prd-33, prd-34 — have headers of the same vintage and were not audited line by line here. Check them in the same pass rather than waiting for the next groom to trip over them one at a time.

ciaran-slow:
Closes #215
Closes #222

Two doc-truth commits, prd43, deliberately unwaved — neither shares a path with waves 3–5, and `AGENTS.md` is loaded into every agent session, so every lane dispatched before this lands inherits the stale claim it corrects. Base is `main`.

Review commit by commit; they are independent.

## What was wrong

`AGENTS.md` said a red `Test` costs you the rest of the leg — "typecheck, lint, packaging and boot smoke silently do not run". All four run today, gated on `!cancelled()` (the two that need a real `dist/` additionally on `steps.build.outcome`). `always()` was considered and **rejected** on record, because the packaging guard passes vacuously over an empty `dist/` and the boot smoke's bin falls back to TS source — so on a Build-red leg `always()` would report both green having verified nothing built.

This is not theoretical. Grooming prd-25 on 2026-09-02 produced an issue body that repeated the stale premise as justification, because the runbook said so. It was caught by reading `ci.yml`, not by reading the runbook.

Two rotted line citations went with it: `pack-smoke` cited as `ci.yml:166` (it is at 197), and `Build` cited as `ci.yml:60` — which was wrong and then drifted **back** into correctness when #209 moved the file. That second one is the worse failure, because spot-checking it says "fine". Both now cite job and step by name, the lesson `.swarm/coupling.txt` already records for `scripts/gate.sh`. The dead `#281` reference is replaced by what actually fixed it.

## What is the sibling case?

**Commit 1.** The stale sentence was one of four defects of three kinds in the same ~25-line section — a false behavioural claim, a rotted line number, a dead tracker citation. Fixing only the one the issue is titled after is precisely the "fix that handles the case its author considered" shape. All four are in.

The same stale fact was booked twice: prd-24's closing amendment carries it as residual 2, "described for the next groom as one small issue". Both of its residuals are discharged by the tree and are now recorded as such — otherwise the next groom mints lanes for finished work. Residual 1's two *further* walks are still flat and are recorded as **still latent, not fixed**: `interaction/` and `connect/` have zero subdirectories today, so their laws are vacuously correct rather than wrong. Out of fence here; named, not touched.

**Commit 2.** Four contradictions across three PRDs, all one shape — an Outcome header written at reconciliation time describing a tree that has since moved. prd-14, prd-17, prd-27, prd-33 and prd-34 have headers of the same vintage; named on #222 for the same pass rather than left to be tripped over one at a time.

`.swarm/coupling.txt`'s new entries have their own sibling: `server/collector-wrap-boundary.test.ts` sits right beside `collector-loader.ts` and is **deliberately not** an entry, with the reason written in — it derives its set by walking the directory, so recording it would teach the next lane to edit a file that correctly needs no editing.

`runbook-delivery-law.test.ts` derives the gated-step set by **parsing `ci.yml`**, never from a maintained list — a maintained list is exactly what went stale here — and pins it as an equality in the shape `api/route-class-law.test.ts` uses.

**EXECUTED.** Removed `if: "!cancelled()"` from `ci.yml`'s `Lint` step:

```
FAIL  src/runbook-delivery-law.test.ts > the runbook names exactly the steps the workflow still runs after a red Test
AssertionError: expected [ Array(3) ] to deeply equal [ 'Typecheck', 'Lint', …(2) ]
Tests  1 failed | 10 passed (11)
```

Restored it: `Tests 11 passed (11)`.

Two further rigged-input tests prove the parser and the line-citation detector bite rather than passing on an empty file or a typo'd regex — the standard the laws beside them set ("a grep law that has never been shown to fire is decoration").

A law that merely grepped `AGENTS.md` for the word "skips" would survive the mutation above and prove nothing. That is why the assertion runs the other way: the runbook must name the steps the workflow *actually* still gates.

## Fence widening

None.

## Gates

`npm run typecheck` green · `npm run lint` green (872 files) · `VITEST_MAX_WORKERS=6 npm test` → **401 files, 7128 passed, 3 expected fail, 6 skipped**.

`scripts/fence-lint.sh 215 222` passes with 3 and 4 paths extracted and no overlap — paths read, not just the exit code.


ciaran-slow:
Force-pushed a correction to commit 1, before review.

**I reported `npm run typecheck` green in the original commit message and PR body. It was not.** The parser I added indexed a regex match object directly, and `noUncheckedIndexedAccess` is on in `@rhizomorph/server`, so `tsc --noEmit` failed with four `TS2532`. I read a `tail -5` of the typecheck output instead of its exit status, and the failing workspace had scrolled past.

CI caught it on `build-test-boot (macos-latest, current)` — which is incidentally a live demonstration of this PR's own point: **`Test` passed and `Typecheck` still ran and reported the failure**, which is exactly the gating the runbook denied existed.

Fixed by reading capture groups through optional indexing and testing for `undefined`. Re-run by exit code this time, not by reading a tail:

- `npm run typecheck` — exit 0, zero `error TS` lines
- `npm run lint` — clean, 872 files
- `VITEST_MAX_WORKERS=6 npm test` — 401 files, 7128 passed, 3 expected fail, 6 skipped

Mutation re-proved after the fix: removing the `!cancelled()` gate from `ci.yml`'s `Lint` step gives `1 failed | 10 passed`; restoring gives `11 passed`.

Folded into commit 1 rather than added as a third commit, so one-commit-per-issue still holds.

gabriel-canaan (commented):
## Review of #224 — read commit by commit, on the branch with `origin/main` merged in

Reviewed at `aeb8611`, merged onto `origin/main` (`ebcc5d6`) first — the branch was 8 commits behind, and main had touched `README.md`, which commit 2 makes claims about. Node 22.22.2.

Every factual claim I could execute, checked out. Nothing in either commit was wrong.

**Commit 1 (#215) — `.github/workflows/ci.yml`, EXECUTED.** `Build` has `id: build` at the step above `Test`, so `steps.build.outcome` actually resolves — the gating you describe is real gating, not a silently-empty expression. `Typecheck` and `Lint` carry `if: "!cancelled()"`; `Packaging guard` and `Boot smoke` carry `if: "!cancelled() && steps.build.outcome == 'success'"`. `pack-smoke` is a second job. Neither job's matrix has a `windows-latest` row, so "absent from the test/typecheck/lint/boot grid **and from pack-smoke alike**" is the correct, stronger statement. `desktop.yml` runs no suite — `grep -n 'npm test\|vitest'` over it is empty.

**Commit 1 — prd-24's residuals, EXECUTED.** `recordings/no-live-fleet-law.test.ts` does walk recursively (`walkSourceFiles`, a separate recursive `realSourceFileNames()` floor), and `recordings/recursion-fixture/walked-marker.ts` is tracked, so the recursive branch is exercised by the committed tree. The two you record as *still latent* are still flat (`interaction/no-model-call-law.test.ts:31`, `connect/index.test.tsx:783`) and both directories have zero subdirectories — `find … -mindepth 1 -type d` returns nothing for either. Recording them as latent-not-fixed is the right call and the right wording.

**Commit 2 (#222) — EXECUTED.** `README.md` reads "Node >= 22.22.2", names `engines` as the source of truth, points at CONTRIBUTING.md. Four support-matrix rows; macOS reads CI-verified with its legs named; Windows reads as an honest partial. `packages/server/bin/rhizomorph.mjs:52` uses `pathToFileURL(distEntry).href`. `connect/wizard.tsx:910-911` renders the detached-launch text you quote, verbatim, and no file under `packages/web/src` calls `POST /api/retarget` (the only hit is a comment in `app/boot-reason-seam.test.ts`). `route-class-law.test.ts:98-99` does pin both counts at 25 independently. `collector-loader.ts` registers by static import; `collector-wrap-boundary.test.ts` does derive its set by walking, and does have the "sixth collector spliced in raw" case — so excluding it from `coupling.txt`, with the reason written in, is right. #210, #211, #212, #216, #220, #221 are all open and all say what you say they say.

**Notably, you did *not* repeat the defect you were fixing.** The prd-25 amendment cites `README.md` by content, not by line — which matters, because main moved that file 18 lines down between your fork point and now. Cited by number, both of those would have rotted before the PR landed.

**Your mutations reproduce.** Independently re-run, both EXECUTED:

- Remove `if: "!cancelled()"` from `ci.yml`'s `Lint` step → `× the runbook names exactly the steps the workflow still runs after a red Test` / `AssertionError: expected [ Array(3) ] to deeply equal [ 'Typecheck', 'Lint', …(2) ]` / `Tests 1 failed | 10 passed (11)`. Restore → `11 passed (11)`.
- Reinstate `ci.yml:166` in `AGENTS.md` → `AssertionError: expected [ 'ci.yml:166' ] to deeply equal []`.

**Suite count.** You reported 401 files / 7128 passed / 3 expected fail / **6 skipped**. I measure 401 / **7132** / 3 / **2** on the same tree — same 7137 total; four tests that skipped in your environment ran in mine. Not a discrepancy in the work.

---

### One finding — the sibling case, and it was live, not hypothetical

**`EXECUTED`.** Commit 1 states the rule generally:

> **Cite CI by job and step name, never by line number.** … a registry entry pinned to a number for a file that moves is a registry entry that rots by construction.

…and then enforces it with a detector scoped to `` /\b[\w.-]+\.ya?ml:\d+/ ``. Scoping the guard to the extension that happened to rot first is the shape `AGENTS.md` warns about — and here the sibling was already wrong on the same tree. After your two CI citations were corrected, `AGENTS.md` still carried exactly two `path:line` citations, and **both were wrong**:

| in `AGENTS.md` | cited | actually at | what is at the cited line |
|---|---|---|---|
| line 83, `split_targets` | `scripts/dev/issues.sh:467` | **511** | inside `cmd_show`'s `gh issue view` pipeline |
| line 380, the load-batches comment | `scripts/gate.sh:91` | **232** | a fence-audit comment about git quoting non-ASCII filenames |

This is the failure `AGENTS.md`'s own orphans note describes: *a wrong pointer to a real command is worse than no pointer, since it reports success*. And the law shipped green over both — `node -e` over the pre-widening pattern returns `null` on `` `scripts/gate.sh:91` ``.

**Fixed in `b9ac433`**, since the fix is smaller than the argument for it:

- both now cite by name (`split_targets` in the script; the `Load gate:` comment above the batches), same treatment the CI citations got;
- the detector is widened to any `path.ext:NN`, and the regex is **hoisted into one `lineCitationsIn()` used by both the assertion and the rigged-input test**. It was duplicated before — two copies of a regex is how a rigged test ends up proving a regex the assertion does not use;
- the rigged test gains the two `.sh` shapes and a negative case (bare filename, a `8/8 at 4x` ratio) so it convicts prose in neither direction.

**Mutation on the widened law, EXECUTED.** Restore `` (`scripts/gate.sh:91`) `` to `AGENTS.md`:

```
× the runbook cites CI by job and step name, never by a line number that rots
AssertionError: expected [ 'scripts/gate.sh:91' ] to deeply equal []
Tests  1 failed | 10 passed (11)
```

Restore the named form → `11 passed (11)`. The broad scan `grep -on '[A-Za-z0-9_./-]\+\.[a-z]\+:[0-9]\+' AGENTS.md` now returns nothing, so the widening convicts no legitimate citation left in the file.

---

### One small correction — `EXECUTED`

prd-30's grooming note reads "97 native tooltips across **15** directories at `26c48c7`". The 97 re-derives exactly; the 15 does not — the set that yields 97 spans **14** top-level directories under `packages/web/src`:

```
29 panels  16 drawer  9 recordings  9 lane-page  8 replay  7 app
 5 why      4 tide    3 scene       2 lab        2 disclosure
 1 trace    1 interaction           1 concierge          = 97, 14 dirs
```

Attribute-for-attribute that matches #220's own breakdown (29/16/9/9/8/7), so both numbers came from one measurement and only the tally slipped. 15 is reachable only by adding `packages/web/src/lib/`, which contributes a `.ts` file the 97 doesn't count. Corrected to 14 in `50c59bd`. #220's body carries the same 15 and isn't editable from this fence.

---

### Gates I ran

On the branch with `origin/main` merged in:

- `npm run typecheck` — exit 0
- `npm run lint` — exit 0, 873 files
- `VITEST_MAX_WORKERS=6 npm test` — exit 0, **402 files, 7180 passed, 3 expected fail, 2 skipped**

On the branch tip as it now stands (pre-merge), after my two commits: typecheck exit 0, lint exit 0, suite exit 0 — 401 files, 7132 passed, 3 expected fail, 2 skipped.

`gh pr checks 224` on `aeb8611`: all 7 jobs pass — 3 × `build-test-boot` + 4 × `pack-smoke`. Both jobs, not one.

**Not run:** `scripts/gate.sh` (operator's step, and it merges), and nothing on Windows or macOS beyond what CI already reported.

---

### Disposition

The two authored commits are correct as they stand and I'd approve them unchanged. My `b9ac433` widens a law's mechanism, which is substantive rather than mechanical, so per the runbook it wants a second pair of eyes rather than my own approval: **@ciaran-slow, confirm the widened detector reads right to you** and this is ready for the operator to land. `50c59bd` is a one-digit correction.

Force-push avoided — both commits fast-forward onto `prd43-docs-truth` (`aeb8611..50c59bd`).

ciaran-slow:
Author's answers, both findings — thank you for fixing rather than annotating.

**Finding 1 — detector scoped to `.ya?ml`: accepted, fixed in `b9ac433`, verified EXECUTED here.** You're right that it's the exact sibling-scoping failure the commit was lecturing about, and worse that it was live: I fixed the two CI citations and shipped a law green over `scripts/dev/issues.sh:467` (really 511) and `scripts/gate.sh:91` (really 231–232) in the same file. Both re-verified against `origin/main` before I pulled your commit.

Re-run on `50c59bd` locally: clean → `11 passed`. Reinstated `` (`scripts/gate.sh:91`) `` after `Load gate:` in `AGENTS.md` → `AssertionError: expected [ 'scripts/gate.sh:91' ] to deeply equal []`, `1 failed | 10 passed`. Restored → `11 passed`.

The widened detector reads right to me, and the hoisting is the more important half — the duplicated regex meant my rigged test was proving a pattern the assertion didn't own, which is precisely the "test that cannot fail for the reason it claims" shape. One edge worth writing down, not blocking: `\.[a-z]+:\d+` will also match a host-with-port like `example.com:8080` if one ever lands in the runbook. Nothing in `AGENTS.md` does today (the law passes on the real file), and a false conviction names the string in its failure message, so the cost is a visible, one-line explanation rather than a silent miss. The asymmetry is the right way round.

**Finding 2 — 15 → 14: accepted, fixed in `50c59bd`.** The tally slipped when I added `lib/` by eye; your derivation is the one the sentence should have carried. **#220's body carried the same 15 and was outside your fence — corrected on the tracker now**, to "across 14 top-level directories", matching the PRD.

**Suite count** — 6 skipped here vs 2 there is environment (four tests gated on something my machine lacks); same 7137 total. Noted, not a discrepancy.

On the second-pair-of-eyes question: `b9ac433` widens a law's mechanism, so per the runbook it's the operator's confirmation that lands it, not mine. My reading is above; @ciaran-slow's is the one that counts.

All 7 checks pass on `50c59bd`; `MERGEABLE / CLEAN`.

PR #227, opened 2026-09-02
docs(prd): prd-33 is shipped, and its glance transfers to prd-30 by ruling 14

ciaran-slow:
Closes nothing — this retires a PRD and files two issues (#225, #226). One commit.

## What it does

prd-33's implementation shipped weeks ago; wave 0 — ruling 5's first-glance act with a real lay viewer — held it open, on a documented deferral. The run happened today and is in `docs/design/glance-2026-09-02.md`: the **instrument half is complete and EXECUTED** (both fixtures at 1440×900, MODE on the real server replaying a real 1425-minute session), and the **lay-viewer half is transferred, not skipped**, to prd-30's first-glance acceptance by **ruling 14**. prd-30 owes that act anyway, on the same screen; the audit's finding 16 named one unrun act blocking two PRDs as the defect.

What is given up is written down: ruling 5's power to **CUT** a scene mark does not transfer.

The PRD moves to `done/`; all four inbound paths follow it; roadmap gets the shipped-form entry.

## What is the sibling case?

**Ruling numbering.** I drafted this as ruling 13. prd-33 already *has* a ruling 13 — the supported-size table `roadmap.md` cites. Caught by grepping the file's ruling numbers before committing, renumbered to 14 in every place I wrote it (amendment, prd-30, results doc, roadmap, #221's body), and nothing pre-existing was touched. The roadmap keeps citing ruling 13 for the size table.

**Inbound references.** Both previous closeouts each missed one (the review fix on #209 was literally *"the move carries the cited line"*). All four were found by grep before the move — `roadmap.md`, `charter.md`, `ui-2.0-progress.md` (backticked, law-checked) and the retained review (relative link) — and a final grep for the pre-`done/` path over `docs/`, `README.md`, `AGENTS.md` returns nothing.

**The protocol's own sections.** `demo.md`'s Setup says `Esc` releases panel focus; the shell says Escape is deliberately unhandled; the camera section names the resulting `1`-key collision as a failure. Executed, and filed as #225 rather than fixed here — out of this change's scope.

## What mutation would this test survive?

None applies — this is documentation and a PRD disposition; no law is added. `doc-citation-law` and `no-personal-paths-law` run over the new document and the moved PRD and are green.

## Fence widening

None.

## Gates

By exit code: `typecheck` 0 · `lint` 0 (873 files) · `VITEST_MAX_WORKERS=6 npm test` 0 — **402 files, 7176 passed, 3 expected fail, 6 skipped**.

Issue #203, opened 2026-09-01
prd43 w1: the exclusion scan survives a tracked-but-deleted file


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/doc-citation-law.test.ts`

## Blocked by
**PR #200 (prd43 w2) landing.** It holds this file as a live fence, so the line
citation below is against its head and must be re-derived.

**Model:** sonnet. **Wave:** 1 — the SAME wave as #16 and #186, deliberately.
This is the law's own file, and the sibling half of a repair #186 already made.

## The defect

`#186` item 5 added `readSweptFile`, which returns `undefined` for a path that
`git ls-files` reports but that is not on disk — a tracked file deleted and not
yet staged. Before that guard the law crashed with `ENOENT` out of `readFileSync`
instead of failing like a law.

`allCitations()` goes through that guard. **The exclusion-honesty scan does
not.** EXECUTED, on PR #200's head:

    :140  return existsSync(filePath) ? readFileSync(filePath, 'utf8') : undefined   <- guarded
    :587  const text = stripFencedCodeBlocks(readFileSync(path.join(REPO_ROOT, file), 'utf8'))

So the same condition `#186` fixed for the main sweep still throws out of the
exclusion scan — and it throws rather than failing, which means the run dies
carrying a stack trace instead of a law's verdict.

## Why it is filed rather than fixed in #186 or PR #200

`#186` fixed the instance it was asked about and left its sibling — the shape
this file has now recorded several times, and worth naming plainly rather than
quietly closing. PR #200 is **deletion-only** in this file (allowlist entries
out, nothing in), so fixing it there would have been unreviewed scope in a
commit reviewed for something else.

## Done when

- Every `readFileSync` in this file that reads a path derived from `git ls-files`
  goes through the same guard, or declares in the input table why it does not.
  There are more than two call sites — enumerate them rather than fixing the one
  named here.
- A test proves the exclusion scan SKIPS a tracked-but-deleted file rather than
  throwing, and reddens if the guard is removed. `#186`'s own content and
  tracked-file pins are the shape to copy.
- The law is run 4x concurrently; it has raced itself twice.

## Provenance

PR #200's verification (Opus `--effort xhigh`), surfaced incidentally while
mutating the exclusion scan, and recorded there as out of fence.

Issue #211, opened 2026-09-02
prd25 w2: a windows-latest leg packs, installs and boots the CLI on every push


ciaran-slow:
## Fence (may touch ONLY)
- `.github/workflows/ci.yml`
- `scripts/pack-smoke.sh`
- `README.md`

## Blocked by
**#23** (owns `.github/workflows/ci.yml` and `README.md`), **#21** and **#24** (own `README.md`). All three are live prd43 lanes. This issue cannot be dispatched until they land — it is groomed and boarded now so the block is explicit rather than discovered at rebase.

**Model:** sonnet. **Wave:** 2.

## Why

prd-25 ruling 1. There is still no `windows-latest` row on either CI job — `build-test-boot` is `[ubuntu-latest, macos-latest]` (`.github/workflows/ci.yml:32`) and `pack-smoke` is the same (`.github/workflows/ci.yml:201`). The desktop installer workflow's Windows row packages the Electron shell; it does not run the native suite or boot the CLI, so it is not this gate.

The leg goes on **`pack-smoke`, never `build-test-boot`**.

prd-25 ruling 1 gives a mechanical reason for that — "a failed `Test` step skips every later step on its leg, so the boot smoke never runs" — and **that reason is now stale**. Verified on `main`: `.github/workflows/ci.yml` gates Typecheck and Lint on `if: "!cancelled()"` (lines 67, 71) and the packaging guard and boot smoke on `if: "!cancelled() && steps.build.outcome == 'success'"` (lines 81, 105), with a comment explaining why `always()` would be worse. All four post-`Test` steps run on a red-`Test` leg today, provided Build was green.

The ruling's conclusion survives its stale premise, on the argument that actually holds: `build-test-boot` on Windows would be **permanently red**, because the suite has known native failures and no committed list yet says which are expected. A leg that is always red teaches everyone to ignore it — the reading this repo legislates against — and giving the suite an honest expected-fail list is wave 3's job, not this one's. `pack-smoke` meanwhile builds, packs, installs the tarball into a project that has never heard of this checkout, and runs the CLI from the installed files. That is exactly the built-clone path that died on win32, and it has no known Windows failures to launder.

The built-clone boot defect itself is **already fixed** — `packages/server/bin/rhizomorph.mjs:52` uses `pathToFileURL(distEntry).href`, so win32 no longer parses the drive letter as a URL scheme. This leg is what witnesses that fix for the first time; no existing leg can.

[...]

PR #229, opened 2026-09-02
test(server): a citing document that declares the tree it is pinned to is a dated artefact, wherever it lives

### LINKED ISSUE #228: a citing document that declares the tree it is pinned to is a dated artefact, wherever it lives
## Fence (may touch ONLY)

## Blocked by
Nothing live. **#203** (prd43 w1, Backlog, not dispatched) claims this same file for a different region — the `readFileSync` guards in the exclusion scan. This issue lands first; #203's own Blocked-by already says its line citations must be re-derived against the current head, and it should rebase over this. Whatever this issue adds reads files through `readSweptFile`, the guard #203 exists to make universal, so it does not widen #203's defect.

**Model:** sonnet. **Wave:** none — unwaved, like #62/#63/#66.

## The defect

`doc-citation-law` excludes dated artefacts as **citing sources** — `docs/research/`, `docs/review/`, `docs/prds/` — on the stated reasoning that *"a dated artefact's own citations are a record of the tree at a commit, not a live claim."* The exclusion is **by directory**.

A dated run record under `docs/design/` gets no such treatment. The first instance is the prd-33 wave-0 glance record landing in PR #227, `docs/design/glance-2026-09-02.md`: pinned to `main` at `0851512` in its second line, eleven code citations, nine of them in the bare form (`app/PanelGrid.tsx:240`) the law cannot see and two rooted. #227's review executed both directions: a dead rooted citation appended to that document goes red; a dead bare one stays green. So the record is *swept* as if it were a live claim, but *protected* only where it happens to use the rooted form — the worst of both: enrolled in a live existence check it should be exempt from, and exempt from it in exactly the places a reader would follow.

Two wrong fixes are available and both are the shape AGENTS.md names:

- **Exclude `docs/design/glance-*.md` by name.** A guard scoped by naming convention misses the files that predate it — the runbook records the fixture-hygiene law doing exactly this to three older captures. The next run record will be called something else.
- **Root the nine citations.** That enrols a historical artefact in a live existence check, which is the very reason the three directories are excluded. Every scene refactor would then redden a record of what the scene looked like on 2026-09-02.

## What becomes true

**A citing file that declares the tree it is pinned to is a dated artefact, wherever it lives** — excluded as a citing source by what it *is*, not by its directory or its name.

- The marker is the form the first instance already uses, near the top of the file: `**Tree:** \`<ref>\` at \`<sha>\``, with a 7–40 hex `<sha>`.
- **The pin must resolve to a real commit** (`git cat-file -e <sha>^{commit}`). A pin that does not resolve is not an exemption; it is a defect the law reports by name — otherwise any live document could write a fake pin and leave the sweep.
- Read through `readSweptFile`, never raw `readFileSync` — #203's rule.
- The header doc comment's "Scope" section states the new criterion beside the three directories, and the input table gains the marker form.

## Definition of done

- `isExcludedCitingFile` (or its replacement) excludes a file with a valid pin in its head, and does not exclude a file with no pin, a malformed pin, or a pin whose sha does not resolve.
- The pin exclusion is shown to do real work, the way the directory exclusion already is (`:573`, "the exclusion is doing real work, not vacuous"): a rigged untracked document under `docs/design/` carrying a valid pin **and** a dead rooted citation is *not* reported; strip the pin and it *is*; replace the sha with `deadbeef0` and it is reported **as a bad pin**, distinctly. Use the file's existing fixture discipline (`cleanUpOrphanedFixtures`, dead-PID names, tracked-file guard).
- `EXCLUDED_DIRS` and its pinned test are untouched — this adds a criterion, it does not loosen one.
- The law is run 4x concurrently, as #203 asks of the same file; it has raced itself before.

## Sibling case

The three excluded directories are excluded *only* as citing sources — a live document citing **into** them still makes a live claim, and the header says so at length. The pin exclusion must keep that asymmetry: a pinned record is exempt from having *its* citations checked, but a live document citing *the pinned record's path* is still checked. A pinned record that moves is still a broken link from wherever cites it.

Change the marker regex so it matches any line containing `Tree:` — a document reading "the *Tree* view: `foo.ts`…" would silently leave the sweep. The rigged-fixture tests must include a prose false-positive that does **not** exempt. And: delete the `cat-file` check — the fake-sha fixture must go from "reported as bad pin" to "silently exempt", and a test must see that.

ciaran-slow:
Closes #228

One commit. prd43, unwaved — the same shape as #215/#222.

## What it does

`doc-citation-law` excluded dated artefacts as citing sources **by directory** (`docs/research/`, `docs/review/`, `docs/prds/`). A dated run record anywhere else — the first is #227's `docs/design/glance-2026-09-02.md`, pinned to `main` at `0851512` — was swept as a live claim, and #227's review showed the consequence with two executed mutations: a dead *rooted* citation appended to it goes red; a dead *bare* one stays green. Enrolled in a check it should be exempt from, exempt in exactly the places a reader follows.

Now a citing document is a dated artefact **by what it is**: its head declares `**Tree:** \`<ref>\` at \`<sha>\`` and the sha **resolves** (`git cat-file -e`). A pin that does not resolve exempts nothing and is reported by name (`badPins()`). Documents only — a `.ts` comment declaring a pin is still a live claim.

The two fixes I did **not** do, and why: excluding `docs/design/glance-*.md` by filename is the guard-scoped-by-naming-convention failure AGENTS.md records twice; rooting the record's nine bare citations enrols a historical artefact in a live existence check, which is the whole reason the three directories are excluded.

## What is the sibling case?

**The marker's false positives.** Prose like *"the *Tree* view: `foo.ts` at `0851512`"*, a pin with no `at`, a pin to `not-a-sha`, a pin past the twelve-line head, and a pin inside a fenced block — each is a rigged-input case and each must **not** exempt. The regex is anchored to a line start and the literal `at`.

**The fake pin.** The exemption is the risk: any live document could write `**Tree:** \`main\` at \`deadbeef0\`` and leave the sweep. So the sha must resolve, and one that does not is a reported defect, not silence — pinned to `[]` against the real tree the way the allowlist is.

**#203** claims this same file (Backlog, undispatched) for the `readFileSync` guards in the exclusion-honesty scan. Nothing here touches that region; every new read goes through `readSweptFile`, which is the guard #203 exists to make universal. #203's own Blocked-by already says its line citations must be re-derived.

**The fixture on disk.** It cites a REAL path — same reasoning as the existing untracked-doc fixture: a dead citation on shared disk reddens a sibling run, a real one cannot. The exclusion is proved by the fixture's *absence* from `allCitations()`, after first asserting the same bytes minus the pin *do* extract that citation. Cleanup recognises the pinned fixture by **shape** (any 40-hex pin over that one line), because an orphan from an earlier `HEAD` carries a different sha and equality would leave it behind forever.

Three, all **EXECUTED**, each restored to `29 passed`:

| mutation | red test | assertion |
|---|---|---|
| regex loosened to any `Tree:` line | *the pin marker is exactly one form* | `expected '0851512' to be undefined` |
| `pinResolves` forced `true` | *a pin that does not resolve exempts nothing* | `expected true to be false` |
| docs-loop skip removed | *a document that declares a resolving tree pin is … excluded* | `expected [ { …(2) } ] to deeply equal []` |

A law that only checked for the marker string would survive the second; one that only tested pure functions would survive the third.

## Concurrency

Run 4× concurrently, as #203 asks of this file: 4/4 exit 0, 29 passed each.

## Fence widening

None. Fence is the one file.

## Gates

By exit code: `typecheck` 0 · `lint` 0 (873 files) · `VITEST_MAX_WORKERS=6 npm test` 0 — **402 files, 7179 passed, 3 expected fail, 6 skipped**.

## Note for #227

When #227 lands, its glance record is the first real document this exclusion applies to — no edit to that record is needed; it already carries the marker in exactly this form.


gabriel-canaan:
Fence widening, recorded before the change (AGENTS.md — on the PR, not after).

**Widening:** `.github/workflows/ci.yml` — one line, `fetch-depth: 0` on the `build-test-boot` checkout, plus the comment saying why. `pack-smoke` runs no suite and stays shallow.

**Why it cannot stay inside the declared one-file fence:** the pin arm asks the *local object store* (`git cat-file -e <sha>^{commit}`). `actions/checkout@v4` with no `fetch-depth` is a depth-1 clone, so no sha but the tip is present and every honest pin to an ancestor reads as a fake one. There is no in-fence way to make an absent object resolve; the alternatives are to stop requiring resolution (which is the fake-pin hole this law closes) or to exempt on an unverified pin (same hole). Evidence and the executed repro are in the review that follows this comment.

The in-fence half — a precondition assertion so a future edit back to the default depth fails with a sentence rather than a bad-pin report — goes in `packages/server/src/doc-citation-law.test.ts` as declared.

gabriel-canaan (commented):
Reviewed commit-by-commit with `origin/main` merged into the branch, which turned out to be the whole review: **#227 merged 12 minutes after this PR's CI ran**, so every one of the seven green checks was computed against a merge-ref that did not yet contain `docs/design/glance-2026-09-02.md` — the one document this PR exists for.

## What verified in the author's favour — EXECUTED

**The exclusion does exactly what #228 asked, on the real record.** With main merged in, I appended a dead rooted citation to the glance record — the same mutation `835df7c` used to prove the law *did* sweep it:

```
appended  `packages/web/src/scene/does-not-exist.ts`  to docs/design/glance-2026-09-02.md
  on main   (835df7c) -> 1 failed  | 25 passed   "glance-2026-09-02.md -> ...does-not-exist.ts"
  on this branch      -> 29 passed (29)          green; the record is out of the sweep
```

The record is now exempt by what it *is*. No edit to it was needed — the marker matches its line 4 (`**Tree:** \`main\` at \`0851512\`. **Disposition:** …`) at the second form the rigged-input test pins, and that test's `# Title / **Protocol:** / **Tree:** … . **Disposition:**` case is that document's real shape, not a convenience.

Everything else I re-derived held:

- **Gates, on the merged tree, by exit code:** `typecheck` 0 · `lint` 0 (873 files) · `VITEST_MAX_WORKERS=6 npm test` 0. Counts came out **402 files, 7183 passed, 3 expected fail, 2 skipped** against the body's 7179 / 6 skipped — the body measured pre-#227 and the drift is main's, not the author's.
- **4x concurrently on the merged tree:** 4/4 exit 0, 29 passed each. Confirms the `#203` condition.
- **The three declared mutations** are each real: I re-ran the `pinResolves`-forced-`true` one and got the named assertion.
- **`docs/*.md` reaches nested directories** — `git ls-files -- 'docs/*.md'` returns 192 paths containing `/`, so `docs/design/` is genuinely in scope and the exclusion is not standing in for invisibility. The test asserts that directly (`the sweep must SEE the file`), which is the right assertion.
- **`badPins()` and the docs arm of `allCitations()` sweep the identical set** — same glob, same `isExcludedCitingFile` skip, same `readSweptFile`, same fence strip, pin checked last. No asymmetry to exploit.
- The two rejected fixes named in the body are the two AGENTS.md records, and rejecting both is right.

## The finding — CONFIRMED, EXECUTED

> **The pin arm asks the local object store, and CI's clone is depth 1.** A fixture that pins `HEAD` resolves at any depth. A real dated artefact pins an **ancestor**, and that is the only form that exists in the tree.

`pinResolves` runs `git cat-file -e <sha>^{commit}`. `.github/workflows/ci.yml`'s checkout is `actions/checkout@v4` with no `fetch-depth`, i.e. **depth 1** — no object but the tip. So on CI the glance record's honest `0851512` does not resolve, and both halves invert:

1. `expected badPins() to equal []` goes **red**, naming a document whose pin is perfectly honest.
2. The record loses its exemption and is swept as a live claim again — the defect #228 fixes, alive on CI only, and *silently* green today purely because its one rooted citation happens to exist.

Proof, on this branch with `origin/main` merged in, in a real `--depth 1` clone of that same tree:

```
$ git rev-parse --is-shallow-repository
true
$ git cat-file -e 0851512^{commit}
fatal: Not a valid object name 0851512^{commit}          (exit 128)

$ npx vitest run packages/server/src/doc-citation-law.test.ts
  x a pin that does not resolve exempts nothing and is reported by name (#228)
    AssertionError: expected [ { …(2) } ] to deeply equal []
    + [ { "file": "docs/design/glance-2026-09-02.md", "sha": "0851512" } ]
  1 failed | 28 passed (29)

$ git fetch --unshallow          # same worktree, nothing else changed
$ npx vitest run packages/server/src/doc-citation-law.test.ts
  29 passed (29)
```

This is the sibling case in the shape AGENTS.md names, one level down: the fixture pins the one sha immune to the failure, so the check verified itself and not the case it was written for. It is not arguable and it is not a flake — it is deterministic on the default checkout.

## Fixed — `0772802`

Fence widening was recorded on the PR before the change (comment above), because no in-fence edit can make an absent object resolve; the alternatives both re-open the fake-pin hole the law closes.

- `.github/workflows/ci.yml` — `fetch-depth: 0` on the `build-test-boot` checkout only. `pack-smoke` runs no suite and stays shallow, so the 10x macOS legs pay nothing.
- `packages/server/src/doc-citation-law.test.ts` — the precondition, in-fence: `repoIsShallow()` and a test asserting it is false, carrying the cause in its own assertion message, plus `pinResolves(HEAD^)` so the ancestor case the depth-1 checkout loses is exercised rather than assumed. Verified by running the file in a fresh depth-1 clone: it now fails with

  ```
  x the pin arm has the history it needs (#228, review of #229)
    this clone is SHALLOW, so `git cat-file -e <sha>^{commit}` cannot see any sha
    but the tip and every tree pin to an ancestor reads as a fake one — run
    `git fetch --unshallow`, or restore `fetch-depth: 0` on the suite leg …
  ```

  rather than as a bad-pin report against an honest document.

Gates after the fix, on the branch with main merged in: `typecheck` 0 · `lint` 0 (873 files) · full suite 0 — **402 files, 7184 passed, 3 expected fail, 2 skipped**; the law 30 passed, and 4/4 exit 0 at 4x.

**This is a substantive fix, so it is not mine to approve** (and I wrote it, which GitHub would refuse anyway). Confirm `0772802` and I will land it; if you would rather the CI half went to `#211`'s milestone as its own change, say so and I will split it — the in-fence precondition test then becomes a red gate until it lands, which is the honest ordering but a worse one for you.

1. **The asymmetry the issue names is preserved by construction but untested.** "A pinned record is exempt from having *its* citations checked, but a live document citing *the pinned record's path* is still checked" — true, because `isPinnedArtefact` is consulted only in the citing loop and `citationExists` never asks. `EXCLUDED_DIRS` has an allowlist entry standing as its live witness (`docs/adr/0029…` → `docs/prds/…`); the pin arm has none. One line in the existing fixture test — a second rigged doc citing the pinned fixture's path and appearing in `allCitations()` — would pin it. **REASONED**, not executed.
2. **`pinResolves(HEAD_SHA.slice(0, 7))` asserts a 7-char abbreviation is unambiguous.** Git's default abbreviation length grows with object count, so this is a real if distant flake. `HEAD_SHA.slice(0, 12)` costs nothing and still exercises the short-sha path the marker allows (`{7,40}`). **REASONED** — it resolves today, at 7.
3. **`badPins()` skips `isExcludedCitingFile` first**, so a fake pin inside `docs/research|review|prds/` is never reported. Harmless — those are already excluded as citing sources — but it means "a fake one landing anywhere under `docs/`" in the test's comment is slightly wider than what the code does. A three-word comment fix, if you want it exact.
4. **Context, not a finding:** after this lands, the law-coverage half of `835df7c`'s rationale ("the repo-rooted form is the load-bearing part … this one citation is now inside it") no longer applies — the record is out of the sweep entirely. The path correction it made stands on its own, and #228 decided this deliberately. Worth knowing when reading that commit later.

## What I did not run

`windows-latest` — there is still no Windows leg in `ci.yml` on either job (prd-25 / `#211`), so nothing here or in the fix has been witnessed on Windows. The fix is a checkout parameter and a `git rev-parse`, both platform-neutral, but that is REASONED.

gabriel-canaan:
CI on `0772802`, 7/7 pass — and this run carries the evidence the previous one could not.

`gh pr checks 229`: `build-test-boot` ubuntu-current 8m43s · ubuntu-min 9m31s · macos-current 3m18s; `pack-smoke` all four green. Run `33602771120`.

Two things that makes load-bearing, both of which the 06:47 run (`33600468285`) missed by twelve minutes:

- **The merge-ref now contains `docs/design/glance-2026-09-02.md`.** #227 merged at 07:00:16Z; the earlier run started 06:47:46Z. So this is the first CI run in which the law has ever met the one real pinned record, on all three suite legs.
- **A green suite leg now *proves* the checkout is not shallow.** The precondition test asserts `repoIsShallow()` is false, so `fetch-depth: 0` cannot silently fail to apply — a depth-1 checkout fails that test by name, with the reason in the assertion message. No log-reading required; the leg passing is the proof.

Nothing outstanding from my side. `0772802` is a substantive fix and it is mine, so the approval and the landing are not — confirm it and it goes in. The three items in the "your call" section of the review are all optional and none of them blocks.

PR #239, opened 2026-09-02
prd25 w1: a CRLF checkout cannot change what the era corpus says

### LINKED ISSUE #210: prd25 w1: a CRLF checkout cannot change what the era corpus says
## Fence (may touch ONLY)
- `.gitattributes`
- `packages/core/src/eras/`

## Blocked by
Nothing. `.gitattributes` does not exist in the tree today — `git check-attr -a packages/core/src/eras/era-1/recording.jsonl` returns no attributes at all — and no open issue claims `packages/core/src/eras/`.

Sibling lanes: prd43 #21, #23 and #24 own `README.md`, and #23 also owns `.github/workflows/ci.yml`. This issue touches neither, and must not.

## Why

prd-25 ruling 4, the line-endings half. Under `core.autocrlf=true` — the Git-for-Windows default — the era corpus checks out with CRLF, and the corpus is bound into the suite as text through Vite's `?raw` (`packages/core/src/eras/corpus.ts:1-2`). That is the whole danger: the corpus arrives as **input**, not as something the suite can re-bless, so a CRLF checkout makes the input lie and there is no in-suite repair. The era golden byte-identity test was one of the 15 files that failed the only native Windows run this project has ever done, and a CRLF checkout is the named suspect.

There is no `.gitattributes` anywhere in the repo, so nothing pins this today on any platform.

## Definition of done

- `.gitattributes` exists at the repo root and pins `eol=lf` for the era corpus and the JSONL/JSON fixtures the suite reads as text.
- A law test under `packages/core/src/eras/` fails if the pin is removed, and fails if a new corpus file is added under `eras/` that the pin does not cover. Scope the check by what a file **is** (a corpus file read through `?raw`), not by what it is named — AGENTS.md records a hygiene guard that missed three older captures because it keyed on a filename prefix.
- The test states which attribute it read and how, so a reader can re-run the check by hand.
- Existing corpus bytes are unchanged by this issue: renormalisation, if any is needed, is shown in the commit as a separate, explained step, and `packages/core/src/eras/eras.test.ts` stays green.

## Sibling case

The corpus is two files (`recording.jsonl` and `session-state.snapshot.json`) and both arrive through the same `?raw` door. A pin that covers `*.jsonl` and forgets `*.json`, or covers `era-1/` by name and forgets the next era directory, is the sibling miss this issue exists to avoid.

ciaran-slow:
Closes #210

prd-25 wave 1. One issue, one claim: **a CRLF checkout cannot change what the era corpus says.** Base is `main`; the lane branch carries the issue's single commit `c56878e` plus a merge-in of `origin/main` (`8f709b9`, 0 conflicts) — merged rather than rebased, so the branch is never force-pushed.

## What becomes true

The era corpus reaches `@rhizomorph/core` as `?raw` text and `eras.test.ts` compares the fold to the committed snapshot byte for byte, deliberately unable to re-bless it from inside the suite. Under `core.autocrlf=true` — the Git-for-Windows default — a checkout rewrote LF to CRLF, the *input* lied, and the golden law failed on a tree nobody changed. That is what the only native Windows run this project has made actually hit.

`.gitattributes` now pins `text eol=lf` for `*.jsonl`, `*.json` and every collector fixture. `packages/server/src/corpus-eol-law.test.ts` holds the pin: three sets derived from the tree (never a maintained list) are each asked of `git check-attr` for the attribute git will *actually apply*, and the law fails if any answer is not `eol: lf`. It also proves the pin renormalised nothing — every pinned file re-hashes under the pin to the blob already in the index.

## What is the sibling case?

Two, and one of them changed the diff.

- **The one the issue named:** a pin covering `*.jsonl` and forgetting `*.json`, when both corpus files arrive through the same `?raw` door. Answered by deriving the corpus set from `corpus.ts`'s own imports rather than listing extensions.
- **The one found during review, and the reason the law was rewritten:** the "changed no committed byte" assertion read `git ls-files --eol` and accepted only `i/lf` / `i/none`. Four `parse-log` fixtures were `i/-text` — git's **binary** heuristic on the `\x01` / `\x1f` separators the collector deliberately emits — and the law reported them as CRLF with a failure message that was simply false. The eol column is a proxy; the sibling miss was trusting the proxy over the property. The assertion now compares `hash-object --stdin-paths` under the pin to `ls-files -s`, which is the claim itself and cannot be fooled by git's classification of the file.

Stated per test, because the answer differs:

- **Remove the pin** → the check-attr test reddens for every derived file (`eol: unspecified`).
- **Widen the pin to a file with CRLF in the index** → the blob-comparison test reddens, naming the file and saying the pin would rewrite it.
- **Make `--path` silently ignored** (a broken probe) → the rigged detector test reddens: `a\r\nb\r\n` under a pinned path must hash differently from the same bytes `--no-filters`.
- **Empty derivation** (an extension nobody imports) → the "three sets non-empty" and "names both era-1 files" tests redden.

The one this suite **cannot** see: the pin doing its job on an actual CRLF checkout. No `windows-latest` leg exists on either CI job; that is #211, wave 2, and the reason wave 1 stops at the pin and the law rather than claiming Windows works.

## Fence

Fence as groomed: `.gitattributes`, `packages/core/src/eras/`. **Two widenings, [recorded on #210](https://github.com/launchpad-26/rhizomorph/issues/210#issuecomment-5509130175) before this PR existed, though after the lane's first edit — the comment says so:**

1. The law lives at `packages/server/src/corpus-eol-law.test.ts`, not under `eras/` as the DoD placed it. It asks `git`, so it needs `node:child_process`, and `@rhizomorph/core` has no Node types in scope (`eras/fold.ts` says so). It sits with the other repo-root laws for the same reason they do.
2. The pin's third pattern reaches `packages/server/src/collectors/*/fixtures/**`. Ruling 4 says *"at minimum for the era corpus and the JSONL/JSON fixtures"*; the `.txt` fixtures are read as utf8 text by `parse-log.test.ts` and are rewritten by autocrlf exactly like the corpus. No file there is edited; each re-hashes under the pin to its index blob.

## Verification

All on the merged HEAD `8f709b9`, node 22.23.2, exit statuses read: the law 8/8 · `eras.test.ts` 15/15 · typecheck 0 · lint 0 (875 files) · **build 0** · `VITEST_MAX_WORKERS=6 npm test` exit 0, 404 files, 7275 passed (main's 7276 + this law's 8 = 7284, reconciled). Executed outside the suite: a CRLF'd copy of a real corpus file hashes under the pin to the index blob `707fdb6c…`; unfiltered it hashes to `870342c5…`.

**What has not been done, so the reviewer does not assume it:** there is **no verify report from a second model** on #210. The false positive above was found and fixed in-session by the same operator's agent that opened this PR. The rewritten assertion is the part that deserves a reviewer's own run — the build report on #210 gives the exact commands. Boot smoke skipped: no server or CLI runtime path changes; the packaging guard excludes test files from the tarball.


gabriel-canaan (approved):
## Approved — reviewed at `8f709b9`, fix pushed as `0ca410c`

Reviewed commit-by-commit in a detached worktree at the PR head. The branch already carried current `origin/main` (`91b974c`), so there was nothing to merge in — the tree I exercised is the tree that lands.

Every gate the PR body claims, re-run here, exit statuses read (node 22.22.2, the declared engine minimum):

| | claimed | measured |
|---|---|---|
| the law | 8/8 | **8/8** |
| `eras.test.ts` | 15/15 | **15/15** |
| `npm run typecheck` | 0 | **0** |
| `npm run lint` | 0, 875 files | **0, 875 files** |
| `npm run build` | 0 | **0** |
| `VITEST_MAX_WORKERS=6 npm test` | 0 · 404 files · 7284 | **0 · 404 files · 7284** |

`gh pr checks` is green on all seven jobs, and I confirmed the run is on this head rather than an older one: run `33628519965`, `headSha=8f709b9`, and the ubuntu-current leg's step list shows Build / Test / Typecheck / Lint / Packaging guard / Boot smoke each `success`, none skipped. The absent eighth leg is `build-test-boot (macos-latest, min)`, which AGENTS.md excludes deliberately.

**Two things I verified that the PR does not claim, and both strengthen it.**

1. **The "changed no committed byte" property holds repo-wide, not just over the derived sets.** The law sweeps three sets; the `*.json` pattern is repo-wide. So I hashed *all* 79 tracked `.json`/`.jsonl` files plus all 76 collector fixtures under the pin and diffed against `git ls-files -s`: zero rewrites. EXECUTED. The wider-than-the-law pin is safe.

2. **The targeted patterns are load-bearing, and the obvious "improvement" would break the DoD.** The first thing a reader will reach for is the blanket idiom `* text=auto eol=lf`. That would renormalise `docs/research/2026-08-04-prd10-gorgeous-spike.md` — 239 CRLF lines in the index, blob `a7c74b7` → `d21a88b` — which is precisely the committed-byte change #210 forbids. EXECUTED. Worth knowing before someone "simplifies" this later; `0ca410c` records it in the file.

### Mutations — the law bites

All EXECUTED against `packages/server/src/corpus-eol-law.test.ts`:

| mutation | result |
|---|---|
| pin removed from index **and** worktree | **4/8 red** |
| drop the `collectors/*/fixtures/**` line | **2/8 red** |
| drop `*.json`, keep `*.jsonl` — *the sibling miss #210 names* | **1/8 red** |
| new unpinned `eras/era-2/recording.txt` — *the DoD's second requirement* | **2/8 red** |
| new fixture at a depth the pin's single `*` misses | **2/8 red** |

The last one is the pleasant surprise and I want it on the record, because it is not obvious and it is easy to break by "tidying": `git ls-files` pathspecs use wildmatch **without** `WM_PATHNAME`, so `*` crosses `/`, while gitattributes patterns use it and `*` does not. Set 3's derivation is therefore strictly **wider** than the pin it audits, and a fixture directory the pin cannot see still reddens the sweep. If anyone later "aligns" that glob to match the pin exactly, the set becomes self-fulfilling and this row goes green forever.

### One correction to the PR body — no code change needed

> **Remove the pin** → the check-attr test reddens for every derived file (`eol: unspecified`).

True only if the pin leaves the **index**. `git check-attr` falls back to the index blob when the working-tree `.gitattributes` is absent, so deleting only the file leaves **7/8 green**:

```
$ rm .gitattributes && git check-attr eol -- packages/core/src/eras/era-1/recording.jsonl
packages/core/src/eras/era-1/recording.jsonl: eol: lf
```

That is git behaving correctly — the index is what a checkout will actually apply — and the law is right to ask it that way. It just means the mutation has to be staged to mean anything, and the table above uses the staged form. Flagging it so the next person to re-derive this does not conclude the law is vacuous, as I briefly did.

### The fix I pushed — `0ca410c`, comments only

Both headers said the law asks git about *"every file the suite reads as text"*. It does not; it sweeps three derived sets, and two live counterexamples sit outside all three:

- `packages/server/src/lab/__fixtures__/*.ts.txt` — read `readFileSync(..., 'utf8')` by `namespace-law.test.ts:137,142`, `eol: unspecified`
- `packages/core/src/events/*.ts` and `packages/web/src/tide/*.ts` — read back as text through `import.meta.glob(… '?raw')`

I CRLF'd both sets and ran their own suites: `namespace-law` 21/21 green, `no-open-payload-law` + `purity` 25/25 green. EXECUTED. **So this is scope, not a live defect** — which is exactly why it wanted a comment fix rather than a change request. The reason it is worth one at all is AGENTS.md's own rule about `cmd_orphans`: a wrong pointer to a real check is worse than no pointer, because it reports success. Someone adding a text fixture beside those `.ts.txt` files would read the old sentence and reasonably believe the law had them.

No pattern changed. Re-verified after the edit: the law 8/8, and `doc-citation` + `no-personal-paths` + `runbook-delivery` alongside it 81/81; typecheck 0; lint 0 (875 files); full suite exit 0 (404 files, 7284); zero rewrites still.

The `.ts` sources read through `import.meta.glob('?raw')` are the one class where a CRLF checkout could plausibly bite a *future* grep law — today's two are line-ending agnostic and I proved it, but nothing holds them that way. That is #211's territory (the Windows leg is what would witness it), not this issue's, and prd-25 ruling 4's *"at minimum"* does not reach it. Mentioning it as a possible w2 sibling, not as a gap in this PR.

I did **not** run `scripts/gate.sh` — that is the operator's landing step, not a reviewer's. Landing is Gabe's.

Issue #241, opened 2026-09-02
prd43 w1: the citation law passes 20/20 under 4x concurrency


gabriel-canaan:
## What becomes true

`doc-citation-law.test.ts` passes 20/20 when run 4x concurrently. Today it is
flaky at roughly 1-2 in 20, and the failing test is the one written to prove the
fixture cleanup is safe.

## The defect

`cleanUpOrphanedFixtures never deletes a file its caller reports as TRACKED`
writes a REAL file into docs/ named for a deliberately dead pid, with
content that matches isFixtureContent, and left genuinely untracked:

    docs/citation-law-untracked-fixture-999997-<hrtime>.md

It then calls cleanUpOrphanedFixtures with an INJECTED isTracked that
reports that path tracked, and asserts the file survives.

The injection protects the call THIS test makes. It cannot protect the file from
a **sibling concurrent run**, whose cleanup uses the default git-backed
isTracked and correctly deletes it: the name matches, the pid is dead, the
content matches, and git reports it untracked. The assertion then fails with

    a file reported TRACKED was deleted by a cleanup step meant only for
    untracked orphans: expected false to be true

The test's own docblock already names this mechanism as one of three earlier
attempts that raced — *"a real, on-disk, dead-pid-named, content-matching file
sitting genuinely untracked for the instant between being written and being
staged, which a SIBLING run's cleanup pass is CORRECT to delete by this guard's
own rules"*. Injecting isTracked was the fix for that, and it closed the
staging race while leaving the file itself as bait.

## Evidence

EXECUTED, 20 runs of 4x-concurrent on each tree:

| tree | failing runs |
|---|---|
| `main` at `6b5f70f` | **1 of 20** |
| the #203 branch | **2 of 20** |

Same test, same assertion both times. **Pre-existing, not introduced by #203** —
that difference is noise at this rate. Surfaced by #203's own third
Done-when bullet, *"The law is run 4x concurrently; it has raced itself twice"*,
which is the bullet working.

## The obvious fix is a trap

Naming the fixture for the CURRENT pid closes the race — a sibling hits
`if (alive) continue` and skips it. It also makes the test **vacuous**: the pid
check runs BEFORE the isTracked check, so the file would survive because its pid
is alive, never because isTracked reported tracked. The test would pass without
exercising the guard it exists for. Do not take that route.

Same for making the content not match isFixtureContent: siblings skip at the
content check, and so does the call under test.

## The fix that works

[...]

Issue #212, opened 2026-09-02
prd25 w3: the Windows suite result is a committed list of files, not a number


ciaran-slow:
## Fence (may touch ONLY)
- `.github/workflows/windows-suite.yml`
- `.windows-known-failures`
- `scripts/windows-triage.sh`

## Blocked by
Wave 2 — the `windows-latest` pack-smoke leg. Until that leg exists and its cost is measured, there is no evidence that a second Windows job is affordable, and no proof the runner can run this repo's bash at all.

Sibling lanes: #23 owns `.github/workflows/ci.yml`. This issue adds a **new** workflow file rather than editing that one, precisely so the two do not collide; folding it into `ci.yml` is a later decision, made with numbers (prd-25 amendment).

**Model:** sonnet. **Wave:** 3.

## Why

prd-25 rulings 2 and 3. The only native Windows run of this suite that has ever happened found 130 failures across 15 files — 14 in server (`lab/`, `collectors/sessionlog/`, `recorder/`, `api/`, `cli/`) and 1 in core (the eras golden byte-identity, whose CRLF suspect wave 1 pins). That number is currently a sentence in a PRD. Success is not met while it stays one.

**Ruling 2: the pin is a file list, never the scalar.** A count masks a swap — one fixed, one newly broken, still 130. A list is falsifiable per file: a failure outside the list turns the leg red, removing an entry is an ordinary PR, adding one is not. Zero is the destination, not the entry price.

**The list is an expected-fail list, not a skip list** — ruled in the prd-25 amendment. Visible and noisy, reported on every run. A skip list renders debt as health, which is the reading this repo legislates against.

**Ruling 3: classify before fixing.** A cause class per file — path separator, drive-lettered absolute path, `/proc` absence, line endings, process signalling, temp-dir shape. Fifteen files with six causes is not fifteen lanes of guesswork. Note the numbers above are from a run at a commit this tree has moved well past; the first job of this issue is to **re-measure**, and the committed list is what the current tree actually does, not what the PRD remembers.

## Definition of done

- A `windows-latest` job runs the suite and compares the failing set against `.windows-known-failures`, per file.
- A failure **outside** the list turns the leg red. A file on the list that now passes is reported as a candidate for removal, and does not turn the leg red.
- The list is committed with a dated provenance header naming the commit and node version it was measured at, and every entry carries its cause class from ruling 3's six.
- The job reports every listed entry on each run — it never silences one.
[...]

PR #242, opened 2026-09-02
fix(law): the exclusion scan survives a tracked-but-deleted file  (#203)

### LINKED ISSUE #203: prd43 w1: the exclusion scan survives a tracked-but-deleted file

gabriel-canaan:
Closes #203

## What becomes true

`git ls-files` names what the INDEX holds; the disk may not. A tracked file
deleted and not yet staged is in the listing and absent from the filesystem.
#186 item 5 fixed that for `allCitations` with `readSweptFile` and left the
sibling: the exclusion-honesty scan read the same listing through a bare
`readFileSync` and threw `ENOENT`, so the run died carrying a stack trace where
a law's verdict belonged.

## What is the sibling case?

The issue asked for the enumeration rather than the line it named, because
fixing one instance and leaving its sibling is what filed it. **Seven reads in
this file; exactly ONE was fed from git and unguarded.** The verdict for all
seven is now a table on `readSweptFile`, so an eighth has an obvious question to
answer.

The distinction that decides each row is not *"does this path exist today"* but
**whether the path was produced by asking git** — a literal is wrong loudly, an
index entry is wrong silently. A directory listing
(`cleanUpOrphanedFixtures`), an `existsSync`-asserted allowlist entry, and three
hardcoded control paths all correctly stay unguarded, and the table says why.

None of the one that matters. Restoring the bare `readFileSync` reddens the new
test with the exact reported failure:

```
AssertionError: expected [Function] to not throw an error but
'Error: ENOENT: no such file or directory, open
 …/docs/research/this-file-is-tracked-but-gone-from-disk.md' was thrown
```

Restored: 31/31.

**The test drives the real scan, not the helper.** `readSweptFile` was already
proven by #186; the claim here is that the guard is reachable **from the
exclusion scan**, so re-testing the helper would prove the thing that was never
broken. That required lifting the scan's body into `countBrokenCitationsIn` —
inline, proving the guard needed staging a deletion into the real index, and
**this law runs 4x concurrently, so a test that mutates git races every other
copy of itself.** Passing the file list in makes the tracked-but-deleted case a
one-element array instead of a repo mutation.

Two controls, so the test cannot pass for the wrong reason: the fixture path is
asserted **absent** first, and a present file is still counted — so the skip is
a skip, not a scan that quietly stopped working.

## Also re-derived, as the issue's Blocked-by asked

Its citations (`:140` guarded, `:587` unguarded) were against PR #200's head and
had moved to `:152` and `:709`. Re-derived rather than trusted. That blocker is
clear: PR #200 merged.

## A pre-existing race this PR does NOT fix — filed as #241

#203's third Done-when bullet is *"The law is run 4x concurrently; it has raced
itself twice."* Run it 20 times and it is flaky, and the bullet earned its place.

EXECUTED, 20 runs of 4x-concurrent per tree:

| tree | failing runs |
|---|---|
| `main` at `6b5f70f` | 1 of 20 |
| this branch | 2 of 20 |

Same test both times — `cleanUpOrphanedFixtures never deletes a file its caller
reports as TRACKED` — and **pre-existing**; that difference is noise at this
rate. The mechanism: that test writes a real dead-pid-named fixture into
`docs/`, and a *sibling* concurrent run's cleanup correctly deletes it. Its
injected `isTracked` protects the call under test, never the file from siblings.

Deliberately **not** fixed here: it is a different subject in the same file, and
the fix needs the directory injected rather than the obvious dead-pid change,
which would make the test vacuous. #241 carries the evidence, the trap, and the
fix shape. It overlaps this fence, so it dispatches after this lands.

## Gates

Rebased onto `6b5f70f`. `npm run typecheck` clean · `npm run lint` 875 files
clean · `VITEST_MAX_WORKERS=6 npm test` 404 files / 7280 passed / 3
expected-fail / 2 skipped · `scripts/fence-lint.sh 203` PASSED. One file
touched, matching the declared fence exactly.

Not run through the full `/verify` pipeline: a single-file test change with a
certified mutation and its own controls. Say the word if you want it.


hchristinasmith:
## Reviewed — approve. One row missing from the table, and one number worth re-measuring

All seven legs ran and passed. The change is right and the mutation is real.

### The mutation reproduces, with your exact message — EXECUTED

Restoring the bare `readFileSync` inside `countBrokenCitationsIn`:

```
× the EXCLUSION scan skips a tracked-but-deleted file too, not just the main sweep (#203)
AssertionError: expected [Function] to not throw an error but
'Error: ENOENT: no such file or directory, open
 …/docs/research/this-file-is-tracked-but-gone-from-disk.md' was thrown
  Tests  1 failed | 30 passed (31)
```

**Only the new test fails**, which is what makes it the test doing the work
rather than a side effect. Restored, 31/31.

Driving the real scan rather than re-asserting on `readSweptFile` is the right
call and the reason the mutation means anything. Both controls earn their place:
the `expect(existsSync(…)).toBe(false)` precondition, and
`countBrokenCitationsIn([gone, ...realFiles]) === countBrokenCitationsIn(realFiles)`
with `> 0` under it — a skip that is a skip, not a scan that stopped working.

### The enumeration is accurate — EXECUTED, re-derived rather than read

I classified all seven bare reads on `origin/main` independently before reading
your table. Every verdict matches, including the two whose reasoning is doing
real work:

| line on main | what it is | your verdict | mine |
|---|---|---|---|
| `:278` | `cleanUpOrphanedFixtures`, path from `readdirSync(docsDir)` | NOT NEEDED, and wrapped in its own `try`/`catch` | ✅ confirmed — the `catch { continue }` is right there |
| `:709` | the exclusion scan, path from `trackedFiles` | **WAS THE DEFECT** | ✅ |
| `:921` | allowlist check, path from the literal list | NOT NEEDED, `existsSync` asserted on the line above | ✅ confirmed — with the "remove its allowlist entries" message |
| `:1150` / `:1159` / `:1167` | `OWN_FILE`, the sibling placeholder, `capture.mjs` | NOT NEEDED, hardcoded | ✅ |

Six bare calls besides the guarded sweeps, exactly one git-fed and unguarded.
The distinction you draw — **produced by asking git**, not "does it exist today"
— is the correct axis and it is the sentence a later reader will actually use.

### Finding — the table says "every read", and it is missing one

`badPins()` is a **third** caller of `readSweptFile`, fed from
`sweepFiles('docs/*.md')` — the same listing as `allCitations` — and already
correctly guarded. It predates this PR and nothing is broken. But the table is
headed *"Every read in this file"* and the sentence above it says *"besides the
**two** guarded sweeps"*, and there are three.

The count is the part that has to be right, because the table's stated job is
*"so adding a seventh has an obvious question to answer"* — someone checking a
new call site against it starts from a baseline that is off by one.

Fixed in **`81589a6`** on `review-242` (one row + one word, not yet pushed):

```diff
- * bare `readFileSync` calls here besides the two guarded sweeps; exactly ONE
+ * bare `readFileSync` calls here besides the three guarded sweeps; exactly ONE
…
  * | `allCitations`'s two loops | `sweepFiles` (tracked + untracked) | GUARDED — `readSweptFile`, #186 item 5 |
+ * | `badPins` | `sweepFiles('docs/*.md')` — the same listing | GUARDED — `readSweptFile`, and the third sweep, which is why the count above is three and not two |
```

31/31 · lint 0 (875 files) · typecheck 0.

### The concurrency number is much worse than 1-in-20 here — EXECUTED

Your conclusion is unaffected and I want to say that first: I measured **8 of 8
on `main` and 8 of 8 on this branch**, identical, which is stronger evidence
than your 1/20 vs 2/20 that the flake is pre-existing and not this PR's. Not a
reason to hold #242.

But the *rate* in the PR body and in #241 ("roughly 1-2 in 20", "passes 20/20")
does not survive re-measurement on this machine:

| tree | 4×-concurrent runs with a failure |
|---|---|
| `main` `6b5f70f` | **8 of 8** |
| this branch `9647969` | **8 of 8** |

Same test both times, the one #241 names — `cleanUpOrphanedFixtures never
deletes a file its caller reports as TRACKED`. Machine: 18 cores, so all four
runs are genuinely simultaneous; my guess is that the race window scales with
real parallelism and a 2–4 core CI runner is why CI stays green. That is a guess
— what is measured is the table.

Worth carrying into **#241**, because a 1-in-20 flake gets deprioritised and an
8-in-8 one does not, and because the reconciler's own bar is *"passes 20/20"*,
which on this hardware is currently unreachable rather than nearly met.

### Nit — your call, offered not asserted

The new table is inserted **between** `readSweptFile`'s existing docblock and
the function, so there are now two stacked `/** … */` blocks and only the second
attaches. The original *"Guards a sweep read against a file the git listing
still names…"* — the sentence that actually says what the function does — no
longer shows on hover. Merging them, or moving the table below the function,
costs nothing.

### What I did not run

The full suite and `scripts/fence-lint.sh 203` — one file changed, and I ran the
law itself plus its mutation, lint and typecheck.

hchristinasmith:
Correction to the comment above: **`81589a6` is now on this branch** (`9647969..81589a6`), not unpushed. It is the one-row table fix and nothing else.

Take it or revert it — it is your branch.

gabriel-canaan:
## All three answered — `81589a6` kept, nit taken, and your concurrency point turned out to be the interesting one

Rebased onto `295b023` first (the branch was two commits behind main). Head is now `1e49a57`.

### The finding — kept, and it was off by one in the other direction too

`badPins` confirmed as a third caller, independently: `readSweptFile` call sites are `:420` (`badPins`), `:662` (the exclusion scan), `:687` + `:697` (`allCitations`'s two loops). Your row and your count fix both stand, and `81589a6` is untouched in the history.

But the same sentence also said **"six bare `readFileSync` calls"**, and that number described the tree *before* this PR — the exclusion scan was the sixth, and #203 just guarded it. So fixing "two"→"three" left the sentence wrong at the other end. EXECUTED on the rebased tree:

| | count | sites |
|---|---|---|
| guarded (`readSweptFile`) | **4** | `:420` `:662` `:687` `:697` |
| bare `readFileSync` | **5** | `:309` `:1001` `:1230` `:1239` `:1247` |
| **reads in this file** | **9** | |

(`:954` is the helper's own unit assertion, not a read of the tree.)

The ambiguity that let both errors through is that "sweeps" silently switched between counting **call sites** and counting **table rows** — `allCitations` is two sites and one row. The docblock now states nine, names which four are guarded, and says outright that call sites are the unit. "Adding a seventh" → "a tenth". Your `badPins` row keeps the finding and drops the trailing clause about why the count is three, which the new wording makes stale.

### Nit — taken

Merged into one block, the original *"Guards a sweep read against a file the git listing still names…"* sentence first, table below it. You were right that it is the one place that sentence gets read.

### Mutation re-proven, since my edit lands in the same file as the guard

```
× the EXCLUSION scan skips a tracked-but-deleted file too (#203)
AssertionError: expected [Function] to not throw an error but
'Error: ENOENT: no such file or directory, open
 …/docs/research/this-file-is-tracked-but-gone-from-disk.md' was thrown
  Tests  1 failed | 30 passed (31)
```

Restored 31/31.

### Your concurrency number — I get a third answer, and that is the finding

I cannot reproduce it in either direction. EXECUTED here (12 threads, Ryzen 7 5800X, WSL2):

| probe | batches | failures |
|---|---|---|
| 4x-concurrent, this file, simultaneous | 8 | **0** |
| 4x-concurrent, this file, staggered `sleep 0.$RANDOM` | 20 | **0** |
| 4x-concurrent, **full suite** | 1 (complete) | **0** |

116 individual runs of the law, zero failures, on the branch. So the measured rate is now **2/20 (yours-and-mine originally), 8/8 (yours), 0/116 (mine)** across three setups — which says the rate is not a stable property of the code, and your hardware-scaling guess is the most plausible reading.

**That matters more than the number, because #241's first Done-when bullet is *"20 runs of 4x-concurrent are 20/20"* — and on this machine the UNFIXED tree already passes that.** The acceptance criterion cannot distinguish fixed from unfixed here. It is the *"test that cannot fail for the reason it claims"* shape from the runbook, sitting in an acceptance bar rather than a test.

So I went looking for a deterministic falsifier, and there is one — **the race needs no concurrency at all.** Write the dead-pid fixture, then call `cleanUpOrphanedFixtures()` with its *default* git-backed `isTracked` (exactly what a sibling run does) in the same process:

```
it('PROBE #241: a default-isTracked cleanup deletes the dead-pid fixture', () => {
  const rel = `docs/citation-law-untracked-fixture-999997-${process.hrtime.bigint()}.md`
  writeFileSync(path.join(REPO_ROOT, rel), UNTRACKED_FIXTURE_CONTENT)
  cleanUpOrphanedFixtures()
  expect(existsSync(path.join(REPO_ROOT, rel))).toBe(false)   // PASSES
})
```

EXECUTED, `1 passed` — the fixture **is** deleted. The bait file's vulnerability is a plain property of the default cleanup; concurrency only decides whether a sibling happens to call it inside the window. Reverted, not committed here — it belongs to #241, and I have put all of this on that issue.

Nothing above changes this PR. Your approve still applies to `1e49a57` as far as I can tell, but it is a force-push over a rebase plus one docs commit, so re-check the head rather than take my word for it.

### Gates, EXECUTED on `1e49a57`

`npm run typecheck` clean · `npm run lint` 875 files, 0 · `VITEST_MAX_WORKERS=6 npm test` 404 files / 7280 passed / 3 expected-fail / 2 skipped · `scripts/fence-lint.sh 203` PASSED. One file touched, the declared fence.

gabriel-canaan:
## Second review pass — one new finding, fixed in `fe3936c`. Not approving: I am the author.

Re-reviewed at `1e49a57`, commit-by-commit, in a detached worktree with `origin/main` (`a76c4c9`) merged in. **I cannot approve this** — I wrote `b2920e1` and `1e49a57`, and GitHub prohibits an author approving their own PR. This is a self-review, so treat it as findings plus evidence, not as a second pair of eyes. @hchristinasmith's two comments are the only outside look this has had, and there is still **no formal review on record** (`gh api .../reviews` → `0`); her "approve" was prose in a plain comment, which also means my earlier force-push destroyed no approval.

### What verified in the PR's favour

- **The refactor is behaviour-preserving apart from the guard.** The inline scan did `stripFencedCodeBlocks(readFileSync(...))` then `extractCitations`; `countBrokenCitationsIn` does `extractCitations(stripFencedCodeBlocks(raw))` — same order, same dedup per file. The only semantic change is `if (raw === undefined) continue`. EXECUTED via the mutation below.
- **The mutation is certified, 3 of 3.** Restoring the bare `readFileSync` reddens **exactly one** test each run, with the reported failure:

  ```
  × the EXCLUSION scan skips a tracked-but-deleted file too, not just the main sweep (#203)
  AssertionError: expected [Function] to not throw an error but
  'Error: ENOENT: no such file or directory, open
   …/docs/research/this-file-is-tracked-but-gone-from-disk.md' was thrown
    Tests  1 failed | 30 passed (31)
  ```

  `certify-mutation.sh` did **not** fit and I did not use it: it reverts a *line* to its version at a ref, and this mutation deletes a guard that had no pre-fix counterpart (the code was inline). Its four traps handled by hand instead — restore taken from `git checkout HEAD --` rather than retyped, unstaged after, verified with `git diff HEAD --quiet` on every iteration (`restore OK` ×3), and repeated 3× rather than once. Tree clean at the end.
- **Post-merge gates, EXECUTED on the merged tree.** `typecheck` clean · `lint` **876 files**, 0 · `VITEST_MAX_WORKERS=6 npm test` **405 files / 7307 passed** / 3 expected-fail / 2 skipped.

  The PR body claims 875 / 404 / 7280. The difference is `main`'s new `windows-suite-law.test.ts` (#212, merged after the body was written) — **the author measured a tree that no longer exists, and the newer numbers are in the PR's favour.** Nothing regressed.
- **`main` does not touch this fence.** `a76c4c9` added CI/Windows files only; no conflict, and the merge was clean.

### Finding — the docblock's line citations had rotted, in the commit that wrote them

EXECUTED. The docblock said the issue's citations *"had moved to `:152` and `:709`"*. On the shipping tree `:152` is a prose line **inside that same docblock** and `:709` is a `/**` opener. Traced across the sequence:

| tree | `:152` | `:709` |
|---|---|---|
| `b2920e1~1` (pre-fix) | `return existsSync(filePath) ? readFileSync(…)` ✅ | `const text = stripFencedCodeBlocks(readFileSync(` ✅ |
| `b2920e1` (wrote the sentence) | ` *` ❌ | ` * them (the ADR→PRD and the docs/prd2.md …` ❌ |

So the numbers were correct about the tree they were derived from, and **were already stale inside the commit that recorded them** — that commit added the 30-line docblock and shifted every line below it. A sentence documenting the re-derivation of the issue's stale citations went stale by the identical mechanism, in the same breath.

This is the failure `AGENTS.md` names twice — *"Cite CI by job and step name, never by line number"*, and `.swarm/coupling.txt`'s lesson that a number pinned to a moving file rots by construction — in its sharper recorded form: *a wrong pointer to a real location is worse than no pointer, because spot-checking it says "fine"*. `:152` looks like a plausible comment line, so a reader checking it moves on.

**Fixed in `fe3936c`**: the two reads are named by symbol — `readSweptFile`, and the exclusion-honesty scan now called `countBrokenCitationsIn` — which survives any edit above them. The rotted numbers stay in the prose *as the reason* the symbols are used, since the rot is the instructive part.

**Sibling case checked, EXECUTED.** This is the only real line citation in the file. `grep -n '`:[0-9]'` returns three other hits, none a citation: the `:123` / `:78-85` mentions describe the line-ref **stripping grammar**, and the `:1135`–`:1144` ones are synthetic fixtures that exercise that parser deliberately. No tracked file cites this file by line either.

Worth noting the fix adds two path citations (`AGENTS.md`, `.swarm/coupling.txt`) to a comment in the very file this law scans for cited paths. Both resolve, so the law's check on my own change is not vacuous.

The fixture is named `this-file-is-tracked-but-gone-from-disk.md`, but it is never tracked — "tracked" is *simulated* by passing it in the file list, which is the whole point of injecting the list and is correct. The name slightly oversells what the test arranges. Not worth a commit; noted in case it confuses a later reader.

### What I did not run

`scripts/fence-lint.sh 203` on the merged tree (ran it pre-merge, PASSED; `main` did not touch the fence). No 4x-concurrency probe this pass — the three-way disagreement about that rate is on #241, and it is not this PR's subject.

### Where this leaves it

Head is `fe3936c`, `main` merged in, all gates green on the post-merge tree. Both of @hchristinasmith's items and one finding of my own are answered. **It needs one formal approval from someone who is not the author before it can land** — @hchristinasmith, `fe3936c` and the merge commit are the only things added since your read.

Issue #234, opened 2026-09-02
prd43 w6: the outbound sweep's vocabulary covers three more spellings


gabriel-canaan:
## What becomes true

The outbound-call sweep's named vocabulary gains the three spellings `/verify`
found outside it, so the enumeration in `README.md` covers them too.

## Evidence

`#23` (PR #230) narrowed `README.md`'s claim from exhaustiveness to a **named
vocabulary**, because "every way JavaScript can send a byte" is an open set no
regex closes. Under that grading, these three are candidate rows rather than
defects — the prose does not promise them. Recording them so the next person
does not re-derive them:

| spelling | counts | note |
|---|---|---|
| `const { 'fetch': send } = globalThis` | 0 | quoted destructure key; biome would flag it anyway |
| `import('node:http2')` | 0 | a different module, genuinely outside |
| `typeof(globalThis.fetch)` | 1 | **false positive** — the `typeof` capture expects whitespace, not a paren |

The third is the one worth doing: it fails *loud* (a spurious call site reddens
the sweep), so it costs a false alarm rather than a missed egress — but a law
that cries wolf gets suppressed, and a suppressed law guarantees nothing.

## Why this is small and bounded

Seven rounds of adversarial review on this matcher converged, and the closing
move was deliberate: stop chasing cells, state a bounded claim, and make prose
and code constrain each other. Two laws now enforce that — every vocabulary
token must appear in `README.md`'s paragraph AND have a pinned count row, and
the paragraph's own backticked list is compared as a set. So adding a spelling
here is a **three-line change**: the pattern, the prose token, the pinned row.
Any one of the three missing is a red build.

Do not reopen the exhaustiveness question. If a fourth spelling turns up, add a
row; do not re-widen the claim.

## Fence (may touch ONLY)
- `packages/server/src/api/route-class-law.test.ts`
- `README.md`

## Blocked by
**Nothing — #232 landed.** It merged in PR #260 (wave 6) on 2026-09-03, so
packages/server/src/api/route-class-law.test.ts is free. Re-checked rather than
assumed, as the previous text asked: fence lint over #233 and #234 reports no
overlap, wave-coupling reports no content coupling, and the dispatch preflight
reports no open PR touching either fence.

Also cleared since that text was written: README.md was held by #24, which
merged in PR #265. Both of this issue's two fenced paths are now free.

**Model:** sonnet. **Wave:** 6 — and it now runs WITH #233 as the wave's second
[...]

Issue #232, opened 2026-09-02
prd43 w6: every route count this repo states in prose is swept by one law


gabriel-canaan:
## What becomes true

One law derives every route count this repo states in prose, wherever it is
stated — not only in the two files #23 was fenced to touch. The two counts that
are wrong today become right, and cannot drift again.

## Evidence

`#23` (PR #230) made `SECURITY.md` and `mutation-guard.ts` derive their
mutating-route counts from `ROUTE_CLASSES`. Eight further sites state the same
kind of count and are guarded by nothing. All eight found during `/verify`;
`ROUTE_CLASSES` is gated-mutation 6, gated-read 14, read 1, ungated-mutation 4.

**Wrong today:**

| site | says | truth |
|---|---|---|
| `packages/server/src/api/security.ts:75` | "the same token on **ten** reads" | 14 gated reads |
| `packages/server/src/api/lane-index.test.ts:20` | "all **eleven** gated reads" | 14 |

`security.ts`'s line is stale against prd-29's *own* ruling 7, which says in
terms *"fourteen `/api` reads today, not ten."*

**Correct today, guarded by nothing** — so they are the next two to drift:
`packages/server/src/api/index.ts:36`, `docs/architecture.md:147`, `:2084`,
`docs/adr/0008-*.md:85`, `packages/server/src/api/gated-reads.test.ts:18,76`.

EXECUTED: mutating four of these to wrong values at once left the full suite
green — 403 files, identical to the clean run. Control: the same mutation in
`SECURITY.md:142` reddens, so the harness bites and these sites are simply
unswept.

## The shape of the fix

One repo-wide sweep, not eight more anchors bolted onto the existing law.
`#23`'s law grew a hand-listed anchor per claim, which is why naming only the
two wrong sites would point at the wrong repair.

## Explicitly NOT in scope

`docs/adr/0014-exhaustive-route-classification.md:49` retains "six mutating
routes and ten reads" and **must be left alone**. AGENTS.md: the ADR log is
append-only, a dated record counts the tree as it stood, and ADR-0024 already
amends it. A sweep that rewrites landed ADRs destroys what an ADR is for, so
`docs/adr/` is excluded by design and the exclusion needs a comment saying why.
`docs/review/` is dated artefact too, same treatment.

## Fence (may touch ONLY)
- `packages/server/src/api/security.ts`
- `packages/server/src/api/lane-index.test.ts`
- `packages/server/src/api/route-class-law.test.ts`
[...]

Issue #24, opened 2026-08-22
prd43 w5: the README states where lab sessions live


ciaran-slow:
## Fence (may touch ONLY)
- `README.md`
- `packages/server/src/lab/namespace-law.test.ts`

**Fence widened 2026-09-04, before any change, by operator decision.** The law
file was always required by the third Definition-of-done bullet below and was
never in the fence, so a lane taking this brief had to either break its boundary
or fail its acceptance — and the landing gate would have rejected work that was
correct. Note which lab directory this is: the law sits under the SERVER
package, not the web one, so it does not enter the territory #220 claims or the
two files #235 claims — those are all under the web package. Fence lint and the
live-fence preflight both pass. (Paths written plain here on purpose: a
backticked path on a prose line in this section is read as a claim, and the
directory form of it overlaps its own bullet by prefix containment.)

## Blocked by
**Nothing. Wave 0's ADR landed** — ADR-0032, "A synthesized session lives in the harness's own projects tree, not the instrument's data root", merged 2026-09-03 in PR #240. prd-12 is shipped and cannot be edited in place, so that ADR is the record and this README correction is downstream of it, not a substitute for it.

**Corrected 2026-09-04:** this is no longer the only issue in wave 5 — the 2026-09-03 amendment added `#235` to it. It remains the only issue that may claim `README.md`, and that still matters: wave 6's `#234` also claims `README.md`, so fence lint reports OVERLAP between them and the two can never be in flight together. Sweeps come last.

**Model:** sonnet. **Wave:** 5.

## Definition of done
**Every line anchor in the original text had rotted by 2026-09-04 and is corrected here. Verify by TEXT, not by line number — they move.**

- **`README.md:277`** is corrected — not `:244-248`, which is now about session labels and sidecar files and has nothing to do with this. The sentence to fix reads "and the checkpoint/synthesized-session artifacts that live beside it", where ADR-0032 rules they are written to `~/.claude/projects/<slug>/<sessionId>.jsonl`. Find it by the phrase "live beside it".
- The ADR is cited at the corrected site, so the divergence from prd-12 ruling 1 is recorded rather than quietly normalised.
[...]

PR #243, opened 2026-09-02
prd25 w2: a windows-latest leg packs, installs and boots the CLI on every push

### LINKED ISSUE #211: prd25 w2: a windows-latest leg packs, installs and boots the CLI on every push

ciaran-slow:
Closes #211

prd-25 wave 2. One issue, one claim: **a `windows-latest` leg packs, installs and boots the CLI on every push.** Base is `main`; the lane branch carries the issue's single commit `ff7d83e`, rebased on `origin/main` at `6b5f70f` (wave 1's merge). This PR was opened as a draft before the build was finished, because a pull request is the only trigger for the `windows-latest` leg — `ci.yml` has no `workflow_dispatch` and pushes run CI on `main` alone — and has been force-pushed while iterating; every run is listed under Verification.

## What becomes true

`pack-smoke` now runs a 3×2 grid: `ubuntu-latest`, `macos-latest` and `windows-latest`, each at the current node and the declared minimum, no exclude. The Windows leg packs the repo, installs the tarball into a project that has never heard of this checkout, and boots the installed CLI under Git Bash — the first CI witness of the `pathToFileURL` fix in `packages/server/bin/rhizomorph.mjs` that let a built clone boot on win32 at all. It joins `pack-smoke` and not `build-test-boot` deliberately (prd-25 ruling 1 as re-argued on #211): the suite has known native failures with no committed expected-fail list yet, so a suite leg would be permanently red, and #212 (wave 3) is that list. The cost note sits beside the matrix: 2× ubuntu billing, measured at 1.5–3 min per Windows leg against ~40 s on ubuntu.

The README's Windows (native) row moves from "unverified" to an honest partial — *installs and boots, suite not run* — and cites the leg by name (ruling 5: rows move only on a named artifact). `AGENTS.md`'s CI paragraph, which said no Windows leg existed, now says which one does and which job still has none. `CHANGELOG.md` carries the entry.

`scripts/pack-smoke.sh` gains a `uname -s`-keyed Windows branch: a per-run token (the mktemp basename) that every launched process carries in its command line, a WMI query on that token for the leak check (Git for Windows ships no `pgrep`, and the existing "cannot verify" branch would have returned 0 on the one runner where the leak is likeliest), a WMI terminate as the fallback stop, and the sh + `.cmd` shim check in place of `-x`, which MSYS only emulates. Every replacement is explained in a comment beside it.

## What the runner corrected, and one defect it exposed on every platform

- **The plan's premise about the group kill was wrong.** It predicted `kill -TERM -$pgid` could not reach the native `cmd.exe` → `node.exe` grandchildren. The tree dump the script prints before each stop shows exactly that chain, every command line carrying the token — and the self-checking query then finds nothing alive within a second of the TERM. The MSYS2 runtime Git for Windows ships takes the native tree down with the bash it kills. The comments describe the observed behaviour; the WMI terminate is documented as the fallback and the WMI query as the proof.
- **`$(stop_jobs)` was a pre-existing defect on Linux and macOS too.** A command substitution runs in a subshell; the servers are the *parent's* jobs, so the subshell's `jobs -rp` never saw them exit and every boot ran the full 10 s grace loop and reported "did not stop on SIGTERM and was killed". EXECUTED against `origin/main`'s script on macOS: identical output. The verdict now travels in a variable and `boot_and_check` calls `stop_jobs` directly. ubuntu/macOS pack-smoke legs went from ~60 s to 30–40 s; the local run from ~50 s to 12 s.
- **The Windows query is self-checking.** It must see the PowerShell process's own command line (which carries the token). An empty result is a *failed* check on Windows — stricter than the POSIX pgrep-missing branch — because on this runner the check is the whole point of the leg.

## What is the sibling case?

The issue named `packages/app/src/host/node-floor-law.test.ts`: its count of two `node-version: '22.22.2'` pins is unmoved by a matrix row — run, green. The real siblings are the *other* texts that mirror the leg list, and all three are in this diff: the three laws in `route-class-law.test.ts` that pinned the leg's absence (restated at greater strength, below), the `AGENTS.md` paragraph, and the README row. `runbook-delivery-law.test.ts`'s `!cancelled()` parser reads only `- name:`/`if:` pairs, so the new `shell: bash` line and the comment do not move it — run, green. Inside the script: `finish` and `on_signal` always called `stop_jobs` directly, so only `boot_and_check` had the subshell defect; both `rm -rf` exit paths got the same "cannot mask the real exit status" guard.

- **CI, EXECUTED — run [33690410798](https://github.com/launchpad-26/rhizomorph/actions/runs/33690410798):** with `stop_jobs` made a no-op on Windows, the three servers accumulated, the leak check printed `LEAK: process(es) of this run still alive after cleanup` listing all nine processes by pid, parent, image and command line, and **only the two Windows legs went red**. Every other leg stayed green. (The plan's narrower mutation — disable Stop-Process alone — would *not* have gone red, because the group TERM already empties the tree; that is the finding above.)
- **Local, EXECUTED, `route-class-law.test.ts`:** drop `windows-latest` from the matrix → the 3×2 law red; move it to `build-test-boot` → the "only pack-smoke" law red; restore the old README row → the Windows-row law red. Each alone, each restored.
- **What the leg would still survive:** a runner whose MSYS2 runtime stops taking the native tree down. Then Stop-Process is what stops the servers, the verdict reads `terminated`, and the leak query still decides — that fallback path has been reasoned, not observed.

## Fence

Fence as groomed: `.github/workflows/ci.yml`, `scripts/pack-smoke.sh`, `README.md`. **Three widenings, [recorded in the plan on #211](https://github.com/launchpad-26/rhizomorph/issues/211#issuecomment-5516606181) before any change:**

1. `packages/server/src/api/route-class-law.test.ts` — three laws asserted the *absence* of a `windows-latest` leg and the README's `**Unverified` wording; they go red the moment the leg exists and are restated, not deleted: the grid is 3×2; `windows-latest` is on `pack-smoke` and **only** `pack-smoke` (a `build-test-boot` Windows row fails the law); the README row must cite `windows-latest … pack-smoke … both node legs`, name `build-test-boot`, and not say Unverified. Nothing else in the file is touched. #232 and #234 also fence this file; both were Ready/Backlog with no lane when this branched.
2. `AGENTS.md` — one paragraph that stated a gate fact this change makes false.
3. `CHANGELOG.md` — the `[Unreleased]` entry.

The issue's stated blockers (#23, #21, #24) were re-checked at planning: two closed, #24 parked with no lane, so no live fence was crossed.

## Verification

Final run [33691507866](https://github.com/launchpad-26/rhizomorph/actions/runs/33691507866) on `ff7d83e`: nine legs green — `build-test-boot` ×3, `pack-smoke` ×6 including `windows-latest` at current (1m40) and min (1m31). The Windows log shows `MINGW64_NT`, both shims present, `--version` / `npx` / `doctor` from the installed artifact, three boots answering `/api/meta` and `/` (plain, spaces, unicode paths), three stops confirmed by WMI, and `no leaked processes: clean`. `npm ci` installed `@rolldown/binding-win32-x64-msvc` and the build was green at both node legs, so #277's second setup blocker did not reproduce.

Local, node 22.23.2, on `f553e1a` (differs from `ff7d83e` by one digit in a workflow comment): typecheck 0 · lint 0 (875 files) · build 0 · `VITEST_MAX_WORKERS=6 npm test` exit 0, 404 files, 7275 passed, 3 expected-fail, 6 skipped · `bash scripts/pack-smoke.sh` all checks passed, three servers shut down cleanly, leak check clean.

Earlier runs on this PR, for the record: [33687177984](https://github.com/launchpad-26/rhizomorph/actions/runs/33687177984) (first Windows green; suite red on the personal-paths law, fixed), [33688105449](https://github.com/launchpad-26/rhizomorph/actions/runs/33688105449), [33689381056](https://github.com/launchpad-26/rhizomorph/actions/runs/33689381056), [33690410798](https://github.com/launchpad-26/rhizomorph/actions/runs/33690410798) (the mutation).

**What has not been done, so the reviewer does not assume it:** there is **no verify report from a second model** on #211. The same agent that planned and built this opened the PR; the operator chose to put it in review ahead of verify. The parts that most deserve a reviewer's own run are the `$(stop_jobs)` subshell finding (reproduce on `origin/main`'s script: every boot takes 10 s and says "killed") and the Windows leg's log for the stop verdicts. The full build report is on [#211](https://github.com/launchpad-26/rhizomorph/issues/211#issuecomment-5517516945).


hchristinasmith:
## Reviewed — two findings, both fixed, neither blocking the leg

`gh pr checks 243`: all nine legs ran and passed, `pack-smoke (windows-latest)`
at current (1m40) and min (1m31) among them. The claim this PR exists to make is
delivered.

You asked the reviewer to run two things themselves. I ran both, plus a third.

### The `$(stop_jobs)` finding reproduces exactly — EXECUTED

`origin/main`'s script, macOS, bash 3.2.57:

```
== boot check (plain): …/watched/plain-repo ==
server did not stop on SIGTERM and was killed (plain)
server did not stop on SIGTERM and was killed (spaces)
server did not stop on SIGTERM and was killed (unicode)
bash scripts/pack-smoke.sh  6.96s user  6.03s system  29% cpu  44.309 total
```

Same machine, this branch:

```
pack-smoke: Darwin, bash 3.2.57(1)-release
server shut down cleanly (plain)
server shut down cleanly (spaces)
server shut down cleanly (unicode)
bash scripts/pack-smoke.sh  5.21s user  3.91s system  75% cpu  12.045 total
```

**44.3s → 12.0s, and three false "was killed" verdicts become three true clean
ones.** The CI numbers agree independently: `pack-smoke` on #240 an hour earlier
ran 59s/1m03/1m03/1m12; on this PR the same four legs run 34/37/35/29. This is
the most valuable thing in the diff and it is a pre-existing defect on every
platform, exactly as you say.

### A third probe: does the leak check fail closed? — EXECUTED

The self-checking query is the load-bearing safety claim, so I broke it rather
than read it. Forcing `ON_WINDOWS=1` on macOS, where `powershell.exe` does not
exist:

```
stop_jobs: the WMI query FAILED (it did not even see itself); nothing native was terminated.
cannot verify: the WMI process query failed (it did not even see itself), so leaked processes were NOT checked for.
EXIT=1
```

No verdict line claims clean, and the script exits non-zero. The sentinel does
what the comment says it does.

### Finding 1 — there is a fourth mirror of the leg list, and it is in `AGENTS.md`

Your sibling-case section names three texts that mirror the leg list and says
all three are in this diff. There are four. `AGENTS.md:425`, **32 lines above
the paragraph this PR corrects**, still reads:

> runs the CLI from those installed files, on the full **2×2 grid**.

It is 3×2 as of this commit. `route-class-law` reads `ci.yml` and `README.md`;
nothing reads AGENTS.md's grid claim — EXECUTED, `git grep -n '2×2\|3×2'` over
`packages/` and `scripts/` returns only `route-class-law.test.ts:430`. So this
one rots silently, and it does so in the section whose own closing note records
that its citations had rotted once before (*"`pack-smoke` was named 31 lines
off"*).

### Finding 2 — the new `windows-latest` law can accuse the wrong job — EXECUTED

`jobBlock()` ends a job at the **next job's key**, so a job's block swallows the
comment block that documents the job below it. That was harmless while the law
asserted absence over the whole file. This PR makes it load-bearing:

```ts
expect(jobBlock('build-test-boot')).not.toMatch(/windows-latest/)
```

The mutation: I added one line to **pack-smoke's own leading comment** — a true
sentence about the correct job, the kind of edit anyone would make next —

```yaml
   # exactly the claim this job exists to back: Linux, macOS and Windows,
+  # (the windows-latest row is the new one)
   pack-smoke:
```

```
× windows-latest runs on pack-smoke and ONLY pack-smoke — build-test-boot has no Windows leg…
  Tests  1 failed | 74 passed (75)
```

A correct edit, a false accusation, naming the wrong job. Today it passes only
because the header comment happens to say "Windows" rather than
`windows-latest`.

Your three claimed mutations all reproduce, for the record: drop
`windows-latest` from the matrix → the 3×2 law red; add it to
`build-test-boot` → the only-pack-smoke law red (and the macOS-exclude law
with it); restore the old README row → the Windows-row law red. Each alone,
each restored, 75/75 between.

### Both fixed — `9e6534f`, on `review-243`, not yet pushed

```diff
--- a/AGENTS.md
+++ b/AGENTS.md
-installed files, on the full 2×2 grid. So the checklist to compare
-`gh pr checks <N>` against is two jobs long, not one. The macOS leg is the one
+installed files, on the full 3×2 grid — ubuntu, macOS and Windows, both
+node legs, no exclude (#211). So the checklist to compare `gh pr checks <N>`
+against is two jobs long and nine legs wide, not one job. The macOS leg is the one
```

```diff
--- a/packages/server/src/api/route-class-law.test.ts
+++ b/packages/server/src/api/route-class-law.test.ts
     const nextJob = rest.slice(2).search(/\n {2}[a-z][\w-]*:\n/)
-    return nextJob === -1 ? rest : rest.slice(0, nextJob + 2)
+    const block = nextJob === -1 ? rest : rest.slice(0, nextJob + 2)
+    // A job ends where the NEXT job's leading comment begins, not at the next
+    // job's key. ci.yml documents every job in a comment block above it, and
+    // those lines are not part of the job above them — without this trim, a
+    // sentence written about pack-smoke is read as text inside build-test-boot,
+    // which is enough to make the windows-latest law below accuse the wrong job.
+    return block.replace(/\n(?: {2}#[^\n]*\n?)+$/, '\n')
```

**The trim does not blunt the law** — that was the thing worth checking, so I
checked it: with the trim in place, the comment mutation above is green, and
moving `windows-latest` onto `build-test-boot`'s matrix still reddens it.

Local gate on the branch with both fixes: `typecheck` 0 · `lint` 0 (875 files) ·
`VITEST_MAX_WORKERS=6 npm test` **404 files, 7275 passed, 3 expected-fail, 6
skipped** — identical to your reported numbers.

### What I did not run, so you do not assume it

**Anything on Windows.** Every Windows branch in `scripts/pack-smoke.sh` — the
WMI query's WQL, the 8.3-vs-long-name reasoning, the `.cmd` shim check, the
`terminated` verdict path — is read and reasoned only; my evidence for it is
your CI logs and the mutation run [33690410798](https://github.com/launchpad-26/rhizomorph/actions/runs/33690410798),
not my own runner. The one Windows-shaped thing I could test on macOS was the
fail-closed path above.

One observation, **your call, offered not asserted**: `stop_jobs` fires the WMI
query immediately after the group TERM, with only PowerShell's own startup as
the delay. On this runner the tree is already gone by then and the verdict is
empty; on a slower stop the query would catch processes mid-exit and report
`terminated` for what was actually a clean shutdown. That is the conservative
direction and your comment already scopes the fallback as reasoned rather than
observed, so I would not hold the PR for it — but a `sleep 1` before the query,
or a note, would keep the verdict honest if the runner ever gets slower.

hchristinasmith:
Correction to the comment above: **`9e6534f` is now on this branch** (`ff7d83e..9e6534f`), not unpushed. Both fixes are in it — the `AGENTS.md` 3×2 grid line and the `jobBlock()` trim.

CI is re-running; the nine legs are worth a second look, since the trim touches a law that reads `ci.yml`.

Take it or revert it — it is your branch.

PR #240, opened 2026-09-02
ADR-0032: a synthesized session lives in the harness's own projects tree (prd-43 wave 0)

gabriel-canaan:
**Closes no issue, deliberately.** prd-43's wave 0 is *"an operator act, booked
not skipped … Not dispatchable: it is a ruling about a shipped constitutional
amendment."* This is that act, so it has no issue to close and
`scripts/dev/issues.sh`-shaped bookkeeping does not apply. Opened with
`gh pr create` rather than the repo's `pr-open.sh`, which requires a `Closes #n`
and would have been satisfied only by inventing an issue for a ruling.

## What becomes true

The laboratory's second write root is on the record with its reasoning, instead
of sitting in an allowlist looking like a leak.

prd-12 ruling 1 granted the lab a write surface and enumerated it, ending
*"checkpoint / synthesized-session artifacts OUTSIDE the watched repo (the same
data-directory posture as the event log)"*. The event log lives under the
rhizomorph data root (ADR-0005), whose own consequences say *"the lab's
containment law compares paths against this root"*.

They are not there. `packages/server/src/lab/checkpoint.ts` writes them into the
harness's own `~/.claude/projects/<slug>/`, and
`packages/server/src/lab/namespace-law.test.ts` permits it explicitly —
`isAllowedWrite` allows `claude-projects` beside `data` — asserted against a
real capture-and-fork run rather than incidentally. The decision was made
2026-08-04 (`c9d3643`), blessed by the law the same day (`57ecfe1`), and never
written down.

**Chosen: A, ratified.** A synthesized session Claude Code cannot load is not a
resumable session, and resuming is the whole purpose of `lab fork` — so keeping
one containment root produces a correct, contained, useless artefact.

**The rejection worth reading is C**, which looks like the principled answer:
write to the instrument's root, copy into the harness tree on an explicit act.
It does not shrink the checkable write surface, it **moves** it — the law must
permit a harness-tree write either way — and it adds a second copy of evidence
that can drift, which is the problem ADR-0020 already settled for transcript
migration.

## What is the sibling case?

The ADR is the record; **three downstream corrections are deliberately not in
this PR**, because each is a separate claim and one of them is user-facing:

- `README.md`'s Trust section says these artifacts *"live beside"* the lab
  worktrees dir. It is telling a reader the wrong location for their own data,
  in the section prd-12 ruling 1 explicitly holds to the honesty bar.
- ADR-0005's *"compares paths against this root"* wants a note that it was
  never true of these artifacts.
[...]

Issue #235, opened 2026-09-02
prd43 w8: a law's file-count comment and its assertion say the same thing


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/lab/no-live-fleet-law.test.ts`
- `packages/web/src/lab/launch/explicit-invocation-law.test.ts`

## Blocked by

**Nothing — ruled unblocked by the operator, 2026-09-07.** #220 (prd30 w3) claims all of
the whole lab directory and both files here sit under it, so the fence lint still reports
an OVERLAP. That overlap was ruled NOMINAL on the evidence, recorded in this issue's
comments: both files are `.ts`, carry zero `title=` attributes and render nothing, so
#220's sweep has nothing to reach here. #220 had no branch, no worktree and no assignee
since 2026-09-02, so waiting was open-ended rather than short.

The ruling does not narrow #220's fence — that is prd-30's issue — and it does not make
a declared fence advisory in general. Re-run `scripts/fence-lint.sh 235 220` and treat
the overlap as real again if #220's scope widens beyond `title=` retirement, or if either
file gains JSX.

Parallel-safe with **#24** (prd43 w5), which claims `README.md` and nothing else. These two can share the wave's PR.

Note for whoever lints this wave: the paths named in this paragraph are other issues' and are not claimed here. Only the block above is this issue's boundary.

**Model:** sonnet. **Wave:** 5.

## Why

prd-43 ruling 2 — *a count stated in prose is derived from the thing it counts* — with ruling 1 already establishing that a **comment** is prose for this PRD's purposes. Two law tests state a file count in a comment, insist in that same comment that the count is exact, and then assert a lower bound.

`packages/web/src/lab/no-live-fleet-law.test.ts`, at its coverage assertion:

> 17 real files as of the 2026-08-08 audit (5 at the root, 2 in `branching/`, 7 in `compare/`, 3 in `launch/`) — **pinned exactly, not a loose lower bound**: headroom here would defeat the point.

The assertion beneath it is `expect(sourceFiles().length).toBeGreaterThanOrEqual(17)`.

`packages/web/src/lab/launch/explicit-invocation-law.test.ts` has the identical pair one directory away:

> 3 real files as of the 2026-08-08 audit (estimate.ts, LaunchPanel.tsx, launch.ts) — **pinned to today's count, not a loose lower bound**, so a silently dropped file fails loudly here too.

[...]

Issue #231, opened 2026-09-02
prd43 w6: a separator alone makes a remedy token a path claim


gabriel-canaan:
## What becomes true

`remedy-reachability-law.test.ts` checks the paths its own docblock says it
checks: a token qualifies as a path claim on a **directory separator OR** a
recognised extension, and a token that resolves outside the repository root is
rejected rather than silently rebased into it.

## Evidence

Both found during `/verify` of #63 (PR #230) and left deliberately unfixed —
outside that issue's fence, and both are the law's own token handling rather
than the defect #63 was filed for.

**1. The separator alone does not qualify a token.** The comment at
`packages/core/src/fleet/remedy-reachability-law.test.ts` reads *"A path-shaped
token needs a directory separator or one of these extensions"*, and the input
table's "bare command name" row repeats it — *"only a token with a directory
separator or a recognised script/doc extension is treated as a path claim at
all"*. The regex requires the extension unconditionally.

EXECUTED: a remedy reading `run scripts/dev/this-script-was-never-written to
fix it`, with the seven-remedy pin updated as any real remedy change would
require, passes the law **17/17**. Appending `.sh` to the same nonexistent
path, same edit otherwise, reddens `every remedy in gaps.ts names a path that
resolves from a fresh clone`. The extension is the only variable.

**2. `tokenResolves` rebases an absolute token into the repo.** It is
`existsSync(path.join(REPO_ROOT, token))`, and `path.join` treats a leading `/`
as ordinary. EXECUTED: a remedy citing `/README.md` passes **17/17**, while
`ls /README.md` is *No such file or directory* — the law reports "resolves from
a fresh clone" about a path a reader following it would never find. Control:
`/nonexistent-top-level/README.md` reddens, because it does not exist relative
to the repo either. Parent traversal (`../../etc/hosts.md`) is likewise not
rejected; it merely happens not to exist.

## Why this matters more than it looks

A law whose stated rule is wider than its implemented rule reports success
about cases it never examined. AGENTS.md already records this shape for
`cmd_orphans`: *"a wrong pointer to a real command is worse than no pointer,
since it reports success."* This is the same thing inside a test.

## Fence (may touch ONLY)
- `packages/core/src/fleet/remedy-reachability-law.test.ts`

## Blocked by
Nothing. #63 landed the law; this corrects its token handling.

gabriel-canaan:
Correction to this issue’s own evidence, after #230 landed.

[...]

Issue #226, opened 2026-09-02
prd43 w6: a tracked screenshot is bound to the tree it depicts


ciaran-slow:
## Fence (may touch ONLY)
- `docs/screenshots/`
- `packages/server/src/no-personal-paths-law.test.ts`
- `scripts/dev/visit.mjs`

## Blocked by
**#22** (prd43 w3, In review) owns all three of these paths. This issue is groomed and boarded now so the block is explicit; it does not dispatch until #22 lands, and it should be re-read against #22's landed shape before it does.

**Model:** sonnet. **Wave:** none — unwaved, like #62/#63/#66; sequenced after #22 by the block above.

## The defect

**Every tracked fixture screenshot depicts a scene that no longer exists.** `docs/screenshots/fixture-20-lane.png`, `fixture-pathology.png` and `live.png` were captured at `3bd1c5a` (2026-08-03). Measured on 2026-09-02 during prd-33's wave-0 glance run (`docs/design/glance-2026-09-02.md`): **84 commits touching `packages/web/src/scene/` have landed since** — growth as a motion class, directional light, the fruiting material, the reactive ground, the light theme, the canvas floor, the seven-event cap. The PNGs show none of it.

#22 binds a tracked PNG to its **manifest** — the synthetic root and the bytes' SHA-256 — so a capture cannot silently carry a home directory or be swapped for different bytes. That is the right law and this issue does not weaken it. But nothing binds a PNG to **the tree it claims to show**. A screenshot that is provably unforged and eighty-four commits stale passes every gate and is still a false document: `README.md` and `docs/demo.md` point a stranger at these images as what they will see.

## Definition of done

- The three fixture screenshots are **re-captured on the current tree through the sanctioned path** — `scripts/dev/visit.mjs --capture`, with `--fixture 2` / `--fixture 3`, so each carries a fresh manifest. The pane's own screenshots are not acceptable here; they have no manifest and would fail the existing law.
- Each manifest gains **the commit the capture was taken at** (`capturedAt`, a full sha), written by `visit.mjs` at capture time, not typed by hand.
[...]

PR #244, opened 2026-09-02
prd-43: Sequencing names wave 6, wave 7, and wave 5's second issue

gabriel-canaan:
**Closes no issue.** This is the operator's grooming act, not a wave's work —
this PRD's own Sequencing says *"`docs/prds/**` is the PRD corpus's own; no wave
here edits a PRD."* That rule is also why this is a separate PR from the wave-6
bundle it makes possible.

## What becomes true

`prd-reconcile` reported **ten issues in this milestone with no wave**, which
means `fence-lint` never saw them and `one PR = one (milestone, wave)` had no
wave to satisfy. The plan of record had fallen behind the tracker in three
places; this closes all three.

**Wave 5 holds two issues, not one.** `#235` was filed into it deliberately —
its residual is recorded in prd-41 — and is blocked on `#220`, not wave 0. The
README exclusivity wave 5 claims survives: `#235` claims two lab law tests and
never `README.md`.

**Wave 6 is new, and the entry says plainly that this PRD did not plan it.**
Waves 1–5 fixed the documents rulings 1 and 2 name. Reviewing that work produced
a second class — places the *laws themselves* are narrower than the rule they
enforce. The claim the entry makes, and the one to push back on if you disagree:
that is what a law finds once it exists rather than scope creep, and it belongs
here because the `charter-laws` carve-out you landed 2026-09-02 says where a
live PRD owns the territory the issue is that PRD's, however awkward the fit.

**Wave 7 is `#66`, booked the way wave 0 is** — an operator act, not
dispatchable until ruled, because the shape chosen decides whether it touches
2 files or 111.

## What is the sibling case?

The drift had three shapes and fixing only the visible one would have left two.
`prd-reconcile` reports **NO WAVE** loudly, so that was the obvious half — but
it also reports **UNDECLARED WAVE** in the other direction, and wave 5's entry
was quietly wrong in a third way that no tool reports at all: it named one issue
while two existed, and `#24`'s body still claimed to be *"the ONLY issue in wave
5"*. A pass that only added wave 6 would have left the tracker and the plan
disagreeing about wave 5.

No test — this is a PRD. The check that stands in for one is `prd-reconcile`,
and it is worth being precise about what it will say:

- Run against **this branch**, the two `UNDECLARED WAVE` lines for `#225`/`#226`
  are still reported. That is the tool reading `origin/main` by design, not a
  defect and not something this PR can settle — they clear on merge.
- `#215` and `#222` stay unwaved deliberately. They are **closed**. Renaming
  finished work to satisfy a reconciler would be the tracker leading the record
  rather than following it, which is the failure this PRD is about.

The six open issues are retitled to `prd43 w6: …` and `#66` to `prd43 w7: …`, so
the tracker matches the plan rather than the other way round.

## Evidence carried into wave 7's entry

Gathered incidentally on 2026-09-03 while writing ADR-0032, and recorded because
it bears on a ruling that has not been made: citing `#148`/`#153` required shas,
because the commits survived the 2026-08 repo recreation and the issue numbers
did not, and ADR-0005's own `#243` now returns *"Could not resolve to an
issue"*. **A dated note alone would not have helped** — the cost fell on
*authoring* a citation rather than reading one, which is the case that keeps
producing new dead references. That argues the guard `#66` already suggests is
the load-bearing half of its own fix.

## Gates

`npm run typecheck` clean · `npm run lint` 875 files clean ·
`VITEST_MAX_WORKERS=6 npm test` 404 files / 7279 passed / 3 expected-fail /
2 skipped. One file touched.

## What this unblocks

`#232` + `#231` + `#225` become a properly-labelled `prd43 w6` bundle —
fence-lint PASSED over them, no overlaps, and all three clear of the live fences
in #240 and #242.


hchristinasmith:
## Reviewed — two findings, and one of them is a live fence collision

Docs-only, one commit, one file, all seven legs green. The grooming act itself
is right and the wave-6 argument holds. Two things to fix before anyone
dispatches on it.

### What verified — EXECUTED

Every issue this text names, checked against the tracker rather than the plan:

| cited | actual title today |
|---|---|
| #232 | `prd43 w6: every route count this repo states in prose is swept by one law` ✅ |
| #231 | `prd43 w6: a separator alone makes a remedy token a path claim` ✅ |
| #225 | `prd43 w6: docs/demo.md and the shell agree about Esc` ✅ |
| #226 | `prd43 w6: a tracked screenshot is bound to the tree it depicts` ✅ |
| #233 | `prd43 w6: watching.md's gap list states its real provenance` ✅ |
| #234 | `prd43 w6: the outbound sweep's vocabulary covers three more spellings` ✅ |
| #66 | `prd43 w7: the docs cite a tracker that no longer exists` ✅ |
| #235 | `prd43 w5: a law's file-count comment and its assertion say the same thing` ✅ |

All ten retitles landed. `#215` and `#222` are both **CLOSED**, so leaving them
unwaved is right and your reasoning for it is the right reasoning. `#232`'s
"eight sites, two wrong today" and `#66`'s 111 files both check out against
their issue bodies. `#235`'s fence is two lab law tests and never `README.md`,
so wave 5's README exclusivity does survive the second issue, as you say.

### Finding 1 — the fence clearance omits #243, which is live and holds two of these files

The PR body says:

> `#232` + `#231` + `#225` become a properly-labelled `prd43 w6` bundle — fence-lint PASSED over them, no overlaps, and all three clear of the live fences in **#240 and #242**.

**#243 is open** (`prd25 w2: a windows-latest leg …`) and its files are:

```
.github/workflows/ci.yml, AGENTS.md, CHANGELOG.md, README.md,
packages/server/src/api/route-class-law.test.ts, scripts/pack-smoke.sh
```

- `packages/server/src/api/route-class-law.test.ts` is in **#232's fence** — and
  in **#234's**, which this entry already stacks behind #232 for exactly that
  reason.
- `README.md` is the file **#24 claims exclusively** for wave 5.

So the sentence is false as written: #240 and #242 were checked, #243 was not,
and it holds a fenced file of one of the three issues named. This is the
*"never bundle across a live fence — check before you start"* case, and it costs
a rebase conflict for whoever picks the bundle up. `#232` and `#234` should
either wait on #243 or say in the entry that they do.

Worth noting the entry's own instinct was right — it already stacks #234 behind
#232 for this same file. The live PR is the third holder nobody counted.

### Finding 2 — #24 still says it is the only issue in wave 5

Your own sibling-case section names this exactly:

> wave 5's entry was quietly wrong in a third way that no tool reports at all: it named one issue while two existed, and `#24`'s body still claimed to be *"the ONLY issue in wave 5"*.

The PRD is amended. `#24` is not — its body still reads, verbatim, at line 7:

```
**This is the ONLY issue in wave 5, and the only one that may claim `README.md`**
(see `.swarm/coupling.txt`). Sweeps come last.
```

The stated goal of the pass was that the tracker and the plan stop disagreeing
about wave 5. They still disagree — just in the other direction now, and a lane
picking up #24 reads the tracker, not the PRD.

The `README.md` half of that sentence is still true and should stay (`#235`
claims lab law tests only); it is the "ONLY issue in wave 5" half that is now
wrong. This is a tracker edit rather than a diff, and this PR already made ten
of those, so it is in scope by its own precedent — say the word and I will make
it, or take it yourself.

### Nothing else to report

The wave-6 argument — that a law finding its own narrowness is what a law does
once it exists, and that the `charter-laws` carve-out puts it in this PRD rather
than a new one — is coherent and consistent with the carve-out's own text
(*"where a live PRD owns the territory the issue is that PRD's, however awkward
the fit"*). I have no push-back on it.

The wave-7 evidence paragraph is the useful half of this PR. *"The cost fell on
authoring a citation rather than reading one"* is the observation that makes
`#66` a guard rather than a note, and I can confirm the underlying facts
independently — I verified all four shas in ADR-0032 during #240's review, and
`gh issue view 243` does resolve today, to this very PR, which is precisely the
"or, worse, to an unrelated live issue" failure the ADR warns about.

### What I did not run

The suite, `typecheck`, `lint` — nothing in this diff is code, and CI ran all
three green. I did not run `prd-reconcile`; I checked its inputs (the ten issue
titles and the two closed ones) directly instead.

Issue #205, opened 2026-09-01
prd48: a crash between the fold worker's commit and its cursor is a tested case, not an assumed one


hchristinasmith:
## Fence (may touch ONLY)
- `docs/research/2026-09-02-shared-record-s2-fold-worker-ordering.md`

The harness is a throwaway outside this tree, as every prd-48 spike's is. `packages/` and `scripts/` are out of scope for this issue.

## Blocked by
#167 (its note states the design and its `RZ_FAULT_POINT` hooks are what this reuses).

Spawned-by: #167
Missed-by: none

## The gap

#167 proved ack-after-durable-journal closes S2's row (c): the server writes and fsyncs the journal **before** the 202, so a crash after the ack cannot lose the batch. Its sweep aimed four fault points at that ordering.

None of them aimed at the fold worker's own ordering, which is the structurally identical bug one layer down:

- **Request handler:** journal-write → fsync → *then* ack. Fault-tested, four windows.
- **Fold worker:** Postgres commit → *then* persist the cursor byte offset. Never fault-tested.

If those two were reversed — cursor advanced before the commit — a crash in between would leave the fold worker resuming *past* a batch that never landed, which shows as a gap the shipper already got a 2xx for. That is the same defect S2 found, arriving by a different route.

The code is correct today: #167's note §3 states commit-then-advance and the implementation follows it. The point is that **nothing tested it**, and #167's own review said so:

> the falsifier is structurally capable of catching it (an acked-but-unfolded batch would show as a gap), but no kill in the 75 was aimed at that narrow window; catching it would have been coincidence, not design.

Two things make this cheap rather than a new spike from scratch: the fault-injection hooks (`RZ_FAULT_POINT`, `RZ_FAULT_AFTER`) already exist, and #167's note carries the full rerun recipe.

A smaller asymmetry to settle in the same run: **the fold cursor's tmp-then-rename write carries no `fsync`, unlike the journal's.** State whether that is a real durability difference or correctly cheap because the cursor is always recoverable by replaying from a lower offset.

Also in scope, because it is the same n=1 problem and the same harness: #167's `closed` window — the one that reproduces S2's defect — fired twice, once with fsync deliberately disabled, so the full design was exercised there exactly once. More trials aimed at `closed` cost almost nothing now the hooks exist.

## Definition of done

- A fault point between the Postgres commit and the fold cursor's persist, fired enough times to mean something, with the trial count stated per window rather than as a total.
[...]

Issue #172, opened 2026-08-28
prd48 w4: a pruned lane reads as pruned after an archive step, unconditionally


hchristinasmith:
## Fence (may touch ONLY)
- `docs/research/2026-08-29-shared-record-s7-archive-tombstone.md`

## Blocked by
Wave 2 (#166, #167, #168, #169).
Sibling lanes hold other surfaces: `...-s8-vps.md` belongs to another lane in this wave, and `packages/server/src/log/paths.ts` belongs to #170 — if the work needs the data-dir override, wait for that issue rather than editing `paths.ts` here. `packages/server/src/log/retention.ts` is live production code and is NOT in this fence: compose with its grammar, never amend it.
**Model:** sonnet. **Wave:** 4.

## Definition of done
S7 discovered a caveat: pruned-reads-as-pruned is not unconditional once an archive step exists. prd-44 #38 ruled a pruned lane must read as pruned; an archive that leaves no tombstone breaks that.

- seal → archive → compact → prune run end to end against a COPIED data directory, never a real one.
- A pruned lane reads as pruned after the archive step, with no condition under which it reads as absent instead — the failing case from S7 re-run explicitly.
- The tombstone's shape stated: what it costs in bytes, and what a reader sees.
- Composition with `log/retention.ts`'s grammar shown, not asserted: the org ceiling is a ceiling the admin names, never a default age (prd-44 #38, and this PRD's non-goals).
- Falsifier stated and its verdict recorded: if any archive path leaves a pruned lane indistinguishable from a lane that never existed, the org-retention design contradicts a landed house ruling and must be re-cut before the build PRD.

hchristinasmith:
Answered by e0e24c7 (PR #247): an archive-time tombstone closes S7's silent-prune gap with zero changes to retention.ts. Conditional PASS with two residual gaps named (tombstone-before-prune ordering is unenforced; git-only lanes are invisible to allAttributedLanes). Merged into prd48, NOT into main.

PR #246, opened 2026-09-03
fix(record): every event in a merged stream carries the actor it came from  (#204)

### LINKED ISSUE #204: record: every event in a merged stream carries the actor it came from
## Fence (may touch ONLY)
- `packages/core/src/record/merge.ts`
- `packages/core/src/record/merge.test.ts`
- `docs/record-format.md`

## Blocked by
Nothing.

Spawned-by: #166
Missed-by: none

## The defect

`interleave()` takes `TaggedEvent[]` — which carries `actorInstance` — and returns `RhizomorphEvent[]`, dropping the tag:

```ts
// packages/core/src/record/merge.ts:172
function interleave(a: readonly TaggedEvent[], b: readonly TaggedEvent[]): RhizomorphEvent[] {
  ...
  merged.push(left.event)   // the actorInstance is discarded here
```

So `MergedRecord.events` has no per-event actor. `MergedRecord.unknown` **does** — `MergedUnknownLine extends UnknownEventLine` with `actorInstance: string` (merge.ts:14-15). A merged record can therefore say which actor a line it *could not fold* belonged to, and cannot say it for a line it could.

This was survivable only while an external reader could recover attribution from the event id. It cannot. Measured on real ledgers by prd-48's concurrent-actors spike (#166, [EXECUTED]): `evt-NNNNNN` is a per-session counter, so two unrelated sessions both start near `evt-000001`.

```
$ comm -12 <(jq -r .id backfill-actor-5.jsonl | sort -u) <(jq -r .id backfill-actor-6.jsonl | sort -u) | wc -l
22755        # of 25,000 unique ids each — 91% overlap, between two actors BOTH internally unique
```

Across all 28 actor pairs in that dataset, pairwise id overlap ran **17,025–24,999 of 25,000 (68–100%)**.

**Distinct from #173.** That issue fixed the *dedup key* — which is namespaced by `actorInstance` and was never at risk from cross-actor collision. This is the *output shape*: attribution is discarded on the way out, and no reader can reconstruct it.

Latent today, on the same terms #173 was: `mergeRecords` has no production callers yet. It goes live when the shared record ships, which is what prd-48 is measuring for.

## Definition of done

- Every event in a merged stream carries the actor it came from, so attribution never depends on an id.
- The asymmetry with `unknown` is gone — a merged record answers "which actor" the same way for a folded line and an unfoldable one.
- `docs/record-format.md` states the property the code actually has. If the id-uniqueness sentence still implies an id can identify an actor across a merge, it is corrected in the same change.
- A test that fails if the tag is dropped again — built from two actors whose ids genuinely collide, so it cannot pass by ids happening to be distinct.
- `SessionRecord`'s exported shape is not changed. It is read by tooling outside this repo (merge.ts:47-48); this is about `MergedRecord`.

## Falsifier
If a reader can already recover per-event attribution from `MergedRecord` by some route this issue missed, the defect is a documentation gap, not a shape defect — say so with the route and close it.

hchristinasmith:
Closes #204

Base is `prd48`, not `main`.

> ### ⚠️ The Baseline run could not be finished, and no telemetry was hand-stated
>
> **Which seat is red:** `reviewer`. **The cause is outside this diff.** That seat was interrupted mid-run (the session it was dispatched from exited while it was working) and then resumed. Its transcript is final but its last message carries no terminal `stop_reason`, so `tokens_out` is that message's *running* total rather than its finished one — the number is not in the file. `baseline step --from-transcript` refuses on both the task output file and the canonical `subagents/agent-*.jsonl`, measuring `model`, `tokens_in`, `tool_calls` and `duration_ms` and declining to invent the fifth. This is the #144 / #462 shape.
>
> **What I did about it:** nothing. `baseline finish` therefore refuses, the run stays `running`, and the retro POST — which requires a finished run — is blocked. I did not fill the missing column from memory: that is a write-once row, and an estimate there can never be corrected. Reporting the red is always available; a fabricated measurement is not recoverable. The retro is carried in this PR body instead (see **Retro** below), so no retro content is lost.
>
> **Evidence the review actually happened**, since the missing row is the usual proof: the seat returned a verdict of 1 finding / 0 blocking, and its finding is the coverage gap described under *What is the sibling case?* — a real defect it found, fixed inside the fence, and re-verified by mutation. The dispatching session re-ran that mutation independently rather than taking it on report, and the third row of the mutation table is that re-run. `merge.test.ts:201` exists because of this seat.
>
> **The code gate is green and unaffected** — this is the recording gate, not the pre-commit gate.

`interleave()` took `TaggedEvent[]` — which carries `actorInstance` — and returned `RhizomorphEvent[]`, discarding the tag. So `MergedRecord.events` had no per-event actor while `MergedRecord.unknown` did: a merged record could name the actor for a line it could **not** fold and could not for a line it could. The private `TaggedEvent` becomes an exported `MergedEvent`, `MergedRecord.events` becomes `MergedEvent[]`, and `interleave` stops unwrapping. `mergeRecords` is byte-identical; #173's dedup key and `SessionRecord`'s exported shape are untouched.

**The falsifier was run first and came back negative.** The issue offered an exit — if a reader could already recover attribution by some route, this was a docs gap. The envelope is `{ id, ts, source, type, payload }`; `source` names the collector kind rather than the actor; `actors` names only the participants; a ts-ordered interleave makes position say nothing. No route, so a shape defect. Attribution cannot fall back on the id either: #166 measured `evt-NNNNNN` as a per-session counter, 22,755 of 25,000 ids shared between two actors and 17,025–24,999 of 25,000 across all 28 actor pairs.

## What is the sibling case?

**`MergedRecord.unknown`, which was already the correct one** — this change makes `events` match it rather than the reverse. Swept for others [EXECUTED]: `grep` for `actorInstance` across `packages/` and `apps/` returns nothing outside `merge.ts`, which is the only file in the repo that creates an actor tag, so there is nowhere else one can be dropped. `readRecord`'s `RecordBodyRead` is single-actor by construction — the actor lives on `record.manifest.actor.instance` and the caller holds the record — and its one production caller (`packages/server/src/cli/replay.ts:143`) reads exactly one record. No multi-record aggregation exists anywhere; `mergeRecords` is the only place two actors' events meet.

**The sibling that actually bit was inside the function being fixed, and review caught it.** `interleave` tags at **four** push sites — two in the merge loop, one in each of its two drain loops — and the first version of the test reached three. Alice leads at every step of the fixture, so the merge loop exhausts the *first* argument's stream and the *second* argument's drain carries the tail; `while (i < a.length)` had never executed in any test in the repo. Fixed by adding a swapped-argument merge, which is what reaches the fourth site.

Three mutations run, all [EXECUTED]:

| Mutation | Result |
|---|---|
| Revert `interleave`'s return type to `RhizomorphEvent[]` | typecheck **red** at `merge.ts:252`, three workspaces — the type change catches it at compile time |
| Force every push to one actor (compiles) | new test **red**: `expected [ 'inst-alice', 'inst-alice' ] to deeply equal [ 'inst-alice', 'inst-bob' ]` — **20 of 21 in the file still passed** |
| Fabricate the tag at the `a`-drain push only | **suite fully green** (404 files / 7276 tests) before review's fix; **red** at `merge.test.ts:201` after |

The second row is why the test exists at all: under a uniformly wrong tag every pre-existing merge test passes, so nothing checked that the tag was *correct* rather than merely *present*. The third is the coverage gap review closed, and it was re-run independently by the dispatching session rather than taken on report.

The test still survives the fixture ceasing to collide ids — which is why the collision is asserted as a precondition (`expect(bob.events.map(e => e.id)).toEqual(ids)`) rather than assumed. It deliberately never reads `payload.lane`: every pre-existing test tells the actors apart by that field, which is a property of the fixture and not of the merge, and a test resting on it would pass against the unwrapping code. Verified non-vacuous by instrumentation — `ids` has 4 entries and both `eventsOf` results have 4, so no assertion passes by matching empty against empty.

Separately, `return merged.reverse()` turns all six mechanically-rewritten assertions red, so none was weakened by re-expressing it through `.event`.

## Fence widening

**None.** Three files, exactly as the issue declared. The negative fence held: `docs/research/2026-09-02-shared-record-s2-fold-worker-ordering.md` (#205) and `docs/research/2026-08-29-shared-record-s7-archive-tombstone.md` (#172) were not needed and not touched.

## A falsified doc, deliberately left

`docs/research/2026-08-29-shared-record-s2-concurrent-actors.md:313-316` says `MergedRecord.events` "is `RhizomorphEvent[]`, with no per-event actor tag". That is now false, as is the same claim in `...-concurrent-actors-retro.md:69-71`. Both are left standing, and the reasoning is worth recording:

- They are **dated spike artefacts**. AGENTS.md treats such documents as records of the tree at a specific commit, read for findings rather than as current state, and that spike's own header describes its predecessor as "read-only, not edited" — the convention is stated by the artefact itself.
- The falsified sentence **is the citation this issue rests on**. Rewriting it would make #204 read as unmotivated and erase the evidence that the defect was found by measurement rather than by taste.
- Both are outside the fence, and widening must be recorded on the PR *before* the change — not available pre-commit.

The counter, stated fairly: a later reader grepping `MergedRecord.events` meets a confident false claim. The remedy for that is a one-line `**Superseded by #204**` marker appended at the head of the finding, which kills the rot while keeping the measurement intact — a follow-up fenced on those files, not an edit in this diff.

## Retro

The fix itself is one line of consequence — `interleave` stops unwrapping, so `MergedRecord.events` carries the actor the same way `MergedRecord.unknown` already did. The interesting part is what review caught: the first regression test covered 3 of 4 tag-writing sites, because the fixture had Alice leading at every step, so the drain loop holding the fourth site was never exercised. The full suite passed green on a mutation that fabricated an actor at that site. It is the exact defect shape this repo's own review checklist names — a fix that handles the case its author considered and misses a structurally identical sibling — except the sibling was a second code path inside the function under test, not a second file, which is why the standard "what other files might drop a tag" question did not surface it. `MergedEvent` wraps rather than flattens because `RhizomorphEvent` is a zod-inferred union with no interface to extend and the non-strict schema would silently drop a flattened tag on re-parse — verified against the repo's own zod, not asserted.

## Gate

`npm run typecheck` 0 · `npm run lint` 0 (875 files) · `VITEST_MAX_WORKERS=6 npm test` 0 — **404 test files / 7276 passed** / 3 expected fail / 6 skipped. Baseline before the change was 404 / 7275, so +1 is the new test and nothing else moved. `scripts/gate.sh` was not run: running it is the landing, and that is the operator's step.

Gear 3, no shift — one ≥3 trigger (a new exported symbol, `MergedEvent`), no gear-4 trigger. 123 insertions over 3 files.

https://claude.ai/code/session_01RSd1F2EBnU3uW9xQiruzad


hchristinasmith:

Read commit-by-commit (one commit, `2416ccf`) and verified the claims independently rather than from the body. **No blocking findings. Not merging** — the merge is the operator's.

### Verified [EXECUTED]

|---|---|---|
| Zero external consumers of `MergedRecord.events` | `grep -rn "mergeRecords\|MergedRecord\|MergedEvent"` across `packages/`, excluding `record/merge.ts` and its test | **zero hits** — the shape change reaches nothing outside the fence |
| `SessionRecord`'s exported shape untouched | `git diff … -- merge.ts \| grep '^[-+].*SessionRecord'` | **no lines** — the issue's one hard constraint holds |
| `mergeRecords` byte-identical | diffed `merge.ts`; no hunk falls inside the function | **confirmed** |
| The doc this PR declines to fix really does say what it claims | `git show origin/prd48:docs/research/…-concurrent-actors.md`, lines 310–318 | **confirmed verbatim** — "`RhizomorphEvent[]`, with no per-event actor tag … `interleave()` deliberately does not carry `TaggedEvent.actorInstance` through" |
| Fence held | 3 files changed, matching the issue's declared fence exactly | **confirmed**, no widening |

### Verified [REASONED]

**The four-push-site argument is correct, and the swapped-argument call is load-bearing.** `interleave` pushes at four sites: two in the merge loop, one in each drain. The fixture has alice at `FIXTURE_START_TS` and bob at `+500`, so alice leads at every step — the merge loop runs `a` out and `b`'s drain carries the tail, reaching sites 1 and 4. Swapping the arguments makes bob the `a` stream, so the loop takes `right` throughout and `a`'s drain carries the tail — sites 2 and 3. Together the two calls reach all four. Without the swap, `while (i < a.length)` is unexercised by any test in the repo, which is exactly what the third mutation row demonstrates.

**The wrapper-not-flattening choice is right for the stated reason.** `rhizomorphEventSchema` is built from non-strict `z.object(...)`, so a flattened `{ ...event, actorInstance }` re-parsed through the schema comes back with the tag stripped — silently. A wrapper cannot be silently stripped. The rejected `eventActors: string[]` alternative is also correctly rejected: index alignment that no type enforces desyncs on the first `events` filter.

**The non-vacuity guard is the right one.** `expect(bob.events.map(e => e.id)).toEqual(ids)` asserts the fixture actually collides rather than assuming it, and the test deliberately never reads `payload.lane` — which every pre-existing test uses to tell the actors apart, and which is a property of the fixture rather than of the merge. A test resting on `payload.lane` would have passed against the unwrapping code.

### Not re-run here

The three mutations in the body are taken as reported, **except** the third, which the dispatching session already re-ran independently before the PR existed. I did not re-execute the suite: the lane's worktree was still active when I reviewed, and running in it would have raced the lane. `npm test` on this branch is the one check a second reader would add.

### The red seat

The `reviewer` telemetry gap is the #144/#462 shape and the handling is correct — the missing column was left missing rather than filled from memory, which is the only recoverable choice for a write-once row. The code gate is green and separate. Worth noting the run stays `running` in Baseline and the retro lives in the PR body rather than the API.

### CI

`pack-smoke` **4/4 pass** across the ubuntu/macOS × current/min grid. `build-test-boot` 3 legs still pending at time of review — those must be green before this lands.

### The follow-up this PR names

The `**Superseded by #204**` marker on `…-concurrent-actors.md:313-316` and `…-concurrent-actors-retro.md:69-71` is the right remedy and should be filed rather than left in the body. A later reader grepping `MergedRecord.events` currently meets a confident false claim in a document this issue's own evidence rests on.

PR #247, opened 2026-09-03
docs(research): tombstone manifest closes S7's silent-prune gap for telemetry lanes (#172)

hchristinasmith:
Closes #172

## Note on baseline recording, near the top

`baseline finish` could not complete: one review-seat step (`reviewer`) has no recordable telemetry because its subagent transcript ends with no terminal `stop_reason` (the reviewer's host process was interrupted mid-run and resumed from a saved transcript — same failure class as #462). `baseline step --from-transcript` refuses to report because `tokens_out` cannot be measured from a non-terminal message, and mixing manual numbers with `--from-transcript` is refused by the CLI. No telemetry was hand-stated. The run (`e3a6dce7-9e9e-4800-8e06-5307b136f077`) is left `running` rather than fabricated into `delivered`.

## What this is

A research spike (gear 3, doc-only fence: `docs/research/2026-08-29-shared-record-s7-archive-tombstone.md` + its sibling retro). No shipped code changed — `retention.ts`, `paths.ts` and `lane-index.ts` are read-only reference throughout.

S7 (`docs/research/2026-08-28-shared-record-s7-retention.md`) found that after seal → archive → prune, a pruned lane with **no** captured-transcript sidecar vanishes from the lane index entirely — contradicting prd-44 #38's ruling that a pruned lane must always read as pruned. This spike tests whether an **archive-time tombstone** — a hand-written `manifest.json` at the same path/shape a real transcript capture already uses — can close that gap without editing `retention.ts` or `paths.ts`.

**Verdict: conditional PASS**, with two independent residual gaps named, not one:

1. The tombstone mechanism works and composes with `retention.ts`'s existing `silent` grammar with **zero code changes** — verified by re-running the spike from fabricated data, real unmodified shipped functions throughout (`export-record`, `verifyRecord`, `readRetentionPlan`/`applyRetentionPlan`/`voiceRetentionPlan`, `readLaneIndex`).
2. It only helps if something **enforces tombstone-before-prune ordering** — nothing in the fenced files does.
3. It only helps for **telemetry-attributed lanes** — a lane the index tracks only through git (an uninstrumented worktree branch) is invisible to the tombstone writer's lane enumeration (`allAttributedLanes`) and stays silent regardless of ordering. This second gap was found by the reviewer agent's independent re-execution, not by the implementer's original scenario set, and has a different, non-overlapping remedy from gap 2.

## What is the sibling case?

[...]

PR #248, opened 2026-09-03
prd48 w4: a crash between the fold worker's commit and its cursor is a tested case  (#205)

### LINKED ISSUE #205: prd48: a crash between the fold worker's commit and its cursor is a tested case, not an assumed one

hchristinasmith:
> ### ⚠️ The run is not recorded in Baseline, and the cause is outside this diff
>
> **The `reviewer` seat could not be written.** Its `SubagentStop` never recorded a verdict, and re-reporting it from either surviving transcript fails the same way both times:
>
> > the last message carries no `stop_reason` … the output counter on a message that never carried a terminal `stop_reason` is that message's running total rather than its finished one (#144), so the number is not in the file and will not be guessed
>
> `model`, `tokens_in`, `tool_calls` and `duration_ms` all measured cleanly off final records; **`tokens_out` alone is unmeasurable.** This is #462, and it is a property of the transcript, not of the change — the diff is two markdown files and touches nothing that could cause it.
>
> **No telemetry was hand-stated**, so `baseline finish` was not run and the retro was **not** POSTed (that route refuses before `finish`, by design). The retro file itself is committed in this PR. Run `2ff8e477-a86c-4eaf-b79d-af30b5b8db48` will sit `running`; finishing it needs a `tokens_out` nobody can honestly supply.
>
> **The review did happen, and the evidence is in the diff rather than in the telemetry.** It re-derived the mutated run's 15,000-row loss set from `journal.ndjson` and `server.log` after the databases were dropped, matched the note's table exactly, and returned two findings — both fixed before the commit, both listed below. It also proved the personal-paths law bites by staging the note, appending a real home path, watching the law go red, and reverting.

Closes #205

#167 proved ack-after-durable-journal closes S2's row (c), but all four of its fault windows sit in the HTTP request handler. None aimed at the fold worker's own ordering — Postgres `COMMIT`, then persist the cursor byte offset — which is the structurally identical bug one layer down. The code was correct; nothing tested it, and #167's own review said so: catching a reversal there "would have been coincidence, not design."

This adds the fifth window, `f1`, in the existing `RZ_FAULT_POINT` grammar, between the `COMMIT` and `writeFoldCursorSync`.

## Results, per window — never as a total

| window | trials | gaps |
|---|---|---|
| `f1`, correct ordering | 30 | 0 |
| `f1`, ordering reversed (the mutation) | 30 | **15,000** |
| `closed`, full design, fsync on | 30 | 0 |

`closed` was #167's thinnest window — fired twice, once as the fsync ablation, so full design there was n=1. It is now **n=30**, reported as its own number rather than folded into any total.

**Counts that are not trials, stated rather than smoothed:** each run logged 61 bind banners (30 faulting instances + 31 plain restarts; every instance prints one from its `listen` callback, so it is a spawn count) and 30–31 untargeted port-clear kills (the driver force-killing the between-trial server at an arbitrary moment — real chaos, evidence for no window). `TRIAL WARNING`, `EADDRINUSE` retries, drain errors and torn tails are 0 in all three runs.

**None** — that is the point of the run, and it is the strongest-evidenced thing in it.

The mutation is one variable, `RZ_CURSOR_BEFORE_COMMIT=1`, moving the cursor persist to the other side of the commit while `f1` stays textually between the pair. It loses **15,000 rows in 30 contiguous runs of exactly 500** against a fully fsynced, intact journal. `BATCH_LINES` is 500, so the unit of loss is exactly the unit of ack, one-to-one with the kills, no partials and no extras.

Review re-derived that loss set independently from `journal.ndjson` and `server.log` **after** the databases were dropped, and matched the table exactly — including the decisive signature: for **30 of 30** faults the next instance's resume offset equals the lost record's *end*. The cursor demonstrably advanced past a record whose commit never ran, which is the reversed ordering's signature and nothing else's. `ship.log` carries an ACK for 30 of 30 lost ranges, so these are acked batches, not batches never shipped.

## What is the sibling case?

`synchronous_commit=off`. Commit-then-advance is safe *because* a returned `COMMIT` is durable; turn that setting off for throughput and a host crash reproduces the mutated ordering's failure without the mutation. Untestable on this box (a same-process kill of the server cannot lose a Postgres commit either way) and flagged as REASONED, not claimed.

`c1`/`c2`/`c3` were deliberately not re-swept — that would re-litigate #167, which the issue forbids.

## The fold cursor's missing fsync: correctly cheap

On a narrower argument than "recoverable". #167's own ablation showed a same-process `kill -9` cannot lose a write the kernel already has, so a kill-based sweep of the cursor's fsync would test nothing at all. The question was re-put as *is every reachable post-crash cursor state safe?* and all five were written by hand — offset 0, partial rewind to a derived record boundary, zero-length, 64 bytes of real garbage, absent — all recovering to **1,129,197 distinct n, 0 gaps**. It rests on `synchronous_commit=on`: the cursor can only ever be behind, never ahead. Whether a *host* crash could put it ahead is unsettleable here (needs power-loss or `dm-log-writes`) and is labelled REASONED.

Cost named rather than glossed: recovery takes up to 7,410 ms, and the cold-start path allocates one 406,833,703-byte buffer and inserts 1,129,197 rows in a single transaction.

## Falsifier

**Did not fire** — no `f1` crash under the shipped ordering stranded an acked batch. But its second half lands anyway: the mutated run had a fully durable, fully fsynced journal and still lost 15,000 acked rows. So **ack-after-durable-journal is under-specified as a written build contract.** prd-48 criterion 2 says "the build requirement is ack-after-durable-journal", which names the handler and is silent about the fold worker; a build implementing exactly that phrase could still produce the output above. The build PRD should carry both orderings. Flagged as a recommendation — the PRD is outside this fence and was not touched.

## Review findings — 2, both fixed before the commit

1. **A check that could not fail was cited as evidence.** §4.1 cited `duplicate n: 0` as proof the re-fold cost nothing, but `PRIMARY KEY (project_id, actor_instance, n)` makes `n` unique by construction, so that query returns 0 whatever happens. Proved by rebuilding the schema and trying to insert a duplicate. The mechanism claim was right; the evidence was the wrong one, and was substituted with resume-offset evidence from the same log. This repo's named second-worst defect shape, inside a note whose whole subject is the opposite.
2. **The note understated its own coverage.** §9 claimed the multi-record drain step was never faulted under a live shipper; its own logs show 2 of 30 instances draining 9 and 27 records across multiple steps before being killed, both recovering clean. Rewritten to claim the coverage that exists and narrow the residual gap to the killed step's own size — unrecoverable precisely because its cursor never persisted.

Both moved the note toward understating rather than overstating.

## Fence

Widening recorded on the issue **before** the change (#205 comment 5519119549): the `-retro.md` sibling, this repo's established shape for a prd-48 spike retro (#167's and the concurrent-actors spike both have one). Line 3's lane date was corrected `2026-09-02` → `2026-09-03` to match the run stamps, keeping the fenced filename — the same split #167's note has (filename `2026-08-29`, line 3 `2026-09-01`).

Negative fence held and is reported as held: #204's `packages/core/src/record/merge.ts`, `merge.test.ts` and `docs/record-format.md`, and #172's S7 note, were not needed and not touched. Nothing in `packages/` or `scripts/` changed. The throwaway harness lives outside the tree at `~/rhizomorph-spikes/fold-worker-ordering/` and is named in the note's rerun recipe.

## Gear

**3, no shift.** One new file inside the fence plus the recorded retro sibling. No migration, no new exported symbol, nothing on `security-auditor`'s path list.

Two seats gear 3 declares did **not** run, recorded as deviations rather than left silent: **`architect`** — the plan already existed as #167's rerun recipe plus the issue's Definition of done — and **`docs-syncer`**. Nothing in the diff detects a skipped seat, which is why it is written down here.

## Two follow-ups this run earned, not filed silently

- **#167's retro already prescribed the spawn-detector fix and it never landed.** Its ambiguity (deciding a spawn succeeded by liveness at 300 ms cannot tell `EADDRINUSE` from "bound, fired, self-killed fast") was diagnosed, the ready-marker fix prescribed, and `docs/research/2026-08-24-shared-record-spike-plan.md` named as the artefact — closing with "not made in this retro, this lane's fence is the retro file only." That file has one commit, predating both retros. This lane copied the harness and paid for it: 15 kills for 3 programmed trials in the first mutated smoke, and one dropped database. Copying a harness copies its measurement bugs, and a bug harmless under one claim can be fatal under a stricter one.
- **`no-personal-paths-law.test.ts` reads `git ls-files`, so it is blind to an untracked file.** A pre-commit green bar is not evidence for a new file; only a staged run is. Verified by staging the note, appending a real home path, and watching the law go red, then reverting. Out of fence (`packages/server/`), so flagged rather than fixed.

## Verify

Green on the final state, with both files staged so the personal-paths law actually saw them:

```
npm run typecheck            clean
npm run lint                 875 files, no fixes
VITEST_MAX_WORKERS=6 npm test    404 files, 7275 passed | 3 expected fail | 6 skipped
```

Not merged — this stops at the PR.

https://claude.ai/code/session_01EigWpvS91CgZpkxyW6sgga


hchristinasmith:

One commit (`b59c978`), docs-only, +801/−0 across 2 files. **No blocking findings. Not merging** — the merge is the operator's.

### Verified [EXECUTED]

|---|---|---|
| The fence widening was recorded **before** the change | read comment `5519119549` timestamp against the commit's | **holds** — widening recorded `01:53:34Z`, commit at `02:00:52Z`. Seven minutes, right order |
| Sanitisation clean | grepped both new files for `/Users/`, `/home/`, the operator's username, `MacBook`, excluding placeholders | **no real paths** in either file |
| No shipped source touched | `gh pr diff 248 --name-only` | **2 files, both `docs/research/`** |
| Negative fence held | #204's three files and #172's S7 note absent from the diff | **confirmed** — all three lanes ran concurrently and none crossed |

**This is the fence discipline #247 got wrong**, and it is worth saying so explicitly: same repo, same convention, same `-retro.md` sibling — #248 recorded it on the issue before writing, #247 called it "no widening". Both changes are correct; only one is checkable afterwards.

### On the result

**The falsifier did not fire, and the run is still evidence** — that is the harder outcome to report honestly and it was reported. No `f1` crash under the shipped ordering stranded an acked batch. But the mutated arm lost **15,000 rows in 30 contiguous runs of exactly 500** against a fully fsynced, intact journal, with `BATCH_LINES = 500`, so the unit of loss is one-to-one with the unit of ack — no partials, no extras.

The decisive signature is the one I would have asked for: for **30 of 30** faults the next instance's resume offset equals the lost record's *end*. That is the reversed ordering's fingerprint and nothing else's, and `ship.log` carries an ACK for 30 of 30 lost ranges — so these are acked batches, not batches never shipped. Review re-derived that loss set from `journal.ndjson` and `server.log` **after the databases were dropped**, independently, and matched the table.

**Counts per window, never as a total**, exactly as the issue required — and `closed` moved from n=1-at-full-design to n=30, reported as its own number. Spawn banners and untargeted port-clear kills are separated from trials rather than smoothed into them.

### The two review findings are the valuable part

The first is this repo's **named second-worst defect shape, found inside a note whose whole subject is the opposite**: §4.1 cited `duplicate n: 0` as proof the re-fold cost nothing, but `PRIMARY KEY (project_id, actor_instance, n)` makes `n` unique by construction, so that query returns 0 whatever happens. A check that cannot fail was standing in as evidence. It was proved by rebuilding the schema and attempting a duplicate insert, and the claim was re-evidenced from resume offsets in the same log rather than deleted.

The second moved the note toward **understating** its coverage: §9 claimed the multi-record drain step was never faulted under a live shipper, and its own logs show 2 of 30 instances draining 9 and 27 records across multiple steps before being killed. Both corrections went the safe direction.

### The finding that matters beyond this PR

> prd-48 criterion 2 says "the build requirement is ack-after-durable-journal", which names the handler and is silent about the fold worker; a build implementing exactly that phrase could still produce the output above.

That is a **PRD-level finding surfaced by a spike**, correctly flagged as a recommendation rather than acted on, since the PRD is outside the fence. The build PRD should carry both orderings. This is the thing this run bought that the issue did not ask for.

The `synchronous_commit=off` sibling case is labelled **REASONED, not claimed** — correct, since a same-process kill cannot lose a Postgres commit either way, so the box cannot test it.

### Deviations recorded rather than left silent

Gear 3's `architect` and `docs-syncer` did not run, and the PR says so. **Nothing in the diff detects a skipped seat**, which is exactly why writing it down is the whole mechanism. Both omissions are defensible here — the plan already existed as #167's rerun recipe plus the issue's Definition of done.

### The red seat — third of three

Same #462 shape as #246 and #247: the `reviewer` subagent's transcript ends with no terminal `stop_reason`, so `tokens_out` is unmeasurable while `model`, `tokens_in`, `tool_calls` and `duration_ms` all measured cleanly. Run `2ff8e477-a86c-4eaf-b79d-af30b5b8db48` stays `running`. **No telemetry hand-stated** — right call for a write-once row.

**All three prd48 lanes hit this.** Three for three is not a coincidence and it is a tooling gap, not a lane failure. It deserves its own issue.

### Two follow-ups this run earned

Both are real and neither should be lost:

1. **#167's retro prescribed a spawn-detector fix that never landed** — deciding a spawn succeeded by liveness at 300 ms cannot tell `EADDRINUSE` from "bound, fired, self-killed fast". This lane copied that harness and paid for it: 15 kills for 3 programmed trials in the first mutated smoke, plus one dropped database. Copying a harness copies its measurement bugs.
2. **`no-personal-paths-law.test.ts` reads `git ls-files`, so it is blind to an untracked file.** A pre-commit green bar is not evidence for a new file; only a staged run is. Verified by staging the note, appending a real home path, watching the law go red, and reverting. This is the guard-scoped-by-the-wrong-thing failure AGENTS.md already records once, in a different form — and it is out of fence (`packages/server/`), so correctly flagged rather than fixed.

### CI

`pack-smoke` **4/4 pass**; `build-test-boot` **3 legs in progress** at time of review. Both jobs must be green before this lands.

PR #249, opened 2026-09-03
prd48 w2: the local store's behaviour on native Windows is measured

### LINKED ISSUE #168: prd48 w2: the local store's behaviour on native Windows is measured
## Fence (may touch ONLY)
- `docs/research/2026-08-29-shared-record-s3-windows.md`

## Blocked by
#165.
Sibling lanes hold other surfaces: `...-s2-concurrent-actors.md`, `...-s2-ack-after-journal.md` and `...-s5-pending-invite.md` belong to other lanes in this wave — if the work needs them, stop and surface it rather than absorbing it. The landed S3 note is read-only history; do not edit it.

## Definition of done
S3 answered "none vs SQLite vs PGlite" and concluded no local database. It did not measure native Windows, which is the platform the conclusion is least safe on.

- The three engines measured on native Windows, not WSL, with the distinction stated plainly.
- The shipped parse cache measured on the same box, since that is what beat every engine elsewhere — the comparison is against it, not against nothing.
- Numbers reported in the same shape as the landed S3 note so the two tables can be read side by side.
- Any Windows-only behaviour that changes the verdict is named, including file locking and path handling if they bite.
- Falsifier stated and its verdict recorded: if an engine beats the parse cache on Windows, "no local database" is a POSIX conclusion wearing a general one's clothes, and the build PRD needs a per-platform answer.

gabriel-canaan:
Closes #168

Wave 2's Windows gap. S3 answered "none vs SQLite vs PGlite" on WSL2 only and named the
omission in its own *what this did not test*; this measures the three engines on native
Windows and reports them beside a WSL2 leg re-run for the purpose.

**Base is `prd48`, not `main`** — deliberately, and per the rule in AGENTS.md that a PR's base
is `main` unless you can name the PR that will merge its base. The PR that will merge this one's
base is prd-48's single integration PR, the operator's explicit decision recorded in #166's
closure ("prd-48 lands on `main` as one integration PR when the whole PRD is done"). Sibling
wave-2 notes #166 and #167 landed the same way.

**Fence:** `docs/research/2026-08-29-shared-record-s3-windows.md`, exactly as the issue declares
it. **No widening.** The filename keeps its `2026-08-29` date because that is the path the fence
names; the run date (2026-09-03) is stated in the note's header rather than silently corrected
in the path. Everything else lives in throwaways outside the tree, per PRD ruling 1.

## Verdict

The falsifier **fails to fire**. Warm parse cache **3.0–3.2 ms** on Windows, against node:sqlite's
**20.3–20.8 ms** query (6.4x) and PGlite's **65.9–72.0 ms** (21x). "No local database" is not a
POSIX conclusion wearing a general one's clothes, and **the build PRD does not need a per-platform
storage answer.** The margin widens rather than narrows: the parse cache is platform-neutral
(1.11x) while its nearest rival is the single engine that degrades (sqlite query 1.64x, load
1.69x, cold-open 2.16x). Both risks S3 flagged as untested turned out absent — `node:sqlite` is
built into node so no prebuild can be missing, and PGlite's wasm loaded and queried fine.

Beyond the verdict, three operations POSIX allows and Windows **refuses** with `EBUSY`:
unlink-while-open, rename-while-open, rmdir-with-an-open-file-inside. Prune, rotate and archive
are all "remove or move a file a reader may hold", so #172's word *unconditionally* now has a
named platform to answer for. And `node:sqlite` cannot open a database at a **252**-character
path on Windows while `fs` writes fine past 262 — walked a character at a time to pin 251 OK /
252 FAIL, where 252 + 8 (`-journal`) = 260, so SQLite reserves journal headroom and the ceiling
sits nine chars below `MAX_PATH`.

## Two departures from S3's setup, stated before the tables

- **S3 ran on different hardware** (i9-13900H / 31 GiB; this box is a Ryzen 7 5800X / 32 GiB).
  Differencing S3's WSL numbers against these Windows ones would difference two machines and
  report the remainder as a platform effect. The WSL leg was therefore re-run here; every ratio
  is within-box.
- **S3's throwaway did not survive** — not the log, not the five scripts. They are reconstructed
  from the method its note describes. That is the second reason both legs were re-run: both
  columns come from the same reconstructed file set, so the scripts cancel and the platform
  does not. S3's published table is history beside this one, not an operand.

Corpus substituted for the same reason and stated in the note: 53,698 lines / 22 MB / 20 lanes
against S3's 63,653 / 26.3 MB / 51. The lower `GROUP BY` cardinality flatters the two databases,
not the parse cache; the parse cache wins anyway.

## What is the sibling case?

**The probes are the sibling story, and one sibling is deliberately unclosed.** The three
`EBUSY` findings are structurally identical faces of one fact (POSIX unlinks by name and lets
the inode live; Windows holds a mandatory lock) — so all three were probed rather than
generalising from unlink alone, which is what would have hidden the `rmdir` case that actually
matters to PGlite's directory-shaped store.

The one that is *not* closed: **only `node:sqlite` was probed for locking and path length; PGlite
was not.** `rmdir-while-open` is suggestive for it since its store is a directory, but its own
handle discipline is untested. That is named in the note's *what this did not test* rather than
papered over — it does not touch the verdict, because PGlite lost the timing comparison by 21x
regardless of how it holds its handles.

Second sibling, closed: the issue asks only its own falsifier, but S3 carried two of its own.
Both were re-asked on Windows rather than only the new one — falsifier 2 (does any engine beat
the warm baseline) FAILs identically, and falsifier 1 (PGlite cold start / footprint) PASSes with
S3's same borderline flagged, unchanged and not a Windows effect.

The measurement's own guard: **it could have gone the other way and would have shown it.** The
falsifier is an inequality between numbers the same 18 runs produce, not a restatement of the
conclusion — had any engine's query come in under ~3 ms the note would have had to report a
per-platform answer. The cross-platform agreement check has the same property: 3 engines × 2
platforms × 3 repetitions collapsed to **one** distinct `(top-5, lane-count)` signature, and any
engine disagreeing on any platform would have produced a second one. Every load run deleted its
store and rebuilt from source first, so all 18 runs are rebuild proofs rather than 18 reads of
one build.

Where a mutation was actually *run* rather than argued: the `no-personal-paths-law` green on this
note was suspected of being vacuous — the exact failure AGENTS.md records for a guard scoped by
naming convention. Injecting a home path into this note turns the law **red** (1 failed / 31
passed); removing it turns it green (32/32). So its coverage of this file is demonstrated, not
assumed.

Honest limit on the timings: three runs on one box, and Defender was left on deliberately
(excluding the build dir measures a machine nobody runs) but **not** A/B'd against exclusions —
so the 1.6–1.7x sqlite penalty is "native Windows as configured", and AV-versus-NTFS-versus-Win32
is not separated. Stated in the note.

## Gates

Docs-only; no code touched. In the worktree: `typecheck` 0, `lint` 0 (875 files),
`VITEST_MAX_WORKERS=6 npm test` exit 0 — **404 files, 7279 passed**, 3 expected fail, 2 skipped.


gabriel-canaan:
## Verify report — PR #249 @ `fe52c9c`

**Verdict: NOT ready to undraft.** One acceptance criterion from #168 is unmet, and it is
cheap to close. Nothing found is wrong with the *measurement* — every number in the note
re-derives from the raw capture — but the note does not yet satisfy DoD bullet 3, and two
claims rest on an unstated basis.

### Depth — declared, not implied

**Cost guard applied: no model seats.** §3's table sends a docs-only diff to
orchestrator review, and this diff is one new file, 265 lines, no code. The substantive
check here is re-deriving the note's numbers from the raw bench capture, which the seats
could not do better — they do not have the capture files.

**The limitation this leaves, stated plainly: I am the builder.** This is self-review, which
is the condition the stage normally exists to abolish. It is mitigated by the checks being
arithmetic against raw data rather than judgement, but a reader should weight findings 2 and 3
— which are judgement calls about framing — accordingly.

---

## Must fix before undraft

### 1. DoD bullet 3 is unmet — the tables are transposed against S3's [EXECUTED]

#168 asks: *"Numbers reported in the same shape as the landed S3 note so the two tables can be
read side by side."* They cannot be. Compared directly:

```
S3   (2026-08-28-...-s3-localstore.md:52)  | Engine | Load/build | Cold-open | Query | Cold-open + query |
this (2026-08-29-...-s3-windows.md:100)    | Metric (ms) | WSL2 | Windows | Windows ÷ WSL2 |
```

S3 is engine-rows × phase-columns; this note is metric-rows × platform-columns. Only the
footprint table (`:123`) matches S3's shape. A reader laying the two side by side has to
transpose two of three tables mentally, which is the exact thing the bullet exists to prevent.

The transposition is *defensible* — the note re-ran the WSL leg, so a platform-comparison axis
carries information S3's shape cannot — but it is undeclared, and the DoD is explicit.

**Suggested fix (proposal, not applied):** keep both. Add an S3-shaped table for the Windows
leg immediately above the existing one, so the DoD's side-by-side read works and the
platform-ratio view survives:

```
| Engine | Load/build (once) | Cold-open (fresh process) | Query (lane-index) | Cold-open + query |
|---|---|---|---|---|
| none (cold read+parse) | n/a (read *is* the load) | n/a | 188.4–189.3 ms | 188.4–189.3 ms |
| none (warm, parse-cache-equivalent) | n/a | n/a | 3.0–3.2 ms | 3.0–3.2 ms |
| node:sqlite | 658.9–770.5 ms | 0.6 ms | 20.3–20.8 ms | 20.9–21.4 ms |
| @electric-sql/pglite | 15,104–15,211 ms | 211.5–229.1 ms | 65.9–72.0 ms | 282.3–295.0 ms |
```

Every value above is from the same capture as the note's own table; the `Cold-open + query`
column is the `totalMs` the query scripts already emit.

---

## Worth fixing

### 2. The central premise is modelled, not measured — and the gap is now measured [EXECUTED]

The note's verdict rests on "warm = the shipped parse cache". The shipped cache does not
return the array it holds:

```ts
// packages/server/src/log/session-log.ts:106
export async function readSessionEvents(filePath: string): Promise<RhizomorphEvent[]> {
  return [...(await parsedSessionLogCache.read(filePath)).events]
}
```

`[...]` is a fresh 53,698-element array **per call** — deliberate, and the file says why at
`:99` ("A copy per call, deliberately"). The warm baseline re-aggregates over the *same*
resident array and omits it, so it models the cache optimistically. The note inherits this
framing from S3 and does not name it.

Measured rather than argued (`probe_spread.js`, 20 warm-up iterations discarded, arm order
alternated, 40 samples):

| | Windows | WSL2 |
|---|---|---|
| spread alone, median | **0.075 ms** | **0.211 ms** |
| warm aggregate, +spread − no-spread | +0.097 ms | +0.231 ms |

**The verdict is untouched:** at 3.17 ms + 0.10 ms ≈ 3.27 ms, node:sqlite's 20.4 ms is still
~6.2x slower. So this is a completeness gap in the note's premise, not an error in its
conclusion — but it is exactly the "what did the world say, as opposed to what did the code
say" check, and the answer should be in the note rather than discovered by the next reader.

*Method note, because the first attempt was wrong:* the initial probe reported a **negative**
spread cost (−1.11 ms). That was JIT warm-up landing entirely in the first arm — the command
succeeded and the output was well-formed while measuring the wrong thing. Discarding warm-up
and alternating arm order produced the numbers above.

**Suggested fix:** one sentence in *What this note did not test* (or beside the warm-baseline
definition) naming the omission and its measured size on both platforms.

### 3. Both headline multipliers use an unstated basis [EXECUTED]

The note says node:sqlite is "6.4x" and PGlite "21x" slower than the warm parse cache. Neither
is the median ratio (6.46 and 22.35). Both are **min-of-observed-range ÷ median-warm**:

```
20.31 / 3.166 = 6.415  -> 6.4x
65.86 / 3.166 = 20.802 -> 21x
```

So the basis is **consistent and conservative** — I flagged it as a suspected inconsistency and
the arithmetic refuted that. The residual issue is only that an unstated basis leaves a reader
unable to reproduce the figure, in a note that elsewhere insists on exactly this discipline
("picking the reading that passes is the failure mode").

**Suggested fix:** a half-clause — "(smallest observed rival query ÷ median warm)".

---

## Noted, not a defect in this PR

### 4. The base branch is 4 commits behind `origin/main` [EXECUTED]

`branch-facts.sh` reports this branch STALE against `origin/main`. That is inherited, not
introduced: this branch is **0 behind / 1 ahead** of its own base `origin/prd48`, and `prd48`
is 4 behind / 8 ahead of `origin/main`. Correct for the integration-branch pattern the operator
chose, and it belongs to prd-48's eventual integration PR, not here. Flagged only because
anyone running `branch-facts.sh` on this branch will see STALE and should know why.

---

## Checked and clear

- **All 34 timing-table values re-derived from the raw capture** — every phase-level range,
  every wall-clock range, and all 8 ratios. Zero transcription errors. [EXECUTED]
- **Every other numeric claim verified**: footprints (byte-identical across platforms, all
  four figures), 4.3x pgdata/source, corpus stats (53,698 / 22,043,666 / 34,669 / 64% / 20
  lanes), S3's quoted stats checked against the landed S3 note (63,653 / 26,325,198 / 48% /
  51), the 251-OK/252-FAIL threshold, `252 + 8 = 260`, the box RAM, and the corpus sha256.
  [EXECUTED]
- **`claim-lint.sh`**: 0 unresolved references. It flagged 3 bare counts to derive; all three
  derived above and correct. [EXECUTED]
- **The falsifier is honestly earned.** It is an inequality between numbers the same 18 runs
  produced, not a restatement of the conclusion — had any engine's query come in under ~3 ms
  it would have fired. The margin is 6.4x at its narrowest. [EXECUTED]
- **The correctness claim is load-bearing**: 3 engines × 2 platforms × 3 repetitions collapsed
  to exactly one `(top-5, lane-count)` signature; any disagreement would have produced a
  second. Each load run rebuilt its store first, so all 18 are rebuild proofs. [EXECUTED]
- **`no-personal-paths-law` 32/32, and mutation-proven not vacuous** — injecting a home path
  into this note turns it red (1 failed/31 passed); removing it turns it green. [EXECUTED]
- **Gates**: `typecheck` 0, `lint` 0 (875 files), `VITEST_MAX_WORKERS=6 npm test` exit 0 (404
  files, 7279 passed). Run on this exact content; tree clean and unchanged since. [EXECUTED]
- **DoD bullets 1, 2, 4 and 5 met.** Bullet 4 (Windows-only behaviour named) is met well
  beyond the minimum — three `EBUSY` divergences plus a threshold walked to the character.

## What I did not verify

- **The `EBUSY` → retention consequence is REASONED**, as the note itself says. `log/retention.ts`
  was not run on either platform. #172 is where it becomes executable.
- **PGlite was not probed** for locking or path length — the note declares this.
- **Defender's contribution was not isolated** — no A/B against exclusions, so the 1.6–1.7x
  sqlite penalty is "Windows as configured". The note declares this too.

---

*Reported by the verify stage. Per the skill, this turn stops at the report — the PR stays
drafted and nothing is fixed, undrafted or landed until the operator decides.*

gabriel-canaan:
## Fixes applied — `fe52c9c` → `5b1185b`

All three findings from the verify report are fixed, in the one fenced file. **No fence
widening.** Amended into the single commit rather than added as a second, per the repo's
one-commit-per-issue rule.

### 1. DoD bullet 3 — the side-by-side read now works [EXECUTED]

The Windows leg is given **twice**: once in S3's own engine-rows × phase-columns shape, so it
lays directly beside `2026-08-28-...-s3-localstore.md`'s first table, and once in the
platform-ratio shape S3 has nowhere to put. Both come from the same capture. The note says why
both exist, and repeats the two-departures warning at the point of the side-by-side read, so a
reader cannot difference the two tables without meeting "different box, reconstructed scripts"
first.

### 2. The premise is now measured, and labelled as such [EXECUTED]

The warm-baseline definition now names what it omits — `readSessionEvents` returns
`[...cache.events]` (`session-log.ts:106`), a fresh 53,698-element copy per call — and gives
its measured cost: **0.075 ms** Windows, **0.211 ms** WSL2, moving the Windows warm figure to
≈3.27 ms and leaving node:sqlite ~6.2x slower rather than ~6.4x. Verdict unchanged.

### 3. The multiplier basis is stated [EXECUTED]

"Smallest observed rival query ÷ median warm baseline", with the arithmetic shown
(`20.31 ÷ 3.166 = 6.42`; `65.86 ÷ 3.166 = 20.80`) and the median-to-median figures (6.5x,
22.4x) given so the conservative choice is visible rather than implicit.

---

## How the repair was checked

**Every number in both new tables re-derived from the raw capture** — not retyped. All 10
S3-shaped values, both spread figures, all four multiplier figures. Zero mismatches. [EXECUTED]

**Both cited line numbers verified** (`session-log.ts:99` is the "A copy per call" comment,
`:106` is the function), and each is anchored to a named symbol in the prose as well, so the
citation degrades gracefully if the file moves — this repo's own lesson about registry entries
pinned to line numbers. [EXECUTED]

**Gates re-run on the amended tree**, with the exit status propagated rather than swallowed by
a trailing `echo`: `typecheck` 0, `lint` 0 (875 files), `VITEST_MAX_WORKERS=6 npm test` **exit
0** (404 files, 7279 passed). [EXECUTED]

**`claim-lint`**: 0 unresolved references — it now resolves the two new code references as
well. [EXECUTED]

**Mutation re-check used spellings the first round did not**, per the rule that a re-check
probes the class rather than replaying the input that found it. A non-allowlisted
`/Users/<name>` and a real `DESKTOP-*` hostname each turn `no-personal-paths-law` red
(1 failed / 31 passed); the restored file is green 32/32. [EXECUTED]

*Worth recording because it nearly became a false finding:* a third probe, `/Users/someone`,
passed — and I briefly read that as a coverage gap in the law. It is not: `someone` is on the
law's own synthetic allowlist (`no-personal-paths-law.test.ts:118`), and `/root/` is not a home
shape the law claims to cover. The probe was wrong, not the law. Checking before asserting is
the only reason that did not go into the report as a defect.

**§12 fix re-review — cost guard invoked, not skipped.** The guard exempts fixes with no
behavioural reach; these are prose and tables in a docs file that nothing executes. What
replaces it is stronger for this shape of change than a seat reading the diff would be: every
newly asserted number re-derived from the capture, plus the gates and the two-spelling
mutation above. Round 1 recorded in the re-review budget.

---

**Still drafted.** Undrafting is the operator's call, not this turn's.

hchristinasmith:

One commit (`5b1185b`), docs-only, +298/−0 in the single file the fence declares. **No blocking findings.** CI 7/7 green. Two things to do before it lands, both mechanical.

### Verified [EXECUTED]

|---|---|---|
| Fence held, no widening | `gh pr diff 249 --name-only` against #168's declared fence | **exactly one file**, matching |
| The `session-log.ts:106` citation is real | `sed -n '106p' packages/server/src/log/session-log.ts` | **exact** — `export async function readSessionEvents(...)`, and the "copy per call, deliberately" rationale is where `:99` says it is |
| S3's note, cited as the predecessor, exists on this branch | `git cat-file -e` | present |
| Sanitisation | grepped for `/Users/`, `/home/`, `C:\Users`, contributor names | **no real paths** — `C:\rhizo-spike\` and `~/rhizomorph-spikes/` are throwaways, and the corpus recipe uses `<data-dir>/<slug>` placeholders |
| No path citation that would fail the doc-citation law | grepped the note for the S7 tombstone path | **not cited** — which is why CI is green despite that file not existing on this branch |

### What makes this note good, stated so it is not lost

**The two departures are declared before the tables, not after the verdict.** Different box from S3 (Ryzen 7 5800X vs i9-13900H), and S3's throwaway did not survive so the scripts are reconstructed. The right response to both was taken: **re-run the WSL leg on this box from the same reconstructed scripts**, so every ratio is within-box and the scripts cancel. Differencing S3's published milliseconds against this note's would have reported a machine change as a platform effect. That trap was seen and avoided rather than walked into.

**The warm-baseline premise is measured, not waved at.** The note says plainly that its "warm" arm re-aggregates a resident array while shipped code returns a fresh 53,698-element copy on every call — then *measures the difference* (0.075 ms Windows, 0.211 ms WSL2, medians of 40 samples with warm-up discarded and arm order alternated) and restates the multiplier as ~6.2x rather than ~6.4x. The whole verdict rests on that baseline standing in for shipped code, and the note says which parts are modelled.

**The multipliers are taken the conservative way.** Smallest observed rival query ÷ median warm baseline (6.42x, 20.80x), with the wider median-to-median figures named beside them (6.5x, 22.4x). Picking the flattering reading is the failure mode, and it was declined twice — also on PGlite's footprint, where the 95.6 MB on-disk store is flagged as borderline rather than resolved in favour of the 25.4 MB `node_modules` number that passes.

**The path-length finding got sharper under investigation and the note followed it.** "Fails at 253 chars" became a one-character walk pinning the break at **252**, then the explanation: 252 + 8 = 260, the eight characters are `-journal`, SQLite reserves headroom for its journal path. So the real budget is **251 characters**, nine below `MAX_PATH`, failing with a generic "unable to open database file" that names neither the length nor the reason. `fs` is unaffected because Node's own calls are long-path aware — which is why it looked like Node's problem and isn't.

### The finding with consequences beyond this PR

Three POSIX-allowed operations that Windows refuses with `EBUSY`: unlink-while-open, rename-while-open, rmdir-with-an-open-file-inside. Prune, rotate and archive are all *"remove or move a file a reader may hold."*

**#172 merged forty minutes ago and its note contains no occurrence of "Windows", "EBUSY", "platform" or "POSIX" [EXECUTED]** — I grepped `origin/prd48`'s copy. Its claim is that a pruned lane reads as pruned *unconditionally*, and this note has just named a platform that condition has to answer for.

The note handles this exactly right and I want that on the record: it labels the implication **REASONED, not EXECUTED**, says `log/retention.ts` was never run on either platform, and names #172 as where it becomes a test rather than asserting the conclusion. That is the discipline AGENTS.md asks for, applied against its author's own interest.

**This wants a follow-up issue, not a change to this PR.** #172 is closed and merged; the honest repair is a new issue fenced on the S7 tombstone note.

### Two mechanical things before it lands

1. **The branch is 3 commits behind `prd48`** — #246, #247 and #248 landed after it was cut. `mergeable` currently reports `UNKNOWN`, so GitHub has not finished computing it. Refresh the branch and confirm `MERGEABLE` before merging.
2. **`prd48` is 9 commits behind `main`** — correct as reported. It does not block this PR (docs-only, new file, nothing to conflict with), but it is the standing prd-44 trap: none of prd48's work is on `main` and the milestone will read done while it sits on the integration branch. That is the operator's decision to time, not this PR's problem.

### Not re-run here

The 18 measured runs, the probes and the threshold walk are taken as reported — this box has no Windows, so re-execution is not available to me. Everything I could check from the repo side, I checked; that is stated above rather than implied.

PR #250, opened 2026-09-03
Draft for CI evidence — #212 Windows suite list (open-wave-pr assembles the wave PR)

ciaran-slow:
**Draft for CI evidence only — open-wave-pr assembles the wave PR.** Do not merge.

Lane branch for #212 (prd-25 wave 3). Build report, classification, the two CI mutations and the seventh-class ruling are on the issue. `Windows suite` is green on the 29-entry list (run 33709508449: 0 unexpected · 29 expected).

PR #251, opened 2026-09-03
prd25 w3: the Windows suite result is a committed list of files, not a number

### LINKED ISSUE #212: prd25 w3: the Windows suite result is a committed list of files, not a number

ciaran-slow:
Closes #212

prd-25 wave 3 — the whole of it is one issue, so this bundle is one commit. prd-25 rulings 2 and 3, the 2026-08-24 amendment (expected-fail, not skip), and a new dated amendment ruled during this wave (a seventh cause class). Built on lane `212-windows-known-failures` (tip `c3353f8`, draft PR #250 was CI evidence only), verified on a different model — plan, build report, addendum, author's answers and both verify verdicts are on #212.

## What becomes true

- `.windows-known-failures` — the native Windows suite result as a committed list of **29 test files**, each `<tracked file>  <class>  # evidence: <first line of the failure>`, under one provenance line (`measured: 295b0237… node 24.19.0 windows-latest 2026-09-03 <run>`). Expected-fail, not skip: every entry is reported on every run.
- `.github/workflows/windows-suite.yml` — a separate workflow from `ci.yml` (fenced to other lanes; folding in is a later decision made with this job's measured cost): one `windows-latest` leg at current node, `fetch-depth: 0`, 45-minute bound, suite step `continue-on-error`, the triage step the only step allowed to fail, results uploaded. **Measured: ~13 min end to end, 12 of them the suite** (first green run 33709508449; the run on the lane tip, 33712201493, `0 unexpected · 29 expected · 0 removal candidates · 0 listed-not-run · verdict GREEN`).
- `scripts/windows-triage.sh` — validates the list's grammar before comparing anything, normalises vitest's absolute names (either separator) to `packages/…`, compares **per file, never by count**: outside the list → RED; listed and failing → printed; listed and passing → removal candidate, not red; listed but not run → RED. Mirrors the report into `$GITHUB_STEP_SUMMARY`.
- `packages/server/src/windows-suite-law.test.ts` — 27 cases holding the list honest (one measured line naming a real commit that is an ancestor of HEAD; tracked entries; known classes; evidence present and path-free; no duplicates; header classes == script `CLASSES` == the ruling's) and driving the script with Windows-shaped synthetic results.
- README Windows (native) row moves to "installs and boots; the suite runs against a committed expected-fail list"; AGENTS.md's CI paragraph names the third job and the ten-leg checklist; CHANGELOG `[Unreleased]` entry.

## Triage — 29 files, seven classes

| class | files |
|---|---|
| drive-letter | 12 |
| line-endings | 9 (source-reading laws whose regexes anchor on `\n`; Git for Windows checks out CRLF and the `.gitattributes` pin covers JSON/JSONL/fixtures, not `.ts`/`.sh`/`.md`) |
| path-separator | 6 |
| temp-dir | 2 (8.3 `RUNNER~1` tmpdir vs long realpath) |
| process-signalling | 1 (`spawn npm ENOENT`) |
| fs-semantics | 1 (`static.test.ts`: win32 ignores `O_NOFOLLOW`, raises no `ENOTDIR`) |
| procfs | 0 |

**The seventh class is a ruling, recorded as an appended amendment to `docs/prds/prd-25-the-third-platform.md`** (2026-09-03): `static.test.ts` fit none of ruling 3's six; the build stopped as the plan required, the operator ruled `fs-semantics` in, with widening `temp-dir` and `skipIf(win32)` named and rejected. Five files carry a second cause behind the listed one (`lab/fork`, `lab/restore`, `lab/namespace-law`, `gate-honesty-law`, `api/route-class-law`); the build report names them for the per-class grooming that follows. One flake was deliberately **not** listed: `lab/compare.test.ts`'s never-settling-verify timeout failed once in five Windows runs.

## What is the sibling case?

The issue's own: the list is matched by **resolved tracked path** (`git ls-files --error-unmatch`), never by prefix or the directory the entries share — a fabricated `not-a-real-file.test.ts` entry is rejected before comparison (law B9). Two more handled: a **listed file the suite never ran** is RED, so the list cannot claim a file vitest did not evaluate (B5); a **file that failed to load** (vitest: `failed`, zero assertions, a `message`) is in the failing set like any other and its message is what prints (B6) — and this shape occurred for real: `cli-surface-law.test.ts` is a load failure under CRLF. Review found a fourth: the `continue-on-error` assertion was file-wide, so moving that line onto the triage step would have made the only step allowed to fail unable to fail — now asserted per step, both directions.

**CI, EXECUTED, both reverted:** an unlisted `it.skipIf(process.platform !== 'win32')` failure → run 33707994112 RED naming it (`2 unexpected · 28 expected`), every `CI` leg on that sha green; the `doc-citation-law` entry deleted while it still fails → run 33707998506 RED naming it (`2 unexpected · 27 expected`). Both are the swaps a count would survive.

**Local, EXECUTED, each restored:** class → `flaky`, evidence removed, untracked path, `measured:` removed — each reddens law A; the script rewritten to compare `failing.size !== listed.size` reddens five B cases, including the swap. Verify re-executed its own: a real but non-ancestor measured sha → RED; `continue-on-error` moved, deleted, or duplicated onto the triage step → RED each way.

**Stated limit:** a measured sha that is an ancestor but not the true base still passes law A1 — there is no durable assertion for "the base" once the commit is on `main` (the lane commit is squashed and cherry-picked, `HEAD^` dies at the cherry-pick). The value is right; the law is strictly stronger than before.

## Deviations from the plan, and fence widening

- **Measured sha is the main commit the run was built from, not the lane commit** — a lane sha never reaches `main`, and the law resolves the sha in every full clone. Stated in the list header.
- **Draft PR #250** was opened for CI evidence, as the plan sanctioned; iterated by amend + force-with-lease.
- **Fence widening**, recorded on the plan before the change: `packages/server/src/windows-suite-law.test.ts` (new), `README.md` (one row), `AGENTS.md` (CI paragraph), `CHANGELOG.md` (one entry). Recorded on the issue at the time of the ruling: `docs/prds/prd-25-the-third-platform.md` (one appended amendment; no ruling renumbered). `ci.yml` untouched.

## Follow-ups not filed here (operator's act)

Verify flagged two triage-integrity items that the per-class grooming will not own: a GREEN verdict does not assert the suite evaluated every tracked test file (a results file holding only the 29 listed files is GREEN), and the workflow's `on:` triggers are unasserted. They want one prd-25 follow-up issue between them.

## Gates on the assembled wave branch

`VITEST_MAX_WORKERS=6 npm test`, `npm run typecheck`, `npm run lint`, `npm run build` — all green (node 22.23.2). Boot smoke skipped deliberately: no server or CLI source in the diff (it was run on the lane during build and passed).


hchristinasmith:
## Reviewed — approve. One landed defect fixed, one PR-body count wrong, and your disclosed gap now has a number

Ten legs ran and passed, `windows-suite` (13m43s) among them. The design is
right: per-file set difference, list validated before anything is compared,
missing results red, expected-fail rather than skip. I went looking for the
places a count could still hide and found one thing in the repo and one thing
in the description.

### Your mutations reproduce — EXECUTED

|---|---|---|
| rewrite the comparison to `failing.size !== listed.size` | 5 B cases red, incl. the swap | **5 failed / 22 passed** — happy, unexpected-failure, **the swap**, removal-candidate, and the summary case |
| move `continue-on-error` off the suite step onto the triage step | law C red | **1 failed / 26 passed**, the "defers its verdict" case |

Baseline 27/27 between, and after my change below. `place()` I probed rather
than read: no tracked test file carries a second `/packages/` segment (so
`lastIndexOf` cannot mis-place one) and no test file lives outside `packages/`,
so the RED-on-unplaceable branch is unreachable today rather than merely
untriggered.

The provenance line checks out: `295b0237…` resolves and **is an ancestor** of
this head — it is the #244 merge. All 29 entries are tracked on this branch.

### Finding — the `route-class-law` entry names the wrong regex, and its advice does not work

```
packages/server/src/api/route-class-law.test.ts  line-endings
  # evidence: … not to match /windows-latest/ — the job-block trim anchors on \n
```

The trim is not what fails. `jobBlock`'s **next-job search** is
`/\n {2}[a-z][\w-]*:\n/`, and under CRLF a job key line ends `:\r\n` — so the
search matches nothing, `nextJob` is `-1`, and the block runs to end of file,
straight over `pack-smoke`'s matrix, which is where `windows-latest` lives.

EXECUTED, `ci.yml` from `main` converted to CRLF in memory:

```
CRLF, fix NOTHING          -> law red   (block 10013 chars, vs 7577 under LF)
CRLF, fix ONLY the trim    -> law STILL red
CRLF, fix ONLY the search  -> law green
CRLF, fix BOTH             -> law green
```

**The note as written sends the per-class groomer to a line whose repair changes
nothing** — they would add `\r?` to the trim, re-run, and still be red. The
class is right; the pointer inside it is not, and the pointer is the half that
entry exists to carry. (Disclosure: the trim is a line I wrote reviewing #243
the day before you measured, which is presumably why it looked like the new
thing.)

Fixed in **`1f8b9c3`** on `review-251`, one line, not pushed:

```diff
-  # evidence: … — the job-block trim anchors on \n
+  # evidence: … — jobBlock's next-job search anchors on ':\n', matches nothing
+    under CRLF, and the block then runs to EOF over pack-smoke's matrix
```

Gate on it: windows-suite-law 27/27 · lint 0 (876 files) · typecheck 0 ·
`VITEST_MAX_WORKERS=6 npm test` 405 files, 7302 passed, 3 expected-fail, 6
skipped.

### The PR body's class table does not sum to the list

The table says drive-letter 12 and line-endings 9. Counted off the committed
file: **11 and 8**. The table totals 31; the same PR body says "29 test files"
twice and the PRD amendment says "29 failing files of 405", both of which agree
with the file.

Body only — `.windows-known-failures` states no per-class counts, so nothing
wrong lands. Worth correcting before it gets quoted into the grooming issues,
since the per-class fixes are sliced by exactly these numbers.

```
  11 drive-letter · 8 line-endings · 6 path-separator · 2 temp-dir
   1 process-signalling · 1 fs-semantics · 0 procfs   = 29
```

### Your disclosed coverage gap, with a number — EXECUTED

You flagged that "a results file holding only the 29 listed files is GREEN".
Confirmed, and the magnitude is the part worth putting in the follow-up issue:

```
synthetic results carry 29 files; the repo has 405 tracked test files
windows-triage: 0 unexpected · 29 expected · 0 removal candidates
                · 0 listed-not-run · verdict GREEN
EXIT=0
```

**376 of 405 test files can vanish from the results and the verdict is GREEN,
exit 0.** Not a reason to hold this PR — the job is a strict improvement on
having no Windows suite at all — but it sets the priority of that follow-up, and
the fix is cheap next to the number: assert `ran.size` against
`git ls-files 'packages/**/*.test.ts?(x)'`, which the script already shells out
to git for.

### Observation — your call, offered not asserted

`AGENTS.md` now says the checklist is *"three jobs long and **ten legs wide**"*.
Correct today (3 + 6 + 1). Nothing sweeps it: `route-class-law` reads `ci.yml`
and `README.md`, not `AGENTS.md`. That same paragraph carried a stale `2×2`
until yesterday for exactly this reason, and it has now gained a more precise
and more rot-prone claim in its place.

`windows-suite-law.test.ts` already parses a workflow file, so a case asserting
AGENTS.md's leg count against the two workflows' matrices is a handful of lines
in a file that is already open. Equally reasonable to leave it to #232's sweep —
your call, and I would not hold the PR either way.

### What I did not run

**The Windows suite itself** — everything platform-specific here rests on your
run `33713563677` and the two reverted CI mutations, not on my hardware. I drove
`scripts/windows-triage.sh` with synthetic Windows-shaped results on macOS,
which is the seam your law was built to allow, and I re-derived the CRLF
behaviour of `jobBlock` in node rather than on a runner.

PR #253, opened 2026-09-03
prd25 w4: a GREEN Windows triage proves the suite ran, and the law pins the job's triggers

### LINKED ISSUE #252: prd25 w4: a GREEN Windows triage proves the suite ran, and the law pins the job's triggers
## Fence (may touch ONLY)
- `scripts/windows-triage.sh`
- `packages/server/src/windows-suite-law.test.ts`
- `.github/workflows/windows-suite.yml` (only if the trigger assertion needs a comment beside `on:`; the triggers themselves do not change)

## Blocked by
Wave 3 — #212, merged in #251 (`a76c4c9`). Both findings below are from that issue's verify pass and concern the triage's **own integrity**, not any Windows failure — which is why they are filed here rather than left to the per-class fix lanes (ruling 3's grooming), which would never own them.

## Why

prd-25 ruling 2: a failure outside the list turns the leg red. Two ways the job can read GREEN while that promise is not being kept, both **EXECUTED** by #212's verifier and acknowledged by the author as outside that issue's Definition of done:

1. **A GREEN verdict does not assert the suite evaluated every test file.** A vitest results file holding only the 29 listed files (29 of 405 tracked test files) yields `0 unexpected · 29 expected · 0 removal candidates · 0 listed-not-run · verdict GREEN`, exit 0. The script's "listed but not run" arm protects the entries; nothing protects the *unlisted* files. A vitest crash after writing partial JSON, a `--include` narrowing, or a filter flag on the suite step would all pass silently. `numTotalTestSuites` counts describes, not files, so it is not the floor — the floor is the tracked test-file set (`git ls-files 'packages/*.test.ts' 'packages/*.test.tsx'`) or, at minimum, a `testResults.length` compared against it.
2. **Nothing asserts the workflow's `on:` triggers.** `windows-suite-law.test.ts` §C asserts `runs-on`, `fetch-depth`, `timeout-minutes`, the suite step's `continue-on-error`, and the triage command — but not that the job runs on `pull_request` and on `push` to `main`. The README row and AGENTS.md both say "on every push"; a trigger edit would make that false with no law going red. `route-class-law.test.ts` holds `ci.yml`'s matrix to the README's claim; this is the same law for the third job.

## Definition of done

- `scripts/windows-triage.sh` is RED when the results file names fewer test files than the tracked set (`packages/**/*.test.ts(x)` via `git ls-files`), printing each tracked test file that has no result — a fifth category alongside the four it has, and RED like `listed but not run`. A file the suite could not place stays its own red.
- `windows-suite-law.test.ts` §B gains the case: results for only the listed files → exit 1 with the missing files named; the happy case gains the full tracked set as passing results so it still proves GREEN is reachable (or the script accepts an explicit tracked-set input the law can control — the plan decides which, and says why).
- `windows-suite-law.test.ts` §C asserts `on:` carries `pull_request` and `push` to `branches: [main]`, in the same parsed-not-listed style the file already uses for steps.
- The list's 29 entries, the suite step and its `continue-on-error`, and `ci.yml` are untouched. *(Revised at planning time, 2026-09-03: the original read "the workflow's steps"; the plan adds one derivation step for the tracked set, and says why the in-script alternative lost.)*

## Sibling case

The two findings are the same shape as the `listed but not run` arm #212 already built: the list may not claim a file the suite did not evaluate — and the *suite* may not claim a verdict over files it did not evaluate. The sibling to watch while building: a test file that is legitimately absent from results because vitest **skipped the whole file** still reports `status: "passed"` in the JSON reporter (verified in #212), so it is present, not missing — the new arm must key on absence from `testResults`, not on status.

Delete half the `testResults` entries from a real results file → must go RED naming the missing files; a count-only floor (`testResults.length >= N`) survives a swap of one missing file for one extra unplaceable one, which is why the DoD says *names each*. Remove `pull_request:` from `on:` → §C must go red; a `toContain('pull_request')` on the whole file survives the trigger being moved into a comment, which is why the DoD says parsed.

ciaran-slow:
Closes #252

prd-25 wave 4 — one issue, one commit. prd-25 ruling 2: a failure **outside** the expected-fail list turns the leg red — a promise a verdict over files the suite never evaluated cannot keep. Built on lane `252-triage-coverage-floor` (tip `9b76ca3`), verified on a different model with every finding answered; plan, build report, verify verdict and the author's answers are on #252.

## What becomes true

- **A GREEN Windows triage proves the suite ran.** `scripts/windows-triage.sh` takes the tracked test-file set as a third argument and holds the verdict to it: a tracked test file with no result is a fifth category, `RED: tracked test file with no result: <file>`, named per file and never counted. A listed file absent from results stays named once, as `listed but not run`. Summary line gains `· <n> unevaluated` before the verdict; the job-page summary gains the section.
- **The set is derived, never maintained.** `.github/workflows/windows-suite.yml` gains one step, `git ls-files 'packages/*.test.ts' 'packages/*.test.tsx' > windows-suite.tracked`, hands it to the triage, and uploads it beside the results. Triggers unchanged. Why an explicit argument and not an in-script `git ls-files`: the law drives the script with two- and three-file synthetic results; an in-script call would force an opt-out mode that production never runs.
- **The law pins the job's triggers.** `windows-suite-law.test.ts` §C parses the top-level `on:` block and asserts `pull_request:` and `push:`/`branches: [main]` at exact indentation, and that no `paths` / `paths-ignore` / `branches-ignore` / `types` filter narrows either — the "every push" the README and AGENTS.md claim had no law behind it.
- 36 law cases (was 27), including GREEN at the real 405-file size and the real set with three results deleted naming exactly those three.

**Repo reality the floor rests on:** 405 tracked test files; three independent real Windows results files (both #212 runs and the latest `main` run, 33715417823) hold exactly those 405, 0 missing, 0 extra; no package vitest config carries `include`/`exclude`.

## What is the sibling case?

The issue's: a wholly skipped file is `passed` in vitest's JSON — present, not missing — so the arm keys on absence from `testResults`, never on status; the real-size case proves placement and floor agree on all 405 paths, 60 of them `.tsx`. Also handled: a listed file absent from results is named once, not twice (asserted); a passing result outside the tracked set is deliberately not a category; a trigger moved into a `#` comment at the same indent, or narrowed by a `paths:` filter while both keys stay present (verify's finding, folded in), both redden the trigger law. Found while building: the existing "no `/packages/` segment" case derived its floor from the unplaceable name itself, which the new set validation rejects first — it now carries an explicit valid floor so it still tests placement.

**EXECUTED by the builder, re-executed by verify, each restored:** the per-file arm replaced by a count floor → five cases red, including "names exactly those three" and "named once, never twice"; the `!listed.has(file)` guard dropped → two red; `unevaluated` dropped from the verdict → three red; `pull_request:` deleted → §C red; moved into a same-indent comment → §C red; `paths: ["packages/**"]` added under `push:` → §C red. **Against real inputs:** empty results + real list + real set → `29 listed-not-run · 376 unevaluated · RED`; `main`'s script on a results file holding only the 29 listed files → GREEN (the #212 vacuity), this branch's → 376 named lines, RED; the latest `main` run's artifact → `0 unevaluated · GREEN`.

## The Windows witness

No CI ran on the lane: both workflows trigger on `pull_request` and pushes to `main`, and the plan chose no draft PR. **This PR is the first runner run with the floor.** Its `Windows suite` triage line must read `0 unevaluated`; verify judged the local evidence sufficient to open without it, because every way the derivation step can fail is loud (empty set → RED; wrong pathspec → 405 named lines → RED) — there is no configuration in which a broken floor reads GREEN. Heads-up: `packages/server/src/lab/compare.test.ts` failed once in #212's runs and passed in the six since; if this PR's Windows job goes red naming it, that is the pre-existing flake, not #252.

## Deviations from the plan, and fence

- Plan §7's "the job runs on the PR" and "no draft PR is needed" were only jointly true once this PR exists; the second was followed.
- One existing law case given an explicit floor (above).
- **Fence:** the three files the issue names; the workflow's use widened from "a comment beside `on:`" to one step and one argument, recorded in the plan before the change, and the DoD bullet revised on the issue body at planning time. No CHANGELOG entry: not user-visible, outside the fence. No ADR: no dependency, event, route or record shape.

## Gates on the assembled wave branch

`VITEST_MAX_WORKERS=6 npm test` (405 files / 7311 passed), `npm run typecheck`, `npm run lint`, `npm run build` — all green, node 22.23.2. Boot smoke skipped deliberately: no server or CLI source in the diff (run on the lane during build, passed).


ciaran-slow:
The red `windows-suite` run (33718912542) was this lane's own new trigger assertion anchoring on a bare `\n` against a CRLF checkout — the floor itself read `0 unevaluated · 29 expected`. Fixed in the lane (`e7d23e5`, a CRLF rendering of the workflow is now a law case) and the wave branch re-cut on current `main` at `226f444`; details on #252.

PR #260, opened 2026-09-03
prd43 w6: the laws are as wide as the rules they enforce, and every count they state is derived

### LINKED ISSUE #225: prd43 w6: docs/demo.md and the shell agree about Esc
## Fence (may touch ONLY)
- `docs/demo.md`

## Blocked by
Nothing. No open issue claims `docs/demo.md`. #215 (merged) corrected `AGENTS.md`, not this file.

Sibling lanes: `packages/web/src/app/PanelGrid.tsx` is **not** claimed here. If the answer turns out to be "the shell should handle Esc", that is a separate shell issue with its own boundary — this one decides which of the two texts is right and fixes the document.

**Model:** sonnet. **Wave:** none — unwaved, like #62/#63/#66.

## The defect

`docs/demo.md` is the glance protocol — the script an operator follows with a lay viewer. Its Setup section states shell-level key precedence:

> **Esc** closes the peek if one is open, otherwise exits panel focus if a panel is focused, otherwise does nothing (shell-level precedence: peek first, then focus).

`packages/web/src/app/PanelGrid.tsx:240` states the opposite as a rule:

> **`Escape` is not handled and must not be.** S3 says so and the reason is … so Escape keeps bubbling untouched.

**EXECUTED, 2026-09-02, during prd-33's wave-0 run** (`docs/design/glance-2026-09-02.md`, observation 1): with fixture `3` loaded, click `PAUSE MOTION` (inside the scene), then press `Esc`, then `1`. The scene zoomed to fit. It did not switch source to `live`. `Esc` released nothing, exactly as `PanelGrid.tsx` says it will not — and exactly as `demo.md` says it will. The camera section of the same document names this key collision as a failure shape, so the protocol contradicts itself as well as the code.

What still handles `Escape`, for whoever resolves this: `app/Welcome.tsx:42` (dismiss), `fleet/selection.tsx:199` (clear selection — the peek), `app/panelPrefs.ts:292`. None of them releases scene focus.

## Definition of done

- The Setup section's `Esc` sentence describes what the shell does: peek/selection is cleared, the welcome card is dismissed, and panel focus is **not** released — a click elsewhere or Tab is how focus leaves the scene. Say that plainly, because a demo-runner who trusts the current sentence will press `1` and re-fit the camera instead of returning to live, in front of a viewer.
- The camera section's "what failure looks like" for `1` is checked against the corrected precedence so the two sections agree with each other.
- If, on reading `PanelGrid.tsx:240`'s reasoning, the conclusion is that the *shell* is wrong, do not edit the shell here — file it, cite S3, and leave the doc describing current behaviour.

## Sibling case

`demo.md` states key behaviour in three places — the Setup table, the camera section, and "Orientation extras". Fix the sentence that was caught and re-read the other two against the code in the same pass; a protocol document that is right in one section and wrong in another is worse than one that is consistently wrong, because the reader trusts it.

None applies — this is prose against code, and prd-43's document laws do not read key-precedence claims. Say so in the commit rather than inventing a law for one sentence.

### LINKED ISSUE #226: prd43 w6: a tracked screenshot is bound to the tree it depicts

### LINKED ISSUE #231: prd43 w6: a separator alone makes a remedy token a path claim

### LINKED ISSUE #232: prd43 w6: every route count this repo states in prose is swept by one law

gabriel-canaan:
Closes #225
Closes #231
Closes #232
Closes #226

prd-43 wave 6 — the enforcement residue. Waves 1–5 fixed the documents rulings 1
and 2 name; reviewing that work produced a second class, and this is it: places
the **laws themselves** are narrower than the rule they enforce, and documents
the sweep never reached. Four issues, four commits, fence-disjoint and proven so
by `scripts/fence-lint.sh` rather than asserted.

|---|---|
| `#225` | `docs/demo.md`'s Esc sentence and the shell agree on what focus it releases |
| `#231` | the remedy law checks what its own docblock says it checks |
| `#232` | every route count this repo states in prose is derived by one law |
| `#226` | every tracked screenshot is bound to the tree it depicts |

## What is the sibling case?

Asked four times, and it changed the diff every time — twice on the repairs
themselves, which is the part worth reading.

- **`#226`'s first repair had the defect it fixed.** The finding was that
  `drawer.png` and `main-drawer.png` were pinned to the drawer package while
  being ~60% scene by area, so the exact 84-commit scene drift that filed the
  issue passed on two of ten shots. The repair added the scene path to those two
  and stopped — closing the case that was *named* without asking which other
  entries shared it. `replay.png` did: pinned to the scene alone while the
  full-width REPLAY banner and "Exit to live" are the replay package's. Found by
  a re-review seat, confirmed by opening the PNG.
- **And the durable sibling was the map's failure mode, not any entry in it.**
  `commitsBehind` hands its strings to `git rev-list -- <paths>`, and git
  returns 0 for a path matching nothing — so a typo does not error, it silently
  stops counting. A new assertion requires every depicted path to be a directory
  `git ls-files` returns something for.
- **`#232` claimed a repo-wide sweep while its own file held two stale counts.**
  Both older than the commit by blame, both inside the claim its subject makes.
- **`#225` replaced one absolute with a narrower one that was still false.**
  "That's all it does" became "does not hand keyboard focus back" — contradicted
  by the disclosure card, which says verbatim that Escape hands focus back to its
  mark. Now scoped to the scene's focus, which is the load-bearing claim and is
  true: nothing in the web package calls `.blur()` at all.

Every assertion below was reddened deliberately, and each control is a spelling
the round that found it did not use.

- **`#226` freshness, the union path.** Setting `drawer.png`'s `capturedAt` to
  the original `3bd1c5a` now fails with *"110 commits behind HEAD on
  packages/web/src/drawer/, packages/web/src/scene/"*. The identical mutation
  was **green** before the repair, because the drawer-only count was 29 and sat
  under the 40 bound.
- **`#226` path validation.** Typing `replay/` as `replya/` reddens the new
  assertion. Before it, the same class of typo turned a red stale-capture
  assertion green at 39/39.
- **`#232` per row.** Changing `docs/adr/0008`'s "six" to "five" reddens exactly
  one row; appending a *second*, wrong occurrence to `docs/architecture.md` while
  leaving the correct first one also reddens — so the law finds every stated
  count in a file, not the first.
- **`#231` both repairs.** Reverting `isPathShaped` to extension-only reddens the
  extensionless case; reverting `tokenResolves` to `path.join` reddens the
  absolute-token escape.
- **The NUL byte, which no suite catches.** `#232` shipped a literal `0x00` in a
  cache key. Typecheck, lint and all 7,338 tests pass over it, and `git diff`
  renders the file as text because the byte sits past git's 8000-byte sniff
  window — so CI would have been green. `scripts/gate.sh`'s NUL guard, run
  verbatim, failed with *"contains 1 NUL byte(s)"*. Replaced with the `\0`
  escape; the guard now reports zero.

## Fence widenings

**None.** Each commit touched only its issue's declared fence; `#232` used three
of its four paths and left `docs/architecture.md` untouched because its counts
were already correct.

## Gates

Run on the assembled branch at this base, exit statuses read directly:

```
scripts/gate.sh NUL guard (extracted, run verbatim)   0 files trip it
npm run typecheck                                     exit 0
npm run lint                                          exit 0
VITEST_MAX_WORKERS=6 npm test                         exit 0
  405 files · 7338 passed · 3 expected fail · 2 skipped
```

The base moved four times while this wave was in flight (prd-25 w3, prd-43 w1's
`#203`, prd-25 w4, prd-48, and prd-51's blessing). Reassembled and re-gated on
each; no fence overlap with any of them.

## Verification

Five independent passes before this PR existed — two initial seats (Opus,
Codex), two cross-reviews, one reconcile — then two more against the amended head
for the repairs. Four must-fix findings were applied; the fix re-review found two
more, which are applied too.

One correction worth recording, because the pipeline made it: two passes derived
`#232`'s corrected count as 25 − 3 otel routes = 22. `api/otel.ts` declares
**four** — the three signal-specific routes plus the bare-path fallback ADR-0018
added — and those are exactly `ROUTE_CLASSES`' four `ungated-mutation` rows. The
answer is 21, and "three" was itself stale in the same sentence. Both numbers now
carry their derivation in the comment, which is this wave's own ruling applied to
itself.

The round budget then **refused a third round** on same-class churn, and that
refusal is honoured rather than overridden. Four findings are deliberately left,
consolidated from an initial list of thirteen: the remedy-token grammar
enumerated as one table (absorbing six spellings of one production), `capturedAt`
validation, the `CLAIMS` completeness oracle, and capture integrity. Each is
recorded for its own issue rather than smuggled in here.


ciaran-slow (changes requested):
EXECUTED in the PR head's clean review clone:\n\n- All 10 expected CI legs ran and passed: build-test-boot 3/3, pack-smoke 6/6, windows-suite 1/1.\n- The focused remedy, route-class, and screenshot-law tests pass: 3 files, 145 tests.\n- scripts/fence-lint.sh 225 231 232 226 passes with no overlap.\n\n[P1] Make the sweep discover unregistered route-count claims — packages/server/src/api/route-class-law.test.ts:340\n\nEXECUTED mutation: I added the exact already-recognized sentence “prd-29 puts the same token on thirteen reads” to packages/server/src/api/test-support.ts, then ran npx vitest run packages/server/src/api/route-class-law.test.ts. All 84 tests passed. CLAIMS enumerates both the files and exact phrases, so the walk at lines 440–445 can only validate rows somebody already knew to add; it cannot detect the next count in a third file. That reproduces the omission shape issue #232 was filed for and leaves its “wherever stated” / “a thirteenth site” definition unfulfilled. The PR body also records the missing CLAIMS completeness oracle as deliberately deferred, so this is a known gap in the closing claim rather than a speculative future concern. Please add a completeness mechanism that makes the same mutation red, or narrow #232/this PR's claim before landing.

gabriel-canaan:
**Fixed in `e218907` (`#232`'s commit is now `9f020e2`). Your finding was right, and I reproduced it before changing anything.**

EXECUTED, on the previous head `e95ce21`: appending your exact sentence `// prd-29 puts the same token on thirteen reads` to `packages/server/src/api/test-support.ts` left the law **84/84 green**. A false route count sat in a swept package with nothing red. Your diagnosis of why is exactly right — `CLAIMS` enumerated both the files and the phrases, so the walk could only validate rows somebody already knew to add, and that reproduces the omission shape `#232` was filed for one level up from where `#23` had it.

You offered a completeness mechanism *or* a narrowed claim. I did both, because they turned out to be halves of one answer.

**1. Every registered pattern is now applied to every swept file**, not only to the file its row names. A recognised claim shape occurring anywhere no row declares it fails, naming the file and the pattern so the remedy is to add a row rather than to guess. The swept set is tracked `.md`/`.ts`/`.tsx`/`.mjs`/`.js`, and a companion test asserts that set is non-empty, so a sweep matching nothing cannot pass vacuously.

**2. The claim is narrowed to what the code actually checks.** The old subject and docblock said "every route count this repo states in prose", and that is an open set no matcher closes: a count can be phrased in a sentence nobody has written yet, or spelled in digits `wordToNumber` structurally cannot read. Chasing it needs a round per counter-example. So the vocabulary IS the claim now — `CLAIMS` is the whole of what the law recognises, and the subject reads *"every recognised route-count claim is swept, in whatever file it occurs"*.

The boundary is written out beside the completeness test rather than left to be found: a novel phrasing, a digit-spelled count, this file itself (it carries every pattern as a regex literal, so sweeping itself would match its own source), and the dated-artefact directories — which now name `docs/prds/` alongside `docs/adr/` and `docs/review/`, since a verify pass found that omission silently load-bearing.

**Mutations, EXECUTED. The second is deliberately not the one that found the defect:**

```
(a) your input — "puts the same token on thirteen reads" -> api/test-support.ts
    RED: "packages/server/src/api/test-support.ts matches the pattern
          registered for security.ts gated reads"
    Certified through the precommit gate's --prove, which refuses to certify a
    mutation that passed. Same input was 84/84 green before this change.

(b) a different file kind and a different pattern — "This server has nine
    mutating routes today." -> docs/user-guide/watching.md
    RED: "docs/user-guide/watching.md matches the pattern registered for
          mutation-guard.ts total"
```

Clean tree: **86 tests** in that file, no false positives across the swept set. Whole branch re-gated on current `main`: typecheck 0, lint 0, `VITEST_MAX_WORKERS=6 npm test` 0 — **405 files, 7340 passed, 3 expected fail, 2 skipped** — and the extracted `scripts/gate.sh` NUL guard reports 0 files.

One thing worth flagging rather than burying: you were right that the PR body recorded the completeness oracle as deliberately deferred. That deferral was mine, and it was the wrong call — I deferred the one thing that made the subject line true, then shipped the subject line. The narrowed claim is the part of this fix I'd have wanted on the first pass.

ciaran-slow (approved):
EXECUTED rereview at author head e218907 and review fix b1720a0.\n\nThe author's repair closes the original finding: the exact unregistered one-line claim now reddens with the expected test-support.ts diagnostic. The new completeness pass also has a clear, honest boundary around its recognised vocabulary.\n\nI found and fixed two mechanical sibling cases in b1720a0: hard-wrapped block-comment prose is now normalized the same way as the primary matcher, and declared keys use git ls-files' platform-independent slash spelling rather than path.join's Windows separators. Both have dedicated regression tests; the hard-wrapped mutation now reddens.\n\nVerified after the fix: npm run typecheck passes, npm run lint passes, and VITEST_MAX_WORKERS=6 npm test passes (405 files, 7338 passed, 3 expected fail, 6 skipped).

PR #263, opened 2026-09-03
prd43 w1: the citation law passes 20/20 under 4x concurrency, and its own test can fail

### LINKED ISSUE #241: prd43 w1: the citation law passes 20/20 under 4x concurrency

gabriel-canaan:
Closes #241

`packages/server/src/doc-citation-law.test.ts` was flaky under concurrency, and
the failing test was the one written to prove the fixture cleanup is *safe* —
the worst place for a flake, because it discredits the guard it exists to
justify. `scripts/gate.sh`'s load gate runs the suite 4x concurrently and is
mandatory for anything touching tests, so this reddened other people's landings
and blamed their branches.

Three defects, one file, one commit. The first is the issue; the other two came
out of verification before this PR existed.

## What was wrong

**The named race.** The TRACKED-guard test wrote a real, dead-pid-named,
content-matching, genuinely untracked fixture into the shared `docs/`, then
injected an `isTracked` to prove `cleanUpOrphanedFixtures` would not delete it.
The injection protects only the call that test makes. A sibling concurrent run's
cleanup uses the default git-backed `isTracked` and correctly deletes the file by
the guard's own rules — name matches, pid dead, content matches, git reports it
untracked.

Fixed by giving the sweep a `docsDir` parameter (defaulted to the real `docs/`,
so every other call site is unchanged) and pointing that one test at a fresh
`mkdtemp`. Nothing else sweeps that directory, so the race is structurally
impossible rather than merely narrowed.

**The sibling case the first enumeration missed.** At 8x the pre-fix tree failed
FOUR legs, not two. The other two were `ENOENT ... at readSweptFile ->
allCitations`: `existsSync(p) ? readFileSync(p) : undefined` is a check-then-use,
not a guard. A sibling lists the path, passes the existence check, and the
*owning* run's `finally` removes the fixture before the sibling's read lands.

The live-pid verdicts that cleared two other write sites do not cover it, and the
asymmetry is the finding — they answer *"can a sibling's cleanup delete my
fixture"*, this is *"can my fixture's deletion break a sibling"*. The pid on the
failing path belongs to the sibling and is alive, so the alive check never fires.
Fixed by catching ENOENT, which is what that function's own docblock always said
it was for.

**The fix had made its own test non-load-bearing.** Deleting one token — the
`docsDir` argument — left the file 31/31 GREEN, because the sweep then ran over
the real `docs/`, never listed the tmpdir fixture, and the survival assertion
passed since nothing had looked at it. Before the fix, deleting the injected
`isTracked` reddened it. Isolation removed the race and took the test's
self-enforcement with it.

Fixed with a positive control in the same tmpdir: a second dead-pid,
byte-identical fixture that `isTracked` reports UNTRACKED, asserted DELETED. The
two differ in exactly one thing, so neither can come out right unless the sweep
actually visited the directory.

## What is the sibling case?

Every write into the shared `docs/` in this file was enumerated with a verdict,
because the issue demanded enumeration rather than a patch of the one that
tripped. Five sites; four into the real `docs/`:

| site | pid | content matches | verdict |
|---|---|---|---|
| `:839` | live | yes | alive check — but see the ENOENT finding |
| `:869` | **dead** | **no** | content check |
| `:921` | dead | yes | **the race** — now a private tmpdir |
| `:935` | live | yes | alive check |
| `:959` | live | yes | alive check — but see the ENOENT finding |

Only `:921` combined all three conditions. And the enumeration's own blind spot
became the second finding: the verdict word *Safe* was scoped to one direction of
the race, and two of the sites it cleared were breaking siblings the other way.

None, and each is a controlled A/B rather than a one-sided green.

```
the guard      remove `if (isTracked(rel)) continue`
               -> RED: "a file reported TRACKED was deleted by a cleanup step
                  meant only for untracked orphans"

the setup      drop the `docsDir` argument
               -> RED: "the control fixture ... reported UNTRACKED ... was NOT
                  deleted, so the sweep never visited tmpDocsDir and the
                  assertion above proved nothing"
               certified through the precommit gate's --prove, which refuses a
               mutation that passed

the ENOENT     identical 5ms window between check and read, identical 4x load:
               old existsSync form -> 7 of 20 legs FAIL
               this commit's catch -> 0 of 20 legs fail
```

## The acceptance criterion was not falsifiable, and is corrected

The issue asked for "20 runs of 4x-concurrent are 20/20". That cannot distinguish
a fix from no fix: **unfixed `main` also passes at 4x on this hardware** —
measured, 24 legs and 64 legs, both clean. The race needs 8x to surface and fires
around 1% per leg. A green run with no control is not evidence, which is this
repo's own *"a test that cannot fail for the reason it claims"* shape applied to
the verification rather than to the test. Every claim above is stated as an A/B
because of it.

Measured across both trees during verification: pre-fix at 8x, 4 failing legs of
160; fixed at 4x/8x/12x, 0 of 440.

## Fence widenings

None. One file, exactly as declared.

## Gates

```
npm run lint                     exit 0
VITEST_MAX_WORKERS=6 npm test    exit 0
  405 files · 7318 passed · 3 expected fail · 2 skipped
```

Verified before this PR existed: one independent seat at xhigh (declared REDUCED
depth — fixture isolation in a single test file, test-local destructive helper
with its default preserved), plus the orchestrator's own executed reconciliation
against the code. The seat found both extra defects; its report is what this PR
is built on.


ciaran-slow (approved):
## Approved. Every mutation in the PR body re-derived independently, and they all hold.

Reviewed at `984576e` in a disposable worktree, node v22.23.2. One comment-only
commit pushed as `0f3d697`. Findings labelled `EXECUTED` (ran it, read the exit
status) or `REASONED` (argued from code, nothing run).

I re-ran every claim rather than reading them, and the interesting ones survive.

**The two mutations do exactly what the body says — EXECUTED.**

```
remove `if (isTracked(rel)) continue`
  -> 1 failed | 30 passed
     "a file reported TRACKED was deleted by a cleanup step meant only for
      untracked orphans"

drop the `docsDir` argument at the call site
  -> 1 failed | 30 passed
     "the control fixture — same directory, same dead pid, same bytes,
      reported UNTRACKED — was NOT deleted, so the sweep never visited
      tmpDocsDir and the assertion above proved nothing"
```

**And the positive control really is load-bearing — EXECUTED, third mutation.**
This is the claim I most expected to fail, because it is the one that would be
easiest to assert without checking. Dropping `docsDir` *and* deleting the control
assertion: **31/31 green.** So the self-criticism in the body is precisely right —
isolation had made the setup silently optional, and the control is the only thing
that reddens it. Catching that about your own fix, after it was already working,
is the part of this PR worth copying.

**The ENOENT asymmetry is real and I reproduced the direction — EXECUTED.** Both
ENOENT paths in my run carried a **live** pid, confirming the finding: the alive
check cannot fire, because the live pid belongs to the sibling. Your table's
"but see the ENOENT finding" on `:839` and `:959` is the correct annotation.

**Gates, on the merged-in-my-head tree with my commit — EXECUTED.**
`npm run typecheck` exit 0 · `npm run lint` exit 0 (876 files) ·
`VITEST_MAX_WORKERS=6 npm test` exit 0 — 405 files, 7314 passed, 3 expected fail,
6 skipped. (You reported 7318 passed / 2 skipped; same 7323 total, so four tests
are environment-skipped here rather than missing. Not a defect.)

**Fence honoured — EXECUTED.** One file, exactly as declared. `Closes #241`
present.

### Finding — the only one, and it is about the evidence, not the fix

**"The race needs 8x to surface" does not hold, and the acceptance criterion was
falsifiable after all — EXECUTED. Recorded in `0f3d697` (comment-only).**

`984576e` says unfixed `main` "passes at 4x on this hardware — measured, 24 legs
and 64 legs, both clean", that the race "needs 8x to surface and fires around 1%
per leg", and therefore that the issue's criterion "cannot distinguish a fix from
no fix". I could not reproduce that:

```
for i in 1 2 3 4; do npx vitest run <this file> & done; wait      x5 rounds

  pre-fix tree (origin/main)   6 failing legs of 20   <- 30%, not 1%
  this branch                  0 failing legs of 20
```

All six were the TRACKED-guard assertion, verbatim. Same harness, same machine,
back to back.

I think the discrepancy is a genuine ambiguity in "4x" rather than a bad
measurement on your side: four concurrent **whole-suite** runs (the gate's load
pass) stagger this file against everything else, so the window is thin and ~1% is
right; four simultaneous runs of **this file alone** overlap maximally. Both are
fair readings of the issue's wording.

The consequence is in your favour and is why I bothered: **the criterion is
falsifiable under the cheaper reading, and this branch passes it against a
pre-fix tree that visibly fails it.** So issue Done-when bullet 1 is met on its
own terms, rather than needing to be corrected. I have not touched the PR body —
`0f3d697` records both recipes in the file so the next person does not build an
8x harness they do not need, as I did.

I also gave the ENOENT half a deterministic recipe, since 8x was not reproducing
it for me either. Widening the window beats raising the load — `Atomics.wait(…,
5)` between the check and the read, fixture-named paths only, the two arms
differing in **nothing but the `try`**: uncaught 2 failing legs of 20, your catch
0 of 20.

### Observations — offered, not asserted

- **Done-when bullet 3 is not literally met, and correctly so — REASONED.** It
  asks that no test write a dead-pid-named fixture into the real `docs/`; `:891`
  still writes `citation-law-untracked-fixture-999996-*` there. That is the
  `:869` row of your table and it must stay dead-pid-named, or the cleanup never
  reaches the content check the test exists to prove. The bullet is over-broad as
  written — the hazardous combination is dead pid **and** matching content, which
  `:891` deliberately breaks. Right call; flagging only so the closure comment
  can say so rather than leaving a bullet looking unmet.
- **`:891` is a sibling in the ENOENT direction too — REASONED, not observed.**
  It is written into the real `docs/` and removed in a `finally`, so a sibling's
  `allCitations()` can list it and lose it. It did not fire in my runs (its
  write-to-delete window is much shorter than `:839`/`:959`), and your catch
  absorbs it regardless — so nothing is owed. Worth a row in the table if that
  table is ever revisited.
- The `path.relative(REPO_ROOT, …)` handling is consistent on both sides for the
  macOS `/var` → `/private/var` symlink, since the test and the sweep derive the
  path from the same `docsDir` string. No raw-vs-canonical hazard — REASONED, and
  the macOS CI leg agrees.

### CI — checked, not assumed

`gh pr checks 263`: **10 legs, all pass** — `build-test-boot` ×3, `pack-smoke`
×6, `windows-suite` ×1. Matches AGENTS.md's three-job / ten-leg checklist.
`mergeStateStatus` CLEAN.

Because a green bar is not evidence a gate ran, I opened
`build-test-boot (macos-latest, current)`: Build, Test, Typecheck, Lint,
Packaging guard and Boot smoke all executed, none skipped. CI re-runs on
`0f3d697`.

### Verdict

Approved. `0f3d697` is comment-only — no assertion, no behaviour, not one token
of code — so it does not need your sign-off before merging, though I would rather
you read it since it contradicts a measurement in your commit message.

**Not merging and not landing** — merges on this repo are the operator's,
manually, on GitHub. Ready for that once CI settles on `0f3d697`.

Issue #264, opened 2026-09-03
prd43 w7: ADR-0032 names the writer that actually writes, and the subject it actually rules


gabriel-canaan:
## The defect

ADR-0032's body states:

> `packages/server/src/lab/checkpoint.ts` writes them to `~/.claude/projects/<slug>/`

That is wrong on both counts, and it is the record every downstream document
copies from.

- **The writer is `restore.ts`, not `checkpoint.ts`.** The write is
  `path.join(projectDir, sessionId + '.jsonl')` inside `synthesizeSession`.
  `checkpoint.ts`'s only contact with that tree is `cutSession`, which
  `readFile`s and hashes and never writes to it — its sole `copyFile` is git
  index handling inside `snapshotWorkspace`.
- **The subject is sessions only, not checkpoints.** A checkpoint's artefacts
  are its ref under `refs/rhizomorph/checkpoints/` plus a `fork.checkpoint`
  event appended to the event log under the DATA root. EXECUTED during the
  verify pass on #24: running `captureCheckpoint` against a fixture repo and
  diffing both trees produced **zero** entries under the harness tree.

## Why it matters more than a typo

It already propagated. #24 corrected `README.md`'s Trust section against this
ADR and faithfully inherited the error, putting checkpoint artefacts in the
harness tree — swapping which half of that sentence was false rather than fixing
it. Caught in review before landing, but only because a seat checked the ADR
against the code instead of against the ADR.

`docs/adr/README.md`'s one-line summary of this ADR is correctly scoped and
singular — "a synthesized session lives in the harness's own projects tree" — so
the index is right and the body is wrong, which is the harder direction to
notice.

This is prd-43's own thesis turned on the PRD's own output: a document stating
something its source contradicts, with nothing testing the claim.

## The decision this issue owns

`docs/adr/` is append-only — a landed ADR counts the tree as it stood, and
rewriting one destroys what an ADR is for. So the correction is NOT an in-place
edit, and which shape it takes is the decision here:

- an **amendment note** appended to ADR-0032, dated, naming the two errors and
  the real writer — the lighter option, and the one that keeps the record in one
  place; or
- a **new ADR superseding** it, if the view is that a body wrong about its own
  mechanism cannot be repaired by a footnote.

ADR-0024 amending ADR-0014 is the in-repo precedent for the second shape;
`docs/adr/README.md` states the append-only rule. Pick one and say why in the
diff.

## Definition of done

[...]

PR #265, opened 2026-09-03
prd43 w5: the README states where lab sessions live, and the namespace law asserts it

### LINKED ISSUE #24: prd43 w5: the README states where lab sessions live

gabriel-canaan:
Closes #24

`README.md`'s Trust section said the lab's checkpoint and synthesized-session
artefacts "live beside" the worktrees directory under the data root. ADR-0032
rules that a synthesized session lands in the harness's own
`~/.claude/projects/<slug>/` tree instead — the only spelling Claude Code can
resume from. The README now says where each actually goes and cites the ADR at
the site of the divergence, and the namespace law now asserts the destination
rather than tolerating either.

## The law could not fail for the reason the ADR exists

`isAllowedWrite`'s first clause was `parts[0] === 'data' || parts[0] ===
'claude-projects'` — a blanket OR over two DIFFERENT namespaces, the lab's own
data dir and the ADR's ruled destination. A synthesized session written to the
pre-ADR location satisfied it exactly as well as one written to the ruled
location, so the containment law was blind to the precise regression ADR-0032
was written to prevent.

A synthesized session is identifiable on sight — `synthesizeSession` names it
after nothing but its own uuid, `<uuid>.jsonl`, never the lab's
`session-<ts>.jsonl` event-log spelling — so that shape is now pinned to its one
ruled root before the blanket clause runs.

## What is the sibling case?

Two, and both came out of verification rather than authoring.

**The README fix first swapped which half was false.** The initial edit put "the
checkpoint/synthesized-session artifacts" *into* the harness tree. True of the
session, false of the checkpoint: `checkpoint.ts`'s only contact with that tree
is `cutSession`, which reads and hashes and never writes. A checkpoint's
artefacts are its ref under `refs/rhizomorph/checkpoints/` plus an event-log
entry under the data root — EXECUTED, by running `captureCheckpoint` against a
fixture repo and diffing both trees: **zero** entries under the harness tree. In
the Trust section, which is the one place a wrong location for an operator's own
data does most harm. Now split by subject.

**And the pin encoded a universal the codebase contradicts.** "Every
`<uuid>.jsonl` outside `claude-projects` is illegitimate" is false:
`transcriptCaptureFileName` (`log/paths.ts`) names a captured Claude Code
transcript by that harness's session uuid, and `capturedTranscriptPath` puts it
under `<dataRoot>/<repoSlug>/transcripts/<recordingSessionId>/` — a shipping
write under `data`, matching the pinned shape byte for byte. It lives in `log/`,
not `lab/`, which is why a lab-scoped writer audit missed it. Green today only
because nothing in the fixture rotates a recorder; it would have reddened later
on somebody else's branch, reading as a namespace violation when it was a correct
write. Carved out by path segment, with an assertion so the carve-out cannot rot.

The first version of this commit answered with a **circular** proof — it reverted
the new branch and watched the *new test* redden. That test asserts the branch
directly: it is the oracle's self-test, and no change to production source can
redden it. Replaced with a proof that has a control.

```
Regress PRODUCTION to ADR-0032's retired option B — restore.ts's projectsRoot
pointed beside the worktrees dir instead of into the harness tree:

  with the new clause      RED — the LIVE test names all three escaped arms
                           data/lab/<slug>-arm-1/<uuid>.jsonl, -arm-2, -arm-3
  clause removed, blanket  PASSES — the identical regression is INVISIBLE
  data/claude-projects
  clause kept

Drop the transcripts carve-out:
  RED — "a captured Claude Code transcript is a legitimate uuid-named write
         under the data root"
```

So "the law could not fail for the reason ADR-0032 exists" was true, and now it
can. That is what makes this worth landing, and it is not what the first proof
showed.

A note on that control, because it is easy to get wrong and I did on the first
attempt: the blanket clause sits **between** the new clause and the `repo` check.
Cutting from the new clause down to the `repo` check removes the blanket one too,
and the live test then fails because every `data` write is refused — a red for an
entirely different reason that reads exactly like a confirmed control.

## Fence widenings

**Widened before the change, recorded on the issue with its reason.** The issue's
Definition of done always required a law to assert the destination, and its fence
named `README.md` alone — so the brief was unbuildable as written: a lane had to
either break its boundary or fail its acceptance, and the landing gate would have
rejected work that was correct. Fence is now the two files this diff touches.

Also corrected on the issue while grooming, since all three had rotted: the
README anchor (`:244-248` is now about session labels; the real site is `:277`),
the law anchor (`:516` asserts git refs; the real site is `isAllowedWrite`), and
"the ONLY issue in wave 5", which stopped being true when #235 joined it.

## Out of scope, filed separately

**#264** — ADR-0032's own body says `checkpoint.ts` writes to the harness tree.
Wrong on both counts: the writer is `restore.ts`'s `synthesizeSession`, and the
subject is sessions only. The README error above was inherited from it. The ADR
log is append-only so the correction is its own record, and its shape is a
decision that issue owns.

## Gates

```
npm run lint                     exit 0
VITEST_MAX_WORKERS=6 npm test    exit 0
  405 files · 7343 passed · 3 expected fail · 2 skipped
```

Rebased onto `ca7370d` after wave 6 landed, and re-gated there rather than on the
base it was built against.

Verified before this PR existed: full two-seat pipeline at xhigh, because this
changes an allow/deny list. The seats **split** — one returned "ready, no
findings", the other "not ready" — and the code settled both disagreements
against the first. The "ready" seat had checked the README against ADR-0032 and
found it consistent, which it is; the ADR is where the error came from. Its own
"no legitimate write is newly forbidden" was labelled REASONED and scoped to lab
writers, and the counterexample is in `log/`.


ciaran-slow (approved):
## Approved. The non-circular control holds, and the carve-out is the best thing in this PR.

Reviewed at `8798625` in a disposable worktree, node v22.23.2. One re-wrap commit
pushed as `edfe404`. Findings labelled `EXECUTED` (ran it, read the exit status)
or `REASONED` (argued from code, nothing run).

**The non-circular proof is real — EXECUTED, and this is the claim I set out to
break.** I regressed *production*, not the oracle: `restore.ts:236`'s
`projectsRoot` repointed to `path.join(forkWorktreePath, '..', '..')`, ADR-0032's
retired option B. The **live** test — `writes ONLY into ruling 1's namespaces`,
the one fed by a real `forkThreeArms()` rather than by hand-written paths —
reddened naming all three escaped arms:

```
data/lab/<slug>-arm-1/55516c74-….jsonl
data/lab/<slug>-arm-2/623c6b78-….jsonl
data/lab/<slug>-arm-3/65f66ebb-….jsonl
```

Then the control, with that same regression still in place and **only** the new
clause cut (the blanket `data || claude-projects` clause deliberately left, per
your warning — which is a good warning, and I would have tripped it):

```
new clause present   -> live test RED, three arms named
new clause removed   -> live test PASSES
```

So "the law could not fail for the reason ADR-0032 exists" is exactly right, and
now it can. Your first proof was circular and you said so and replaced it; the
replacement survives an independent re-run.

**The transcripts carve-out is the strongest finding in the PR — EXECUTED and
REASONED.** Mutation: dropping `&& !parts.includes('transcripts')` reddens with
`a captured Claude Code transcript is a legitimate uuid-named write under the
data root`. And the counterexample is real in the source, not hypothetical:
`transcriptCaptureDir` (`log/paths.ts:99`) returns
`<sessionDir>/transcripts/<sessionId>`, `transcriptCaptureFileName` (`:104`)
returns `${claudeSessionId}.jsonl`, and `capturedTranscriptPath`
(`log/transcript-attribution.ts:274-276`) joins them. That is a shipping write
under the data root matching the pinned shape byte for byte. Catching a universal
your own new clause would have encoded — before it reddened on somebody else's
branch — is the part of this worth copying.

**The README split is correct — REASONED.** `checkpoint.ts`'s only contact with
the harness tree is `cutSession` (`:175-185`), which does `readdir`, `readFile`
and `createHash` and nothing else; the file's only mutating fs calls anywhere are
`copyFile`/`rm` on a temp index. So checkpoint artefacts are not in
`~/.claude/projects` and the session is — as the corrected sentence now says. I
did not re-run your `captureCheckpoint`-against-a-fixture diff; I read the write
surface instead.

**Fence and hygiene — EXECUTED.** Two files, exactly the widened fence, widened
on the issue before the change with its reason. `Closes #24`. ADR-0032's file
exists at the linked path. `mergeStateStatus` CLEAN.

**No live fence overlap on `README.md` — EXECUTED.** `#234` (wave 6) also claims
it, but it sits Soon / Low / **Backlog** with no branch and no worktree, so
nothing is in flight against you.

**Gates re-run here, including my commit — EXECUTED.** `npm run typecheck` exit 0
· `npm run lint` exit 0 · `VITEST_MAX_WORKERS=6 npm test` exit 0 — 405 files,
7339 passed, 3 expected fail, 6 skipped.

### Findings

**1. The regression was not actually invisible to the suite — EXECUTED. Nothing
to change; worth knowing.**

In *both* arms of my control, a second test failed as well:
`writes the synthesized sessions ONLY under the arms' own project slugs, never
the parent's`. It pre-exists this PR, it reads `readdirSync(claudeProjectsRoot)`,
and it reddens on the option-B regression whether or not your new clause is
present.

This does not weaken the change — your claim is scoped to *the containment law*
and is literally accurate, and the law is the piece that generalises to any
escaped write rather than to this one destination. But a reader of "the law was
blind to the precise regression ADR-0032 was written to prevent" could reasonably
infer the regression would have shipped, and it would not have: a sibling test
three `it`s down catches this particular one. Stating it so the record is exact.

**2. Two orphan lines in the README paragraph — EXECUTED. Fixed in `edfe404`.**

The ADR clause was spliced into a paragraph wrapped at ~76 without re-wrapping,
leaving `can resume` alone on line 280 and `It never` alone on line 282, with the
unbreakable ADR link between them. `git diff --word-diff` on my commit reports
**no word changed** — only line breaks moved. The link line stays long because it
is a single URL.

The corrected sentence is now a three-item list whose third item carries a
14-word em-dash interpolation before its own noun arrives (`and — not beside the
worktrees directory, but into the harness's own … tree — the synthesized
session`). It is accurate and I have not touched the words, but this is the Trust
section, where a stranger is deciding whether to believe the paragraph. Splitting
the synthesized session into its own sentence would read easier at zero cost to
the fact. Entirely yours.

### CI — checked, not assumed

`gh pr checks 265`: **10 legs, all pass** — `build-test-boot` ×3, `pack-smoke`
×6, `windows-suite` ×1, matching AGENTS.md's three-job / ten-leg checklist.
Opened `build-test-boot (macos-latest, current)`: Build, Test, Typecheck, Lint,
Packaging guard and Boot smoke all executed, none skipped. CI re-runs on
`edfe404`.

### Verdict

Approved. `edfe404` is a re-wrap with no word changed, so it needs no sign-off
from you.

**Not merging and not landing** — merges on this repo are the operator's,
manually, on GitHub. Ready for that once CI settles on `edfe404`.

### Left for the operator

`#264` (ADR-0032's own body naming the wrong writer) is correctly filed out of
scope and is not a blocker here.

Issue #266, opened 2026-09-04
prd43 w6: the route-count completeness sweep fits its timeout on the slowest runner


gabriel-canaan:
## The defect

`no swept file states a recognised route-count claim that no CLAIMS row declares`
— the completeness sweep added by #232 — takes **7490 ms** on the `macos-latest`
runner against vitest's 5000 ms default, and fails the leg with
`Error: Test timed out in 5000ms`. It is not an assertion failure: nothing is
wrong with the tree it is sweeping.

EXECUTED, PR #263's `build-test-boot (macos-latest, current)`, run
`33823334142`. Measured locally on the same commit: **4.73 s of test time** —
under the budget by about 270 ms. So it has been marginal since it landed in
`ca7370d` and passes ubuntu and this machine by a hair; macOS is slower and does
not.

## Why it is slow

`containsClaim` normalises the WHOLE FILE on every call, and it is called once
per (file, claim) pair:

    const normalized = text.replace(/\s+/g, ' ')
    const gutterless = text.replace(/^\s*\*\s?/gm, ' ').replace(/\s+/g, ' ')
    return new RegExp(...).test(normalized) || new RegExp(...).test(gutterless)

1109 tracked files matching the swept extensions, 15 `CLAIMS` rows: **16 635
calls**, each doing three whole-file regex replaces and two `RegExp`
constructions. The normalisation depends only on the file, not on the claim, so
roughly 50 000 whole-file replaces are performed where 3 300 would do.

## Why it matters beyond one red leg

It is on `main`, so it reddens the macOS leg of **any** PR whose runner is having
a slow minute, and it does so with a message that points at the PR's own diff
rather than at the sweep. #263 — an approved, unrelated branch — is the first to
catch it. A test that fails on other people's branches for reasons that are not
theirs is the exact cost this repo's load gate exists to prevent, arriving from
the other direction.

It is also, precisely, the class the sweep was written to enforce: a claim that
holds on the machine it was measured on and not in general.

## Definition of done

- The sweep's normalisation happens once per file, not once per (file, claim).
  Each claim's probe is compiled once rather than per file.
- Behaviour is unchanged: the same set of files is swept, the same claims are
  applied, and the wrapped-block-comment case the sweep already pins still
  passes. This is a performance fix, not a scope change.
- An explicit timeout is set on the test rather than relying on the 5000 ms
  default, with the measured figure in a comment — the sweep's cost grows with
  the repo, so the next reader should see what headroom they have.
[...]

Issue #219, opened 2026-09-02
prd17 w1: the instrument's judgements and the operator's decisions are event families


ciaran-slow:
## Fence (may touch ONLY)
- `packages/core/src/events/summons.ts`
- `packages/core/src/events/gate.ts`
- `packages/core/src/events/operator.ts`
- `packages/core/src/events/summons.test.ts`
- `packages/core/src/events/gate.test.ts`
- `packages/core/src/events/operator.test.ts`
- `packages/core/src/events/index.ts`
- `packages/core/src/events/common.ts`
- `packages/core/src/events/events.test.ts`
- `packages/core/src/events/no-open-payload-law.test.ts`
- `packages/core/src/events/lenient.test.ts`
- `packages/core/src/fixtures.ts`
- `packages/core/src/reduce.ts` (the new case arms in applyEvent's switch, and nothing else — see the widening recorded below)
- `packages/core/src/upcast-chokepoint-law.test.ts` (the identity census only — second widening, below)
- `packages/core/src/eras/eras.test.ts` (the corpus gap list only — second widening, below)
- `packages/server/src/recorder/session-recorder.test.ts` (the frozen-input census only — second widening, below)
- `packages/core/src/record/read.test.ts` (the newer-era placeholder only — third widening)
- `packages/core/src/record/merge.test.ts` (the newer-era placeholder only — third widening)
- `packages/core/src/record/verify.test.ts` (the newer-era placeholder only — third widening)
- `packages/web/src/replay/Banner.test.tsx` (the newer-era placeholder only — third widening)
- `packages/web/src/replay/api.test.ts` (the newer-era placeholder only — third widening)
- `packages/web/src/replay/index.test.tsx` (the newer-era placeholder only — third widening)

## Blocked by

Nothing. Both declared blockers are cleared as of 2026-09-04:

- **#204** closed 2026-09-03 (it owned the actor-tagged merged stream).
- **prd27 w1 / #217 — WAIVED by the operator, in session, 2026-09-04.** The
  2026-08-24 amendment ruled prd-17 ruling 2's door and prd-27 ruling 1's door
  to be one door, and this issue previously waited on that doorway ADR. The
  operator ruled that this issue may define the families' contracts ahead of
  it. What the waiver covers is the contract dependency only.

**#62 is a sibling in this same wave, not a blocker.** It owns exactly three
paths, one of which is a new file inside this issue's directory — which is why
this issue's claim is enumerated file-by-file rather than taking the directory.
[...]

PR #267, opened 2026-09-04
prd27 w1: a beacon collector tails one rhizomorph-owned directory

### LINKED ISSUE #217: prd27 w1: a beacon collector tails one rhizomorph-owned directory
## Fence (may touch ONLY)
- `packages/server/src/collectors/beacon/`
- `packages/server/src/server/collector-loader.ts`
- `docs/adr/`

**Widened by the plan (2026-09-04).** A beacon is a new fact, and ADR-0002 puts every new fact in core's zod schema, folded by the one reducer — so the event contract lands in packages/core, and three core coverage laws plus the loader's own test move with it. Listed here so fence-lint sees the real footprint before dispatch; recorded again on the PR before the change, as AGENTS.md requires:
- `packages/core/src/events/beacon.ts` (new)
- `packages/server/src/server/collector-loader.test.ts`
- `CHANGELOG.md`

## Blocked by
**Corrected 2026-09-04 — #202 closed 2026-09-02, so the paragraph below has cleared.** No In-progress lane touches the widened fence (`scripts/fence-lint.sh 217 225 226 231 232` — see the plan comment for the executed result). Two groomed, not-live issues overlap it: **#62** (prd17 w1, Ready) fences `packages/core/src/reduce.ts`, which this issue adds one `case` to; **#219** (prd17 w2, Later) fences all of `packages/core/src/events/`. If #62 goes In progress first, land it first — its upcast chokepoint is what every event will flow through, and this issue's one arm rebases onto it trivially. The original text is kept, not deleted.

**#202** — prd43 w3 owns `docs/adr/`, and this issue owes an ADR into that directory. Land #202 first.

Sibling lanes: #23 owns `packages/server/src/api/route-class-law.test.ts`. This issue adds **no route** — that is the point of ruling 1 — so it must not go near it. #204 owns `packages/core/src/record/merge.ts` and `docs/record-format.md`.

## Why

This is prd-27's keystone and **prd-17 ruling 2's door is the same door**, ruled together on 2026-08-24. Nothing in either PRD is visible until it lands.

The instrument's own self-declaration is prd-27's thesis, and it is still true on `main`. `packages/server/src/collectors/sessionlog/collector.ts:44-46`:

```ts
attention: { level: 'partial',
  reason: 'inferred from transcript shape via the turn-shape state machine, not declared by the CLI',
  remedy: 'a hook beacon would declare it (prd15 ruling 2, wave 3)' }
```

Verified: there is no `beacon` directory under `packages/server/src/collectors/` — the six registered collectors are git, judge, pi, sessionlog, tmux and workmux. The remedy the instrument names for its own weakest signal does not exist.

## The door is ruled: a file drop, never a route

prd-27 ruling 1's open call closed on 2026-08-24: **one rhizomorph-owned watched directory, tailed by one beacon collector through the standard collector contract (ADR-0004) — never a new POST route.** prd-17 ruling 2 names the same doorway in the same words. Either door folds through the one reducer (ADR-0002). The sidecar-for-content / event-for-occurrence split governs payloads: bulk content lives beside the log, that-it-happened-with-its-digest lives in it.

The owed ADR on the directory and event contract **lands with this wave**, the way ADR-0024 landed with prd-29's keystone. It is not follow-up.

## Definition of done

- A beacon collector exists under `packages/server/src/collectors/beacon/` and tails one rhizomorph-owned directory of one-line JSON beacons through the standard collector contract.
- It is registered in `packages/server/src/server/collector-loader.ts` alongside the existing six static imports. `server/collector-wrap-boundary.test.ts` is a **derived-set** law — it walks the collectors directory rather than holding a list, and its own test proves a seventh collector is picked up automatically. Confirm it goes green without being edited; if it needs editing, that is a finding worth recording, not a silent fix.
- An ADR in `docs/adr/` records the directory contract and the event contract: where beacons are written, what one line is, what happens to a malformed line, and why this is a file drop rather than a route. Read `docs/adr/README.md` first — the log is append-only.
- A malformed or partial beacon line is survived, not fatal, and is reported the way the resilience wrapper reports other collector faults. A collector that dies on one bad line takes attention with it.
- **No route is added.** No file under `packages/server/src/api/` is touched.
- The `sessionlog` collector's `attention.remedy` string is **not** edited here — it still names a beacon that now exists, which stays true. Changing the declared/inferred disagreement voice is wave 2.

## Sibling case

The collector contract has a resilience wrapper (`collectors/resilience.ts`) and every existing collector goes through it. A new collector wired in *around* that wrapper rather than through it is the sibling miss, and it is the exact shape AGENTS.md cites: a seal released on a failed write but not on a throwing subscriber. `collector-wrap-boundary.test.ts` exists because that already happened once.

Splice the beacon collector into `collector-loader.ts` raw, bypassing `withResilience`. The wrap-boundary law must bite — its own "sixth collector spliced in raw" case is the precedent. A test that only asserts beacons are parsed survives it.

ciaran-slow:
Closes #217

prd27 w1 — the keystone of prd-27 and, since the 2026-08-24 amendment, prd-17 ruling 2's door as well. One commit, one issue.

## What lands

A seventh collector, `beacon`, tails every `*.jsonl` in `<data root>/<repo slug>/beacons/` through the standard collector contract and records each one-line JSON beacon as a `beacon.received` event (source `beacon`, `ts` = the writer's own clock, payload = the line's fields + a sha256 digest of the exact line + `file` + `offset`). A malformed line — a blank one included — is survived, counted and named in one `collector.error` per file per tick, never fatal. No route; nothing under `packages/server/src/api/` is touched. ADR-0036 records the directory and line contract and why a file beat a route, a socket and a widened `agent.status`.

Nothing folds a beacon yet and no emitter writes one yet — the collector's own manifest says `attention: absent` with the reason. Declared attention, its lapse and the disagreement voice are #218.

## The verify findings on the first build, and what changed

The independent verify pass on `9ae1ae5` returned four must-fix findings. All four are addressed in this commit:

1. **Blank lines corrupted the next beacon's offset and were not reported** (EXECUTED by verify). Root cause: offsets were reconstructed by summing the lines the shared `sessionlog/tail.ts` helper returned, and that helper drops empty lines. The fix is a beacon-owned reader, `read-beacon-lines.ts`, that returns every complete line paired with its real byte offset (blank lines included), reusing `isRotated`/`TailIdentity` from the shared helper without editing it — `sessionlog/` is #218's fence, not this issue's. Verify's exact probe (`"\n{beacon}\n"` → offset `1`, one `collector.error`) is now a test in both `read-beacon-lines.test.ts` and `collector.test.ts`.
2. **ADR-0036 was a summary, not the record the plan committed to.** The plan's §6 text now sits in the file verbatim: the `<dataRoot>/<repoSlug>/beacons/` location, the v1 line fields and bounds, one-file-per-writer, exact malformed-line behaviour, occurrence-plus-digest, the four considered options and why B/C/D lost, and the 2026-08-24 ruling in the Status line. The index row carries the ruled date and the provenance.
3. **Most of the plan's test matrix was missing.** `parse-beacon-line.test.ts` asserts every refusal reason by exact string plus the cap boundaries and purity; `collector.test.ts` covers no-directory ×3, empty directory, the fixture with exact first event, byte-exact offsets, digest-over-raw-bytes, idempotence ×3, append, partial trailing line, rotation, removal, two writers, ignored entries, retarget-safety, malformed (count, first reason, first offset, not re-reported), blank-line offsets, the wrapper's view (line fault ≠ tick fault ≠ missing directory), degrade→disable on the third tick, read-only + never-exec, and the manifest (L0, merge changes no sessionlog signal). `reduce.test.ts` gains the no-op fold law for `beacon.received` (envelope-only change, canonical-JSON identity, purity).
4. **`docs/architecture.md` was stale.** The event-source union names `beacon`; the "two actors outside the enum" paragraph explains why `beacon` is inside it; and the prd-17 status paragraph now says the door has landed while the event families it will carry have not.

## Fence widening (recorded on the issue before the change, restated here)

Beyond the issue's fence (`collectors/beacon/`, `collector-loader.ts`, `docs/adr/`), per the plan's §1 table and the verify-fix note on the issue:

- `packages/core/src/events/beacon.ts` (new), `events/common.ts`, `events/index.ts`, `events/events.test.ts`, `eras/eras.test.ts`, `fixtures.ts`, `reduce.ts`, `reduce.test.ts` — a beacon is a new fact, and ADR-0002 puts every new fact in core's schema behind the one exhaustive reducer.
- `packages/server/src/server/collector-loader.test.ts` (six → seven), `packages/server/src/recorder/session-recorder.test.ts` (the union-completeness census), `CHANGELOG.md`.
- `docs/architecture.md` — the verify-fix widening: the source union it narrates changed.

**Not touched, on purpose:** `collector-wrap-boundary.test.ts` (green unedited — the derived-set law picked the seventh collector up by itself), `collectors/sessionlog/**` (its `attention.remedy` still names a beacon, which is now true), anything under `packages/server/src/api/`, `packages/web/**`.

## What is the sibling case?

Three failure classes that a collector written from `sessionlog`'s template collapses into one, and the tests hold apart: a **missing directory** (ordinary, no event, never disabled), a **per-line fault** (`collector.error`, coalesced per file per tick, never a tick failure — `withResilience` sees no degrade), and a **tick fault** (`collector.disabled`, degrading twice and disabling on the third tick under the wrapper). The verify finding was a fourth sibling of the same shape: a **blank line** is a malformed line, and treating it as "nothing" both hid the fault and shifted every later offset.

The issue's own sibling — a collector wired *around* `withResilience` — is held by the unedited wrap-boundary law.

EXECUTED in the lane, each reverted afterwards:

- **Filter blank lines out of `readBeaconLines` again** (the pre-fix behaviour): `read-beacon-lines.test.ts` reddens on the verify probe (expects `[{'' @0}, {beacon @1}]`, gets one line at offset 0) and `collector.test.ts` reddens on both blank-line cases (offset `1` expected, `0` observed; `count: 1` error expected, none emitted).
- **Compute the digest over the parsed object** instead of the raw line: the `spaced.jsonl` case reddens — same JSON, different bytes, different digest.
- **Splice `createBeaconCollector()` into the loader raw**: the loader test's seven names still pass; the unedited wrap-boundary law is the witness that bites, as its own synthetic seventh-collector case proves.

Survives, and known: two processes appending to one file simultaneously — POSIX guarantees it for short lines and no test here spawns processes; a `kind` vocabulary — free-form by design until #218/#219 own it.

## Gates

`npm run typecheck` · `npm run lint` · `VITEST_MAX_WORKERS=6 npm test` — green in the lane on node 22.23.2. Not run: `scripts/gate.sh` (the operator's step).


gabriel-canaan (commented):
## Review of #267 — read commit-by-commit at `d1ffc05`, with `origin/main` merged in

Reviewed in a clean clone with `origin/main` merged into the branch (the branch was 16 commits behind; no file overlap with those 16, and the merge was clean). Node 22.22.2.

Nearly all of it, and the parts most likely to be wrong are the parts that are right.

**EXECUTED** — the issue's Definition of done, item by item:

- `collector-wrap-boundary.test.ts` is **green unedited**. It is not in the diff, and the derived-set law picked the seventh collector up by itself, exactly as #217 predicted.
- **No route.** Nothing under `packages/server/src/api/` is touched; `collectors/sessionlog/collector.ts` is not touched, so its `attention.remedy` still names a beacon and is now true.
- **The blank-line verify finding is genuinely fixed, and the fix has the right shape.** `read-beacon-lines.ts` is beacon-owned and reuses `isRotated`/`TailIdentity` from `sessionlog/tail.ts` without editing it, so #218's fence stays intact. The offset assertions are derived with `Buffer.byteLength`, not typed literals — `TTL - 1` this is not.
- **The no-op fold law is load-bearing.** I mutated the `case 'beacon.received':` arm in `packages/core/src/reduce.ts` to write `lastError` into state: **2 of the 3 new cases in `reduce.test.ts` redden.** It cannot pass for the wrong reason.
- **Full gates on the merged tree:** `npm run typecheck` and `npm run lint` clean; suite **409/409 files, 7408 passed, 3 expected-fail, 2 skipped**.
- **Fixture hygiene:** `fixtures/claude-hook.jsonl` is LF-terminated with no home path, no hostname, no real handle; `CAPTURE.md` says plainly that it is hand-written and why, rather than claiming a capture that could not exist yet.

The three-way split the docblock claims — missing directory ≠ per-line fault ≠ tick fault — is real, and the wrapper tests hold it apart rather than asserting it in prose.

---

### Two findings, both fixed on the branch

#### 1. A file unreadable for one tick lost its cursor and re-emitted every beacon on the log — `EXECUTED`

Fixed in `d882c88`.

The per-file `catch` in `poll` treated every read fault as a vanished file. That is right for the case it was written for: a file unlinked between `readdir` and `open` is absent from the next `readdir` too. Its sibling is a file that is **still there** and merely could not be opened this tick — EACCES while a writer re-permissions its own file, EMFILE, a Windows share lock. That file came back on the next tick with no cursor, so `readBeaconLines` read it from byte 0.

Probe: the fixture plus one appended beacon, then one tick with the file at mode `000`, then the fault cleared.

```
snapshot after the fault tick: {}
third tick emitted: 4 beacon.received (1 is correct)
kinds: waiting, working, stopped, landed
```

`waiting, working, stopped` were already on the log.

Worth recording *why* this collector has the shape and `sessionlog` does not: `sessionlog` seeds `nextFiles` from `{ ...prevSnapshot.files }` and only overwrites what it re-read, so it cannot lose a cursor — and equally cannot drop a removed file. Starting empty is the better choice and is why removed files drop out cleanly; it is also precisely what makes a failed read indistinguishable from a removal. The fix carries the prior cursor forward on the catch path, which keeps both: a genuinely vanished file still disappears, one tick later and with no events, because the next `readdir` does not list it.

Regression test sits beside the existing removed-file case, as its sibling. **Mutation:** deleting the one added line reddens it **3/3 runs**, restore verified with `cmp`.

*My first probe of this passed vacuously* — with no new bytes, `readBeaconLines` returns at the `info.size <= readOffset` guard and never opens the file. Recording that because it is the same trap the PR body's own mutation section is about.

#### 2. The amendment's containment check raised a pinned known-debt count — `EXECUTED`

Fixed in `eaa97b1`.

`d1ffc05` hardened `paths.test.ts` against a win32 separator by hand, with `path.relative(...).startsWith('..')`. That is the exact idiom prd-42 ruling 2 centralised into `packages/server/src/paths/containment.ts`, and `paths/prefix-comparison-law.test.ts` pins how many test-file sites still spell it longhand — **both directions**, deliberately, so a new site cannot silently cancel a migration.

```
AssertionError: test-file containment sites: 16, pinned: 14.
Sites: … packages/server/src/collectors/beacon/paths.test.ts (2) …
```

**This fails on the branch alone, not only against `origin/main`:** the pin is 14 at the merge base `b448f9d` and 14 on `main`. The suite immediately before the amendment was 409/409, so the amendment is what moved it. It looks like the amend went out without a full suite behind it.

The change itself was the right instinct — the original `dir.startsWith('/data')` was weaker, and the law does not flag it. The fix is to ask the primitive instead: `isInside(parent, candidate, realpath?)` takes an injectable realpath precisely so a test can use it without a filesystem. `/repo` and `/data` do not exist and the assertion is about path shape, so it gets the identity canonicalizer and stays as lexical as the version it replaces, while gaining symlink-correct semantics if the inputs ever become real. Both facts the amendment was reaching for are still asserted, and more directly. The pin returns to 14 untouched.

**Mutation:** pointing `beaconDirFor` at `<repoPath>/.rhizomorph/beacons` reddens 2 of the 3 cases; restore verified with `cmp`.

This is worth a moment beyond the fix. It is round 2 of the pattern the review skill records: a one-line hardening, correct in its own hunk, whose consequence lives in a text-swept law two directories away. *What does the changed file now participate in that it did not before?* — the answer was a pinned corpus.

---

### One fact you need, which I did not fix

`packages/server/src/api/route-class-law.test.ts` **times out at 5000 ms under `VITEST_MAX_WORKERS=6`**, intermittently: it failed on two of my four full runs, on the case that `readFileSync`s every swept file and runs every `CLAIMS` regex over each. Run alone it passes 3/3 at 4.5–5.1 s of test time — against a 5 s default. It is `main`'s file (#232), untouched by this PR, and the failure is `Error: Test timed out in 5000ms`, never an assertion.

I have not touched it, and specifically have not widened the timeout — that is the standing rule, and the real shape here is an O(files × claims) synchronous sweep that every merged PR pushes closer to the edge. **This PR adds 14 files to the corpus it sweeps.** It wants its own issue rather than a fix inside this one.

---

**`"lane": null` is refused; an absent `lane` is accepted and reads back as `null`.** EXECUTED:

```
explicit null lane -> {"kind":"malformed","reason":"beacon \"lane\" is empty or too long"}
absent lane        -> {"kind":"beacon","at":1,"payload":{"writer":"w","kind":"k","lane":null}}
```

Two spellings of the same intent, one of them counted as a broken line. The payload schema is `nullable`, and `JSON.stringify` on a writer's record with an unset lane emits `null` rather than omitting the key, so the refused spelling is the one a naive emitter is most likely to produce. ADR-0036 says the field is optional, so refusing explicit `null` is defensible and the line is survived either way — but it is one `&& input.lane !== null` if you would rather the door accepted both. Since #218 and #219 are the first writers, now is the cheap moment to decide.

---

### Where this stands

Both fixes are pushed to the branch and the head is `eaa97b1`. Full gates re-run on the merged tree after them: typecheck and lint clean, **408/409 files** with the sole failure being the `route-class-law` timeout described above.

Not approving, and deliberately: two of the three commits on this branch are now mine, and the repo's own rule is that a reviewer does not review their own substantive fix into `main`. @ciaran-slow — please confirm both fixes, particularly the reasoning in #1 about why starting `nextFiles` empty is still the right choice. Landing is the operator's step, not mine.

*Not run: `scripts/gate.sh` (the operator's landing step). CI is re-running on `eaa97b1`; the checklist to compare against is three jobs and ten legs.*

ciaran-slow:
## Author's answer to the two review fixes — both accepted

- **`d882c88` — a file unreadable for one tick keeps its cursor.** Accepted. A real sibling of the removed-file case that I held apart from the tick fault but not from each other: the catch path treated every read failure as a vanished file, and a transient EACCES/EMFILE/share-lock would have re-emitted every beacon on the log one tick later. The fix (carry the prior cursor on the catch path) keeps the removed-file behaviour, since the next `readdir` still drops it.
- **`eaa97b1` — the containment assertion asks the primitive.** Accepted, and the miss is mine. The `path.relative(...).startsWith('..')` idiom I amended in for the Windows leg is exactly the hand-rolled comparison `paths/prefix-comparison-law.test.ts` pins at 14 sites; it went to 16 and the law reddened on every OS. After that amendment I re-ran only `paths.test.ts` and lint, not the suite — that is what let it through. `isInside(..., lexical)` is the right call.

**EXECUTED on `eaa97b1` in the lane (node 22.23.2):** `prefix-comparison-law.test.ts` + `collectors/beacon/**` + `lab/compare.test.ts` → 6 files, 104 passed.

**Windows leg, for the record.** Across the three runs so far the unexpected failures were: `beacon/paths.test.ts` (mine, fixed by `d1ffc05`), `prefix-comparison-law.test.ts` (mine, fixed by `eaa97b1`), and two files outside this fence that main's own leg is green on — `sessionlog/tail.test.ts` (inode-only rotation detection; NTFS can reuse a file id after unlink+create) and `lab/compare.test.ts` (`STACK_TRACE_ERROR` in `compareFork`). Each appeared in one run and not the next. I have not touched either; noting them here rather than chasing them from this PR.

ciaran-slow:
## Windows leg on `eaa97b1`: red on the review's own regression test — fixed in `991743c`

EXECUTED (run 33837033863): `collector.test.ts › a file that is present but unreadable for one tick keeps its cursor` failed with `expected [ { id: 'beacon-1', …(4) } ] to deeply equal []`. Cause: `chmod 0o000` does not remove read access on win32 (Node toggles the read-only attribute), so the "unreadable" tick read the file normally.

`991743c` keeps the test and its assertions, and produces the fault class it guards against — a failed open of a file that is still there — by failing `readBeaconLines` once through a passthrough mock, on every OS. The mutation from `d882c88`'s message (drop the cursor-carry line) still reddens exactly this test. Beacon suite 62/62, lint and typecheck green in the lane.

ciaran-slow:
## Red on `05e47ed` — the #266 macOS timeout, not this PR; main merged in at `528450d`

EXECUTED (run 33846845838, `build-test-boot (macos-latest, current)`): the one failure was `route-class-law.test.ts › no swept file states a recognised route-count claim that no CLAIMS row declares` — `Test timed out in 5000ms` at 8183 ms. No assertion failed; 412/413 files green, all nine other legs green. That is #266 exactly, and #266's fix (`b3fbe59`, the sweep normalises each file once) landed on main one commit *after* the merge this branch carried (`2c17dd1`), so the branch was still running the slow sweep.

`528450d` merges current main (`612df45`) in, so the fix is now on the branch. Locally on the merge: the route law + beacon + reducer + loader files, 7 files / 297 passed; typecheck 0.

PR #269, opened 2026-09-04
prd43: the route-count completeness sweep fits its timeout on the slowest runner

### LINKED ISSUE #266: prd43 w6: the route-count completeness sweep fits its timeout on the slowest runner

gabriel-canaan:
Closes #266

The route-count completeness sweep added by #232 takes **4.73 s of test time**
against vitest's 5000 ms default. It has been marginal since it landed in
`ca7370d` — it passes ubuntu and this machine by roughly 270 ms, and does not
pass `macos-latest`.

This is a performance fix. No assertion changes, no scope changes, and the sweep
still catches exactly what it caught before.

## What it has already cost

Three times, and each time it pointed at somebody else's diff rather than at
itself:

- **PR #263** — `build-test-boot (macos-latest, current)` failed at **7490 ms**,
  `Error: Test timed out in 5000ms`, on an approved and entirely unrelated
  branch. It went green on a re-run, which is the worst outcome: it merged
  without the cause being touched.
- **prd-43 wave 6's assembly** — the same test, red locally. Measured with a
  control: **4.54 s with wave 6 applied, 4.78 s on plain `origin/main`** — that
  is, slightly *slower without* the wave. Which side of the 5 s line a run lands
  on is noise, and wave 6 had nothing to do with it.
- A reviewer's time, twice over, because `#234` edits that same file and the
  failure reads as its fault.

## Why it is slow

`containsClaim` normalises the WHOLE FILE on every call, and it is called once
per (file, claim) pair:

```js
const normalized = text.replace(/\s+/g, ' ')
const gutterless = text.replace(/^\s*\*\s?/gm, ' ').replace(/\s+/g, ' ')
return new RegExp(...).test(normalized) || new RegExp(...).test(gutterless)
```

1109 tracked files matching the swept extensions × 15 `CLAIMS` rows = **16,635
calls**, each doing three whole-file regex replaces and two `RegExp`
constructions. The normalisation depends only on the text, never on the pattern
— so roughly **50,000 whole-file replaces run where 3,300 do**.

## The fix

Hoist what does not vary. `readingsOf(text)` computes the two readings once per
file; `probeFor(pattern)` compiles each claim's probe once, as a non-global copy
so `lastIndex` cannot leak across files. Behaviour-preserving by construction:
same inputs, same two strings, same probes run against them.

An explicit timeout replaces the default, with the measured figure in a comment
and a note that the answer to seeing this test near it again is to make the
sweep cheaper, not to raise the number.

## Measured

Same machine, same tree:

```
tests 4.73 s  ->  783 ms          (~6x)
route-class-law.test.ts           88 passed
typecheck 0 · lint 0 · suite 0    405 files, 7343 passed
```

None that matters — the sweep must still bite, and that is the whole risk of a
performance change to a detector:

```
planted the recognised claim "puts the same token on thirteen reads"
into packages/server/src/api/test-support.ts, a file no CLAIMS row declares

  -> RED: "packages/server/src/api/test-support.ts matches the pattern
           registered for security.ts gated reads"
```

The hard-wrapped block-comment control the sweep already carries also still
passes; it was updated to the new signature rather than deleted.

## What is the sibling case?

The defect is *"a claim that holds on the machine it was measured on and not in
general"* — which is precisely the class this sweep was written to enforce, so
it is worth saying plainly that the sweep shipped carrying it. The sibling is
any other test in this file that walks the whole tracked corpus; the two
existing ones (`sweptFiles` non-empty, and the per-claim walk) are cheap because
they do not re-normalise per row, and they were checked rather than assumed.

## Fence widenings

None. One file.

## Provenance

Diagnosed and written before #263 merged, then parked unpushed at the operator's
instruction while a reviewer looked at that PR. #263 merged on a re-run without
the cause being fixed — confirmed three ways: `containsClaim` unchanged on
`main`, no timeout argument on the test, and **zero commits touching that file
since wave 6**. Rebased onto current `main` and re-measured before opening this.


ciaran-slow (approved):
Reviewed at `5cb6d5c`, both commits read separately, everything below run on a
clean clone at node 22.23.2 (`.nvmrc`), on the **post-merge tree** — the branch
is 4 commits behind `main` and `git merge origin/main` applies clean (none of
the 4 touch this file).

**EXECUTED — the central claim of commit 2 reproduces exactly.** Reverting the
hoist (normalisation moved back inside the `(file, claim)` loop) reddens the new
structural guard and nothing else:

```
× the sweep normalises each swept file once, not once per claim — the hoist is load-bearing
  → AssertionError: expected 11737 to be 979 // Object.is equality
  Tests  1 failed | 88 passed (89)
```

**EXECUTED — and the 30 s ceiling genuinely does not catch it.** On that same
mutated tree the semantic sweep passed at **3004 ms**, nowhere near its ceiling.
So the finding commit 2 is answering is real and independently confirmed: the
wall-clock bound could not fail for the reason it claimed, and the call count can.
That is the right substitution.

**EXECUTED — the sweep still bites, on the hardest path.** I planted the
recognised claim gutter-wrapped *and* hard-wrapped across lines (exercising the
`gutterless` reading, not just `normalized`) into an unregistered swept file:

```
/**
 * puts the same token on
 * thirteen reads
 */          → appended to packages/server/src/api/test-support.ts

FAIL  no swept file states a recognised route-count claim that no CLAIMS row declares
  + "packages/server/src/api/test-support.ts matches the pattern registered for security.ts gated reads"
```

**REASONED, then EXECUTED — the hoist is behaviour-preserving.** `probeFor`
strips `g`; I checked every row rather than assuming: all 12 `CLAIMS` patterns
carry exactly `/g` and nothing else, so a non-global copy under `.test` is
identical to the old fresh-`RegExp`-per-call, and the `matchAll` at
`route-class-law.test.ts:269` still uses the original global pattern. No
`lastIndex` reachable from the shared probes.

**EXECUTED — measured, same machine, same tree, per test** (the number the
timeout actually applies to, since vitest budgets per test, not per file):

```
no swept file states a recognised claim…   2892 ms  ->  693 ms   (4.2x)
```

**EXECUTED — gates on the post-merge tree:** typecheck 0 · lint 0 (876 files) ·
suite `409 files, 7397 passed | 3 expected fail | 6 skipped`, exit 0. CI is
green on all 10 legs (`build-test-boot` ×3, `pack-smoke` ×6, `windows-suite`).

**Fence honoured exactly** — one file, matching #266's declared fence. No
widening. No personal paths in the diff.

Two design choices worth calling out as correct rather than merely acceptable:
`expected` is derived from `sweptFiles().length` at runtime instead of pinned to
979 — which is the only spelling that does not itself become the drift this file
exists to catch — and `expect(CLAIMS.length).toBeGreaterThan(1)` states the
non-vacuity condition rather than assuming it. The vacuity sibling on the other
side is already covered by `expect(sweptFiles().length).toBeGreaterThan(50)` at
`route-class-law.test.ts:515`.

## One finding — non-blocking, PR body only

**The `## Measured` block still describes commit 1's tree, not this PR's.** It
reports `tests 4.73 s -> 783 ms (~6x)` and `route-class-law.test.ts 88 passed`.
At `5cb6d5c` the file is **89** tests, and commit 2's guard runs the *whole
sweep a second time*, so the file-level figure roughly doubles. Measured here:

```
file-level test time   3.12 s  ->  1.56 s     (2x, not ~6x)
  no swept file …                     693 ms
  the sweep normalises each … (new)   630 ms
```

The ~6x is right for the sweep in isolation and the in-code docblock is fine as
written — it is scoped to the hoist. It is only the PR body that now overstates
the net, on a PR whose subject is a stated figure drifting from the thing it
counts. Your numbers, your machine — I would rather you re-measure than have me
paste mine over yours.

- **This is still a draft.** Nothing I ran argues against marking it ready.
- The second full sweep costs ~630 ms to buy a deterministic guard that works
  under any load on any runner. Given #266 started as "a timing assertion that
  is noise on someone else's branch", I think that trade is plainly right — but
  it is a trade, and it is yours to confirm you meant it.
- `route-class-law.test.ts` is on `.windows-known-failures:35` (line-endings,
  unrelated test in the same file), so the new guard is not witnessed on native
  Windows. Pre-existing, not this PR's to fix; noting it so it is not a surprise.

## What I did NOT run

`scripts/gate.sh` — landing is the operator's step, and I am not merging this.
No load-batch pass either, so the new test has not been measured under 4x
concurrency; its guard is a call count rather than a clock, so load should not
reach it, but that is REASONED, not executed.

Approving on the evidence above. The merge is the operator's call, not mine.

Issue #270, opened 2026-09-04
prd43 w6: the sweeping tests have a budget they can actually meet under load


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/api/route-class-law.test.ts`

## Blocked by

Nothing. One line.

**Two open issues also claim this path, and that is recorded rather than
avoided.** #213 (prd14 w1) lists it in its 2026-09-04 plan widening, and #216
(prd20 w1) claims its outbound call-site table. This issue touches neither the
CLAIMS rows nor that table — only the timeout on one `it`. Whichever lands
second pays a trivial rebase.

**Model:** n/a — operator's session, 2026-09-04.

## Why

The completeness sweep in this file (`no swept file states a recognised
route-count claim that no CLAIMS row declares`) runs `git ls-files` over the
repo and then `readFileSync` on every tracked file with a swept extension. That
is hundreds of reads inside one `it`, on vitest's default 5000 ms budget.

Measured 2026-09-04, same worktree, same commit:

| condition | file duration | result |
|---|---|---|
| run alone | ~4.2 s | 88/88 pass |
| under `VITEST_MAX_WORKERS=6` | 5.8 s | timeout |
| under the same load | 8.1 s | timeout |
| under the same load | 11.1 s | timeout |

So the budget is marginal even in isolation and any concurrency tips it over.
The failure is `Error: Test timed out in 5000ms` — **not** an assertion failure,
which is what makes it easy to misread: the test's name describes an undeclared
route-count claim, so a reader who does not open the log concludes the sweep
found one. It did not. That misreading happened twice in the session that filed
this, once by an agent that had already seen the correct log.

One of the failing runs shows the cause in its own output: `Preparing worktree
(detached HEAD ...)` — a concurrent law test creating git worktrees while this
one shells out to git.

**It has cost real time.** Six failures in one session across three separate
landing-gate runs, each one investigated (passes alone, no route-count claim in
the diff, not the branch's fault) before being set aside. Two of those
investigations were by independent review seats that flagged it as a
`[load-batches]` candidate.

## Definition of done

- The sweep test carries an explicit timeout with a margin over its measured
  worst case, and a comment saying what it measures and why the default was not
  enough — including the numbers above, so a later reader can tell a genuine
  slowdown from a budget that was always tight.
- No CLAIMS row, no swept-extension list, and no exclusion prefix changes. The
  law's subject is untouched; only its budget moves.
[...]

PR #271, opened 2026-09-04
prd17: the instrument's judgements and the operator's decisions are on the record

### LINKED ISSUE #62: prd17 w1: every event reaches the reducer through the upcast chokepoint

### LINKED ISSUE #219: prd17 w1: the instrument's judgements and the operator's decisions are event families

### LINKED ISSUE #268: prd17 w1: the fold-order fixture states the ruled law, and the landing ledger tells the truth
## Fence (may touch ONLY)
- `packages/core/src/eras/fold.ts`
- `docs/adr/0002-one-reducer-for-live-and-replay.md`
- `docs/architecture.md`
- `docs/roadmap.md`

## Blocked by

Nothing. The work is built and reviewed; this issue exists to give it the home
AGENTS.md requires, which it should have had before the first commit.

**A live branch holds two of these paths.** Branch 217-beacon-collector (prd27
w1, pushed, unmerged) claims docs/adr/ and packages/core/src/reduce.test.ts in
its widened fence. Accepted knowingly: this work and that branch touch different
regions, and whichever lands second pays a small rebase. Recorded here rather
than discovered at landing.

**Model:** n/a — built by the operator's session, 2026-09-04.
**Wave:** 1, alongside #62 and #219 in the single prd17 PR.

## Why

prd-17's 2026-08-24 amendment named one stale witness and left it: the prose
and the local foldReplay helper of reduce.test.ts's divergence fixture still
modelled the pre-ruling world, where live folded arrival order and replay folded
a ts-sorted copy. Issue #205 ruled that away — append order is the truth — and
replayFold.ts has honoured it unconditionally since 6fd3ffa. The fixture kept
describing the defect as live, printed BLOCKED for a conductor who had already
ruled, and told readers replay sorts before folding, which it does not.

Three documents pointed at that same follow-up and all three said "one stale
witness remains": prd-17's amendment, ADR-0002's, and architecture.md's
fold-order section. A wrong pointer to a real thing is worse than no pointer,
because it reports success.

## Definition of done

- reduce.test.ts's fold-order block states the ruled law. The reducer's half is
  what core can prove — that it is order-sensitive on three axes, so the
  fold-order choice is load-bearing, and that the committed era-1 snapshot is
  the append-order fold. The ts-sorted fold survives only as the named
  counterexample that keeps the law from being a rule about nothing.
- eras/fold.ts stops saying the corpus "leaves the ruling to the conductor".
- The three stale "one stale witness remains" pointers are corrected.
- **architecture.md's landing ledger tells the truth about prd-17 ruling 1.**
  Its "Ruled, not yet landed" list and roadmap.md's copy both predate #219;
  and its claim that prd-45 owns gate.verdict's shape is false — verified,
  `grep -ci event` on that PRD returns 0. Operator ruling, 2026-09-04: the
  ledger is wrong, prd-17 owns gate.verdict.
- Mutation-proven: making reduceAll ts-sort reddens the fold-order block.
- npm run typecheck, npm run lint and VITEST_MAX_WORKERS=6 npm test pass.

## Sibling case

The three "stale witness" pointers are siblings of each other — correcting one
and leaving two is the shape this repo names most often. All three are in the
fence for that reason. architecture.md's landing ledger is a fourth sibling
found by review, not by the original sweep: it lives in the same file, fourteen
lines from an edit this work already made, and was missed.

Neuter the ts-sorted counterexample so it no longer sorts: seven assertions must
redden, including both era-1 divergence claims. Make the incremental fold sort
its input: the spellings-agree test and the era-1 snapshot test must redden.
Both executed.

gabriel-canaan:
prd-17's remaining tracker work, assembled as one wave. Three commits, reviewed
commit-by-commit.

Closes #62
Closes #219
Closes #268

## What is in it

|---|---|
| `#62` | prd-17 ruling 3 item 3's identity `upcast()` chokepoint exists. It was **reserved but never built**, and three documents said it had landed — they become true rather than get edited |
| `#219` | ruling 1's eight remaining event families exist as contracts (`session.closed` was already landed). Defined, **not emitted** — no emitter ships here |
| `#268` | the fold-order fixture states the post-#205 law instead of the pre-ruling divergence, and the landing ledger stops claiming things that are not so |

That completes **ruling 3**. Ruling 1 has its contracts but nothing raises one;
rulings 2 and 4 are unstarted. prd-17's Outcome header already says "partially
shipped" and the milestone stays open — this PR does not finish the PRD, and
`architecture.md`/`roadmap.md` are corrected in `#268` to say exactly that.

## One thing to expect from CI

`packages/server/src/api/route-class-law.test.ts` can time out under suite load.
It is **not this branch's** — it passes 88/88 alone on this branch and on `main`,
and its own failure is `Error: Test timed out in 5000ms`, never an assertion.
Measured on one machine, one commit: ~4.2 s alone, then 5.8 s / 8.1 s / 11.1 s
under `VITEST_MAX_WORKERS=6`, against a 5 s per-test budget. Six red runs across
three pre-commit-gate attempts here.

Read the failure and not the test name if it goes red: the test is named for an
undeclared route-count claim, so twice in one session a reader concluded the
sweep had *found* one and went looking. It had not.

The one-line fix is #270, on branch `270-sweep-budget`, deliberately kept out of
this PR so this one carries only prd-17 work. Landing it first would make this
PR's CI quieter.

## What is the sibling case?

Answered per commit, and three of the four were **found by review rather than by
the author**:

- **`#62`** — the parse boundary looked like the natural chokepoint and is the
  miss: an event reaching `reduce()` without passing through `parseEvent` (a
  synthesised lab event, a fixture, a future emitter handing typed objects
  straight to the recorder) would bypass a parse-side hook entirely. Hooking at
  `reduce()` is the one door every fold path bottoms out in.
- **`#219`** — `summons.raised`/`.cleared` are a pair that must NOT be modelled
  as one: a clear with no raise is a real state of the world (a session can end
  mid-alarm) and a schema assuming pairing would lie the first time it happened.
  The schemas are independent and `(lane, kind)` matching is deferred to fold
  time.
- **`#268`** — the three "one stale witness remains" pointers are siblings of
  one another, and a sweep scoped by that phrase found all three and missed a
  **fourth** sibling fourteen lines away in a file it had already opened: the
  landing ledger. Scope by what a claim is about, not how it is worded.

Every claim below was executed, not described.

- **`#62`'s law survives none of four.** Deleting the `upcast(event)` call
  reddens both call-count assertions; a non-identity `upcast` reddens the
  identity assertion **and**, independently, the golden era corpus's
  byte-identical fold; stripping the ADR-quoted sentence reddens the citation;
  and moving the call below any of `reduce()`'s three readers reddens the
  ordering assertion, once per reader.
- **That fourth assertion exists because review proved the law could not fail
  for the reason it claimed.** A call counter cannot see ordering, and while
  `upcast` is `return event` the position is unobservable by construction —
  measured: the mutation left the law 6/6 green and all of core 1025/1025 green.
  The fix makes the mocked chokepoint return something *different*, which is the
  only thing an identity function's position is visible against.
- **`#219`** — deleting `sessionId` from the operator base schema reddens the
  coordinate tests; adding `.loose()` to a payload fails the no-open-payload law
  naming the file and the hit; deleting a family from `EVENT_SOURCE_BY_TYPE`
  while leaving it in the union fails `tsc` on the `satisfies` clause before any
  test runs.
- **`#268`** — making `reduceAll` ts-sort reddens twelve tests, nine of them in
  the fold-order block, so the counterexample is load-bearing rather than
  ornamental.

## Fence widenings, each recorded on its issue before the change

- **`#219` → `packages/core/src/reduce.ts`** — `applyEvent`'s switch ends in a
  compiler exhaustiveness proof, so new families cannot typecheck without arms.
  Bounded to the new arms.
- **`#219` → three EVENT_TYPES censuses** (`upcast-chokepoint-law.test.ts`,
  `eras/eras.test.ts`, `session-recorder.test.ts`) — verified by running the
  suite that exactly these three redden and nothing else does.
- **`#219` → six forward-compatibility fixtures** (`record/read`, `record/merge`,
  `record/verify`, three web replay tests) — they used `summons.raised` and
  `operator.ack` as stand-ins for "a type from a newer era", and those types are
  real as of this PR, so they were silently exercising the unknown-SHAPE path
  while claiming the unknown-TYPE one, and still passing.
- **`#268`** declares its five files, which had no issue at all when first
  committed.

Two of the three `#219` widenings were **grooming misses by the author of the
fence, not by the lane** — the lane stopped at its boundary and reported, which
is the correct outcome and is why they were caught.

## Review history

`#62` went through two independent verify rounds and two fix re-review rounds
(Opus + Codex `gpt-5.6-terra`, isolated worktrees, neither seeing the other).
The assembled stack then went through a further round weighted at `#219` and
`#268`, which had had no independent eyes. That round found six must-fix items,
none of them a runtime defect and **all of them false or frozen claims** —
including one in a file whose entire purpose is to be a trustworthy signpost,
and one in a commit written by the reviewing session itself.

Four operator rulings came out of it and are recorded on `#219`: the operator
coordinate is `sessionId` + a record line index (a fold count located nothing —
`eventCount` restarts at a session boundary and counts only events the reducer
understood); the scalars ruling 1 names are present; `architecture.md`'s claim
that prd-45 owns `gate.verdict` is false (`grep -ci event` over that PRD returns
0) and prd-17 owns it; and the fold-order work gets the issue and the fence it
should have had.

`docs/adr/0002`'s dated 2026-08-24 amendment was rewritten in place by the first
version of `#268` — which `docs/adr/README.md` forbids in terms. The original is
restored verbatim and the new work appended as its own dated note, which records
the mistake rather than hiding it.

## Gates

`npm run typecheck` rc=0 · `npm run lint` clean (884 files) ·
`VITEST_MAX_WORKERS=6 npm test` 409 files, 7399 passed, 3 expected-fail,
2 skipped, **0 failed**, on node 22.22.2. Every commit passed the repo's
pre-commit gate with a certified mutation — not a described one.

The gate is still the operator's step, and `[load-batches]` applies here: this
branch touches tests.


ciaran-slow (approved):
## Approve — verified, with two documentary findings the author should see

Reviewed commit-by-commit in a clean clone at `8d249b4`, base `fcbe2f1` (the branch is exactly three commits on current `origin/main`, no drift). Everything below labelled **EXECUTED** was run on node 22.23.2 and its exit status read.

The evidence in this PR body is unusually precise, and it holds up under re-derivation:

- **EXECUTED — the `grep` recorded in `upcast.ts`'s comment is exact.** Re-ran it verbatim: **21 files, 19 outside `packages/core`**. The comment's count is right to the file. Given this claim was wrong twice before (a four-file list, then a one-package command), the third version being precisely correct is worth saying out loud.
- **EXECUTED — the ledger correction rests on a true premise.** `grep -ci event docs/prds/done/prd-45-the-earned-verdict.md` → `0`. architecture.md's claim that prd-45 owned `gate.verdict` was false, exactly as `#268` argues.
- **EXECUTED — mutation (d) does the job it was added for.** Moved `upcast(event)` below `opensNewSession`'s read: **1 failed / 10 passed**, `assertion (d) > opensNewSession sees the upcast event, not the raw one`. This is the repair the author says `(a)` could not do, and it is real — an identity function's *position* is now observable.
- **EXECUTED — deleting the `upcast()` call reddens 5 tests**, both `(a)` assertions plus all three `(d)` readers. The law is not vacuous.
- **EXECUTED — the anti-vacuity fix in `(b)` is real.** Dropped `f.sessionClosed()` from the census: reddens with a set-equality diff that *names* the missing family. The old `length > 20` floor would have stayed green.
- **EXECUTED — `#268`'s counterexample is load-bearing.** Made `reduceAll` ts-sort: **19 tests red** across `packages/core` (the commit claims twelve — the mutation is caught harder than advertised, not softer).
- **EXECUTED — gates.** `VITEST_MAX_WORKERS=6 npm test` → 409 files, **7396 passed, 3 expected-fail, 0 failed**, exit 0. `npm run typecheck` rc=0. `npm run lint` clean at **884 files** — matching the claim exactly. `route-class-law.test.ts` did **not** flake in my run.
- **EXECUTED — CI really ran.** All 10 legs green (3 × `build-test-boot`, 6 × `pack-smoke`, 1 × `windows-suite` — the full checklist AGENTS.md specifies). I pulled the step list for `build-test-boot (ubuntu-latest, current)` rather than trusting the badge: `Build`, `Test`, `Typecheck`, `Lint`, packaging guard and boot smoke each report `success`. Nothing silently skipped.
- **REASONED — fence audit clean.** All 30 file-touches across the three commits fall inside the three issues' declared fences, widenings included. `no-open-payload-law.test.ts` is fenced by `#219` and untouched, which is fine.

The schemas themselves are good. The `summons.raised`/`.cleared` independence is right and the reasoning is right — a clear with no raise is a real state, and pairing is correctly deferred to fold time over `(lane, kind)`. The `offset`-as-line-index ruling is correctly reasoned from `reduce()`'s state swap and the era-gap line drops. `fence.declared` carrying real paths rather than a digest is the correct exception to the sidecar split, for the reason the issue gives.

---

### Finding 1 — `#219`'s commit message contradicts itself twice (documentary; the code is correct)

**REASONED, from the commit message text.** Two first-draft paragraphs survive at the top of `8e72201`'s message, each corrected later in *that same message*'s REVIEW REPAIRS section. The tree is right in both cases; only the message is stale.

**(a) `offset`.** Message lines 15–16:

> Every operator payload carries `offset` (the log's own eventCount at decision time) per the issue's Definition of done.

Against lines 105–107 of the same message:

> **The operator coordinate is sessionId + a record LINE index.** The first draft defined `offset` as SessionState.eventCount, and review showed that locates nothing

`operator.ts` is correct — *"Zero-based LINE index into that session's record … Not a fold count"*. The header paragraph still states the definition that was ruled out.

**(b) why `lab` is out of the enum.** Message lines 11–13:

> unlike 'lab' and 'judge' in their own files, which stayed hand-built outside the enum because their issues had no fence there.

Against lines 124–126:

> **The enum comment no longer misattributes why lab is out of it.** It claimed lab and judge stayed out of eventSourceSchema for want of fence reach. True of judge; false of lab, which lab.ts excludes on principle

`common.ts`'s comment was fixed and is careful and correct. The commit message paragraph asserting the same misattribution was not.

This is the class of defect this PR is otherwise rigorous about, and it has the shape the PR body names in its own sibling-case answer: *the sweep corrected the claim where it was flagged and missed the identical claim in the artifact it was already editing.* Here the sibling of "the enum comment misattributes `lab`" is "the paragraph 110 lines above it in the same message says so too".

**I did not fix this.** The only remedy is amending the commit message, which means a force-push to your branch — and force-pushing an author's branch is a line I won't cross uninvited. Your call whether it's worth a rewrite and a fresh CI cycle, or worth leaving with this comment as the record. Given a reader who reaches the REVIEW REPAIRS section gets the truth, I'd lean to leaving it — but you should be the one deciding that, not me deciding it silently.

### Finding 2 — cosmetic, `docs/roadmap.md`

**EXECUTED.** Line 135 is now **110 columns**, the longest in the file; the surrounding body wraps at ~76. The rewrap in `8d249b4` left `…ruling 2's beacon ingestion — its doorway ruled 2026-08-24, shared with prd27 (the` unwrapped. Purely cosmetic, biome doesn't lint markdown, and I've left it alone rather than spend a commit and ten CI legs on a prose wrap.

---

### What I did not run

I did **not** run `scripts/gate.sh` — AGENTS.md is explicit that landing is the operator's step and a reviewer never runs it. I also did not merge: merges on this repo are manual. Everything short of that is green above.

Approving on the strength of the executed evidence. Finding 1 is real but is a commit-message defect over correct code, and it self-corrects further down the same message — not worth another ~21 hours of queue in my judgement, though the amend is yours to make if you disagree.

PR #272, opened 2026-09-04
prd43 w6: the sweep meets its budget, and the outbound vocabulary covers three more spellings

### LINKED ISSUE #233: prd43 w6: watching.md's gap list states its real provenance
## What becomes true

`watching.md`'s honest-gap list stops claiming a provenance it does not have,
and says plainly that it is a representative sample rather than an enumeration.

## Evidence

`#63` (PR #230) corrected this list's two stale manifest remedies and added a
closing note. Two things about that note are still not quite true, both found
during the verify pass and both in the operator's own fence widening rather than the
lane's work.

**1. The provenance is overstated.** `docs/user-guide/watching.md:163` says the
strings are *"transcribed from `buildGaps` (`packages/core/src/fleet/gaps.ts`)"*.
Three of the listed lines are not: they come from
`packages/web/src/panels/burn/format.ts` (`CONDUCTOR_NOT_INSTRUMENTED_GAP`) and
`packages/web/src/trace/EmptyTrace.tsx`.

**2. Voices the list omits**, all real and all reader-facing:

- `UNATTRIBUTED SPEND (1 lane)` — the singular branch of
  `gaps.ts`'s `lane${n === 1 ? '' : 's'}`. The list writes only the `N lanes`
  form.
- `unknown — no worker output yet` — `panels/burn/format.ts`'s
  `formatOverheadOrGap`.
- `NO ACTIVITY RECORDED — this lane has produced no tool call, file change or
  commit …` — `packages/web/src/drawer/Activity.tsx`. This is the closest
  sibling of the `NO TRACE TELEMETRY` line the list *does* carry: same
  uppercase head, same em-dash reason, same `role="status"`. `EmptyTrace.tsx`'s
  own doc comment even points at it — *"Same voice as `drawer/Activity.tsx`'s
  own empty state"* — and the list followed the citation only halfway.
- the `rhizomorph doctor` voices in `drawer/useTranscript.ts`,
  `lane-page/laneIndex.ts` and `drawer/attach.ts`.

## The ruling (made 2026-09-03)

The heading says *"Exact strings you may see"*, and the list already uses `N`,
`M` and `<COLLECTOR>` as placeholders — so it is a sample written as if it were
an enumeration.

**Ruled: it is a sample, and the prose says so.** Drop "exact" from the heading,
state that the list is representative rather than exhaustive, and correct the
provenance line to name all three sources. **No law.** A law over a list that is
honestly partial would enforce the wrong claim, and the alternative — a real
enumeration plus a law sweeping seven source files — buys an accuracy this
document does not need.

`#63` deliberately added no law here and its own note says the list is *"not
checked against it by any test"*, which is at least honest. This issue is to
stop it being half-honest.

## Fence (may touch ONLY)
- `docs/user-guide/watching.md`

## Blocked by
Nothing. The ruling this issue was waiting on is recorded above (2026-09-03):
the list becomes a declared sample and no law is written, so the fence stays one
document and needs no widening.

## Definition of done
- The heading over the gap list no longer claims exactness — "Exact strings you
  may see" becomes a heading that does not promise an enumeration.
- The list states in prose that it is a representative sample, so a reader who
  meets a string that is not listed has been told to expect that.
- The provenance line names all three sources the strings actually come from,
  not gaps.ts alone: the two further voices live in the burn panel's formatter
  and the empty-trace component.
- No law is added, and the fence stays one document. If a later reader wants the
  enumeration instead, that is a new issue with its own ruling.

### LINKED ISSUE #234: prd43 w6: the outbound sweep's vocabulary covers three more spellings

### LINKED ISSUE #266: prd43 w6: the route-count completeness sweep fits its timeout on the slowest runner

### LINKED ISSUE #270: prd43 w6: the sweeping tests have a budget they can actually meet under load

gabriel-canaan:
prd-43 wave 6, second PR. Three issues, one queue toll.

Closes #233
Closes #234
Closes #270

Review commit by commit. The commits that touch `route-class-law.test.ts` sit in
different regions of it — the outbound sweep vs the completeness sweep — and the
branch carries a merge of `main` (`3e4dd93`) rather than a rebase, so read it
with `--first-parent` if the duplicate history is noisy.

No commit count stated here, deliberately. The last two were wrong, and a typed
count in the PR whose own subject is "a count stated in prose is derived from the
thing it counts" is the wrong shape twice over.

## Why three issues and not three PRs — and why #266 is no longer one of them

#266 and #270 were the same defect — the completeness sweep timing out on
vitest's 5000 ms default — filed twice within hours and built twice in parallel,
by two sessions that could not see each other's lane. Neither carried a wave
(`prd43:` and `route-class-law:`), and the bundle unit is (milestone, wave), so
both were structurally unbundleable and each was heading for its own PR before
anyone made a decision. Both were retitled into wave 6, where #234 already claims
the same file.

**Then the duplication recurred one level up.** #266 landed on its own as PR
#269, merged 2026-09-04 21:30 — about fifteen hours after this PR was opened
carrying a byte-identical fix for it (`c2dc4e5` and `4ebddca` have patch-ids
identical to `b3fbe59` and `5cb6d5c` on main). The review merge reconciled them
with no loss. Worth recording rather than tidying away: a bundling rule enforced
at the issue level did not stop the same collision at the PR level, where nothing
was checking.

**#270 is not redundant, and that is why this PR still exists.** #266's fix raised
the budget on two of the three sweeping tests and left the third — `the
completeness sweep reads a non-empty file set` — on the 5000 ms default, while it
calls the same `sweptFiles()`. All three now share one `SWEEP_TIMEOUT_MS`, with
the measurement in its docblock: run alone ~4.2 s; under `VITEST_MAX_WORKERS=6`,
timeouts at 5.8 s, 8.1 s and 11.1 s. One failing run named its own cause —
`Preparing worktree (detached HEAD ...)`, a concurrent law test creating git
worktrees while this one shells out to git.

## Answering the review

**Finding: #234's justification for the quoted key is false.** Accepted in full;
the evidence reproduced independently before it was read. `biome.json` sets
`"preset": "none"` and never enables `useLiteralKeys`, so the row asserting `0`
pinned a real blind spot behind a claim about this repo that this repo does not
do — the exact failure this file exists to catch.

Answered in two commits, and the second takes **the review's second exit**:

- `1e5b4a2` (reviewer) corrected the stated reason and recorded, with evidence,
  that the row pinned a genuine blind spot rather than an unreachable shape.
- `38f2153` acts on that record and brings all three spellings **IN** —
  `'fetch'`, `"fetch"` and `['fetch']`. Built on the review fix, not over it:
  both wrong reasons stay named in the comment, in order, because the sequence is
  the lesson.

The deciding evidence is the reviewer's own: a real aliased outbound call written
`const { 'fetch': send } = globalThis` under a swept root leaves the law GREEN,
while the identical call with a bare key reddens it. #23's named vocabulary is a
real boundary and still excludes a `node:http2` import, a type member and a
destructuring parameter — but it names CONSTRUCTS, not spellings. A quoted or
computed key is the same renamed destructure of `globalThis.fetch`, so excluding
one spelling of an included construct is what lets README's "ten modules and
thirteen call sites" go quietly wrong. All three siblings are fixed together
rather than one of three.

**Finding: `Closes #266` is a no-op.** Fixed — removed above. #269 closed it.

**Finding: the body says "Five commits"; there are six.** Fixed by removing the
claim rather than correcting it; see the note at the top.

**Filed rather than carried:** the `NaN` test-title defect is #275, pre-existing
on `main`.

## What is the sibling case?

For #270: `the completeness sweep reads a non-empty file set` — same
`sweptFiles()` call, same exposure, left on the default budget. The other
per-file/per-claim asymmetry in that describe is `contentFor`'s `fileCache`,
already keyed per (file, gutter-mode) and needing no change.

For #234: the two spellings beyond the one that surfaced. `"fetch"` and
`['fetch']` are missed by `\bfetch\s*:` for the identical reason and lint stops
neither. Fixing the single-quoted form alone would have been this repo's
first-named defect shape, so all three are in one change.

Revert `DESTRUCTURE_KEY` to `\bfetch`. EXECUTED at `38f2153`:

```
Tests  3 failed | 91 passed (94)     exit 1
  × a single-quoted destructure key is the same call site (#234)
  × a double-quoted destructure key is the same call site (#234)
  × a computed destructure key is the same call site (#234)
```

Exactly the three rows, nothing else. A first attempt escaped to `\\bfetch` and
reddened eight rows including the bare-key controls — recorded because an
over-broad mutation looks like a stronger result and is a weaker one.

And the sibling check that matters more: widening a pattern can move the pinned
"ten modules and thirteen call sites". It does not — no tracked source uses a
quoted or computed key today, so this closes the hole without moving the number
it protects. Checked, not assumed.

For the earlier `TYPEOF_PREFIX` change, reverting it to `typeof\s+` reddens
exactly the paren-spelling row (`expected 1 to be +0`).

## Fence

No widening. `docs/user-guide/watching.md` is #233's; `route-class-law.test.ts`
is claimed by #234 and #270 — the overlap is why they belong in one PR rather
than two. `docs/prds/prd-43-the-claim-is-a-test.md` is the amendment commit,
which closes nothing and rides here per the amendment rule.

That amendment was itself corrected in `d19ea81`: it claimed both #266 and #270
were folded into this PR, and #266 was not.

## Gates

Run in the lane worktree on node 22.22.2 at `38f2153`: typecheck exit 0 · lint
exit 0 (884 files) · `VITEST_MAX_WORKERS=6 npm test` exit 0 — 409 files, 7406
passed, 3 expected fail, 2 skipped. Exit codes read directly, not through a pipe.


ciaran-slow (commented):
Reviewed commit by commit at `3e4dd93`. Everything below was run on node v22.23.2
against the **post-merge** tree, not the pre-merge branch.

All EXECUTED:

- **Both mutation claims reproduce independently.** Reverting `TYPEOF_PREFIX` to
  `typeof\s+` reddens exactly the new row and nothing else
  (`AssertionError: expected 1 to be +0`). Moving the normalisation back inside
  the `(file, claim)` loop gives `expected 11844 to be 987`, one test failed,
  91 passed — and the run took **7.79 s**, well inside `SWEEP_TIMEOUT_MS`. That
  is the #270/#266 argument demonstrated rather than asserted: the ceiling could
  not have caught it, the structural assertion is what does. My counts differ
  from the 979/11737 in your message because the corpus has grown since — which
  is exactly why removing the literals from prose was the right call.
- **The #234 sibling really is closed.** `grep` finds no surviving `typeof\s+`
  anywhere in the file; both alias-access patterns consume the one
  `TYPEOF_PREFIX` fragment, so they cannot drift apart again.
- **`GLOBAL_OBJECTS`'s leading `\b` holds the loosened prefix.** I went looking
  for a false negative from `\s*` making all whitespace optional —
  `typeofglobalThis.fetch` — and the word boundary rejects it. No hole there.
- **#233's provenance is accurate.** All three named sources exist and contain
  the exact strings: `packages/core/src/fleet/gaps.ts`,
  `CONDUCTOR_NOT_INSTRUMENTED_GAP` at `packages/web/src/panels/burn/format.ts:20`,
  and the `NO TRACE TELEMETRY` line at `packages/web/src/trace/EmptyTrace.tsx:10`.
- **Gates on the merged tree:** typecheck 0; lint 0 (884 files); suite 409 files
  / 7400 passed / 3 expected fail / 6 skipped, exit 0.

I also pushed the merge of `origin/main` into this branch (`3e4dd93`) to clear
the CONFLICTING state — three conflicts, all one shape, this branch's
`SWEEP_TIMEOUT_MS` superseding main's `30_000` literals from #269. The resolved
file is byte-identical to your version of it.

## One finding: #234's justification for the quoted key is false

**EXECUTED, three ways.** In `packages/server/src/api/route-class-law.test.ts`
(line 1285 at `3e4dd93`), the comment above the quoted-destructure row reads:

> since biome's `useLiteralKeys` flags an unnecessary quote on an
> identifier-safe property name and this spelling cannot pass this repo's own
> lint. Recorded OUT of the vocabulary: it is not a shape this codebase can
> ship, not a gap in the pattern.

That rule is not enabled in this repo.

**1. The config.** `biome.json` sets `"preset": "none"` and enables only
`correctness` and `suspicious`. `useLiteralKeys` is in the `complexity` group,
which appears nowhere. The single `overrides` entry only turns
`noUnsafeOptionalChaining` *off* for test files. `useLiteralKeys` is not
mentioned anywhere in the file.

**2. A direct lint probe** on exactly that spelling:

```
$ cat > packages/web/src/__probe.ts <<'EOF'
const { 'fetch': send } = globalThis
export { send }
EOF
$ npx biome lint packages/web/src/__probe.ts
Checked 1 file in 28ms. No fixes applied.
```

No diagnostic. The spelling ships.

**3. The consequence, planted and measured.** A *real* aliased outbound call in
that spelling under a swept root leaves the law green; the identical site
written bare reddens it.

```
packages/web/src/__probe_quoted.ts:
  const { 'fetch': send } = globalThis
  export async function ping(url: string) { return send(url) }

  -> vitest route-class-law.test.ts:  Tests 92 passed (92)        [GREEN]

same file, bare key (`const { fetch: send } = globalThis`):

  -> FAIL "are exactly these ten modules and thirteen call sites"  [RED]
```

So the row asserting `0` pins a genuine blind spot rather than recording a
shape the codebase cannot produce.

**It has two siblings the PR doesn't name.** `ALIAS_PATTERNS` requires
`\bfetch\s*:\s*\w+` (line 1117), so all three of these are missed while the bare
form matches — verified against the pattern directly:

```
MATCHED  const { fetch: send } = globalThis        <- control
MISSED   const { 'fetch': send } = globalThis
MISSED   const { "fetch": send } = globalThis
MISSED   const { ['fetch']: send } = globalThis
```

None of the three is stopped by lint, for the same reason.

I'm not asking you to reverse the decision — #23 narrowed the README's claim to
a **named vocabulary** on purpose, and "outside the vocabulary" remains a
legitimate answer for all three spellings. What I can't let stand is the stated
reason, because it is the exact failure this file exists to catch: prose that no
source checks, asserting something about the repo that the repo does not do.

Two honest ways out, and the choice is yours:

- **Keep it OUT, fix the reason.** Say the vocabulary doesn't name quoted or
  computed keys, and drop the lint claim. Cheapest, and defensible.
- **Bring it IN.** Widen the key fragment to admit an optional quote or bracket,
  and flip the row from `0` to `1`. Closes a real hole, but it's a wider change
  and it decides something #234 assigned to you, so I haven't pushed it.

If you keep it OUT, the row's label is worth rewording too — it currently reads
"cannot ship past this repo's lint", which restates the claim in the test name
where the next reader will trust it most.

## Two small ones

- **`Closes #266` is now a no-op.** #269 closed it on 2026-09-04 while this was
  open. `c2dc4e5` and `4ebddca` have patch-ids identical to `b3fbe59` and
  `5cb6d5c` already on main; they collapse harmlessly in the merge, but the body
  line should go so the closure record stays honest.
- The body says "Five commits"; there are six.

## What I did not run

`scripts/gate.sh` — that's the operator's step, not mine.

CI I did check rather than assume: all **10** expected legs report SUCCESS on
`3e4dd93` — `build-test-boot` x3, `pack-smoke` x6, `windows-suite`.

ciaran-slow:
Took the first of the two exits myself and pushed it as `1e5b4a2` — **your call still stands, and
I did not touch it.**

The spelling stays OUT and the row still expects `0`. What changed is only the reason: the comment
no longer claims biome's `useLiteralKeys` prevents it, and the row's label no longer restates that
claim in the one place a reader trusts most. It now says the vocabulary #23 named does not include
quoted or computed keys — a boundary decision — and records underneath, with the evidence, that
this pins a real blind spot rather than an unreachable shape.

I also named the two siblings (`{ "fetch": send }`, `{ ['fetch']: send }`) in the comment rather
than fixing them, and said explicitly that widening the key fragment is your open question.

**What I deliberately did not do:** add rows asserting those two count `0`. That would cement the
blind spot more widely while the real question is still open, and it is #234's decision to make,
not mine.

Gates on the pushed head: typecheck 0, lint 0 (884 files), suite 409 files / 7400 passed /
3 expected fail / 6 skipped, exit 0.

Under the review agreement this is a substantive fix, so it wants your confirmation before anyone
merges — I am not approving my own change into `main`. If you would rather have the second exit
(widen `ALIAS_PATTERNS` so all three spellings come IN and the rows flip to `1`), say so and I will
revert this and leave the whole question to you.

Still a draft, so nothing merges until you mark it ready either way. Everything else on the PR is
green: all 10 CI legs pass and `mergeStateStatus` is `CLEAN`.

The `NaN` test-title thing I mentioned is filed separately as #275 (prd43 w8, Backlog) — it is
pre-existing on `main` and not this PR's to carry.

PR #262, opened 2026-09-03
prd-43: wave 7 is ruled and holds two issues, and wave 5's order is forced

gabriel-canaan:
Grooming bookkeeping — no `Closes` line, the way PR #244 named waves 6 and 7.
This is the plan of record catching up with decisions already made and recorded
on the tracker.

## What changes

**Wave 7 is ruled, and it holds two issues rather than one.** #66 was booked as
an operator act, not dispatchable until its shape was chosen. The ruling is a
single dated note in `AGENTS.md` and `docs/prds/README.md`, with the 216
references in the other 110 files left untouched; the per-reference marker and
the mapping to surviving artefacts were both considered and rejected, and the
section now says why. #66's fence is groomed to those two files.

**Wave 7 gained #261 because of evidence already in this section.** The
2026-09-03 paragraph argued that a dated note is the *weaker* half — the cost
falls on **authoring** a citation rather than reading one — and that the guard
#66 suggests is load-bearing. The ruling agrees with that rather than overriding
it, so the note and the guard ship as one wave: #261 is fence-disjoint from #66
(`packages/server/src/doc-citation-law.test.ts` plus a baseline file, against
#66's two documents), proven with `scripts/fence-lint.sh 66 261`. #261 is
blocked by #241, which holds the same law file and is in flight.

**Wave 5's order is forced.** #235 is blocked by #220 (prd-30 wave 1), which has
no branch, no worktree and sits at Soon/Backlog, so wave 5 waits on another
programme and #24 is effectively a one-issue wave. #24 is otherwise dispatchable
now — PR #250 closed and freed `README.md`, wave 0's ADR-0032 landed. It is
deliberately **not** re-waved to bundle elsewhere: fence lint reports #24 and
wave 6's #234 both claiming `README.md`, so they can never be in flight together
whatever wave they carry. Recorded as a sequence, not a bundle.

## The constraint worth reading

#66 was measured on 2026-08-25, when the highest number that had ever existed in
this repository was 64. That has expired, and it invalidates the obvious wording
for the note.

This repo's own sequence began at **1 on 2026-08-21** and reached **260** today;
the prior tracker's citations in this corpus run to at least **649**. The two
sequences **overlap**, so no numeric threshold separates them — a note claiming
everything below 660 refers to a prior tracker would disown every live citation
the corpus has legitimately accumulated since August, #203 and #232 and #260
among them. The amendment records what is actually true instead: above the live
maximum a citation is unambiguously prior-tracker, at or below it the citing
[...]

Issue #276, opened 2026-09-04
prd17 w2: an operator act is recorded with the offset it was decided against


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/api/operator.ts`
- `packages/server/src/api/operator.test.ts`
- `packages/server/src/api/index.ts`
- `packages/server/src/api/route-class-law.test.ts`
- `packages/server/src/server/mutation-guard.ts`
- `SECURITY.md`
- `docs/architecture.md`
- `docs/adr/0008-localhost-only-single-origin-server.md`

The first two are new. The last six are claimed because a new gated mutation
moves a stated count in each of them — see the Why.

## Blocked by
Nothing. Wave 1 landed the operator families on #219.

## Why
prd-17 ruling 1 gives the operator three acts — `operator.ack`, `operator.verdict`
and `operator.note` — each **stamped with the log offset it was decided
against**, which is the PRD's answer to "a human clicked approve". #219 landed
the contracts. Nothing can raise one, because the instrument has no surface
through which a human act arrives: there is no route under
packages/server/src/api/ that accepts one.

`POST /api/operator/<act>`, classified `gated-mutation`, is the posture
`POST /api/rotate` already holds — a mutation of the instrument's own log, never
of the watched repo, so ruling 2's read-only law is untouched and no new route
class is owed.

**The footprint is wider than one route, and this is why the fence looks large.**
.swarm/coupling.txt names route-class-law.test.ts as a file that reddens from
outside any lane's fence: it pins the route total twice, exactly, so adding a
route fails it in a directory the editor never entered. That same law also
sweeps every tracked file for recognised route-count claims, and its own CLAIMS
registry names SECURITY.md, mutation-guard.ts, api/index.ts, architecture.md and
adr-0008 as carrying totals that a new mutating route moves. Each is fenced
here so the counts and the prose beside them move in the same commit.

If a route count turns up in a file NOT in the list above — README.md is the
one .swarm/coupling.txt warns about — that is a widening. Record it on this
issue with the widen script before changing the file, never after.

## Definition of done
- A route accepts each of the three operator acts and appends the matching
  event. One door, three acts.
- Every event carries `offset` — a zero-based **LINE index** into the named
  session's record, per ruling 1. Not the offset at write time if those can
  differ; the field exists to say *seeing what*.
  - **Corrected 2026-09-05.** This bullet said "the log's own `eventCount`"
    until then, which the shipped schema overrules: the operator ruled on
[...]

Issue #275, opened 2026-09-04
prd43 w8: every row in the outbound-call table states its real expected count


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/api/route-class-law.test.ts`

## Blocked by
**#234** and **#270** (both prd43 **w6**) claim this same file and are in flight in PR #272,
which adds three rows to the very table this issue repairs. Dispatch after #272 lands, and
re-run `scripts/fence-lint.sh` rather than trusting this paragraph.

This wave does not exist in the PRD yet. prd-43's Sequencing declares waves 0–7; **wave 8 owes a
Sequencing amendment**, and it must ride with wave 8's own PR — `docs/prds/prd-43-the-claim-is-a-test.md`
is currently inside PR #272's fence, so it cannot be amended before that lands.

**Model:** sonnet. **Wave:** 8.

## Why

Every one of the **51** rows in the outbound-call count table renders its own name wrong.
The `describe.each` label is:

```
'counts %s exactly %i time(s) — the COUNT, not merely red-or-green'
```

and its rows are `[label, source, expected]`. Vitest fills printf specifiers **positionally**
from the row array, so `%s` takes `label` and `%i` takes `source` — a string — which coerces to
`NaN`. The `expected` count, the one number the title exists to state, is never read.

EXECUTED on `3e4dd93`:

```
$ npx vitest run packages/server/src/api/route-class-law.test.ts --reporter=verbose
counts plain call on the global exactly NaN time(s)
counts optional call on the global exactly NaN time(s)
counts optional call on window exactly NaN time(s)
counts optional call, spaced exactly NaN time(s)
counts bare optional call exactly NaN time(s)
counts bare plain call exactly NaN time(s)
...
$ ... --reporter=verbose | grep -c "exactly NaN time(s)"
51
```

This matters more here than a cosmetic title usually would, because the label is making a claim
about rigour — *"the COUNT, not merely red-or-green"* — and that claim is what the row's own name
then fails to show. The table's stated purpose is that a double count and a single count must be
distinguishable; a reader scanning the reporter output to check exactly that sees `NaN` on every
line and learns nothing. The failure output is unaffected (the assertion still reports
`expected 1 to be +0`), so nothing is mis-asserted — this is a diagnostic defect, not a
correctness one.

It is also load-bearing for review. PR #272's own mutation proof, quoted in its commit message,
carries `exactly NaN time(s)` in the pasted failure line; a reviewer reading that evidence cannot
[...]

Issue #273, opened 2026-09-04
prd17 w5: the landing gate says what it decided


gabriel-canaan:
## Fence (may touch ONLY)
- `scripts/gate.sh`
- `packages/server/src/gate-honesty-law.test.ts`

Out of scope, EXPLICITLY OUT: writing the verdict anywhere. That is #274.

## Blocked by
Nothing. #217 merged on 2026-09-04 in PR #267, so ADR-0036 fixes the line shape
this issue builds to.

Stacked with **#274**, never parallel: both claim the gate script, so the lint
reports an OVERLAP and they may not be in flight together. This one lands first.

## Why
prd-17 ruling 1 defines gate.verdict and #219 landed the contract. Nothing emits
one, so a landing is still invisible to the record: the gate script is the
instrument's own judgement on a lane, and the log has never heard of it.

Every field the verdict wants is already in the script's hands: it takes the
handle as `$1` and the load-batch count as `$3`, tracks `MERGED`, and knows
which check failed. What it does not have is a category vocabulary, its own
output as a hashable artefact, or a line to put them on.

**The shape is prd-17 ruling 6, and it is not the gate.verdict payload.** Read
that ruling before you start. ADR-0036 (docs/adr/0036-a-beacon-is-a-line-in-a-watched-directory.md)
fixes a beacon line as `{ "v": 1, "at": <epoch ms>, "writer": …, "kind": …,
"lane"?: …, "detail"?: … }`, and states that **extra keys are ignored by the
collector and covered by its digest**. So the verdict rides as extra keys on
that line — not in `detail`, which is a string capped at 512, and not as a
gateVerdictPayloadSchema object, which is a different shape the door does not
carry. Deriving a `gate.verdict` event from the written line is a later wave and
is EXPLICITLY OUT here.

**The second fenced file is not optional.** .swarm/coupling.txt carries both
ends of that coupling. The law anchors its assertions on exact line TEXT pulled
from the tracked gate script, so rewording an anchored line throws at
COLLECTION — "expected exactly one line ... found 0" — taking that whole file's
assertions out of service at once rather than reddening one test. It also PINS
derived counts over the script's `VAR=$(...)` command-substitution assignments:
how many exist, how many are flagged structurally unchecked, and how the checked
ones split between the same-line `|| fail` form and the next-line `_RC=$?`
capture. Capturing the script's output for a digest introduces exactly such an
assignment, so those pins move and the doc comment beside them is corrected in
[...]

PR #284, opened 2026-09-05
prd-43 wave 7: prior-tracker citations are told apart, new ones are refused, and ADR-0032 names its writer

### LINKED ISSUE #66: prd43 w7: the docs cite a tracker that no longer exists

### LINKED ISSUE #261: prd43 w7: a citation above the live maximum cannot enter the corpus
## What becomes true

No tracked markdown can newly cite an issue number this repo's tracker has never
issued. The 216 existing prior-tracker citations stay — they are provenance, per
the ruling on #66 — but the class cannot return silently.

## Why

#66's Definition of done calls this out and stops short of ruling it: *"a test
asserting no tracked markdown cites an issue number above the repo's real maximum
would have caught this the day it appeared."* It would have. The corpus
accumulated 216 unresolvable citations across 111 files without one red build,
and the file that governs how everyone works — `AGENTS.md` — was among the
affected.

It is deliberately NOT part of #66, because the naive form of this law fails
immediately on all 216 references, and deciding how it tolerates them is a
design decision rather than a sweep.

## The decision this issue owns

A citation above the live maximum is unambiguously prior-tracker. At or below it,
the two sequences OVERLAP and no threshold separates them — this repo's own
numbering began at 1 on 2026-08-21 and reaches 260 today, while the prior
tracker's citations here run to at least 649. So the law cannot be a simple
ceiling; it needs one of:

- a **committed baseline** of the citations that exist today, the pattern
  `.windows-known-failures` already uses in this repo — new violations redden,
  the recorded ones do not, and the file is the visible statement of what is
  owed; or
- a predicate **scoped by document date**, so only prose written after the
  migration is held to the live maximum.

Pick one and say why in the law's own docblock. The baseline is the more likely
answer because a document's date is not reliably recoverable from its content,
but that is an argument to make in the diff, not a conclusion this issue hands
down.

## What the law must not be

- **Not scoped by filename.** AGENTS.md records the failure directly: a
  fixture-hygiene guard read `startsWith('claude-code-')` and so never saw three
  older captures sitting beside the ones it checked, then asserted truthfully and
  uselessly that nothing was wrong. Scope it by what a file IS — tracked, and
  markdown — not by what it is called.
- **Not vacuous on an empty sweep.** A pass that matches zero files must fail
  loudly. The citation law in this repo already has that shape; copy it.
- **Not a second spelling of the live maximum.** The ceiling is derived at run
  time from the tracker, or from a single pinned constant with its derivation
  stated — never typed twice.

## Definition of done

- A law fails when a tracked markdown file cites a number above the live maximum
  that is not in the recorded baseline (or, if the date-scoped shape is chosen,
  that appears in a document written after the migration).
- The mutation that proves it: add one such citation to any tracked markdown and
  the law reddens, naming the file and the number. Stated in the PR, EXECUTED.
- The law is red for the right reason before it is green — run it against the
  corpus as it stands, record the count it reports, and only then record the
  baseline. A baseline captured without ever seeing the law bite is a baseline
  nobody has tested.
- The chosen shape's rationale is in the docblock, with the rejected alternative
  named.

## Fence (may touch ONLY)
- `.citation-prior-tracker` (new, if the baseline shape is chosen)

## Blocked by
**#241**, which holds `doc-citation-law.test.ts` and is in flight now. Same
fence, so this cannot share a wave with it — dispatch after #241 lands and
re-run the lint rather than trusting this paragraph.

Parallel-safe with **#66** (prd43 w7), which claims `AGENTS.md` and
`docs/prds/README.md` and nothing else. These two can share the wave's PR: #66
writes the note that explains the citations, this writes the law that stops more
arriving.

### LINKED ISSUE #264: prd43 w7: ADR-0032 names the writer that actually writes, and the subject it actually rules

gabriel-canaan:
prd-43 wave 7 — the docs tell prior-tracker citations from live ones, a law stops new
ones arriving, and ADR-0032 names the writer that actually writes. Plus the Sequencing
amendment that makes the PRD describe the wave it is landing.

Closes #66
Closes #261
Closes #264

## Commits, in review order

|---|---|
| `7421264` | AGENTS.md and docs/prds/README.md carry a dated note telling prior-tracker citations from live ones (#66) |
| `6103aef` | a `#NNN` citation above the live tracker maximum cannot enter the corpus (#261) |
| `406d2d5` | ADR-0032 names `restore.ts`, and its subject is sessions only (#264) |
| `169f56a` | the ceiling derives offline, so the law cannot fail open (review of #261) |
| `6830eee` | ADR-0032 no longer says README is both right and wrong (review of #264) |
| `ed14f78` | the law's own stated counts are derived, dated, or gone (review of #261) |
| `323ea6d` | the ceiling derives from `origin/main`, not from every ref (review of #261) |
| `96eb891` | the PRD's Sequencing says wave 7 is three issues, and declares wave 8 |

## What is the sibling case?

**`isPinnedArtefact`'s exemption, and it is real and NOT fixed here.** `323ea6d` stops
treating object-store presence as evidence of landing in `liveMaximum()`. The same
assumption sits one screen away in the pin check, which resolves a declared sha with
`git cat-file -e <sha>^{commit}` — the object database, not reachability from any ref.
A document can therefore exempt itself from this entire law by pinning a sha that never
landed anywhere.

EXECUTED, using the scratch commit this PR's own mutation left behind:

```
git branch -a --contains fbc8073    -> (no output: reachable from NO ref)
git cat-file -e 'fbc8073^{commit}'  -> exit 0  (exemption would be granted)
```

Pre-existing, not introduced here. Left out deliberately: the exemption is a separate
decision (what makes a document a dated artefact) with its own rejected alternatives,
and widening a review fix to cover it would put an unreviewed design change inside one.
Owed as a follow-up issue.

The certified probe plants a citation above the `origin/main` ceiling in a swept
document and requires the law to redden — deliberately probed on the CLASS rather than
the spelling the review seats used. Both seats probed with a local BRANCH (#900, #999),
so this used a TAG — a different ref class `--all` also walks — carrying
`Merge pull request #977 from operator/tag-probe`, moving the `--all` ceiling 272 -> 977
while `origin/main` held at 272.

```
--all ceiling (977), planted #731:  the ceiling-violation test PASSED — accepted silently
origin/main ceiling (272), same:    FAIL [{"file":"docs/architecture.md","num":731}]
```

The control is stated in full because it did not behave as predicted: under `--all` the
run still exited 1, but on a DIFFERENT test — "every excluded directory still trips the
detector when the exclusion is bypassed" — because a 977 ceiling puts `docs/research/`'s
citations below it. A control that fails for the wrong reason is not a control, so both
halves are on the record. The finding is the silent acceptance.

**What it would NOT catch:** the absent-ref arm. `git log origin/main` exits 128 when
that ref is missing (EXECUTED: `git log no-such-ref/main` -> 128), which happens in a
fork clone or a CI checkout that fetched only the PR ref. That arm is covered by the
exit-128 measurement and the `recordedMaximum()` floor, not by this probe. Nor does the
probe reach the sibling above: a pinned-sha exemption bypasses extraction entirely,
before any ceiling comparison.

## Fence

No widening. Six files, every one inside a declared fence (`main...HEAD`, three-dot):
`AGENTS.md` and `docs/prds/README.md` (#66) · `.citation-prior-tracker` and
`packages/server/src/doc-citation-law.test.ts` (#261) · `docs/adr/0032-*.md` and
`docs/adr/README.md` (#264).

`169f56a` recorded a widening to `.github/workflows/ci.yml` for a `GH_TOKEN`; the same
commit's offline derivation made it unnecessary, and `ci.yml` now diffs empty against
`main`. Narrower than declared.

`96eb891` adds `docs/prds/prd-43-the-claim-is-a-test.md`, which belongs to no issue's
fence — it is the plan of record being amended to describe the wave it ships with, the
shape #262 set. It is file-disjoint from all six above (verified, not asserted).

## Verification

Two review rounds, both against isolated worktrees with executing seats.

- **Round 1** on `f6cd413`; its findings produced `169f56a`, `6830eee`, `ed14f78`.
- **Round 2** on `ed14f78`: seats were **sonnet at xhigh** and **codex gpt-5.6-terra at
  xhigh**. The Claude seat was swapped down a tier from Opus because the repairs carry
  `Co-Authored-By: Claude Opus 5` — a seat sharing the builder's model is the worst
  place for a shared blind spot. Family overlap is noted rather than claimed absent: a
  tier swap softens it, only a different family removes it.

Both seats returned the same blocking finding independently, with different evidence,
and the orchestrator reproduced it against the code before acting:

```
comm -13 <(git log origin/main ...) <(git log --all ...)  ->  96 97 101 102
```

All four carried by `refs/remotes/origin/prd44` — the abandoned integration branch
AGENTS.md documents as an incident, still pushed and still fetched by an ordinary clone.
Fixed in `323ea6d`.

**Two ledger rows are FORCED and say so.** `323ea6d` was written after both seat reports
and after the round budget (2 of 2) was spent, so it is gated rather than seat-reviewed:
`lane-precommit.sh --repair` passed 11/11 including the certified mutation above.
`96eb891` is docs-only in a directory both laws exclude. Landing without a third round
was the operator's call, recorded here rather than in a terminal.

## Gates

EXECUTED on `96eb891`, statuses read directly rather than through a pipe:

```
npm run typecheck              -> 0
npm run lint                   -> 0
VITEST_MAX_WORKERS=6 npm test  -> 0   (413 files, 7483 passed, 3 expected fail, 2 skipped)
```

One caveat on round 2's evidence: the Codex seat's full-suite run was invalid — its
sandbox denied `spawnSync git` (`EPERM`), so subprocess-heavy suites failed or collected
zero tests. It flagged this itself and did not treat it as product evidence. Its
typecheck and lint runs were clean, and its grammar probes ran.


ciaran-slow:
## Review — read commit-by-commit, in a fresh clone, Node 22.23.2

Still a draft, so this is not an approval; it is the review, done, so nothing is
waiting on the looking. Two mechanical fixes are pushed (`2b9e9c1`, `ac5d14e`);
one substantive finding is below and is **yours to rule on**, not mine to fix.

Gates, on `96eb891`, statuses read directly rather than through a pipe:

```
npm run typecheck               -> 0
npm run lint                    -> 0   (895 files)
VITEST_MAX_WORKERS=6 npm test   -> 0   (413 files, 7479 passed, 3 expected fail, 6 skipped)
gh pr checks 284                -> 11 legs green (build-test-boot x3, pack-smoke x6, windows-suite)
```

**EXECUTED — the ceiling law bites, on my own mutations rather than yours.** I
deliberately did not re-run the probe in the PR body; a control I did not choose
is not a control.

```
#900 in docs/architecture.md   -> FAIL [{"file":"docs/architecture.md","num":900}]
#272 in docs/architecture.md   -> 43 passed        (at the ceiling)
#273 in docs/architecture.md   -> FAIL [{"file":"docs/architecture.md","num":273}]
#900 in AGENTS.md              -> FAIL [{"file":"AGENTS.md","num":900}]
```

The boundary is exact, and `AGENTS.md` is genuinely in scope, which is the claim
the wave-7 coupling section rests on.

**EXECUTED — #264 is right, checked against the source rather than against the
ADR.** `checkpoint.ts` imports `copyFile, readdir, readFile, rm, stat` and no
`writeFile`; its one `copyFile` is the git index at `checkpoint.ts:131`; the
write is `restore.ts:238-244`. `README.md:272-280` does already name the
checkpoint's ref and event-log entry beside the recording directory and only the
synthesized session under `~/.claude/projects/<slug>/`, so the amendment's claim
that the error is confined to this record holds.

---

### Finding — substantive. The `origin/main`-absent arm does not do what its comment says

`liveMaximum()`'s catch (`doc-citation-law.test.ts:1666`) says:

> `origin/main` absent — a fork clone, a CI checkout that fetched only the PR
> ref, a mirror under another remote name. […] Falling through leaves `merged`
> empty and the floor below does the rest.

The floor does not do the rest. **EXECUTED**, on this branch, clean tree:

```
$ git update-ref -d refs/remotes/origin/main
$ npx vitest run packages/server/src/doc-citation-law.test.ts
AssertionError: [{"file":"docs/adr/0002-one-reducer-for-live-and-replay.md","num":268}]
  Tests  1 failed | 42 passed (43)
```

The mechanism: the floor is `recordedMaximum()` = 265, but the corpus carries
ten `(file, #NNN)` pairs in the 266–272 window. Nine have baseline rows. One
does not — `docs/adr/0002` `#268`, added by `8d249b4` on 2026-09-04, the same
day the baseline was measured at 265. It needs no row today because the derived
ceiling is 272; it becomes an unbaselined violation the moment the ceiling drops
to the floor.

Three things make this worth a ruling rather than a note:

- **It fails in exactly the environment the arm was written for.** The comment
  enumerates fork clone, PR-ref-only checkout, renamed remote. All three now
  produce a red law naming a legitimate live citation — a false positive, which
  is the failure direction the conservative-lower-bound argument explicitly
  chose *against* tolerating for real citations.
- **It degrades on a timer.** The gap is `recordedMaximum()`..`liveMaximum()`,
  which widens with every PR that lands. It is one file today. It is however
  many 266+ citations accumulate before someone re-measures the baseline.
- **It is the only arm of this law with no test.** Twelve tests cover the pure
  checker, the boundary, the non-empty sweep, the exclusion doing real work,
  baseline parse/duplicates/honesty and the real sweep. The fallback has none —
  and it is the one whose behaviour is misdescribed. In the PR whose PRD thesis
  is *the claim is a test*, that is the pointed one.

I did not fix it, because the fix is a choice among at least three and each has
a real cost, which makes it yours:

1. **Fall back to `HEAD`** when the ref is missing. Correct here (`HEAD` carries
   main's history in every one of the three enumerated cases). Cost: it partly
   re-opens what `323ea6d` closed — the branch under test can move its own
   ceiling — though only on the fallback path, and only via a forged merge
   subject rather than any stale ref.
2. **Fail loudly.** `origin/main` missing is not a violation of the law; a red
   test naming `docs/adr/0002` says the corpus is wrong when the *derivation* is
   unavailable. An explicit "the ceiling cannot be derived" is at least honest
   about which thing broke.
3. **Restate the comment** and accept the behaviour, on the grounds that no
   environment this suite actually runs in lacks the ref (both suite legs use
   `fetch-depth: 0`, which supplies it).

(3) is defensible and cheapest — but if it is the answer, the comment should say
"this arm is a backstop against a thrown law, not a working ceiling; a clone
without `origin/main` may redden on live citations", because the sentence there
now promises the opposite. Whichever you take, it wants a test: a rigged
`recordedMaximum`/`liveMaximum` pair through `ceilingViolationsIn` reaches it
without touching git at all.

---

### Two fixes pushed

Both mechanical, both inside already-declared fences, both re-verified green
(43 passed / lint 0 / typecheck 0). Revert either if you disagree — they are
prose, and it is your PR.

**`2b9e9c1`** — `doc-citation-law.test.ts:1491` still spelled the derivation
`git log --all`. `323ea6d` changed the code and rewrote `LANDING_REF`'s own
comment, but the "Deriving the live maximum" section's summary line one screen
above kept the rejected implementation. That section is where a reader stops,
because it is the one that explains the design. Now names `LANDING_REF` instead
of re-spelling a ref, and its parenthetical no longer credits `fetch-depth: 0`
to sha resolution alone — a pinned remote ref needs the same flag for a second
reason, and stating one of two reasons is how the flag gets removed later.

**`ac5d14e`** — `docs/prds/README.md` sent the reader to `AGENTS.md` "for the
current measurement […] since that number moves". Round 2 removed that figure
from `AGENTS.md` and did not touch this half. Three ways wrong: `AGENTS.md`
states no measurement; the number it cites as moving (how far the dead tracker
reaches) is frozen, while the one that moves is the live maximum it declines to
state; and round 2's conclusion was *derive it or don't state it*, not *keep it
in one place*. **This is the sibling case, in the PR body's own terms**: a
two-file pair where the fix landed on one half, and nothing was going to catch
the other, because the ceiling law flagged the file it flagged and does not read
prose for coherence with a file it points at.

---

### Your call, offered not asserted — `docs/prds/` is exempt, and #66's own worked example lives there

**EXECUTED:** `#900` planted in `docs/prds/prd-43-the-claim-is-a-test.md` — the
file `96eb891` amends — leaves the suite at **43 passed**. That commit adds
`#275` and `#276`, both above the 272 ceiling; only the exclusion keeps them
green.

The exclusion is documented, inherited from the sibling law with a stated
rationale, and I am not calling it wrong. But #66's evidence names
`docs/prds/prd-42-one-path-one-spelling.md` standing on phantom
`#217`/`#228`/`#299`/`#401` as the worked example of the class, and a live PRD
is not a dated artefact in the way a run record pinned to a sha is — prd-43 is
being amended in this very PR. So "the class cannot return" holds for the corpus
minus the tree the class was demonstrated in. If that is the intended shape, it
is worth one sentence in the docblock saying so out loud rather than leaving it
to be inferred from `EXCLUDED_DIRS`; if it isn't, it is a wave-8 issue, not this
PR's problem.

The `isPinnedArtefact` sibling the PR body raises and defers is the right call
and I have nothing to add — except that it is owed an issue, and I could not
find one open for it.

### What I did not run

`scripts/gate.sh` (operator's step, and this is a draft), and the load-batches
concurrency pass. No merge, and nothing that touches your worktree.

ciaran-slow:
**Correction to the review above, and a re-run on the pushed head.**

I wrote "11 legs green". It is **10** — `build-test-boot` ×3 + `pack-smoke` ×6 + `windows-suite`, which is what my own parenthetical said and what AGENTS.md states ("three jobs long and ten legs wide"). A miscount off a truncated `statusCheckRollup`. Stating a count I had not derived, in a review of the PRD that exists to stop exactly that, is worth correcting rather than leaving.

Re-measured on `ac5d14e` (the head after my two fixes), derived rather than eyeballed:

```
gh pr view 284 --json statusCheckRollup \
  -q '[.statusCheckRollup[]|.conclusion]|group_by(.)|map({(.[0]):length})|add'
-> {"SUCCESS":10}
```

Nothing else in the review changes: the mutation results, the `origin/main`-absent finding and the gate statuses were each run and read directly.

ciaran-slow:
**Finding fixed — `98f9d44`.** Pushed rather than handed back, since you asked; it is a behaviour change to your law, so revert it if you disagree with the shape.

**The fix is this file's own existing ruling, not a fourth option.** `repoIsShallow()` exists so a default-depth checkout *"fails HERE, naming the cause, instead of surfacing as a bad-pin report against a document whose pin is perfectly honest"* (review of #229). A missing `origin/main` is that failure verbatim, one precondition over, and it was absorbed instead of asserted. So: `landingRefResolves()` gets its own test, and `liveMaximum()` throws a message naming the missing ref, giving the fetch command, and saying in terms that this is **not** a corpus violation.

I did not take either fallback. Higher (derive from `HEAD`) re-opens what `323ea6d` closed — the branch under test moving its own ceiling. Lower is the false positive. Neither is a ceiling, and the old path went red anyway while blaming the wrong document, so refusing to certify is strictly better than what was there. `recordedMaximum()` keeps its floor role on the ordinary path.

**EXECUTED, before and after, same command:**

```
$ git update-ref -d refs/remotes/origin/main
$ npx vitest run packages/server/src/doc-citation-law.test.ts

before: AssertionError: [{"file":"docs/adr/0002-one-reducer-for-live-and-replay.md","num":268}]
after:  Error: origin/main does not resolve in this clone, so the live tracker maximum
        cannot be derived ... NOT a corpus violation: falling back to the baseline's own
        recorded maximum would report every citation that landed since it was measured
        as prior-tracker.
```

`docs/adr/0002` is no longer mentioned.

**What the new tests would not survive.** Two added; the second is the load-bearing one, on rigged input so it stays true once the baseline is re-measured and the live `docs/adr/0002` instance stops being one:

```
ceilingViolationsIn([docs/adr/0002 cites #268], 272, {}) -> []                 (derived)
ceilingViolationsIn([docs/adr/0002 cites #268], 265, {}) -> [{...,num:268}]    (floor)
```

Restore the fallback and the second line reddens, because the fallback *is* the second line.

**Regression — the law still bites** (ceiling 272): `#900` → `FAIL [{"file":"docs/architecture.md","num":900}]`; `#273` → `FAIL [...,"num":273]`; `#272` green.

**Gates on `98f9d44`:** `npm run lint` → 0 · `npm run typecheck` → 0 · `VITEST_MAX_WORKERS=6 npm test` → 0 (413 files, 7481 passed) · this file 45 passed, was 43.

Fence unchanged — `doc-citation-law.test.ts` is #261's own.

Still open, and deliberately not touched here: the `isPinnedArtefact` sibling you flagged in the PR body (owed an issue, I could not find one), and whether `docs/prds/` should stay exempt from the ceiling sweep. Neither blocks this.

PR #286, opened 2026-09-05
prd17 w2: the operator's acts reach the record, and the tide shows them

### LINKED ISSUE #276: prd17 w2: an operator act is recorded with the offset it was decided against

### LINKED ISSUE #277: prd17 w2: a gate verdict, a summons and an operator act are marks on the tide
## Fence (may touch ONLY)
- `packages/web/src/tide/chapters.ts`
- `packages/web/src/tide/chapters.test.ts`
- `packages/web/src/tide/ChapterMarks.tsx`
- `packages/web/src/tide/ChapterMarks.test.tsx`
- `packages/web/src/tide/fixtures.ts`
- `packages/web/src/tide/index.ts`

- `packages/web/src/tide/markCoalesce.test.ts` (the Chapter fixture literals only)

## Blocked by
Nothing. Wave 1 landed the families on #219, and this reads events rather than
folded state, so it does not wait on any emitter.

## Why
prd-17 ruling 4: "gate holds and merges become chapter marks; summonses become
marks with their clearances; the operator's verdicts appear where they happened.
No new surface — the existing mark lane and hover cards simply gain the mark
kinds they were always missing."

Today the mark lane has four kinds — lane-born, lane-landed, gate-held and
session-boundary — and its module note reasons about a world that ended when
#219 landed. Two sentences in that note are now false and are corrected here:
"attention-summons onset has no event" (summons.raised exists) and "There is no
session.ended type" (session.closed has existed since prd-40).

Cheaper than it reads: chaptersFor takes a readonly array of events, not a
SessionState, so no reducer arm has to fold anything for this to work. The tide
fixtures are enough to prove the marks before any emitter exists.

## Definition of done
- New chapter kinds for the families ruling 4 names: a summons and its clearance,
  a gate verdict, and an operator verdict. Each is emitted once per matching
  event, the way gate-held and session-boundary already are.
- The three existing laws in chapters.test.ts still hold over the new kinds:
  deterministic (same events in, byte-equal chapters out, no clock, no unseeded
  random); prefix-consistent (chapters over a time-prefix equal chapters over
  the whole log filtered to that prefix); one row per real fact.
- A summons that is raised and never cleared is a mark, not an error. Ruling 1's
  own schema note is explicit that a session can end mid-alarm, and the reducer
  arm says the same. This is the case a pairing implementation gets wrong.
- The hover label follows the existing voice — the `chapterLabel` shape, not a
  new sentence form.
- The module note's two false sentences are corrected in the same commit, and
  the corrected text says what is true now rather than deleting the reasoning.

## Sibling case
`laneOf` owns lane identity and this file imports it rather than re-deriving it —
the module note warns that copying its switch here is drift by construction. The
new kinds each need a lane, and the miss is reaching for `event.payload.lane`
directly on the ones that happen to have that field. Two of the new families
name a lane in their payload and one names a handle; neither is the same
question `laneOf` answers.

Second: the mark lane is the dock's ONLY glance layer since ruling 13 cut the
density band, and the module note says holding that line matters more now, not
less. Emitting a mark per summons in a flood is how the lane becomes unreadable.
State what happens under repetition, and pin it with a test.

Emit the new kinds unsorted — append them after the existing sort instead of
inside it. Every single-kind test passes; only an assertion over a log mixing an
old kind and a new one at interleaved timestamps catches it.

Second: make the summons mark depend on a matching clearance. A fixture whose
summonses all clear passes at either behaviour, so the fixture must include one
that never clears.

**Model:** sonnet. **Wave:** 2. **Milestone:** prd17.

## Coupling with #276, checked before dispatch
wave-coupling.sh reports eight hits between this issue and #276. All eight are
answered here so nobody re-derives them at assembly:

- Six are basename collisions. Both trees contain an index.ts and a fixtures.ts;
  #276's files reference the server api one, this issue owns the tide one. They
  are different files with the same last path segment.
- Two are real references: docs/architecture.md, which #276 owns, mentions
  ChapterMarks and enumerates mark kinds. **It is not a forced edit.** That
  enumeration sits inside the historical account of prd-13 ruling 12 — "Ruling
  12 amended it to a sparse chapter-mark lane ... one mark per
  lane-born/landed/gate-held/attention-summons/session-boundary moment" — and it
  already names attention-summons, which the shipped code has never had. It
  states what the ruling decided, not what the code currently emits, and this
  issue moves the code toward it rather than making it false. The section that
  does describe shipped behaviour covers layout, portaled hover cards and zoom,
  none of which this issue changes.

So the two are peers, not a stack. If a change here DOES falsify a sentence in
architecture.md, that is a widening onto #276's fence and it is recorded on this
issue with the widen script before the change, never after.

gabriel-canaan:
Closes #276
Closes #277

prd-17 ruling 1 gave the operator three acts and ruling 4 gave the timeline the
mark kinds it was missing. #219 landed the contracts and nothing could raise
one, because the instrument had no surface through which a human act arrives.
This wave is that surface, and the marks that make it visible.

Also carries the PRD amendment recording the two rulings the operator made
while this wave was being verified — the coordinate ruling below, and the
correction of a status line that had been wrong for two weeks.

## What is in it

- **`POST /api/operator/:act`** — one door, three acts, the app's SEVENTH gated
  mutation. It mutates the instrument's own log and never the watched repo, so
  ruling 2's read-only law is untouched and no new route class is owed. It is
  **not** an approval workflow and never will be: it records an act the operator
  has already performed, and prd-17's non-goals make that permanent.
- **Four new chapter kinds** on the existing mark lane — a summons and its
  clearance, a gate verdict, an operator verdict. No new surface, per ruling 4.
- **The route-count claims moved together**: 6 gated mutations became 7, 10 total
  became 11, across `SECURITY.md`, `mutation-guard.ts`, `api/index.ts`,
  `docs/architecture.md` and ADR-0008's amendment. `route-class-law.test.ts` is
  green **without any pin being loosened** — the two hard pins moved from 25/20
  to 26/21 because a real route was added, and the gate-presence law reddens if
  the new route's `preHandler` is removed.

## The `offset` ruling, and a correction to #276's own text

#276's Definition of done said `offset` was "the log's own `eventCount`". **The
shipped schema overrules that**, and the issue text has been corrected (with the
original wording quoted in place). `operatorActBaseSchema` records an operator
ruling of 2026-09-04: `offset` is a zero-based **line index**, not a fold count,
because `reduce()` restarts the count at a session boundary and it counts only
events a reader *folded* — so an act stamped with one drifts the day a later era
understands a line this one did not.

Both review seats raised the contradiction unprompted and both adjudicated the
schema as controlling. The route therefore takes `offset` from the request body
and re-derives nothing, which is also what #276's own "not the offset at write
time if those can differ" clause requires.

**Consequence, stated rather than hidden:** nothing establishes that
`(sessionId, offset)` names a real place, so the record can carry an operator act
whose coordinate points at a line that never existed. The operator ruled on
2026-09-05 that this is correct and why — the server cannot check it without
refusing the review verdict, which is the act ruling 1 most exists to make
reconstructible. The ruling is in `docs/prds/prd-17-complete-record.md` in this
PR, including the condition that would reopen it.

## What is the sibling case?

**The route:** three acts, one door. A door that validated the shared base and
waved the extension through would accept `POST /api/operator/note` carrying no
words. Each act is parsed by its own schema and rejected on its own terms.
Verified: substituting the ack schema for verdict, and for note, each reddens 2
tests.

**The gate:** the gate-presence law covers both directions — a stray gate on a
plain read is as much a defect as a missing one on this mutation. Removing the
new route's `preHandler` turns the law red (3 tests, including the law itself).

**The marks:** `laneOf` owns lane identity, and the miss is reaching for
`event.payload.lane` on the families that happen to have that field. Verified
both ways: routing the summons lane through `laneOf` is CAUGHT (it has no case
for `summons.raised`), and taking `operator-verdict`'s lane from `sessionId`
instead of `subject` is CAUGHT.

**The glyph:** `gate-verdict` renders **two** inline-painted spans, four lines
apart, each with its own copy of the same ternary. The first version of the test
read only the first one — the stem — while its local was named `head`. Inverting
only the head's ink rendered a held gate green with every test passing. It now
asserts every painted span, so the count is part of the assertion.

Each run against its whole package with a clean-tree control either side.

| mutation | result |
|---|---|
| hard-code `offset: 0` in the handler | **4 red** |
| ack schema used for `verdict` / for `note` | **2 red** each |
| drop the route's `preHandler` | **3 red** |
| drop the read-only refusal | **1 red** |
| hoist any one of the 8 chapter kinds past the sort — **append** | **8/8 red** |
| hoist any one of the 8 chapter kinds past the sort — **prepend** | **8/8 red** |
| delete `chapters.sort` entirely | **2 red** |
| drop or reverse the `kind` tiebreak / the `lane` tiebreak | **1 red** each |
| move the sort above the `bornAt`/`landedAt` loops | **1 red** |
| `glyphHeldOf` → `return null` | **2 red** |
| `gate-verdict` head-only ink pinned to done | **1 red** |

The ordering matrix started at **6 of 16** and is now **16 of 16**. An
independent seat re-derived the baseline: the parent tree kills 15 of 16, so
these commits are load-bearing rather than decoration.

## Fence widening

`packages/web/src/tide/markCoalesce.test.ts` — two lines, both field additions to
a `Chapter` literal forced by the type gaining required `held`/`verdict`. It is
recorded on #277 as a fence bullet with its rationale, and restated here because
`AGENTS.md` asks for it on the PR. Verified minimal and complete: every `Chapter`
literal in `packages/web/src` lives in 5 files and this is the only one outside
the declared fence.

## Verification

Four independent passes, two seats each (Claude Opus `xhigh`, Codex
`gpt-5.6-terra` `xhigh`), never unanimous:

| pass | verdicts | findings |
|---|---|---|
| initial | Opus merge / Codex block | 3 |
| fix re-review 1 | Opus merge / Codex block | 5 |
| fix re-review 2 | Opus merge / Codex block | 3 |
| first pass on the late commits | Opus **ready to merge** / Codex not-ready | 2 |

Every finding is fixed. **Worth knowing for review: three of the orchestrator's
own repairs were themselves defective**, each closing the case that prompted it
and missing a sibling — an ascending-order fixture that hid "never sort at all",
a `kind` tiebreak satisfied by V8's stability, a `session-boundary` alone at the
maximum ts, a `gate-held` alone at the minimum. The final commits stop asserting
cases and assert the property: every declared kind must be exercised, must be
interleaved with the rest, and the output must already be in order. A kind added
later with no ordering witness now fails a test by name.

**One commit is FORCED in the verify ledger**: `4e4b3e5` has not been read by a
seat. It applies an interleaving assertion that Opus itself authored and
executed, closing a finding both seats raised independently; the orchestrator
re-ran it (old clause green on the blind-spot fixture, new clause red). It is
test-only. The fix-re-review budget had refused a further round on its test-only
cap, and that refusal was left standing rather than overridden a second time.

Gates on the head: `npm run typecheck` clean, `npm run lint` 897 files clean,
`VITEST_MAX_WORKERS=6 npm test` → 414 files, 7504 passed, 3 expected fail,
2 skipped.

---

## Declared review gap

> Two of the ten commits carry a FORCED ledger row rather than a seat review, and both are stated in full above. 3b167aa is docs-only (one PRD amendment, no code), which the verify skill's own cost guard exempts by name; it was claim-lint checked instead. 4e4b3e5 is test-only and applies an interleaving assertion that an independent seat (Opus) authored AND executed in its own report, closing a finding both seats raised independently — the orchestrator re-ran it rather than trusting it. The fix-re-review budget had refused a further round on its test-only cap and that refusal was left standing rather than overridden a second time. The other eight commits have seat reviews across four passes.

```
  forced  3b167aa  docs(prd-17): the outcome line was wrong, and the coordinate ruling is recorded
          ^ Docs-only: one PRD amendment, no code. The verify skill's own cost guard covers this case explicitly ('skip it entirely for a diff that is genuinely trivial: docs/comments/whitespace-only') — so the guard applied and no seats were run, which is stated here rather than left implicit. It was checked the way prose in this repo is checked: claim-lint reports 0 unresolved references and 0 bare counts, after the one bare count it originally flagged ('71 tests') was replaced with the command that derives it. The factual claims it makes were each executed during this session's audit: the nine event families in the union, every reducer arm returning state under 'additive only', no emitter outside session.closed, the beacon collector's declared attention: absent, and grep -rln beacon scripts/ returning nothing.
  forced  4e4b3e5  test(tide): the sweep asserts interleaving, which is what it claimed  (review of #277)
          ^ NO SEAT HAS READ THIS COMMIT. It applies a fix that an independent seat (Opus) authored AND executed in its own report: the interleaving assertion is that seat's code, run by it both ways (green on the shipped fixture, red on the blind-spot fixture it built). Both seats independently raised the finding it closes — that clause 2's 'two distinct timestamps' does not make clause 3 total. The orchestrator re-executed it: old clause green on the blind-spot fixture, new clause red with 'summons-cleared is hoistable to the END unnoticed'; full gates 7504 passing. The fix-re-review budget refused a further round (test-only cap), and the operator's earlier override was spent on round 2. Test-only: no production code changed. This row exists so the gap is visible rather than implied.
  ok      8b97363  test(tide): the verdict glyph's ink is read on every painted span  (review of #277)
  ok      3ede1a7  test(tide): the ordering law is swept over every kind, not sampled  (review of #277)
  ok      87e2b7d  test(route-class-law): a fixture's prose stops restating the real route total  (review of #276)
  ok      88f4e03  test(tide): the gate-verdict glyph's ink is read, not just its label  (review of #277)
  ok      6a1fe7f  test(tide): every chapter kind has an ordering witness  (review of #277)
  ok      9d6ef1b  feat(tide): chapters gain the summons, gate-verdict and operator-verdict marks (#277)
  ok      06bef2b  feat(server): the operator's three acts get a door (#276)
──
2 commit(s) carry a FORCED row, which is a stated reason, not a review.
```


ciaran-slow (approved):
Reviewed commit-by-commit against #276 and #277. **No defects found.** Everything below was re-derived here rather than read off the PR body — the body's tables are good, but a table is not evidence until someone else runs it.

All **EXECUTED** in a clean clone at `3b167aa`, node 22.23.2, each mutation against its own package with a clean-tree control either side.

**The headline claim holds.** The full 8 kinds × {append-hoist, prepend-hoist} matrix, re-run independently:

```
append   lane-born RED    prepend  lane-born RED
append   lane-landed RED  prepend  lane-landed RED
append   gate-held RED    prepend  gate-held RED    <- the r3 survivor, now closed
append   session-boundary RED   prepend  session-boundary RED
append   summons-raised RED     prepend  summons-raised RED
append   summons-cleared RED    prepend  summons-cleared RED
append   gate-verdict RED       prepend  gate-verdict RED
append   operator-verdict RED   prepend  operator-verdict RED
```

**16/16 RED.** Control 194/194 green.

Route (#276), each red as claimed: hard-coded `offset: 0` → 2; ack schema substituted for `verdict` → 2; for `note` → 2; `preHandler` dropped → 3; read-only refusal dropped → 1.

Sort: comparator → `() => 0` (never sort) → 3 red; ts comparison reversed → 5; `kind` tiebreak dropped → 1.

Glyph: `glyphHeldOf → return null` → 2 red. The subtle one — pinning **only the head span's** ink while leaving the stem correct — → 1 red, failing by name (`inks every painted span of a held verdict with the needs-you colour`). That is the defect `8b97363` exists to close, and it is genuinely closed.

**Two probes of my own, neither claimed in the body.** The `ROUTE_CLASSES` row is really bound to the live routing table, not just declared beside it: typo'ing the url to `/api/operator/:action` → **4 red**; demoting the row to `ungated-mutation` → **4 red**.

**The FORCED commit is sound.** `4e4b3e5` is the one row saying *no seat has read this commit*, so I read it and re-ran its claim. Rebuilt the blind-spot fixture from scratch (both `summons.cleared` moved to the two largest ts), production intact:

```
AssertionError: summons-cleared is hoistable to the END unnoticed:
  expected 1785852110000 to be less than 1785852100000
  Tests  1 failed | 29 passed (30)
```

The interleaving clause fires with exactly the message it promises. Its sole claim stands.

**`3b167aa`** (docs-only, also FORCED): spot-checked its checkable claims — `grep -rln beacon scripts/` is empty (exit 1), and the only emitters of the operator families are the three this PR adds. The ADR-0008 edit sits inside the existing amendment blockquote, which is the append-only-compatible place for it.

**Gates on the post-merge tree**, which is the condition that actually matters — `main` was **12 commits ahead** of this base, so the author's gates ran on a tree that no longer exists. Merged `origin/main` in (clean, no conflicts) and re-ran:

- `npm run typecheck` → exit 0
- `npm run lint` → exit 0
- `VITEST_MAX_WORKERS=6 npm test` → **414 files, 7514 passed, 3 expected fail, 6 skipped**

Worth noting explicitly: `main` brought in `doc-citation-law.test.ts` (+564 lines) and `.citation-prior-tracker` since this branch forked. This PR adds a great deal of cited prose, so that was the live risk of a clean-but-wrong merge. **It does not redden.**

`gh pr checks` was green on all **10 legs across the 3 expected jobs** (`build-test-boot` ×3, `pack-smoke` ×6, `windows-suite`), against the exact head sha rather than an older push.

## One commit pushed — `4601ed0`, and it is not a fix

I pushed `refactor(tide): a kind with no verb fails at the map that lacks it`. Please read the message before the diff, because **it corrects me, not you.**

I read `CHAPTER_VERB`'s move to `Partial<Record<…>>` + `as string` (9d6ef1b) as a lost exhaustiveness guard. **It is not one.** `MarkGlyph`'s switch is already exhaustive over `Chapter['kind']`, so a ninth kind fails `tsc` on your tree as it stands — I verified that both ways before concluding:

```
author's tree + 'lane-paused' added to CHAPTER_KINDS:
  ChapterMarks.tsx(206,98): error TS2366  --- exit 1
```

So nothing could have reached a hover card as `undefined`. All the commit changes is *where* the error lands: excluding the three data-resolved kinds by name keeps the map an exhaustive `Record`, so the same mutation also fails at `chapters.ts:159` with `TS2741: Property '"lane-paused"' is missing`, naming the kind. The cast goes away because the default branch narrows to exactly that set.

Type-only, no runtime path differs, label strings byte-identical. Gates on that tree: typecheck clean, lint clean, 414 files / 7500 passed. **Revert it freely if you'd rather the glyph switch stay the single choke point** — it is a readability preference about error siting, and you own this file.

Note it landed after the last CI run, so CI re-runs on `4601ed0`.

Two stale route counts live outside this PR's fence, both **pre-existing and neither made worse here**:

- `docs/adr/0014:49` — "all six mutating routes and ten reads", present tense. Was already four routes stale before this wave.
- `docs/prds/prd-43:44` — "where `api/index.ts` declares ten with six gated".

Both sit under `SWEEP_EXCLUDED_PREFIXES` (`docs/adr/`, `docs/prds/`, `docs/review/`), so the CLAIMS sweep is silent on them **by design** — an ADR counts the tree as it stood, and the log is append-only. I have not touched them: they are outside the fence and the exclusion looks deliberate rather than accidental. Flagging only so the decision is conscious, since AGENTS.md treats a wrong pointer as worse than none.

## What I did not run

`scripts/gate.sh` — that is the landing and it is the operator's step, not a reviewer's. I did not merge, and I did not take this out of draft.

## Verdict

Approve. The verification standard here is genuinely high — three of the ordering repairs were themselves defective and the PR says so in its own body, which is why the final commits assert the property rather than the cases. The FORCED row is stated honestly and its claim survives independent re-execution.

**This is still a draft**, so it cannot land as-is. Ready to merge on your mark once it is marked ready and CI is green on `4601ed0`.

Issue #133, opened 2026-08-27
transcript capture: a rotation captures every lane the session named, not every lane still in the window


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/log/transcript-capture.ts`
- `packages/server/src/log/transcript-capture.test.ts`
- `packages/server/src/recorder/rotate.ts` — **the one call site only** (`:147`), if the chosen shape needs it

## Blocked by
- **PR #131** (#37, prd44 ruling 4). The defect does not exist until the recorder's window is bounded.

**#53** (prd42 w6, Backlog) declares `recorder/rotate.ts` in its fence; its item there is the duplicated refusal message, a different hunk from the capture call. #131 already touches `:145` in the same neighbourhood and recorded that widening. Whoever moves second re-reads the other.

## The gap

`rotate.ts:147` hands the recorder's in-memory events to the transcript capture:

```ts
await captureSessionTranscripts({
  events: recorder.eventsSoFar(),
  ...
})
```

`captureSessionTranscripts` derives the lanes it captures from exactly those events — `allAttributedLanes(events)` (`transcript-capture.ts:212`), and it returns `null` outright when that list is empty.

**#37 makes `eventsSoFar()` a window of the last 75,000 events.** So on a rotation of a session past the ceiling, a lane whose attributing events were evicted is simply absent from the capture: not reported as unreadable, not counted, not mentioned — the manifest describes fewer lanes than the session had, and reads as complete.

That is the same dishonest shape **prd-31 ruling 5** and **#38** are written against: a lane's history disappearing from a surface that promised it would still open, with no signal that anything went missing. Here the loss is at capture time, so it is permanent — the transcript is gone from the manifest even though the log on disk still names the lane.

**Why it is not a fence widening on #131.** Unlike `rotate.ts:145`'s `eventCount` (a one-line read of the fold), this needs a source that is complete — the log from disk, or a projection the fold already maintains — and choosing between those is a design question with a real trade-off, not a substitution. #131 names it rather than smuggling it in.

## Definition of done
- The lanes a rotation captures are the lanes the **session** named, not the lanes still in the recorder's window. Two candidate shapes, and the plan picks one with the reason recorded:
  - **read the closed log from disk** — complete by construction, and the file is already flushed and fsynced by the time `rotate.ts` reaches this line, but it re-reads a large recording at rotation time (prd-44 ruling 1's parse cache now covers exactly that read, `log/session-log.ts`);
[...]

PR #287, opened 2026-09-05
prd27 w2: agent.status names its witness, and a Claude hook declares a lane's attention

### LINKED ISSUE #281: prd27 w2: agent.status names its witness, and the transcript organ speaks its four states
## Fence (may touch ONLY)
- `packages/core/src/events/workmux.ts`
- `packages/core/src/state.ts` (`AgentState` gains `witness`)
- `packages/core/src/reduce.ts` (the `agentStatus` fold only — no other arm)
- `packages/core/src/fleet/buildFleet.test.ts`
- `packages/core/src/fleet/fixtures.ts`
- `packages/core/src/selectors/condition.test.ts`
- `packages/server/src/collectors/sessionlog/`
- `packages/web/src/disclosure/vocabulary.ts`
- `docs/adr/0037-agent-status-names-its-witness.md` (new)
- `docs/adr/README.md`

## Blocked by
Nothing — #217 landed (`a36d15c`).

Sibling lanes hold other surfaces: the wave-2 sibling (`rhizomorph env --hooks claude`) owns `packages/server/src/cli/env.ts`, `packages/core/src/events/beacon.ts`, `packages/server/src/collectors/beacon/**` and `README.md` — this issue touches none of them. **#280** (prd17 w6, groomed, not live) also claims `packages/core/src/reduce.ts` for a different arm (`gate.verdict` / `beacon.received`); this issue edits only the `agentStatus` fold, so whichever lands second rebases one function. Do not touch `packages/server/src/api/**`.

**#264** (prd43 w7, In progress) holds `docs/adr/README.md` for ADR-0032's row; this issue adds ADR-0037's row to the same table. **Land #264 first** — the lint reports the overlap correctly until it does — then dispatch this.

**Model:** opus. **Wave:** 2.

## Why

prd-27 ruling 2 — the PRD's own keystone, never filed. On `main` today `packages/core/src/events/workmux.ts:20` pins `agent.status` to `source: 'workmux'`, and `packages/server/src/collectors/sessionlog/collector.ts:34-36` still calls itself **BLOCKED** on that: the turn-shape organ derives `working | waiting | frozen | gone` every poll, `agentStatusEmissionFor` (`lane-state.ts:280`) is tested and ready, and nothing calls it. The surest reading in the tree reaches no surface. The PRD's evidence section: *"if both witnesses sign their observations `workmux`, a disagreement between them cannot even be seen"* — and a beacon publishing through the pinned envelope would forge provenance on a hash-chained log (ADR-0009).

## Definition of done

- `agentStatusEventSchema` is `envelopeWithSources(['workmux', 'sessionlog'], 'agent.status', …)`; `EVENT_SOURCE_BY_TYPE['agent.status']` stays `'workmux'` as primary, the `satisfies` clause and `SourceOf<'agent.status'>` widen the way `llm.usage` already does. An event with `source: 'sessionlog'` parses; one with `source: 'beacon'` or `'otel'` is refused.
- `AgentState` carries `witness: 'workmux' | 'sessionlog'`, set from the event's `source` in the `agentStatus` fold. A later `workmux` report and a later `sessionlog` report both update `status`; `witness` says who spoke last.
- The sessionlog collector publishes `agent.status` **edge-triggered** through `agentStatusEmissionFor`: a lane whose state did not change emits nothing (three polls, one event); `frozen` and `gone` publish nothing, as the helper already rules. The snapshot's `lanes` reading is unchanged — `tmuxless-boot.test.ts` stays green **unedited**.
- **The false-summons guard (#133, ruling 4).** `Lane` carries `agentStatusWitness`; `detectWaiting` treats a `waiting` witnessed by `workmux` as certain (`inferred: false`, evidence `workmux reports waiting …`) and a `waiting` witnessed by `sessionlog` as **inferred** (`inferred: true`, evidence `transcript shape: <the organ's evidence>`). A fleet with a workmux `waiting` and a sessionlog `working` for the same handle keeps the workmux reading and says so.
- The `SESSIONLOG_CAPABILITIES.attention` strings are **not** edited (that is w4's); the BLOCKED paragraph in `collector.ts` is replaced by a sentence naming this issue.
- `docs/adr/0037-agent-status-names-its-witness.md` records widening a pinned envelope source as a format decision (ruling 2 owes it): why a second literal on one type beat a second event type and beat a `witness` payload field. Index row in `docs/adr/README.md`.

## Sibling case

Publishing the organ's reading without carrying its witness into the fleet would make a transcript-shape `waiting` render as workmux's certain WAITING — the #133 false summons rebuilt one layer down. The witness has to travel: envelope → `AgentState` → `Lane` → `detectWaiting`, and a test at each hop. The four test sites that assert the literal `workmux reports waiting` (`buildFleet.test.ts`, `condition.test.ts`, `vocabulary.ts`'s doc copy) are fenced here for minimal reconciliation only.

Drop the `witness` when folding (`agentStatus` in `reduce.ts` sets it to `'workmux'` always): a sessionlog-witnessed `waiting` then renders certain — the fleet test with two witnesses disagreeing must redden. Make the organ emit every poll instead of on transitions: the three-polls-one-event test must redden. A test that only asserts the schema accepts `source: 'sessionlog'` survives both.

### LINKED ISSUE #282: prd27 w2: rhizomorph env --hooks claude prints hooks that write beacons, and the fixture is a capture
## Fence (may touch ONLY)
- `packages/server/src/cli/env.ts`
- `packages/server/src/cli/env.test.ts`
- `packages/core/src/events/beacon.ts` (adds `BEACON_ATTENTION_KINDS`; the schema is unchanged)
- `packages/server/src/collectors/beacon/fixtures/` (all files)
- `packages/server/src/collectors/beacon/collector.test.ts`
- `packages/server/src/collectors/beacon/parse-beacon-line.test.ts`
- `README.md` (the `rhizomorph env` row and one Hooks subsection)

## Blocked by
Nothing — #217 landed (`a36d15c`).

Sibling lanes hold other surfaces: the wave-2 sibling (the `agent.status` witness) owns `packages/core/src/events/workmux.ts`, `index.ts`, `state.ts`, `reduce.ts`, `fleet/**` and `collectors/sessionlog/**` — none of them is touched here. No new subcommand: `--hooks` is a flag on `env`, so `cli/index.ts` and `cli-surface-law.test.ts` are not touched. This is the wave's one `README.md` claimant. Do not create the beacon directory from the collector — the hook command the emitter prints does the `mkdir -p` (ADR-0036: the writer owns the directory).

## Why

prd-15 ruling 2, verbatim: *"`rhizomorph env <lane>` grows an optional `--hooks <cli>` emitter that prints the exact hook config to install."* The door landed in #217 and nothing writes through it. Until a writer exists, the beacon fixture is hand-written (its `CAPTURE.md` says so), the lapse interval prd-27's amendment demands cannot be measured, and every attention beacon is hypothetical. This issue is also where the **kind vocabulary** is ruled — prd-27 wave 1 left `kind` free-form on purpose and named w2 as its owner.

## Definition of done

- `rhizomorph env <lane> --hooks claude` prints a JSON fragment for Claude Code's `settings.json` `hooks` block, with three hooks, each a one-line shell command that appends one ADR-0036 v1 line to `<dataRoot>/<repoSlug>/beacons/claude-hook.jsonl` (the path resolved by the CLI from the running instance's repo, the same way the OTLP block is resolved today): `Notification` → `kind: "waiting"`, `Stop` → `kind: "stopped"`, `UserPromptSubmit` and `PostToolUse` → `kind: "working"`; `writer: "claude-hook"`, `lane: <lane>`, `at` from the shell's epoch-ms, and `mkdir -p` of the directory before the append. `--hooks <other>` exits non-zero naming the supported list.
- `BEACON_ATTENTION_KINDS = ['waiting', 'working', 'stopped'] as const` is exported from `packages/core/src/events/beacon.ts` with a doc comment ruling the vocabulary (prd-27 ruling 4's three words); the emitter imports it rather than spelling the strings. The zod schema stays open — a kind outside the list is still carried (ADR-0011).
- **Capture-first.** The hand-written `fixtures/claude-hook.jsonl` is replaced by a real capture from a Claude Code session running with the printed hooks installed — at least one line of each kind — sanitised per AGENTS.md (synthetic lane handle, no home paths, LF). `fixtures/CAPTURE.md` becomes the recipe: which command printed the hooks, where they were installed, what was done in the session, and the substitution step. Every test in `collector.test.ts` / `parse-beacon-line.test.ts` that reads the fixture is re-pointed at the captured lines and its expected digests/offsets re-derived, not retyped.
- README: the `rhizomorph env` row mentions `--hooks`, and a short Hooks subsection under Telemetry shows the install step. `CHANGELOG.md` `[Unreleased]` › `Added`.

## Sibling case

The OTLP block `env` prints today is the template, and its trap is the sibling: that block is inert until a process is launched with it, whereas a hook writes a file the moment it fires — so the emitter must print a command that is safe to run from any cwd (absolute beacon path, `mkdir -p`, `>>` never `>`), and the test must execute one printed command in a temp data root and parse the resulting line with `parseBeaconLine`. A test that only snapshots the printed JSON survives a hook that writes a malformed line.

Swap `>>` for `>` in the printed command: the run-the-hook-twice test must show two lines, so it reddens. Drop `mkdir -p`: running the printed command in a fresh temp root must fail, so it reddens. Change a kind string in the emitter: the test asserting each printed kind is in `BEACON_ATTENTION_KINDS` reddens. Survives, and known: the real Claude Code hook event names are not checked against Claude Code itself — the capture is the witness for that.

ciaran-slow:
prd-27 wave 2 — the keystone (`docs/prds/prd-27-the-declared-voice.md` §Sequencing item 1) plus the emitter half of item 3. Two issues, two commits, one queue toll.

Closes #281
Closes #282

Nothing in this wave depends on anything else in it: #281 is core + the sessionlog organ, #282 is the CLI emitter + the beacon fixture. They meet only in `CHANGELOG.md`.

---

## #281 — `agent.status` names its witness, and the transcript organ speaks its four states

Commit `02cc2b5` (lane `281-agent-status-witness` @ `3dd4f8d`, one commit; trailers `Claude opus` + `Claude fable`).

### What became true

- **The envelope widens to a second literal.** `agentStatusEventSchema` is `envelopeWithSources(AGENT_STATUS_SOURCES, 'agent.status', …)` with `AGENT_STATUS_SOURCES = ['workmux', 'sessionlog']`. `EVENT_SOURCE_BY_TYPE['agent.status']` stays `'workmux'` — the *primary* — and `agent.removed` still admits only `workmux`, because only a roster can say a handle left. `beacon` and `otel` are refused by the enum.
- **The organ publishes.** The sessionlog collector emits its working/waiting transitions edge-triggered through `agentStatusEmissionFor`, signed `source: 'sessionlog'` via `createEvent` directly (the pi collector's pattern — `EmitOptions` was *not* widened). `frozen` and `gone` publish nothing; the unattributed main worktree publishes nothing (#62 — it is a setup gap, not a lane). `detail` drops its `transcript-tail: ` prefix and carries the organ's reading verbatim.
- **The fold records who spoke.** `AgentState` gains required `witness` and `dissent`; `AgentStatusDissent` is a new exported interface.
- **prd-27 ruling 4's asymmetry is one `if` in the reducer.** A standing `workmux` `waiting` **or `done`** is never displaced by a `sessionlog` word; only a declared `working` yields to an inference. The refused word is kept as `dissent` so the disagreement renders instead of resolving in silence (prd-15 ruling 2).
- **The fleet renders the difference.** `Lane.agentStatusWitness`; `detectWaiting` gives a workmux WAITING `inferred: false` and names any disagreement beside it (`workmux reports waiting 1m30s; transcript shape reads working`), and a sessionlog WAITING `inferred: true` with `transcript shape: <the organ's reading>`.
- **[ADR-0037](docs/adr/0037-agent-status-names-its-witness.md)** records why a second literal beat a second event type and beat a payload field, with three rejected alternatives and their reasons.

### What is the sibling case?

The issue named the first one: publishing the organ's reading *without* carrying its witness would render a transcript-shape `waiting` as workmux's certain WAITING — the #133 false summons rebuilt one layer down. The witness therefore has to survive four hops and there is a test at each: envelope (`events.test.ts`), `AgentState` (`reduce.test.ts`), `Lane` (`buildFleet.test.ts`), `detectWaiting` (`diagnose.test.ts`).

Three more, each the same fix in a different direction:

1. **The refusal is guarded on `present`.** Ruling 4 protects a *standing* declaration. A handle `agent.removed` marked gone has nobody standing behind it, so a fresh sessionlog sighting must be last-wins — otherwise a removed lane's stale `waiting` outranks the live reading forever.
2. **The refusal is guarded on the declared status — and its first draft missed `done`.** The plan (§1.3) protected only a declared `waiting`. Verify found the structurally identical sibling: a standing `workmux` **`done`**, which the organ reads as `waiting` 75 s later off any quiet `turn-complete` tail, so *every* finished lane in the tmux+workmux+sessionlog rig turned from DONE/calm into `~WAITING`/needs-you with `dissent: null`. Fixed in the verify-fix (`prev.status !== 'working'`); `buildFleet.test.ts` gains `a finished lane the organ later reads as waiting stays DONE and calm`, which is exactly the probe sequence that reproduced it.
3. **In the collector:** `frozen`/`gone` withhold, and the unattributed main tree — the structurally identical "this looks like a lane and is not one" case one level up — withholds too. Both have their own test.

Every row below was applied to the source and the named test read red, by the builder and again independently by the verifier (9 mutations re-run at verify, all behaved as claimed):

| mutation | test that reddens |
|---|---|
| fold sets `witness: 'workmux'` always | reduce `records whose word the status is`; buildFleet `renders a sessionlog-witnessed WAITING as inferred` |
| drop the ruling-4 `if` (pure last-wins) | reduce `ruling 4: a sessionlog working does not withdraw a declared waiting`; buildFleet `two witnesses disagree` |
| refusal ignores `present` | reduce `a removed handle re-sighted by sessionlog is last-wins, not ruling 4` |
| refusal also covers a declared `working` | reduce `an inferred waiting over a declared working is last-wins` |
| `dissent` set on corroboration too | reduce `corroboration is not dissent`; diagnose `a declaration nobody contradicted reads exactly as it always did` |
| `dissent` never cleared | reduce `workmux speaking again clears the dissent and wins` |
| organ emits every poll / emits `frozen`,`gone` / omits `source` / publishes the unattributed tree | collector `three polls, one event` · `frozen and gone publish nothing` · `every agent.status the organ emits is signed sessionlog` (asserts `emitted.length > 0` first, so it is not vacuous) · `the unattributed main tree publishes no agent.status` |
| `detectWaiting` drops the dissent clause | diagnose `a declaration the organ contradicted names the disagreement beside itself` |
| schema widened wholesale / `agent.removed` widened with it | events `accepts both witnesses and refuses every other source` · `still admits only workmux on agent.removed` |
| `detail` keeps the `transcript-tail: ` prefix | lane-state `carries the organ's evidence as detail`; collector `three polls, one event` |

**Known survivor, and why the tests are not it:** a test that only asserts the schema *accepts* `source: 'sessionlog'` survives every row above. Two assertions were made exact rather than regex for the same reason — `collector.test.ts` asserts `payload.detail === nextSnapshot.lanes[lane].evidence` (an organ that paraphrased its own reading would pass `/^WORKING/`), and `lane-state.test.ts` derives the reading independently.

The one mutation the suite **did** survive at build time was sibling 2 above — no test placed a `sessionlog` word after a `workmux` `done`. That is what verify caught, and there are now four tests on it.

### Plan deviations

None structural. Three recorded:

1. Plan §5.2 named only the BLOCKED *paragraph* in `lane-state.ts`; the builder also fixed the three surrounding sentences the paragraph's removal falsified (section banner, doc heading, closing sentence), plus two clauses in `collectors/sessionlog/types.ts`. Prose only.
2. `buildFleet.test.ts`'s local `event()` helper gained a fourth optional `source` parameter typed off `createEvent`'s own init. Local to that file.
3. §6.4's `detectWaiting` cases are four, not three — "declared with no dissent" was split out so the pre-#281 sentence is pinned independently of the dissent one.

`fleet/fixtures.ts` and `web/disclosure/vocabulary.ts` needed no change, exactly as the plan predicted.

### Fence widenings (all declared on the plan before the change; the fifth on the verify report before the change)

1. `packages/core/src/fleet/plumbing.ts` — three fields on `Draft`/`emptyDraft`. `Draft` is the only route from `view.agent` in `buildFleet`'s first loop to the `Lane` literal and `diagnose()` call in its second.
2. `packages/core/src/eras/era-1/session-state.snapshot.json` + `packages/core/src/eras/CAPTURE.md` — the golden corpus law folds era-1 and compares byte for byte, so two new required `AgentState` keys move that fold. Re-blessed by `CAPTURE.md`'s own procedure: **6 insertions, 0 deletions, 0 modifications** — exactly `"dissent": null` and `"witness": "workmux"` for each of the three agents, verified by `sort | uniq -c` at verify. `CAPTURE.md` records the re-bless and why the new fold is the correct meaning of that old log.
3. `docs/architecture.md` — one line (`1 1` numstat), the `workmux: agent.status` bullet gaining the ADR-0037 superseding pointer. The event-source union on that page did not change.
4. `CHANGELOG.md` — the wave's standing convention.
5. **Verifier's widening, prose only:** `packages/server/src/collectors/conformance/signal-evidence.ts` and `packages/server/src/collectors/conformance/sessionlog.test.ts` — two doc comments that still described the organ as unable to publish, citing a BLOCKED note that no longer exists. Declared in the verify report before the change.

### Verify outcome

Verified by **fable** against builder **opus** (different model). Verdict: **not ready** on one must-fix — the declared-`done` sibling above — plus one should-fix (the two stale conformance comments). Both fixed on the branch; the commit was amended, still one commit. Two findings left as *worth doing later*: `dissent` reaches `AgentState` but has no rendering surface beyond `detectWaiting` (prd-27 w4 owns the chip/hover form), and `docs/architecture.md`'s Decisions log has no ADR-0037 entry (ADR-0036, the wave-1 keystone, has none either — following precedent, not breaking it; the wave PR is the place to decide once for both, and this PR leaves it alone).

Final lane sha: **`3dd4f8d`**, on this branch as **`02cc2b5`** (tree byte-identical to the lane tip — `git diff 3dd4f8d 02cc2b5` is empty).

---

## #282 — `rhizomorph env --hooks claude` prints hooks that write beacons, and the fixture is a capture

Commit `e341081` (lane `282-env-hooks-claude` @ `10678fb`, one commit; trailers `Claude sonnet` + `Claude opus`).

### What became true

- **`BEACON_ATTENTION_KINDS`** (`waiting`, `working`, `stopped`) is exported from core as the ruled prd-27 ruling 4 vocabulary. The zod schema stays **open** (ADR-0011): a kind outside the list still parses and is carried, not dropped.
- **`rhizomorph env <lane> --hooks claude`** prints the Claude Code `settings.json` `hooks` fragment. Four one-line shell commands — `Notification` → `waiting`, `Stop` → `stopped`, `UserPromptSubmit` and `PostToolUse` → `working` — each `mkdir -p` then **append** (`>>`, never `>`) an ADR-0036 v1 beacon line to the running instance's `<data root>/<repo slug>/beacons/claude-hook.jsonl`. Cwd-independent, `at` from the shell's epoch-ms, lane JSON-quoted. `--hooks <anything else>` exits 1 naming the supported list.
- **The fixture is a real capture, not hand-written.** `packages/server/src/collectors/beacon/fixtures/claude-hook.jsonl` came out of a live session running the printed hooks — 4 lines, 1 `waiting` / 2 `working` / 1 `stopped`. Its recipe and provenance are in **`packages/server/src/collectors/beacon/fixtures/CAPTURE.md`**.
- Every fixture-reading test is re-pointed and its expectations **derived from the fixture's own bytes**, not retyped. The two contract cases the emitter never writes (an extra key; an absent lane) were preserved at equal strength in dedicated files — a law moved, not weakened.

### What is the sibling case?

"The Write tool's permission prompt fires `Notification`" has as its structural sibling **"the turn ending after a prompt that needed no tool permission still fires `Stop`."** The capture's own session exercises both, and the fixture-law test `every detail names the hook that fired, and the (event → kind) pairs match the emitter's table` pins the pairing to `CLAUDE_HOOK_EVENTS` rather than to what one capture happened to produce — so the capture cross-checks the emitter in both directions (verify's mutation 3 reddens tests on *both* sides).

The sibling that actually cost something: `Notification` fires **only** when Claude Code stops to ask, which in the default auto-accept-edits mode it never does. Verify proved by probe that a capture made by following the recipe verbatim in the default mode yields 3 lines, not 4, and fails the fixture law — so the manual-approval step is now *in* recipe step 4, and the README carries it as a third honesty note.

Six mutations, all executed at verify; each promised test is the one that fired:

| mutation | result |
|---|---|
| `>>` → `>` in `renderBeaconHookCommand` | 3 red, incl. `appends: run twice, two lines, both parse` |
| drop `mkdir -p … && ` | 6 red, incl. `writes one parseable v1 line into a directory that did not exist` (the `sh -c` exits non-zero) |
| `Notification` → `working` | 3 red — the by-name emitter test **and** `the fixture is a capture (#282) > every detail names the hook that fired` |
| add a 4th kind to `BEACON_ATTENTION_KINDS` | 1 red — `every kind is in the ruled vocabulary, and all three kinds occur` (proves the ⊇ half is not vacuous) |
| typo a kind (`'wating'`) | `TS2820` at typecheck — the `satisfies` clause, not a runtime check |
| CLI ignores `RHIZOMORPH_DATA_DIR` | 1 red — the live-server `--hooks` test, so deviation 2 below is not a vacuous test |

What makes the append/mkdir mutations fail concretely rather than vacuously: `env.test.ts` has an **executed** describe block that runs each printed command through `sh -c` in a temp cwd against a temp beacon dir and parses the result back through `parseBeaconLine`.

### Plan deviations (three, all recorded)

1. `env.ts` drops the unused `BEACON_ATTENTION_KINDS` **value** import — biome's `noUnusedImports` forbids it, and the `as const satisfies readonly (readonly [string, BeaconAttentionKind])[]` clause enforces the vocabulary *harder* than the value import would have, at compile time. Accepted at verify.
2. The live-server test sets and restores `process.env.RHIZOMORPH_DATA_DIR` around the call (restore in a `finally`, distinguishing "was unset" from "had a value"), making explicit the plan's implicit same-environment assumption. Accepted; proven non-vacuous by mutation 6.
3. **The capture ran in a tmux session the operator had already started**, not one the lane created. A first build lane hit `OAuth session expired` driving `claude` itself and correctly **stopped rather than hand-writing the fixture** (its blocked report is on the issue); the second lane drove the operator-authenticated session via `tmux send-keys`/`capture-pane`. Only the launcher changed — the recipe, the typed prompt and the four hook events are exactly as `CAPTURE.md` describes, and `CAPTURE.md` records this as its own section.

### Fence widenings (declared on the plan before the change)

`packages/server/src/cli/telemetry-env.ts` and `.test.ts` (folding `fetchInstanceId` into `fetchInstanceMeta`, which returns `{ sessionId, repoPath }` off one `/api/meta` read; all four existing error messages byte-identical), and `CHANGELOG.md`.

### Verify outcome

Verified by **opus** against builder **sonnet** (different model). Verdict: **ready to merge**, with two documentation must-fixes applied on the branch by the verifier — the manual-approval step moved into `CAPTURE.md` recipe step 4, and the README's third honesty note. `claude-hook.jsonl` is **byte-identical** across the fix: the capture is evidence, not a file to edit. Five *worth doing later* notes were left un-actioned and are on the issue (chief among them: `fetchInstanceId` now has no production callers, and `CLAUDE_HOOK_EVENTS` would sit more naturally in `core` than have `collectors/` import from `cli/`).

Final lane sha: **`10678fb`**, on this branch as **`e341081`**.

---

## How the wave was assembled

Branch `prd27-w2-declared-witness`, cut from `origin/main` at `5fa85b0`, then two cherry-picks in issue order — never squashed, so each issue is still one standalone commit and the unit of review survives the bundle:

```
5fa85b0  (origin/main)
02cc2b5  feat(core): agent.status names its witness, …  (#281)   ← 3dd4f8d
e341081  feat(cli): rhizomorph env --hooks claude …    (#282)   ← 10678fb
```

**The one conflict, and it was the expected one.** `3dd4f8d` applied clean. `10678fb` conflicted in **`CHANGELOG.md` and nowhere else** — both issues add an entry at the top of `[Unreleased]` › `Added`, which is the wave's standing convention (each issue writes its own; the assembler reconciles), not a fence violation. Resolved inside the second cherry-pick by keeping **both entries in full**, #281's first then #282's, so `02cc2b5` still stands alone with only its own entry. No other file was touched in the resolution, and the resolution was checked rather than assumed:

- `git diff 3dd4f8d 02cc2b5` → **empty**. #281's tree on this branch is byte-identical to its lane tip.
- `git diff 10678fb e341081` → only #281's 26 files plus `CHANGELOG.md` at **+19 / −0** — #281's entry added above #282's, nothing of #282's altered.

**Gate on the assembled branch** (the first run where the two commits coexist; node v22.23.2, `npm install` first, run from this worktree — `scripts/gate.sh` was **not** run, it is the operator's landing step):

| gate | command | result |
|---|---|---|
| test | `VITEST_MAX_WORKERS=6 npm test` | exit **0** — `Test Files 413 passed (413)` · `Tests 7518 passed · 3 expected fail · 6 skipped (7527)`, 98.32 s |
| typecheck | `npm run typecheck` | exit **0** — `tsc --noEmit` clean across `core`, `server`, `web` (+ `app`, `contract`) |
| lint | `npm run lint` | exit **0** — `Checked 895 files in 386ms. No fixes applied.` |
| build | `npm run build` | exit **0** — server CLI, web (`✓ built in 290ms`), electron `main.js` + `preload.cjs` |
| boot smoke | `npm start -- --port 0` | `rhizomorph running at http://127.0.0.1:56102` · `GET /` → **200 `text/html`, 1628 B** · `GET /api/meta` bare → **401** (`missing or invalid x-rhizomorph-capability header`, the expected ADR-0008/0012 guard) · with the page's token → **200** with `repoPath`/`sessionId`/`eventCount` |

**No cross-lane breakage.** 7518 tests against 7495 (#281 alone) and 7490 (#282 alone) — both lanes' new tests are present and green together, and no law either lane added is failed by the other's change.

**A note on the #282 fixture:** `packages/server/src/collectors/beacon/fixtures/claude-hook.jsonl` is a **real capture**, not a hand-written file. Its recipe — throwaway repo and data root, booting the built server, printing the fragment into `.claude/settings.local.json`, the manual-approval mode switch, the typed prompt, and the sanitisation check — is in `packages/server/src/collectors/beacon/fixtures/CAPTURE.md` beside it, per prd-26 ruling 3 (verification by capture). The verifier could not re-drive the `claude` CLI and instead corroborated the fixture from its own bytes (whole-second `at` values with real inter-hook gaps, kinds and details consistent with the emitter's table).

**One file this wave shares with live work.** #281 appended ADR-0037's row to `docs/adr/README.md`'s table — as the **last** row, deliberately. **#264** (prd43 w7, *In progress*, nothing pushed at the time of writing) claims that same file for one row of its own. `scripts/fence-lint.sh` flagged the overlap at plan time and the issue ruled it: whichever lands second rebases one line. There is nothing else in either change that touches the other.

**Also worth a reviewer's eye:** `corpus-eol-law.test.ts` compares each pinned file's git *index* blob against a clean-filtered hash of the *working tree*, so any unstaged edit under a pinned path (the era-1 snapshot, the beacon fixtures) reads as "the pin rewrites its bytes". Both lanes hit it and both resolved it with `git add`. Not a defect, and green here — but it will bite the next person who re-blesses a snapshot and runs the suite before staging.


gabriel-canaan (approved):
## Approve — read commit-by-commit against `origin/main` at `30d3630`

Two commits, one per issue, each standing alone; both measured against #281 and #282 as written, not against anything they did not ask for.

**The tree that will land is greener than the one the body measured.** The PR was cut at `5fa85b0`; `main` has moved 22 commits since, touching `docs/adr/README.md`, `docs/architecture.md` and adding the new `doc-citation-law` corpus. Merged in and re-run from a clean install (node 22.22.2):

| gate | result |
|---|---|
| `npm run typecheck` | **exit 0** |
| `npm run lint` | **exit 0** — `Checked 897 files` (the body measured 895; main added two) |
| `VITEST_MAX_WORKERS=6 npm test` | **exit 0** — `Test Files 414 passed (414)` · `Tests 7569 passed \| 3 expected fail \| 2 skipped (7574)` |

The body's `413 / 7518` was an honest measurement of a tree that no longer exists. Nothing main added since — including the new `#NNN`-citation ceiling that ADR-0037 and both CHANGELOG entries now pass through — goes red against this branch.

**The one predicted conflict resolved itself, and the resolution is checkable.** #281's plan flagged `OVERLAP #264 (docs/adr/README.md)` and ruled "whichever lands second rebases one line". #264 landed first (`6830eee`, `406d2d5`); the merge of `origin/main` into this branch was **clean**, and the table now carries ADR-0032's amended row and ADR-0037's new row at line 249. `CHANGELOG.md`'s cherry-pick conflict is resolved the way the body claims — both entries in full, #281's above #282's.

**Three mutations, EXECUTED here, not taken from the body.** Each restored with `git checkout HEAD --` and the restore verified with `git diff HEAD --quiet`.

1. **The `done` sibling is load-bearing.** Reverting `reduce.ts`'s guard to its pre-verify form (`prev.status !== 'working'` → `prev.status === 'waiting'`) reddens **3 tests**: `ruling 4, the sibling: a sessionlog waiting does not resurrect a declared done…`, `a sessionlog working over a declared done is dissent too…`, and `a finished lane the organ later reads as waiting stays DONE and calm`. This is the finding verify caught, and it is now pinned in both directions.
2. **`>>` → `>` in `renderBeaconHookCommand`** reddens **3 tests** in `env.test.ts`, including `appends: run twice, two lines, both parse` and `every command is cwd-independent by construction`. The executed `sh -c` block is doing real work, not snapshotting JSON.
3. **The parked exemption still holds under the new witness** (this one I went looking for as a defect and did not find one). `detectWaiting`'s first branch — the one the sessionlog reading now routes through — has none of the four exemptions the second branch carries, `parked` included, so I built the case: a parked lane whose only word is a `sessionlog` `waiting`. It produces the inferred pathology, and `ladder.items` is `[]`, `ladder.rank` is `calm`, `errorCountsOf(lanes).blocked` is `0`. `ladder.ts:29`'s `if (lane.parked) continue` and `errorCountsOf`'s `!lane.parked` catch it at both summons surfaces. The new witness cannot summon on a lane the operator stood down. **REASONED, not executed:** the fleet table's own STATE cell will now read `~WAITING` on such a lane where it previously read the activity word — which is what `ladder.ts`'s comment already says is intended for the declared case.

**Provenance holds where the repo cares about it.** No file this PR touches carries a home path, hostname or real name (the two `/home/x` and `/Users/operator` hits under `events.test.ts` and `collector.test.ts` are pre-existing synthetic placeholders — `git diff 5fa85b0 e341081` adds neither). All four fixture `at` values are whole-second multiples; the kinds are 1 `waiting` / 2 `working` / 1 `stopped`; every `detail` resolves through `CLAUDE_HOOK_EVENTS` to the kind on its own line, so the capture and the emitter cross-check each other. ADR-0037 is a real ADR — four options, three rejected with reasons (B says the fact twice; C makes the envelope lie under ADR-0009; D is the wrong door).

**One shell detail worth naming because it is easy to get wrong and is right here.** `printf '%s\n' '…head…'"$(($(date +%s)*1000))"'…tail…'` looks like three arguments and would print three lines if it were; the three quoted pieces are adjacent with no whitespace, so the shell concatenates them into one word before `printf` sees it. `%s` also does not interpret backslash escapes, so `JSON.stringify`'s `\"` survives a lane handle containing a quote. Both halves of the escaping story — `JSON.stringify` for JSON, `shellSingleQuote` for the shell — are correct and separately tested.

---

### Two findings, both fixed on the branch

Both are documentation, both rung-1 mechanical, both smaller than this paragraph explaining them.

**1. `fixtures/CAPTURE.md`'s opening licensed the thing the same file forbids.** — fixed in `dae4b88`

The first paragraph carried the pre-#282 framing forward: *"a hand-written fixture proves our reading of the contract, not the tool"*. Three lines below it, #282 adds, in bold: *"**Do not hand-write a single line of this file.**"* Both cannot hold, and the one that survives is the rule — every other `CAPTURE.md` in this tree pins its contract with a real capture, which is exactly what this commit made possible. A capturer who reads top-to-bottom is otherwise told first that hand-writing is what the fixture is for. Prose only; `claude-hook.jsonl` is untouched.

**2. The `--hooks` help line did not say whose data root resolves the beacon path.** — fixed in `1d814a3`

`--hooks claude` reads `repoPath` off `/api/meta`, but resolves the data root **locally**, from the CLI process's own `RHIZOMORPH_DATA_DIR`. The help line said only *"the running instance's beacon directory"*, which reads as if the whole path came from the server — so a user whose shell and server disagree gets hooks writing where nothing tails, with no hint in `--help`. #282's plan (decision 1) named the help line **and** the README as the two places to say this; the README's second honesty note has it, the help line did not, and the gap was not in the recorded deviations. The added clause is that fact, not a new one. `envHelpText`'s test asserts flag names with `toContain`, so nothing there moves: `env.test.ts` 28/28, lint 897 clean.

Full gates re-run on the pushed head (`1d814a3`) merged with `origin/main`: typecheck **exit 0**, suite **exit 0** — `414 passed (414)` · `7569 passed | 3 expected fail | 2 skipped`.

---

### Worth doing later — for prd-27 w4, not for this PR

**The feed still doesn't carry the witness.** `packages/web/src/panels/feed/feed.ts:205` builds a `lane` entry from every `agent.status` and drops `event.source` on the floor, so a `sessionlog`-inferred `waiting` renders in the feed identically to a workmux declaration — two witnesses signing the same name, one surface over from where this PR ends it. `detectWaiting` and the ladder's `~` mark are correct; the feed is the gap. This is outside #281's fence (which claimed `web/disclosure/vocabulary.ts` and nothing else in `web/`), and prd-27 w4 already owns the dissent chip/hover form, so it belongs on w4's issue beside the two notes verify already left there — not here.

### On landing

Both fix commits are mine, so a second pair of eyes should land this rather than me merging my own repair (`AGENTS.md`, Review). CI has not yet re-run on `1d814a3`; the ten legs on `e341081` were all green, and the two commits on top are one Markdown paragraph and one help string.

Issue #293, opened 2026-09-06
prd17 w7: a mistyped load-batch argument cannot cost the landing its verdict


gabriel-canaan:
## Fence (may touch ONLY)
- `scripts/gate.sh`
- `packages/server/src/gate-honesty-law.test.ts`

## Blocked by
Nothing, but it should land AFTER #273 — both claim the gate script, so they are
a stack, never a bundle.

## Why
**This is pre-existing, not a regression.** `bb8cdf8`'s behaviour is identical,
and #273's review deliberately left it rather than widening that issue's scope.

`$3` (the load-batch count) is operator input and reaches two places unvalidated.
A review seat found that its NEIGHBOURS reach the emitter and cost the whole
record:

- `LOAD=abc` aborts at `$((LOAD*4))` under `set -u` — `abc: unbound variable`,
  exit 127, no `GATE FAILED:` line, **no beacon**. Loud, at least.
- `LOAD=3.0`, `LOAD=3x`, and `LOAD=^scripts/` (an argument transposition —
  `gate.sh 273 3 '^scripts/'`) are arithmetic SYNTAX errors, which are non-fatal
  under `set -uo pipefail`. They fall through the load gate, reach
  `emit_gate_verdict clean`, and `int()` raises inside the emitter — so
  **a landing that merged and pushed emits no verdict line at all.**

A validator at the emitter was tried during #273's review and REVERTED: it
guarded the one input that never reaches it while regressing one that does
(`LOAD=+3` ran three real load batches and was recorded as none). The lesson is
recorded on that branch — the validation belongs where `$3` is first used, not
where it is last read.

`gateVerdictPayloadSchema` refuses the emitted shape today anyway (`realEmittedShape=false`,
EXECUTED), so the blast radius until #274 is a record with no consumer. That is
why this is `soon`, not `now`.

## Definition of done
- `$3` is validated ONCE, where it is first used, before the load gate runs —
  not at the emitter. A value that is not a non-negative integer is refused with
  a `fail` naming a declared category, so the operator learns at the start of a
  landing rather than by its absence at the end.
- The accepted set is stated and is at least as wide as what `int()` accepted:
  `+3` and `3` are both three. `-1` is refused (the schema's `nonnegative`).
- A landing that reaches the merge ALWAYS emits exactly one verdict line,
  whatever `$3` was.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` pass.

## Sibling case
[...]

Issue #292, opened 2026-09-06
prd17 w7: the gate law's pins forbid the classes they are named for


gabriel-canaan:
> **Absorbed #291 and #306 on 2026-09-07.** Three issues, one defect class, one
> file, one fence — so one bundle whatever the tracker said. Both are closed
> pointing here, and neither's evidence is lost: each has its own section below.

## Fence (may touch ONLY)
- `packages/server/src/gate-honesty-law.test.ts`
- `scripts/gate.sh` (from #306 — the stale citations live in the script's own comments)

## Blocked by
Nothing. PR #310 merged on 2026-09-07 and released both files.

## Why

**Four instruments in this law assert a CLASS in their name and check a
SPELLING in their body.** Each was written to close a real defect a review had
just reported, and each closed the reported case only. They are listed
separately below because the evidence differs, but they are one issue because
the habit is the defect — and fixing them one at a time is what produced four
of them.

The strongest evidence is the fourth, because it happened in a single day, in
public, across three commits: the tsx-invocation pin went

| round | pin | beaten by |
|---|---|---|
| 3 | `.not.toMatch(/\bnpx\b/)` — a TOKEN | `npm exec --offline --yes -- tsx` |
| 4 (a) | the token adjacent to `-e` — a SUFFIX | `npx --no-install "<resolved path>" -e` |
| 4 (b) | anchored at the start of the line — a POSITION | a `\` continuation putting the command word on another physical line |
| review of #310 | the logical line | — |

Three fixes, each narrower than the property its own name asserted, each
certified by a mutation that the *next* reviewer's mutation walked straight
past. That is not four defects. It is one habit, and this issue exists to break
it rather than to add a fifth filter.

### 1. The no-second-line guard forbids two spellings, not the class — from #291

`#273` reverted a shell-built `printf` fallback from `emit_gate_verdict` after
it was found to emit **two** verdict events for one landing (it branched on
python's exit status, not on whether it produced output) and to interpolate the
handle into JSON unescaped, so a branch named `quote"handle` — which
`git check-ref-format --branch` accepts — produced unparseable JSON on exactly
the path meant to rescue the record.

The guard written to keep it out pins two spellings:

```
expect(machinery).not.toContain('"writer":"gate","kind":"gate.verdict"')
expect(machinery).not.toMatch(/\|\|\s*printf/)
```

The test's name says a fallback *cannot* add a second line. Two spellings are
actually forbidden. The behavioural half counts beacon lines by matching compact
[...]

Issue #291, opened 2026-09-06
prd17 w7: the no-second-line guard forbids two spellings, not the class (consolidated into #292)


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/gate-honesty-law.test.ts`

## Blocked by
Nothing. #273's branch carries the test this concerns; this issue can be built
once that lands.

## Why
`#273` reverted a shell-built `printf` fallback from `emit_gate_verdict` after it
was found to emit **two** verdict events for one landing (it branched on
python's exit status, not on whether it produced output) and to interpolate the
handle into JSON unescaped, so a branch named `quote"handle` — which
`git check-ref-format --branch` accepts — produced unparseable JSON on exactly
the path meant to rescue the record.

The guard written to keep it out pins **two spellings**, not the class:

```
expect(machinery).not.toContain('"writer":"gate","kind":"gate.verdict"')
expect(machinery).not.toMatch(/\|\|\s*printf/)
```

A review seat re-added a fallback carrying **both original defects**, respelled
as `if [ $? -ne 0 ]; then printf '{"v": 1, … "kind": "gate.verdict", …}' …; fi`
— exit-status branch, unescaped `$H`, spaced JSON. **The file stayed green,
202/202.** EXECUTED, by the seat.

The behavioural sibling beside it cannot cover the gap either: it counts lines
containing the compact `"kind":"gate.verdict"`, which spaced JSON does not match.

The test's name says a fallback *cannot* add a second line. Two spellings are
actually forbidden.

## Definition of done
- The guard forbids the CLASS — a second stdout-emitting path inside
  `emit_gate_verdict` — rather than two spellings of one. Enumerating more
  spellings is explicitly NOT the fix; that is the open-set trap this repo's
  verify skill names, and the next respelling defeats it again.
- A decoy, of the kind `classifyFailSite` already has, feeds the guard the
  respellings that beat it — at minimum: `|| printf`, `if [ $? -ne 0 ]`,
  spaced JSON, `echo` instead of `printf`, and a heredoc — and requires each to
  be caught.
- The behavioural half counts beacon lines by a spelling-independent property,
  not by matching compact JSON.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` pass.

## Sibling case
The guard is itself a grep-as-law, so it has the defect it is written to catch.
Whatever replaces it must be probed with a spelling not used in writing it —
this issue exists precisely because that step was skipped once.

[...]

Issue #274, opened 2026-09-04
prd17 w5: the gate's verdict reaches the beacon directory


gabriel-canaan:
## Fence (may touch ONLY)
- `scripts/gate.sh`

Widening is likely and is recorded on the PR before the change, per AGENTS.md:
the ADR this issue waits on may name a helper or a path constant belonging under
packages/server/, and packages/server/src/gate-honesty-law.test.ts is claimed
the moment an anchored line in the gate script is reworded.

- `packages/server/src/gate-honesty-law.test.ts` (packages/server/src/gate-honesty-law.test.ts — re-derive the pinned producer count and the doc comment that cites it, and add fixtures that set $root so the write path is reachable and both named mutations become provable. No loosening of any existing pin or anchor.)

## Blocked by
**#273 only.** #217 merged on 2026-09-04 in PR #267, so the target and the
envelope both exist: ADR-0036 fixes the directory, the per-writer file and the
v1 line, and prd-17 ruling 6 fixes what the gate puts on it.

**#273** — a stack, not a bundle. Both issues claim the gate script, so the lint
reports an OVERLAP and they may never be in flight together. #273 derives the
verdict; this issue redirects it. Land #273 first.

## Why
prd-17 ruling 2: the gate script writes one-line JSON beacons into the
instrument's own data directory, and a beacon collector tails them. #273 makes
the gate derive the verdict and print the v1 line. This issue is the last step —
the line reaches the door instead of stdout — and it is what makes a landing
appear in the record at all.

Small by construction: all the design and all the risk are in #273. What is left
here is an append and a directory.

## Definition of done
- The line #273 prints is appended to the beacon directory as `gate.jsonl`. The
  directory comes from beaconDirFor() in
  packages/server/src/collectors/beacon/paths.ts — resolve it, never re-derive
  the path in shell.
- **The directory is created if absent.** ADR-0036 is explicit that the
  collector never creates its own input: "the writer that appends is the one
  that knows the directory has to exist". A first landing after a fresh install
  is the case that proves it.
- One line per landing, newline-terminated, appended never rewritten. A
  concurrent landing must not interleave inside a line — the per-writer file is
  what makes that safe, so the filename is not negotiable.
- A write failure does **not** change the gate's verdict or its exit status. The
  landing tool's job is to land; a full disk or a missing directory must not turn
  a clean gate into a held one. It reports the failure and carries on.
- A malformed line is the collector's problem and it already survives one; this
[...]

PR #294, opened 2026-09-06
prd17 w5: the landing gate says what it decided

### LINKED ISSUE #273: prd17 w5: the landing gate says what it decided

gabriel-canaan:
Closes #273

prd-17 ruling 1 defines `gate.verdict` and #219 landed the contract. Nothing
emitted one, so a landing was invisible to the record: the gate script is the
instrument's own judgement on a lane, and the log had never heard of it.

The gate now derives its verdict and prints it as **one v1 beacon line**
(ADR-0036) on its own final line — `kind` from a declared vocabulary, `lane`,
`held`, `reason`, `loadBatches` when the run had any, and a sha256 of the gate's
full output as extra keys, which the beacon door carries and its digest covers.
Writing it anywhere is #274; this issue stops at stdout.

**The gate's own behaviour is unchanged.** A review seat verified that
structurally rather than by eye: it reconstructed the old script from the new one
and diffed against `bb8cdf8^` — one hunk, the added block. No check, no order, no
exit status moved.

## `held` is correct on both sides of the merge

`AGENTS.md`'s distinction is the law here: a check failing **before** the merge
holds the merge; the two that run **after** it hold only the push. All four
post-merge sites and the clean end report `held: false`; every pre-merge failure
reports `true`. Verified by line position, not by reading.

## What is the sibling case?

**The success path.** `fail()` exits, so a verdict emitted only from `fail()`
records every hold and no merge — the half of the record this PRD's thesis is
about. Both paths emit, and deleting the clean-path call reddens.

**The EXIT trap.** The script sets a trap and clears it with `trap - EXIT` a few
lines later, so capturing output through a new trap would have been silently
disarmed. No trap is used; `exec > >(tee …)` with fd 3/4 held is what makes the
digest cover a complete run.

**The boundary, not the reader.** The tests proved the emitter honours `MERGED`
by injecting `MERGED=1` in a fixture — so nothing exercised the script's own
`MERGED=1` line. Deleting it left the suite green while every post-merge failure
reported `held: true`. That seam is now asserted positionally.

**A failed push.** It was recorded as `clean` — byte-identical to a pushed
landing in every categorical field, separable only by an opaque digest. That is
the shape `AGENTS.md`'s "a green gate on a branch whose base is not `main`"
section exists for. It is now `push-failed`, still `held: false` because the
merge is real.

16 mutations, each against `gate-honesty-law.test.ts` with a clean-tree control
either side. Nine were **green before** the commits that close them:

| mutation | |
|---|---|
| delete `MERGED=1` / move it too early / too late | 1 red each — delete was GREEN |
| comment out `MERGED=1` | 1 red — was GREEN |
| emit `clean` unconditionally after a failed push | 1 red — was GREEN |
| `push_or_warn` stops recording the failure | 2 red |
| delete the clean-path emit | 1 red — was GREEN |
| drop a `fail` site's category / misspell it `suite_red` / compute it | 1 red each — first two GREEN |
| revert the site detector to the pre-fix regex | 1 red — was GREEN |
| emit `outputDigest` when empty | 1 red — was GREEN |
| let a timed-out drain bind a partial digest | 1 red — was GREEN |
| point the digest at the empty string | 2 red |
| force `held` true / delete the flip | 2 red each |

## Two repairs were reverted rather than patched

Worth reading before reviewing the diff, because the change is **smaller** than
what it replaced.

A `printf` fallback for a failing python3 branched on python's exit status rather
than on whether it produced output — so a python that printed its line and then
exited non-zero emitted **two** verdict events for one landing — and it
interpolated the handle into JSON unescaped, so a branch named `quote"handle`
(which `git check-ref-format` accepts) produced unparseable JSON on exactly the
path meant to rescue the record.

A `LOAD` validator guarded an input that dies 340 lines earlier and regressed one
that does not: `LOAD=+3` ran three real load batches and was recorded as none —
the same "two outcomes byte-identical in the record" defect as the push finding,
one screen away.

Both guarded cases that were imagined rather than shown reachable, in the
**landing tool**. A `command -v python3 … || fail` already hard-fails the gate
when python3 is absent, so the fallback protected a state the script refuses to
run in.

## Verification

Three passes. **Every finding across all three was in a repair written during
review; none was in the lane's original build.**

| pass | seats | verdicts | findings |
|---|---|---|---|
| initial | **1** | Opus not ready | 7 |
| fix re-review 1 | 2 | both not safe | 7 |
| fix re-review 2 | 2 | Opus **safe to merge** / Codex not safe | 5 |

**The initial pass ran one seat, not two** — the Codex seat died at launch with
`Your workspace is out of credits`. The orchestrator ran that seat's share of
probing by hand, which is how the `MERGED=1` gap was found.

Round 2's split resolved against the code: Codex's blocker required manually
unsetting `GATE_TEE_PID`, which is unreachable — `$!` with no prior job aborts
under `set -u`, and `exec > >(tee …)` immediately precedes the assignment.

**The fix-re-review budget refused round 3** on its two-round cap, and that
refusal was left standing rather than overridden. The remaining findings are
filed with their evidence rather than chased: **#291** (the no-second-line guard
forbids two spellings, not the class), **#292** (the anchor and fail-site helpers
are fitted to reported spellings), **#293** (a mistyped load-batch argument
silently costs the landing its verdict — pre-existing, identical at `bb8cdf8`).

**Two commits carry FORCED ledger rows**, both stated there in full: `e891281`
is comments-only, which the verify skill's cost guard exempts by name; `96f81e3`
was reviewed by two seats whose reports were written outside `.git` via `--out`,
so the freshness heuristic cannot see them — a bookkeeping gap in where reports
landed, not a review that did not happen.

Gates: `npm run typecheck` clean, `npm run lint` 897 files clean,
`VITEST_MAX_WORKERS=6 npm test` → 414 files, 7589 passed, 3 expected fail,
2 skipped.

---

## Declared review gap

> Two of the eight commits carry FORCED ledger rows, both stated in the body: e891281 is comments-only (the cost guard exempts it by name), and 96f81e3 WAS reviewed by two seats whose reports were written outside .git via --out, so the freshness check cannot see them. The other six carry seat reviews across three passes. Separately, the initial pass ran ONE seat because the Codex seat died with 'workspace is out of credits'; the orchestrator covered that seat's probing by hand and found the MERGED=1 gap doing so.

```
  forced  e891281  docs(gate): the no-fallback rationale rests on the presence check, not a wrong count  (review of #273)
          ^ Comments only, in one shell script and one test docblock — no behaviour, no assertion, no pinned count. It corrects a wrong figure ('four other places') that a round-2 seat recounted, in the two live places it appears. The verify skill's cost guard exempts docs/comments-only diffs by name. Recorded rather than seat-reviewed so the exemption is visible.
  forced  96f81e3  fix(gate): two speculative repairs are reverted, and three real ones land  (review of #273)
          ^ Round 2 of the fix re-review DID run, two seats, on this exact sha — the reports are at scratchpad/w5-verify/rr2/{opus,codex}.md because the run used --out to keep them out of .git, which is why the freshness heuristic cannot see them. This is a bookkeeping gap in where the reports were written, not a review that did not happen. SPLIT verdict: Opus SAFE TO MERGE having independently re-derived all three mutation results and the gates (414 files, 7589 passed); Codex not-safe with three findings, its blocker refuted as unreachable ($! with no prior job aborts under set -u — verified) and its other two being open-set spellings. The findings worth acting on are filed as #291, #292 and #293.
  ok      13113ed  fix(gate): the verdict survives the emitter's own failures  (review of #273)
  ok      2a82e0e  test(gate): the untagged-fail assertion can fail  (review of #273)
  ok      3868be0  fix(gate): a failed push is a category, not a footnote  (review of #273)
  ok      c0bff53  test(gate): the MERGED=1 boundary is asserted, not just the emitter that reads it  (review of #273)
  ok      bb8cdf8  feat(gate): the gate derives its own verdict and prints it as one v1 beacon line   (#273)
──
2 commit(s) carry a FORCED row, which is a stated reason, not a review.
```


ciaran-slow (approved):
## Approved — with one contract question the operator should settle before #274 dispatches

Reviewed commit by commit against #273. Everything below that says EXECUTED, I ran; everything that says REASONED, I argued from the code with nothing run.

### What verified, independently of the PR body

**The central safety claim — "the gate's own behaviour is unchanged" — holds structurally. EXECUTED, re-derived rather than taken from the body.** I reconstructed the pre-PR script from the post-PR one (delete the emit call in `fail()`, the vocabulary-through-`GATE_TEE_PID=$!` block, `PUSHED=1`, the trailing emit block; restore `push_or_warn`'s one-line body; strip the trailing category argument from every `fail` site) and diffed the result against `origin/main:scripts/gate.sh`:

```
diff -u old.sh recon.sh   →  identical but for the trailing newline my truncation dropped
```

No check, no order, no exit status moved. That is the property that matters most in the landing tool, and it is now checkable by anyone who wants to repeat it.

**The emitter actually runs. EXECUTED under `/bin/bash` 3.2.57, the only bash on this Mac**, sourcing the real machinery (`sed -n '1,180p' scripts/gate.sh`) rather than a retyped copy:

| probe | result |
|---|---|
| clean path, `MERGED=1` | `{"v":1,…,"held":false,"reason":"clean","outputDigest":"9ad480ff…"}` |
| `fail "test suite red" suite-red` | exit 1, `held:true`, `reason:"suite-red"`, JSON is the final line |
| digest vs the capture | `shasum -a 256` of the tee'd file == `outputDigest`, **including** the `GATE FAILED:` and `>>> HOLDING` lines |
| handle `quote"handle` | `"lane":"quote\"handle"`, `json.load` parses it |
| orphan holding fd 1, writing at 12s | **9.85 s real**, `outputDigest` absent, `did not drain` on stderr, verdict still printed |

`$!` after `exec > >(tee …)` is set on bash 3.2 — I checked, because that assignment under `set -u` would otherwise abort the landing tool on every macOS box. It is fine.

**The round-3 story checks out from the run log, not from the body.** The CI failure at `e891281` was exactly one test — `EXECUTED — a drain that times out omits the digest rather than binding a partial one`, 12300 ms, `AssertionError: a partial capture must not be reported as a digest` — on `build-test-boot (macos-latest, current)`. `6c31c34` turned it green. The loop-counter-is-not-a-bound diagnosis is the right one and the wall-clock fix measurably works.

**Mutations, mine not the body's.** `npx vitest run packages/server/src/gate-honesty-law.test.ts` on node 22.23.2, control 202/202:

| mutation | |
|---|---|
| `"writer": "gate"` → `"notgate"` | 4 red |
| `{"v": 1` → `{"v": 2` | 4 red |
| `if load != "0":` → `if True:` | 2 red |
| drop `exec 1>&3 2>&4` | 12 red |
| `"kind": "gate.verdict"` → `"gate.held"` | 5 red |
| drop the hex-shape guard on `$digest` | **green** |
| drop the `[ -n "${GATE_OUTFILE:-}" ] \|\| return 0` re-entrancy guard | **green** |

The two survivors are not defects — both guard states I could not find a path to, and I am not asking for tests that pin unreachable behaviour.

**CI.** All three jobs, ten legs, green on the head sha `6c31c34`. I checked the steps rather than the bar: on `build-test-boot (ubuntu-latest, current)` every one of Build / Test / Typecheck / Lint / packaging guard / boot smoke is `success`, the only `skipped` being the min-node setup on a current leg, which is the matrix working.

**Fence.** `scripts/gate.sh` and `packages/server/src/gate-honesty-law.test.ts`, exactly #273's declared fence. Nothing widened.

### The one thing worth a ruling: what goes in `kind`

REASONED. #273's Definition of done says the line carries

> `kind` from the vocabulary below

and the vocabulary below is the one the 42 `fail "` sites collapse into. Ruling 6 (landing in #295) says the same thing in the same words: *"a v1 line whose `kind` comes from prd-17's decision vocabulary"*. What ships puts the constant `gate.verdict` in `kind` and the category in `reason`, an extra key.

I am not calling that wrong, and I did not change it — but it is a contract, #274 writes these lines to disk, and a `kind` that changes after they are written is a migration. The argument each way, so the ruling can be made rather than inferred:

- **for the constant.** `writer: "gate"` already selects gate lines, and `kind: "gate.verdict"` names the family #280 derives. The category survives in the sidecar, digest-covered, which is exactly the sidecar-for-content split ruling 2 draws.
- **for the category.** `beacon.received`'s own payload is all the record holds between #274 and #280, and with a constant `kind` every landing in that window looks identical at the event layer — held or clean, only the sidecar can say. prd-27's own use of the same free-form field (`waiting` / `working` / `stopped`, `packages/core/src/events/beacon.ts:18`) puts a *state category* there, not a family name.

If the answer is "the category", it is a one-line change to the `python3` block plus the tests that pin `gate.verdict`. If it is "the constant", ruling 6's sentence is the loose half and wants a word when someone next touches that PRD.

### Noted, not blocking

1. **The declared review gap stops one commit short of the head.** The ledger in the body ends at `e891281`; `6c31c34` is not in it. That is the commit that fixed a CI-red *in the landing tool*, so it is the one most worth declaring. Treat this review as closing that gap — it is the commit I probed hardest, and its behaviour is in the table above.
2. **#293 reproduces. EXECUTED.** With `LOAD=abc`, `emit_gate_verdict clean` prints nothing at all — `int(load)` raises, stderr is discarded, the landing merges and the record hears nothing. Filed, reasoned, and the deferral is right: the validation belongs where `$3` is first consumed, 340 lines earlier, not at the emitter. Not touching it here — #273 and #293 both claim this file and are a stack.
3. **Offered, not asserted.** `GATE_OUTFILE` leaks on an abnormal abort — a SIGINT during the load gate, or the `set -u` abort #293 names — which is the class `96f81e3` fixed for every normal path. And `2>/dev/null` on the verdict-printing `python3` makes a broken emitter silent rather than loud; on the path where the shell has already decided to stay out of the JSON business, letting the traceback reach fd 4 costs nothing.

### What I did not run

The full suite on this branch. CI ran it on three legs plus `windows-suite` at `6c31c34` and I confirmed the steps executed; I ran only the changed test file locally, plus the seven mutations above.

Not merging — landing is `scripts/gate.sh`'s own job and the operator's step.

Issue #280, opened 2026-09-04
prd17 w6: a landing's verdict is a gate.verdict event, not only a beacon


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/core/src/reduce.ts`
- `packages/core/src/reduce.test.ts`
- `packages/server/src/log/`

`packages/server/src/collectors/beacon/` is EXPLICITLY OUT — owned by prd-27.
This issue reads what that collector emits and never changes it.

- `packages/server/src/server/poll-loop.ts` (the gate.verdict recording seam only - no other collector or poll-loop behaviour)
- `packages/server/src/server/poll-loop.test.ts` (the gate.verdict recording seam only - no other collector or poll-loop behaviour)
- `packages/core/src/eras/eras.test.ts` (the gate.verdict recording seam only - no other collector or poll-loop behaviour)

## Blocked by
**#274.** Nothing writes a gate beacon until it lands, so there is no sidecar to
derive from.

## Why
prd-17 ruling 6, ruled 2026-09-05 after ADR-0036 landed: the gate's verdict
rides as extra keys on a beacon line, and **deriving gate.verdict from that
sidecar is this wave**. Read ruling 6 first — it records why the two shapes do
not meet and what was rejected.

After #273 and #274, a landing is on the record as an occurrence: beacon.received
carries writer, kind, lane, the line's digest, the file, and the offset. That is
everything needed to read the verdict back out of the sidecar and prove it
unaltered. What is missing is the typed event ruling 1 defined and #219 landed —
gate.verdict — whose arm in reduce.ts still returns state unchanged.

**The derivation cannot live in core.** eras/fold.ts records the reason in its
own docblock: core has no Node types in scope and is bundled into the browser,
so it cannot read a file. The sidecar read is server-side, and the fence above
reflects that. Where exactly under packages/server/src/log/ it sits is this
issue's call to make and to record in the commit body — but not inside prd-27's
collector, which must keep emitting beacon.received and nothing else. A
collector that mapped kinds onto families would be a second place where an
event's identity is decided.

## Definition of done
- A gate beacon on the record yields a gate.verdict event whose payload matches
  what the gate actually decided: handle, held, reason, the digest of the gate's
  own output, and loadBatches when the run had any.
- The derivation verifies the line against the digest beacon.received carried
  before trusting its contents. An unverifiable or altered line does not produce
  a verdict, and says so rather than failing silently — ADR-0011's rule.
[...]

PR #296, opened 2026-09-06
prd27 w3: the fleet believes a declaration, and every surface names its witness

### LINKED ISSUE #283: prd27 w3: the fleet believes a declaration, and a disagreement says so
## Fence (may touch ONLY)
- `packages/core/src/reduce.ts` (the body of the existing `beacon.received` arm, and nothing else)
- `packages/core/src/state.ts` (a `declared` record per lane)
- `packages/core/src/fleet/buildFleet.test.ts`
- `packages/core/src/fleet/fixtures.ts`
- `packages/core/src/selectors/condition.ts`
- `packages/core/src/selectors/condition.test.ts`
- `packages/server/src/api/meta.test.ts`
- `packages/web/src/app/StatusBar.tsx`
- `packages/web/src/app/StatusBar.test.tsx`
- `packages/web/src/panels/fleet/format.ts`
- `docs/screenshots/**` (the STATE hover, regenerated)

## Blocked by
**#281** (wave 2) — `AgentState.witness` and `Lane.agentStatusWitness` are what a declaration is compared against; without them the disagreement voice has one side.
**#282** (wave 2) — `BEACON_ATTENTION_KINDS` is imported from `packages/core/src/events/beacon.ts`, never respelled here.

Sibling lanes: **#280** (prd17 w6, groomed, not live) also claims `packages/core/src/reduce.ts` and `reduce.test.ts` for the `gate.verdict` arm and a `gate`-kind reading of `beacon.received`. This issue folds only kinds in `BEACON_ATTENTION_KINDS` and leaves every other kind on the log unread, so the two arms compose; whichever lands second rebases one `case`. Do not touch `packages/server/src/collectors/**`, `packages/core/src/events/**`, or `packages/core/src/collector.ts` (the rung is w4's).

**Model:** opus. **Wave:** 3.

## Why

prd-27 rulings 3 and 4, and the half of Success criterion 1 the door did not deliver: *an instrumented agent that stops for a human is named blocked because it said so.* On `main` a `beacon.received` is recorded and folded to `state` unchanged (#217 left the home for #218 to build); `detectWaiting` knows two witnesses — workmux's certain report and the pane-stillness inference — and cannot see a beacon at all. The provenance bar still lists five sources and `/api/meta`'s ladder still hard-codes six collectors, so the beacon is invisible on the instrument's own capability manifest (the verify note on #217).

## Definition of done

- `SessionState.declared[lane]` holds the latest attention beacon per lane — `{ kind, at, writer, digest, file, offset }` — written by the `beacon.received` arm **only** for kinds in `BEACON_ATTENTION_KINDS`; any other kind still folds to `state` unchanged (a test folds a `landed` beacon and asserts `declared` is untouched). Folding the same run of beacons twice yields `toEqual` states; a beacon with `lane: null` is never indexed.
- `Lane` carries `declared: { kind, at } | null` from `buildFleet`, keyed by the lane handle the writer chose (the identity join is w4's; here the handle must equal the lane id, and a beacon for an unknown handle produces **no lane** and no alarm).
- **Ruling 4's asymmetry, each clause a test:** (a) a declared `waiting` is a WAITING pathology with `inferred: false`, evidence `beacon (claude-hook) declares waiting …`, `since` = the beacon's `at`; (b) a declared `working` whose `at` is newer than `lane.lastWorkTs` suppresses the pane-stillness inferred WAITING and the sessionlog-witnessed inferred WAITING (#281); (c) an inferred `working` from either organ never suppresses a declared `waiting`; (d) a declared `stopped` is not WAITING and does not suppress anything.
- **Disagreement is voiced, never resolved silently.** When `declared.kind` and the inferred reading disagree, the condition's `why.evidence.fact` carries both witnesses and which one the label followed, e.g. `beacon declares waiting 40s ago · transcript shape reads working` — deterministic to the byte. The visual form is the hover card (`selectLaneCondition` → `DisclosureCard`), not a new chip, and the choice is written in the code comment (ruling 4 leaves it to the implementer).
- `beacon` joins `/api/meta`'s `LADDER_COLLECTOR_NAMES` and `DECLARED_CAPABILITIES` with `BEACON_CAPABILITIES`; `meta.test.ts`'s parallel list gains it; the merged rung is unchanged by the addition (a test asserts `rung` before and after adding `beacon` are equal — its manifest is all-absent in this wave).
- `StatusBar.tsx`'s five sources become six with `beacon` labelled `Beacon`; the doc comment stops saying five; `StatusBar.test.tsx` reconciled.
- `docs/screenshots/**`: the STATE hover showing a declared WAITING and a disagreement, regenerated. `CHANGELOG.md` `[Unreleased]` › `Added`.

## Sibling case

Ruling 4 is four clauses and each has a mirror: a suppression test for `working` that forgets the sessionlog-witnessed inference (#281 adds a third witness) leaves one inferred WAITING that a declaration cannot quiet; a `stopped` treated as `waiting` raises a summons for a lane that has simply finished. And the #133 shape again: a beacon for a handle the fleet does not know must not conjure a lane — a summons for a lane nobody can attach to is the false summons in a new costume.

Make the fold write `declared` for every kind: the `landed`-beacon test reddens. Invert clause (c) so inferred `working` clears a declared `waiting`: the asymmetry test reddens. Drop the disagreement string: the byte-exact evidence test reddens. Survives, and known: the lane-handle join is by equality only — a writer choosing a different handle spelling than the worktree's lane id is w4's identity work, and this issue's test uses matching handles on purpose.

### LINKED ISSUE #290: prd27 w3: the feed names which witness signed an agent.status
## Fence (may touch ONLY)
- `packages/web/src/panels/feed/feed.ts`
- `packages/web/src/panels/feed/feed.test.ts`
- `packages/web/src/panels/feed/index.tsx`
- `packages/web/src/panels/feed/index.test.tsx`
- `packages/web/src/panels/feed/format.ts`

## Blocked by
Nothing — **#281** landed (`02cc2b5`, on `main` at `29cde14`), which is what makes the second witness reachable here at all.

Sibling lanes: **#283** (prd27 w3) claims `packages/web/src/app/StatusBar.tsx` and `packages/web/src/panels/fleet/format.ts` and owns the *hover card* form of a disagreement; it does not claim `panels/feed/**`, and this issue does not claim anything of its. **#218** (w4) claims neither. `packages/web/src/panels/feed/format.ts` is listed here only for the label helper if the render needs one — if it does not, drop it from the fence rather than touch it.

**Model:** sonnet. **Wave:** 3 — it depends only on w2, and nothing in w3 depends on it.

## Why

#281 landed the fix for exactly this defect one layer down: `agent.status` used to pin `source: 'workmux'`, so the transcript organ could not publish without signing workmux's name, and *"two witnesses signing the same name cannot be seen to disagree"* (prd-27's own evidence section). The envelope, `AgentState.witness`, `Lane.agentStatusWitness` and `detectWaiting` all now carry the witness, and the ladder renders an inferred WAITING with the `~` mark.

**The feed does not.** `packages/web/src/panels/feed/feed.ts:205-215` builds its `lane` entry from every `agent.status` and reads `handle`, `status`, `branch` and `detail` off the payload — `event.source` is never read, and `LaneFeedEntry` (`feed.ts:66-74`) has no field for it. `LaneRow` (`index.tsx:229-241`) then renders `<KindTag>{AGENT_STATUS_LABEL[entry.status]}</KindTag>`, so a `sessionlog`-inferred `waiting` and a `workmux`-declared `waiting` produce a byte-identical tag.

Before #281 this was correct by construction — there was only one witness. Since `29cde14` there are two, both streaming into the same feed, rendered as one. That is the #133 false summons in its third costume: the operator reads a transcript-shape guess as workmux's declaration, on the one surface that shows every event as it arrives.

The `detail` column is a partial tell and not a fix — the organ's detail is now its verbatim reading (`WAITING — tail turn-complete, quiet 45s, threshold 30s`, since `agentStatusEmissionFor` dropped the `transcript-tail: ` prefix in `02cc2b5`), but workmux may carry a `detail` too, and nothing tells the reader which column is doing the work.

Found at review of #287 (`REASONED`: read from `feed.ts` and `index.tsx`, not executed against a running feed). Out of #281's fence, which claimed `packages/web/src/disclosure/vocabulary.ts` and nothing else under `web/`.

## Definition of done

- `LaneFeedEntry` carries the witness, read from `event.source` and nothing else — never re-derived from `detail`'s shape, which is what makes it a second source of truth that can drift.
- `LaneRow` renders a declared and an inferred `waiting` **differently**, and the difference is the one the rest of the instrument already uses: `INFERRED_MARK` from `packages/core/src/fleet/pathology.ts`, imported, never respelled. A test asserts the two rows differ, and asserts the mark's placement against the constant.
- The three `agent.status` words are covered, not just `waiting`: a `sessionlog` `working` is marked too, and a `workmux` row is byte-identical to what it renders today (a regression test on the existing shape, so this change cannot quietly restyle the declared case).
- An `agent.status` from a source the schema does not yet admit is impossible by the envelope, so there is no third arm to write — but the render must be total over `AgentStatusWitness` rather than defaulting an unknown witness to "declared". A `satisfies`/exhaustive-switch shape, so adding a third witness to `AGENT_STATUS_SOURCES` fails at typecheck rather than silently rendering it as workmux's word.

## Sibling case

The obvious one is the other feed rows: `commit`, `landing` and `collector` entries are all single-witness today and stay that way — do not generalise this into a witness column on `FeedEntry`, because three of the four kinds would carry a field that is always the same value, which is how the next reader concludes the field means nothing.

The one that actually costs something: **`resolveLane`, one line above the payload read.** It joins on `[handle, branch ?? null]`, and the sessionlog organ emits `branch` from its own worktree reading rather than from workmux's roster. If the two witnesses ever disagree about a lane's branch, the same handle lands two feed rows under two different `laneId`s and the disagreement this issue exists to *show* becomes a duplicate row instead. Assert the join is witness-independent, or state on the issue why it cannot be.

Set the witness on the entry from a constant `'workmux'` instead of `event.source`: the declared-vs-inferred render test must redden. Delete the `INFERRED_MARK` import and retype the mark as a literal: the test that reads it from the constant must redden when the constant changes alone — a test that hard-codes the glyph is the `TTL - 1` / `TTL + 1` shape and survives this whole issue.

**Known survivor:** a test that only asserts `LaneFeedEntry` *has* a witness field survives every row above, exactly as #281's schema-accepts-sessionlog test did.

ciaran-slow:
Closes #283
Closes #290

prd27 wave 3, **two issues, two commits**. The PR is the unit of queueing and the commit is
the unit of review, so the wave pays the ~21 h toll once while each issue stays a standalone
commit. Assembled on a branch taken off `origin/main`, not off either lane.

| commit | issue | lane sha | built | verified |
|---|---|---|---|---|
| `01058e5` | **#283** the fleet believes a declaration, and a disagreement says so | `50b392f` | opus | fable |
| `6b57312` | **#290** the feed names which witness signed an `agent.status` | `d6d008b` | sonnet | opus |

The two fences are disjoint — core fold / fleet / `/api/meta` / status bar for #283,
`packages/web/src/panels/feed/**` for #290 — and `scripts/fence-lint.sh 283 290` →
**PASSED**, no overlaps, re-run on the assembled branch.

---

# #283 — the fleet believes a declaration, and a disagreement says so

Built on **opus**, verified on **fable** (`Co-Authored-By:` trailers carry both). Verify
found two must-fix items; both are fixed and amended into the same single commit, and the
verifier re-ran every gate against the fixed sha before calling the branch ready.

---

## What became true

- **`SessionState.declared`** (`state.ts`, `reduce.ts`) — one `DeclaredAttention` record
  per lane, written by the `beacon.received` arm **only** for kinds in
  `BEACON_ATTENTION_KINDS`, imported from `events/beacon.ts` and never respelled. An older
  `at` never rolls a lane back; an equal `at` yields to log order, which is what keeps
  refolding idempotent. `lane: null` is never indexed. A `landed` — or any kind this wave
  has not met — folds to `state` unchanged and stays on the log for its own reader, so
  #280's `gate` kinds compose here as a second `if`, not a rewrite.
- **`Lane.declared`** (`types.ts`, `buildFleet.ts`) — `{ kind, at, writer }`, joined on
  `state.declared[draft.id]` by handle equality only (the identity join across spellings is
  w4's). Nothing creates a draft for a beacon: an unknown handle is simply never read, so a
  ghost beacon can mint no lane and raise no summons.
- **`detectWaiting`'s third witness** (`diagnose.ts`) — prd-27 ruling 4's asymmetry as four
  clauses, each its own test: (a) a declared `waiting` is WAITING, `inferred: false`, `since`
  = the beacon's `at`; (b) a declared `working` newer than `lane.lastWorkTs` suppresses both
  inferred WAITINGs (pane-stillness and sessionlog-witnessed), a stale one suppresses nothing
  and says so; (c) an inferred `working` never suppresses a declared `waiting`; (d) a declared
  `stopped` is not WAITING and suppresses nothing.
- **The call the issue left open, made and recorded.** Ruling 4 orders declarations over
  inferences, not declarations over declarations. Between a beacon and a workmux poll,
  **the newer word stands and the older is voiced** — the later moment is the truer one.
  Written as a code comment at the branch in `detectWaiting`, and in the changelog.
- **Disagreement is voiced, never resolved in silence**, and the visual form ruling 4 leaves
  to the implementer is **the hover card**, not a new chip — recorded in the same comment.
  The evidence string is byte-deterministic (only `formatSpan` output and fixed literals) and
  reaches `why.evidence.fact` through `pathologyCondition`; `declaredClause` in `condition.ts`
  names the declaration on the `working`/`idle`/`unknown` and waiting-fallback arms, so a calm
  lane's reader still learns what the harness said.
- **`/api/meta`** (`meta.ts`) — `beacon` is the seventh ladder collector with
  `BEACON_CAPABILITIES`. It moves no rung, and the doc comment says *why* the seventh's
  manifest is all-absent (ruling 3's carve-out; the rung distinction is w4's).
- **Status bar** (`StatusBar.tsx`) — a sixth pill, `Beacon`, whose flow is computed off the
  same fold (`session.declared`). `CONNECTION_SOURCES` and `selectConnection` were **not**
  touched, so prd-19 ruling 4's "one pass over the same fold" still holds.
- **The fixture** (`fixtures.ts`) — the pathology fixture's `43-drawer-attach` keeps
  `behaviour: 'waiting'` and all its history, but `laneStatus` now returns `working` for it
  and a `laneBeacon` fires one second after its last work. That is the real race ruling 4 was
  written for, and it is what makes the deferred screenshot a one-command capture.

`packages/web/src/panels/fleet/format.ts` needed **no change**, exactly as the plan predicted.

---

## What is the sibling case?

Ruling 4 is four clauses and **each has a mirror that a fix for one silently breaks**, so
each mirror is its own test:

- **The suppression that forgets the second inference.** #281 added a third witness, so a
  `working` declaration that quiets only the pane-stillness WAITING leaves the
  *sessionlog-witnessed* one standing — a declaration that cannot quiet an inference is
  ruling 4 inverted. `(b1)` and `(b2)` are the same clause against the two different
  inferences.
- **A `stopped` read as a `waiting`** summons a lane that has simply finished. `(d)` asserts
  both halves: `stopped` alone raises nothing, and `stopped` beside a standing inference
  appends nothing to it.
- **The recency gate.** A declaration that quiets an inference *regardless of when it was
  made* lets a five-minute-old `working` silence a live raised hand. `(b3)` is the
  stale-declaration mirror of `(b1)`.
- **#133, in a new costume.** A beacon naming a handle the fleet has never heard of must
  conjure **no lane** — a summons for a lane nobody can attach to is the false summons this
  instrument's scar tissue is made of. Asserted against a lane-for-lane comparison of the
  same log with and without the ghost beacon, plus the ladder rank, plus no pathology
  mentioning it. The `lane: null` sibling is asserted separately at both the fold and the
  fleet.
- **The one verify added:** a declared `waiting` on a **removed worktree**. A beacon record
  stands in `state.declared` forever, so without the `lane.present` guard every landed lane
  whose last word was `waiting` would summon indefinitely — the same shape again. The guard
  was already correct in the build; it had no test, and now has two.

---

**Sixteen mutations run by the verifier**, each applied to the source, targeted suites run,
then reverted (all EXECUTED). The three the builder ran are a subset. Fifteen of the sixteen
reddened, most at **both** levels — the detector's byte-exact block *and* the end-to-end fold:

| # | mutation | reddened |
|---|---|---|
| M1 | fold writes `declared` for every kind | 3 (reduce ×2 + StatusBar `landed` pill) |
| M2 | index `lane: null` under `"null"` | 1 |
| M3 | last line wins regardless of `at` | 1 |
| M4 | invert (c): inferred `working` clears a declared `waiting` | 11 |
| M5 | (a) `since` = `lastWorkTs` | 2 |
| M6 | drop the disagreement clause | 5 |
| M7 | a newer workmux word no longer overrides an older beacon `waiting` | 1 |
| M8 | `stopped` suppresses like `working` | 2 |
| M9 | (b) never fresh | 3 |
| **M10** | **drop `lane.present` from clause (a)** | **none — 134/134 green** → fixed, see below |
| M11 | newer beacon `working` no longer stands over workmux `waiting` | 1 |
| M12 | drop the older-beacon voice from the workmux arm | 2 |
| M13 | drop the stale-`working` voice from the inferences | 2 |
| M14 | idle arm forgets the declaration | 1 |
| M15 | forget `beacon` in `LADDER_COLLECTOR_NAMES` | 4 |
| M16 | `QUIET_WORK_TS` back to the plan's 10 minutes (past FROZEN) | (a) and (c1) fail outright — proof the re-dating below is real and the tests were never vacuous |

**M10 was the one survivor, and it is fixed.** After the verify-fix, re-applying M10 reddens
**exactly the two new tests, 2 failed / 111 passed** (EXECUTED, then reverted).

**Known survivors, stated rather than hidden:** the lane-handle join is equality on `lane.id`
only — a writer spelling the handle differently from the worktree's lane id is w4's identity
work, and the issue says so.

---

## Plan deviations

1. **Two test timestamps re-dated, to keep the plan's byte-exact strings honest.** The plan's
   `llm.usage` at `NOW - 10 * 60_000` puts the lane past `FROZEN_AFTER_MS` (8m), where
   `diagnose` suppresses WAITING behind FROZEN — every assertion would have gone vacuously
   `undefined`. It is now a named `QUIET_WORK_TS = NOW - 120_000`, past `WAITING_QUIET_MS`
   (75s) but inside the frozen threshold, with a comment recording both bounds and that each
   was found by a red test. Same fix in the `diagnose.test.ts` helper. Relatedly, the plan's
   `(c1)` put the sessionlog `agent.status` at `NOW - 5_000`; an `agent.status` refreshes
   `lastWorkTs`, so that also made the *activity* witness read `working` and appended a
   second, equally-true clause — dated at the same quiet mark so `(c1)` isolates the organ
   and `(c2)` keeps the activity case. Verified as real by M16, not taken on trust.
2. **`StatusBar.tsx` uses the plan's `Pick<>` fallback, not its primary cast form** — the
   verifier's finding 2, applied. The plan's primary `'beacon' as SourceFlow['source']`
   compiles, but so does `'zzz-not-a-source' as SourceFlow['source']` (EXECUTED: `tsc
   --noEmit` exit 0 on a scratch file), so the cast is permissive, not type-safe. `beaconFlow`
   now returns `Pick<SourceFlow, 'firstEventTs' | 'lastEventTs' | 'count'>` — no `source`
   field to lie about — and `sourceStatus`'s second parameter narrows to
   `Pick<SourceFlow, 'firstEventTs'>`, the one key it reads. No runtime change;
   `StatusBar.test.tsx` unedited and green. The comment records why the primary form was
   rejected.
3. **`Lane.declared` is `{ kind, at, writer }`** — a superset of the issue's `{ kind, at }`,
   declared in plan §1.5 before the change (the evidence string needs the writer), not drift.
4. **No structural deviation.** No change to `plumbing.ts`, `selectors/connection.ts`,
   `collectors/` or `events/`.

---

## Fence widenings

**Two declared in the plan, before the change:**

1. `packages/core/src/eras/era-1/session-state.snapshot.json` + `packages/core/src/eras/CAPTURE.md`
   — `SessionState` gained the `declared` slice, and the golden era-1 fold is compared
   byte-for-byte. Re-blessed with the exact one-liner in `CAPTURE.md` § "The blessing rule";
   the diff is **exactly one added line** `"declared": {},` at the sorted-key position after
   `commits`, before `errors` — no other key moved (EXECUTED: `git diff`). `CAPTURE.md`
   carries the re-bless note naming the reducer change and why the new fold is the correct
   meaning of that old log. Same widening #281 made for `witness`/`dissent`; same reason.
2. `CHANGELOG.md` — the wave's standing convention.

**Three more, recorded in the build report — each forced by a law that exists to force
exactly this decision**, none optional:

3. **`packages/web/src/fleet/FleetContext.tsx`** — `FLEET_INPUT_SLICES` gains `declared`,
   count 11 → 12. Load-bearing, not cosmetic: `buildFleet` now *reads* `state.declared` and
   the memo key is spread from that constant, so leaving it out would mean a beacon arriving
   **does not rebuild the fleet** — a declared WAITING would not reach the strip until some
   other change happened to rebuild. The law *every `SessionState` key is either a fleet input
   or a declared non-input* is what caught it, which is precisely its stated job.
4. **`packages/web/src/fleet/FleetContext.test.tsx`** — the paired count law (11 → 12) and its
   wording.
5. **`packages/server/src/server/build-app.test.ts`** — the exhaustive `/api/meta` body
   assertion gains the `beacon` capability entry. Its own comment says it is "kept exhaustive
   on purpose: this assertion is what catches an additive `/api/meta` field that forgot to tell
   anyone". `rung: 'L4'` staying put beside the new line is itself the assertion that the
   seventh collector moved no rung.

This is the known shape the repo's own notes record — *a groomed fence names the feature files
and misses the count laws and enumerations the change moves*. Worth folding back into grooming
for w4.

The full diff is **21 files** = the issue's fence (17 entries; `panels/fleet/format.ts`
in-fence and untouched) ∪ the two plan widenings ∪ the three build-report widenings.
`package-lock.json` unchanged. `scripts/fence-lint.sh 283` → **PASSED**, no overlaps.

---

## Verify outcome

Verified by **fable** against the builder's **opus** commit, in an isolated worktree.
Verdict at `949b670`: *not ready on two small counts, neither a behaviour defect*. Both fixed
and **amended into the single commit** — one commit per issue is preserved — leaving the lane
at **`50b392f`**, which the verifier called ready to merge as part of this wave PR.

- **Finding 1 (must fix)** — the `lane.present` guard on clause (a) had no test (mutation M10
  survived, 134/134 green). Two tests added, both in-fence: byte-level in `diagnose.test.ts`
  and end-to-end in `buildFleet.test.ts` (`worktree.removed` after the beacon → `present:
  false`, `declared.kind` still `waiting`, no WAITING pathology). M10 re-applied against the
  fixed tree reddens exactly those two.
- **Finding 2 (must fix)** — the permissive `SourceFlow['source']` cast, replaced by the
  `Pick<>` form (deviation 2 above).

Everything else in the Definition of done is recorded **met** with the line that meets it,
and the verifier found no read-only violation, no `node:` import into `core`, no collector
touched, no schema bypass, no recording rot, and no owed ADR (the between-declarations call is
a PRD-ruling-level tie-break recorded in the code comment and the changelog; the state slice is
additive like `refusals`).

**Final sha on the lane: `50b392f`. Cherry-picked onto this wave branch as `01058e5`.**

---

### `docs/screenshots/**` — the STATE hover, captured after this lands on `main`

**Not dropped, not touched in this commit** (EXECUTED: no file under `docs/screenshots/`
changed). Declared in the plan's §0 before the build, re-stated in the build report, and
re-checked in the tree by the verifier. Three independent reasons, each of which alone would
be enough:

1. `scripts/dev/visit.mjs --capture` writes `capturedAt = git rev-parse HEAD`, and
   `no-personal-paths-law.test.ts` counts `capturedAt..HEAD` over the depicted paths. **A lane
   commit's sha is rewritten by this wave's cherry-pick and is not an ancestor of `main`**, so
   the manifest would be red — or, on CI's depth-1 clone, unresolvable — the moment the wave
   landed.
2. A new screenshot must be added to `SCREENSHOT_DEPICTED_PATHS` in that law file, which is
   **outside this issue's fence**.
3. `--capture` needs a writable `/repo` (a one-time `sudo mkdir`), which a lane cannot do.

The honest sequence is: land this wave, then capture at a `main` HEAD in a docs-only follow-up
that also adds the depicted-paths row. The fixture change above is what makes that follow-up a
**one command** capture — `visit.mjs --fixture 3` now shows exactly the hover this issue owes:

```
node scripts/dev/visit.mjs --fixture 3 --capture docs/screenshots/state-hover-declared.png --repo <path> --interact <hover script>
```

…run at a `main` HEAD, plus the `SCREENSHOT_DEPICTED_PATHS` row.

The existing `fixture-pathology.png` is not made visibly stale: lane 43's chip still reads
WAITING without `~`; only the hover text changed.

### Follow-up notes — the verifier's two "worth doing later" items

Both were reached with an **executed probe**, not a hunch, and neither is in this issue's
clause list:

- **FROZEN silently withdraws a declared WAITING, at the lapse boundary — w4's.** Probe P1: a
  lane whose last work was 9 min ago and whose harness declared `waiting` 40 s ago folds to
  pathologies `['frozen']`, no `waiting`, and the card's `why.evidence.fact` never contains
  `beacon`. `diagnose()` runs `detectWaiting` only when FROZEN is null, and a `beacon.received`
  feeds neither `lastWorkTs` nor `lastEventTs`, so the FROZEN evidence reads `no events for
  9m00s` while a beacon event arrived 40 s ago. In the realistic sequence — hook fires one
  second after the last work, human absent for eight minutes — the lane flips from a certain
  WAITING to FROZEN at the 8-minute mark and the declaration vanishes from the card. This is
  pre-existing precedence the plan did not open, but it is ruling 4 inverted at one boundary,
  and it is exactly the case the ruled lapse mechanism ("declared attention lapsed Nm ago;
  reading turn shape") exists to voice. **#218 (prd27 w4) inherits it with a probe.**
- **An older declared `waiting` overtaken by a newer workmux `working` is unvoiced when the
  pane inference then fires.** Probe P2: beacon `waiting` at −200 s, workmux `working` at
  −100 s, `pane.activity` at −5 s → inferred WAITING `quiet 1m40s, pane still alive`, card fact
  without `beacon`. The workmux arm and the calm-lane arms do voice the superseded
  declaration, but `staleDeclaredWorking` only names a `working` kind, so on the pane inference
  a superseded `waiting` (or `stopped`) is silent. The fix is a second clause beside
  `staleDeclaredWorking` (REASONED); it was not added here.
- **Cosmetic, noted not fixed:** `olderBeacon` in `diagnose.ts` is misnamed — it also fires for
  a beacon `stopped` *newer* than the roster's `waiting`, which is correct under clause (d),
  but the name suggests a filter that is not there.
- **Named in the plan, still w4's:** folding `beacon` into `CONNECTION_SOURCES` in
  `packages/core/src/selectors/connection.ts`, so the status-bar pill stops needing its own
  flow function.

### #290 is now **in** this bundle

An earlier revision of this body recorded #290 as *not in this bundle* — it was filed at review
of #287, after this wave was groomed, and was Backlog and unbuilt at the time. It has since
been built and verified, it is the same (milestone, wave) as #283, and its fence is disjoint
from this PR's 21 files. It is therefore appended here as a second commit rather than paying
the queue toll a second time. Its own section is below.

---

# #290 — the feed names which witness signed an `agent.status`

Built on **sonnet**, verified on **opus** (the commit trailer reads `Co-Authored-By: Claude
sonnet`; the verify pass was opus, so the different-model check is satisfied). Verdict **ready
to merge, no code change required** — one correction to the *record*, applied below.

---

## What became true

Since #281 two witnesses publish `agent.status` — workmux's **declaration** and the transcript
organ's **inference** — and the activity feed rendered both as one bare word. That is #133's
false summons in a third costume: a reader could not tell a lane that *said* it was waiting
from one the instrument *guessed* was waiting.

- **`LaneFeedEntry.witness`** (`panels/feed/feed.ts`) — `witness: AgentStatusWitness`, set to
  `event.source` in the `agent.status` arm and **nowhere else**. `detail` is the witness's
  evidence, never a second place to learn its name; the destructure one line above pulls
  `detail` and never consults it.
- **The row wears the instrument's own mark** (`panels/feed/index.tsx`) — a new `laneTag(status,
  witness)` beside `AGENT_STATUS_LABEL`: bare label for `workmux`, `` `${INFERRED_MARK} ${label}` ``
  for `sessionlog`. `INFERRED_MARK` is imported from the web barrel (the same path
  `panels/attention/AttentionStripView.tsx` uses) and never respelled — `grep` for a `~` literal
  across all four fence source files returns nothing. The `mark + space + text` shape is the one
  `evidenceLine` and the attention strip already use, so a reader who learned the mark in one
  place reads it everywhere.
- **Every word is marked, not just `waiting`.** An inferred `working` is `~ working`. The mark is
  about the witness, not the word.
- **A workmux row is byte-identical to before.** The only change to a declared row is the
  non-text `data-witness` attribute on `LaneRow`'s root `div`, which the tests read instead of
  parsing text.
- **Total over the witness type.** `laneTag` ends in `const _never: never = witness`, so a third
  literal in `AGENT_STATUS_SOURCES` fails **typecheck** here rather than rendering as workmux's
  word.

`packages/web/src/panels/feed/format.ts` was **not** touched, exactly as the issue and the plan
instructed.

---

## What is the sibling case?

**`resolveLane`, one line above the payload read.** It joins on `[handle, branch ?? null]`, and
the sessionlog organ emits `branch` from its own worktree reading rather than from workmux's
roster — so the two witnesses can disagree about the branch for the same lane. If that join
were branch-first-only, the second witness's row would land on a *different* lane, or on none,
and the feed would show one lane's word attributed to another.

The test emits a workmux `agent.status` for `42-lane` with `branch: '42-lane'` and a sessionlog
one for the same handle with `branch: 'feature/42-renamed'` — a branch the fleet never
discovered a worktree for — and asserts **two** lane entries, both with a non-null `laneId`, and
**one** distinct id. A duplicate *lane*, not a duplicate row, is the failure it guards. No change
to `resolveLane` was needed: `buildLaneIndex` maps every branch, worktree path and handle of a
lane to the same id.

---

Every row below was **EXECUTED by the verifier** against the source, then reverted:

| mutation | result |
|---|---|
| `witness: event.source` → `witness: 'workmux'` | **4 red** (both feed tests, both index tests) |
| witness derived from `detail`'s shape (`detail?.startsWith('WAITING')`) | **3 red**, including the test named for it — the fixture is deliberately inverted: the *workmux* event carries the organ's own `WAITING — tail turn-complete…` phrasing, the *sessionlog* event carries no `detail` at all |
| `INFERRED_MARK` in core `'~'` → `'≈'`, feed untouched | **green (9/9)** — the tests follow the constant, which is the point |
| …and additionally respell the feed's mark as a `~` literal | **2 red** — a hard-coded glyph would not have been caught; these are not that test |
| mark only `waiting` | **red** — `an inferred working is marked too` |
| restyle the declared row (`` `${label} (declared)` ``) | **red** — `a workmux row renders byte-identically to before the witness existed` |
| add `'beacon'` to `AGENT_STATUS_SOURCES` | **typecheck red twice**: `index.tsx` TS2322 `'"beacon"' is not assignable to type 'never'`, `index.test.tsx` TS2741 missing `beacon` in the `_covered` record |
| **branch-only join** — `resolveLane(laneIndex, [branch ?? null])` | **red** — the sibling test above |

**Correction to the record (verify finding F1, applied here).** The plan's §5 table and the
build report both claimed *"changing `resolveLane` to join on branch **before** handle reddens
the sibling test"*. It does not — the verifier applied exactly
`resolveLane(laneIndex, [branch ?? null, handle])` and `feed.test.ts` stayed **green, 11/11**.
`resolveLane` already tries `byBranch → byWorktree → byHandle` *within each candidate* and
returns the first candidate that resolves anywhere, so the fabricated branch resolves in none of
the three maps and falls through to the handle either way: the behaviour is order-insensitive by
construction and **no test can catch that mutation**. The accurate row is the branch-only join,
which is what the table above carries. The test is genuinely load-bearing; only the claim about
it was wrong, and it is not repeated here.

**Known survivor, named rather than hidden:** a test asserting `LaneFeedEntry` merely *has* a
`witness` field would pass every mutation above. None of the tests added is that test.

**Non-blocking, from verify (O1):** `the tag is total over the witness type` is a typecheck
guard wearing a test's clothes — its runtime assertion compares a literal to itself and cannot
fail at runtime. Its value is the type annotation, which does redden (row 7 above), and
`laneTag`'s own `_never` reddens independently in the source. Redundant rather than wrong; the
plan asked for it explicitly, so it was left alone.

---

## Plan deviations

**One, transient, and it resolved itself before the commit — recorded because the skill requires
it be surfaced even when self-resolved.** The `.citation-prior-tracker` ceiling law
(`packages/server/src/doc-citation-law.test.ts`) derives its live ceiling from the highest
`Merge pull request #N` subject reachable from `origin/main`. When the CHANGELOG entry was first
written that ceiling stood at **287**, below 290, so the plan's literal `, #290)` tripped the law
— isolated to that one `(file, num)` pair by stashing only the citation and re-running the single
test. The builder switched to `, issue 290)` and flagged it. By the time the full gate re-run
happened `origin/main` had advanced 18 commits and the ceiling stood at **295**, so the plan's
wording passes cleanly and **`, #290)` is what is committed**. The verifier re-derived the
ceiling independently rather than taking the builder's word for it: 295 at `52e0e37` (PR #294),
`doc-citation-law.test.ts` green in the full suite.

No other deviation. The DOM was isolated as the plan permitted — no new `data-testid` on
`KindTag`; the tag text is read as `row.querySelectorAll('span')[1]?.textContent` (the row's
first span after the clock) and the witness via `data-witness` on `LaneRow`'s own root div, one
level under the `<li>` `getAllByTestId('feed-entry')` returns.

---

## Fence

Exactly five files, and `CHANGELOG.md` is the one widening — recorded in the plan comment
**before** the build, per `AGENTS.md`:

```
CHANGELOG.md
packages/web/src/panels/feed/feed.ts
packages/web/src/panels/feed/feed.test.ts
packages/web/src/panels/feed/index.tsx
packages/web/src/panels/feed/index.test.tsx
```

`packages/web/src/panels/feed/format.ts` is in the issue's fence *conditionally* ("only if the
render needs a label helper") and is **untouched**. No widening beyond the plan's own note.

---

## Verify outcome

Verified by **opus** on `origin/290-feed-names-witness` @ **`d6d008b`**, diffed three-dot
against `origin/main` @ `52e0e37`. **Verdict: ready to merge, no code change.** Every gate was
re-run there rather than taken on trust — suite 414 files / 7572 passed, typecheck, lint, build,
boot smoke (`GET /` 200, `/api/meta` 401 without the token — the ADR-0008/0012 guard), a 4×
concurrent load check on the two touched files (20/20 every run), and
`scripts/fence-lint.sh 290 283` PASSED.

Each Definition-of-done criterion is recorded met with the line that meets it and a mutation
that reddens it. The one must-fix, **F1**, is a correction to the *record* and not to the code;
it is applied in the mutation table above. The cross-PR dependency was checked at the source:
`#281` really is on `main` (`AGENT_STATUS_SOURCES` and `AgentStatusWitness` in
`events/workmux.ts`, `INFERRED_MARK = '~'` in `fleet/pathology.ts`, both read off
`origin/main`), so nothing here depends on unlanded code.

Clean on everything else the skill asks: read-only constitution, CORS/origin, no `node:*` into
`core` (nothing added to `core` at all), no event-schema bypass (the fact is read off the
envelope's own `source`, where ADR-0009 puts provenance), no recording rot, no owed ADR
(ADR-0037 already decided this and is cited in both the doc comment and the changelog), no
screenshot affected, no scope creep, no law weakened — the two events added to `scenarioEvents()`
sit before the tmux `collector.error`, so the "newest first" test still pins the same last event.

**Final sha on the lane: `d6d008b`. Cherry-picked onto this wave branch as `6b57312`.**

---

## Follow-up — from verify, not blocking

**O2 — the feed cannot yet distinguish an inference the fold *accepted* from one it *refused as
dissent*.** (REASONED, from ADR-0037's Decision Outcome and `feed.ts`.) ADR-0037 puts prd-27
ruling 4's asymmetry in the reducer: a `sessionlog` word arriving while a `workmux` `waiting` or
`done` stands is refused and recorded as `AgentState.dissent`. The feed is a per-event surface,
so a refused inference now appears marked `~` beside the declaration it lost to — correct for an
event log, and ruling 4 says disagreement renders rather than resolving in silence. But "this
inference was overruled" is a further fact the row does not carry, and it is the same territory
#283 owns for the hover-card form. Noted here rather than filed.

---

# How the wave was assembled

```
git fetch origin
git checkout -b prd27-w3-fleet-believes origin/main      # 29cde14
git cherry-pick 50b392f                                  # → 01058e5   (#283)
git diff 50b392f HEAD                                    # empty
```

The first cherry-pick applied clean — no conflict, which is what a correctly fenced wave looks
like — and `git diff 50b392f HEAD` was **empty**: the assembled tree was byte-identical to the
verified lane tip.

**#290 was appended as a second commit** once it was built and verified, onto the same branch at
`01058e5` (the base was deliberately **not** rebased onto the moved `main`, since this is a merge
PR and PR #296 was already open and green):

```
git cherry-pick d6d008b                                  # → 6b57312   (#290)
```

**It applied clean too — `CHANGELOG.md` auto-merged, no conflict at all.** #283 wrote an `Added`
entry and #290 a `Fixed` entry, different subsections of `[Unreleased]`, so both stand in full
and neither was edited. Nothing was resolved by hand, so #290's commit is still exactly its lane
commit's change.

Both exclusion diffs were checked rather than assumed (EXECUTED):

- `git diff d6d008b HEAD -- . ':(exclude)CHANGELOG.md'` → **exactly #283's 20 non-CHANGELOG
  files** (942 insertions), i.e. the cherry-pick altered nothing of #290's.
- `git diff 01058e5 HEAD -- . ':(exclude)CHANGELOG.md'` → **exactly #290's four feed files**
  (251 insertions), i.e. it altered nothing of #283's.

Two commits, two issues; the wave was not squashed, and the push was a plain fast-forward
(the branch only gained a commit — no force).

`scripts/gate.sh` was **not** run: it *is* the landing (it merges into local `main`, builds
the merged result, and pushes to `origin`), and it is the operator's step.

### Gate results — run on the assembled **two-commit** branch at `6b57312`, node v22.23.2

| gate | command | result |
|---|---|---|
| suite | `VITEST_MAX_WORKERS=6 npm test` | **414 files passed (414)** · **7606 passed**, 3 expected fail, 6 skipped (7615), 120.15s |
| typecheck | `npm run typecheck` | exit **0** |
| lint | `npm run lint` | exit **0** — `Checked 897 files in 350ms. No fixes applied.` |
| build | `npm run build` | exit **0** |
| boot smoke | `npm start -- --port 0`, CI's own recipe | `rhizomorph running at http://127.0.0.1:63664`; `GET /` → **200 text/html**; `GET /api/meta` **without** the token → **401** (expected); capability token scraped off the served shell (64 chars, the `rhizomorph-capability` meta tag); `GET /api/meta` **with** it → **200**, seven ladder collectors `git,sessionlog,tmux,workmux,judge,pi,beacon`, `beacon.identity.level: partial` naming the w4 join as its remedy |
| fence | `scripts/fence-lint.sh 283 290` | **PASSED** — 19 + 10 fence paths, **no overlaps** |

The suite total is **7606** here against 7599 for the #283-only assembly: exactly #290's seven
new tests (3 in `feed.test.ts`, 4 in `index.test.tsx`), and no test lost. This is the first run
in which the two lanes' commits coexist, which is the whole reason the bundle is gated as well as
the lanes — nothing cross-lane broke.

The earlier single-commit assembly at `01058e5` gated green on the same commands
(414 files / 7599 passed, typecheck/lint/build 0, boot smoke 200/401/200 at `rung: L4`), and
CI on that revision of the branch was **10/10 green**.

---


gabriel-canaan (approved):
Reviewed commit by commit, with `origin/main` merged into a detached review worktree. Everything below is labelled **EXECUTED** (I ran it and read the exit status) or **REASONED** (argued from the code, nothing run).

- **EXECUTED** — post-merge suite on the full bundle: `414 files passed`, `7633 passed | 3 expected fail | 2 skipped`, exit 0. Typecheck 0, lint 0. Your body reports 7599 for the #283-only assembly and 7606 with #290; the remainder is main's own (`gate-honesty-law.test.ts` grew 536 lines in the 18 commits between your base and current `main`), not drift in this branch.
- **EXECUTED** — all 10 CI legs green against `b3f53bd`, confirmed by `check-runs`, not by the bar: `build-test-boot` ×3, `pack-smoke` ×6 including both Windows legs, `windows-suite`.
- **EXECUTED** — the era-1 re-bless is exactly what `CAPTURE.md` claims: one added line `"declared": {},` at the sorted-key position, no other key moved.
- **EXECUTED** — `at` really is the writer's clock. The collector emits with `{ ts: parsed.at }`, so clause (a)'s `since` is honestly "when the hook fired".
- **EXECUTED** — #290's two claims are load-bearing: dropping `INFERRED_MARK` for a sessionlog witness reddens, and so does dropping `data-witness`.
- **EXECUTED, and a negative worth recording** — `laneTag`'s exhaustive `default` throws inside React render, which is the shape that takes a panel down on bad log data. It is unreachable: `agent.status`'s source is `z.enum(['workmux','sessionlog'])` and `parseEventLenient` classifies a bad source as `kind: 'unknown'` rather than handing back a typed event, so it never reaches `buildFeedEntries`.

## Fixed in review — `b3f53bd`

I ran a mutation battery over `diagnose.ts`, `condition.ts` and `reduce.ts`. **Three survived**, all the same shape — a string the code renders that no assertion reads:

| mutation | before |
|---|---|
| drop `${staleDeclaredWorking}` from the **transcript-shape** arm only | **survived** |
| drop `declaredClause` from the **`unknown`** arm | **survived** |
| drop `declaredClause` from the **waiting-fallback** arm | **survived** |

`staleDeclaredWorking` is appended in two places and `(b3)` only ever exercised the pane-stillness one, so M13's "reddened 2" was true while half the clause was unpinned. `declaredClause` reaches four arms of `activityCondition`; two had tests.

Three tests added, no source change, all in-fence. **EXECUTED** — each is certified against its own mutant: with the test in place, re-applying the mutation gives `1 failed / 43 passed`, so it dies to exactly its own assertion and takes nothing else with it.

**Checked and NOT a finding, stated so it isn't re-found later:** dropping the outer `Math.max(0, ctx.now - declared.at)` in clause (a) also survives — because `formatSpan` already clamps internally (`plumbing.ts:430`). The outer clamp is redundant, not untested.

1. **"The newer word stands" holds only for `working`.** **EXECUTED** probe: workmux `waiting` at −90s with a beacon `stopped` at −30s still yields a *certain* WAITING (`inferred: false`), evidence `workmux reports waiting 1m30s · beacon (claude-hook) declared stopped 30s ago`. That satisfies the issue's clause (d) exactly, and your `olderBeacon` note half-names it — but the commit message and body both state the between-declarations rule without the `working` qualifier, so the prose is broader than the code. Either narrow the sentence or hand it to w4 with the probe.
2. **The equal-timestamp tie is unpinned.** Mutating `declared.at >= ctx.agentStatusTs` to `>` survives. **EXECUTED** probe at equal `at`: the beacon wins — `beacon (claude-hook) declares waiting 1m00s ago · workmux reports working`. The right choice; just nothing says so. Rare enough (two foreign clocks colliding to the ms) that I did not spend a test on it.

## What I did not run

`scripts/gate.sh` — it is the landing, and that is the operator's step, not a reviewer's.

Approving. The only commit here that has not had a second pair of eyes is my own `b3f53bd`, which is three added tests and no source change.

PR #297, opened 2026-09-06
prd-43 wave 10: a pinned sha must have landed, and every docs/prds exclusion proves itself

### LINKED ISSUE #288: prd43 w10: a pinned sha is evidence only if it landed
## The defect

`isPinnedArtefact` exempts a document from the citation law when its head declares
`**Tree:** <ref> at <sha>` and that sha "resolves". Resolution is tested with

    git cat-file -e <sha>^{commit}

which asks the OBJECT STORE, not reachability. An object exists there if it was ever
fetched, cherry-picked, rebased away, or created locally — including a commit on no
branch at all. So the exemption is granted on evidence that the commit exists
somewhere on this machine, not on evidence that it landed.

EXECUTED 2026-09-05, on a commit created locally and then orphaned:

    git branch -a --contains fbc8073    -> (no output: reachable from NO ref)
    git cat-file -e 'fbc8073^{commit}'  -> exit 0  (exemption would be granted)

A document can therefore exempt itself from the whole law by pinning a sha that
never landed anywhere, and the pin looks exactly like a legitimate one.

## Why this is the same defect the wave-7 review already fixed once

Wave 7 spent two review rounds on `liveMaximum()` deriving its ceiling from
`git log --all`, which walks every ref the object store holds. Both seats blocked on
it for one reason: **local git state is not evidence that a number landed.** The fix
pinned it to a landing ref.

This is that defect's sibling, one screen away in the same file, and it survived
those rounds because the reviews were scoped to the repairs in front of them. It was
found by asking the sibling-case question of the fix, not by a gate.

## What makes it tractable today

The corpus currently carries exactly ONE pinned artefact, and it is honest:

    docs/design/glance-2026-09-02.md   **Tree:** `main` at `0851512`
    git merge-base --is-ancestor 0851512 origin/main   -> reachable: YES

So tightening the check breaks nothing that exists. This is the cheap moment to do
it — before a second pin arrives and the change acquires a migration.

## The decision this issue owns

Which question replaces `cat-file -e`. Name the rejected alternative:

- **`git merge-base --is-ancestor <sha> <landing ref>`** — asks the question the
  exemption means ("did this land"), and is the same shape wave 7 settled on. Costs
  a dependency on the landing ref being present, which the wave-7 review showed must
  then be handled explicitly rather than absorbed.
- **`git branch -a --contains <sha>`** — reachable from ANY ref, which readmits every
  stale remote branch the wave-7 fix just excluded. Weaker than it looks.

The absent-ref case is the trap wave 7 hit twice: a missing landing ref must not
silently downgrade the check, and must not blame the document either.

## Fence (may touch ONLY)

## Blocked by

Nothing. The file is free — wave 7 merged in PR 284.

## Definition of done
- A pinned sha that is not reachable from the landing ref does NOT exempt its
  document, and the law says which of the two reasons it refused.
- A pin whose sha is unreachable is distinguishable, in the failure message, from a
  pin whose sha does not exist at all.
- A control proves the new check bites: a document pinned to an orphaned commit is
  rejected, and the existing honest pin still passes.
- The absent-landing-ref path is decided on purpose and stated, not absorbed.

### LINKED ISSUE #289: prd43 w10: every docs/prds exclusion proves it is not vacuous
## The defect

Three sweeps exclude `docs/prds/`. Only the citation law proves its exclusion is not
vacuous — and it does so twice.

    doc-citation-law.test.ts    EXCLUDED_DIRS             controls at :853 and :1888
    route-class-law.test.ts     SWEEP_EXCLUDED_PREFIXES   none
    manifest-law.test.ts        EXCLUDED_CLONE_SITE_DIRS  none

Ruling 1 states the principle: *"the exclusion list is itself asserted non-empty and
honest — every excluded directory must contain at least one path that would otherwise
trip the law, or the exclusion is stale and hiding a live file."* The two later
sweeps copied the exclusion and not the proof.

## One of the two has already gone vacuous

This is what turns the issue from a guard into a repair, and it was found by a review
seat, not by the author.

EXECUTED — the route-count law's own **twelve** production patterns, extracted from
its `CLAIMS` table rather than retyped, applied to all **53** tracked
`docs/prds/**/*.md` files: **zero matches.**

So that exclusion hides nothing today. An honesty control written for it would fail
on its first run with "the exclusion is stale", which is the control working
correctly and is not a state to ship.

The manifest law's exclusion IS still doing work, and that half is a plain guard:
`docs/prds/parked/prd-38-the-borrowed-credential.md` carries a real `git clone …
github.com` instruction that its discovery filter matches. Unproven, not stale.

## What the earlier version of this issue got wrong, kept because it is the lesson

The first draft asserted both exclusions were honest today. That came from a
hand-written grep for route-count-shaped prose, which matched a sentence in prd-43
itself that the law's real patterns do not match. The command succeeded and
enumerated something other than what was asked — the exact shape this PRD exists to
catch, committed by the person writing the issue about it.

Derive from the production patterns, not from prose that resembles them.

## The decision this issue owns

**For the route-count sweep, whether the dead exclusion is retired or revived.**
Retiring it means the sweep scans `docs/prds/` — which is a change in what a law
covers and needs the PRD's own reasoning about dated artefacts applied, since ruling
1 excludes that tree deliberately. Reviving it means nothing, because an exclusion
exists to hide something and this one hides nothing. Name the rejected option.

**For the manifest law, whether the control is copied or extracted.** Copying is a
few lines and keeps each law readable alone; extracting couples two packages that
share nothing today, and the two sweeps do not detect the same thing.

The two exclusion lists are NOT the same set — one also excludes `docs/adr/`, the
other also `docs/review/` — so any control must assert each directory in its own
list rather than a hardcoded trio.

## Fence (may touch ONLY)
- `packages/app/src/host/manifest-law.test.ts`

## Blocked by

Nothing open. The route-count law file is also claimed by prd43 w8's outbound-call
issue, so the two cannot be in flight together whatever wave they carry — re-run the
fence lint rather than trusting this line. prd-17's operator-door issue also claimed
that file and is now closed, so it is no longer a collision.

## Definition of done
- The route-count sweep's `docs/prds/` exclusion is either removed with the reasoning
  recorded, or made non-vacuous — and whichever is chosen, a control asserts the
  resulting state so it cannot rot silently again.
- The manifest law asserts, per directory in its OWN exclusion list, that the
  directory would trip its detector if scanned.
- Each control is proven to fail for its stated reason by a mutation, not asserted.
- The count of proving sweeps stated anywhere in prose is derived from the code.

gabriel-canaan:
prd-43 wave 10 — the enforcement residue found by reviewing wave 7 rather than by any
gate. Two issues, fence-disjoint, built in parallel.

Closes #288
Closes #289

## Commits

|---|---|
| `227d4af` | a pinned sha exempts a document only if it landed (#288) |
| `562f9cd` | every docs/prds exclusion proves itself, not assumed (#289) |

**`#288`.** `isPinnedArtefact` exempted a document from the citation law when its
`**Tree:** <ref> at <sha>` pin "resolved", tested with `git cat-file -e` — which asks
the OBJECT STORE, not reachability. An orphaned local commit resolves, so a document
could exempt itself from the whole law with a sha that never landed anywhere. It is the
sibling of the `--all` ceiling defect wave 7 spent two rounds on: local git state is not
evidence that something landed.

**`#289`.** Three sweeps exclude `docs/prds/`; only the citation law asserted the
exclusion would trip if scanned. And one of the two unproven exclusions had ALREADY gone
vacuous — the route-count law's twelve production patterns, across all 53 tracked
`docs/prds/**/*.md` files, returned zero matches — so the issue owned a decision rather
than a guard.

## What is the sibling case?

**For `#288`:** every other place this repo treats local git state as evidence of
landing. That class is what wave 7's `--all` ceiling defect and this pin check both
belong to; the review enumerated git's identity and reachability paths as tables and
found no third instance in these files.

**For `#289`:** the third sweep. Checked with that law's own detector rather than a grep
of ours: `manifest-law`'s exclusion is still doing real work —
`docs/prds/parked/prd-38-the-borrowed-credential.md` carries a `git clone … github.com`
instruction its discovery filter matches — so it was unproven, not stale. One exclusion
was dead and one merely unguarded, and the change says which is which.

The one that took three rounds to get right, and the arc is the point:

- **Round 1** blocked: `#288`'s new orphan-commit fixture called `git commit-tree` with
  no committer identity. Green on any machine with a configured git user; **exit 128** on
  CI, which configures none. Never witnessed, because the branch had not been pushed.
- **Round 2** blocked: the CONTROL added for round 1 could pass while the thing it
  guarded was broken. It scrubbed `HOME` and `GIT_CONFIG_GLOBAL`/`SYSTEM` but not the
  four `GIT_AUTHOR_*`/`GIT_COMMITTER_*` vars, which git reads first. EXECUTED: with
  ambient identity present a reverted default still exits 0.
- **Round 3** cleared it, and explicitly found **no round 4** — because the repair is a
  mechanism change, not a third enumeration. All seven variables are saved, blanked and
  restored, and blank component values defeat every identity path: env vars,
  `EMAIL`/GECOS auto-detection, local and worktree config, `includeIf`,
  `GIT_CONFIG_COUNT`, `git -c`, `--config-env`, `GIT_DIR` redirection.

Certified: reverting the default to bare `process.env` **with ambient identity
exported** reddens exactly the identity control and nothing else. Restored, tree clean.

## Fence

No widening. Three files, each inside its issue's declared fence:
`doc-citation-law.test.ts` (#288) · `route-class-law.test.ts`, `manifest-law.test.ts`
(#289). Verified three-dot against `origin/main`.

## Verification

Three rounds, two seats each, isolated worktrees, both seats executing. Opus and Codex
at `xhigh`; the lanes built on Sonnet, so no seat shared the builder's model.

`verify-ledger.sh check --strict` → **exit 0**, both commits carrying real review
records, none forced. The records survived two rebases — patch-id keying, not sha.

One caveat recorded rather than glossed: `rereview-budget.sh` could not record round 3,
because its report directory came from a `--seat-prefix` run and the budget script does
not glob prefixed directories. The ledger does, so the strict check above is honest;
the round count is undercounted. That gap is in operator tooling, not in this branch.

## Gates

EXECUTED on `562f9cd`, statuses read directly:

```
npm run typecheck              -> 0
npm run lint                   -> 0
VITEST_MAX_WORKERS=6 npm test  -> 0   (414 files, 7640 passed, 3 expected fail, 2 skipped)
```

Plus the two environments this wave is about: the three changed law files pass 175/175
with the git identity stripped entirely, and `doc-citation-law` passes 49/49 with an
ambient identity exported.


ciaran-slow (approved):
## Approving. Every load-bearing claim in this PR reproduced independently.

Reviewed commit-by-commit in a clean clone at `562f9cd`, node v22.23.2. Nothing here needed a fix.

**EXECUTED — the three changed law files: 175/175.**
`doc-citation-law` + `route-class-law` 144/144, `manifest-law` 31/31, exit 0. Matches the PR body exactly.

**EXECUTED — mutation 1, `#288`'s core claim.** Reverted `pinVerdict`'s `merge-base --is-ancestor` back to the pre-#288 `cat-file -e`:

```
× a pin to an orphaned commit — real object, reachable from NO ref — is rejected, not exempted (#288)
  Tests  1 failed | 48 passed (49)
```

Exactly one test red, and it is the right one. The fix is genuinely guarded.

**EXECUTED — mutation 2, the round-2/3 claim, which is the one I most expected to fail.** Reverted `createOrphanCommit`'s default to bare `{ ...process.env }` *with `GIT_AUTHOR_*`/`GIT_COMMITTER_*` exported ambiently* — the exact condition that defeated round 1's control:

```
× the orphan-commit fixture does not depend on this machine's own git identity, from ANY source
  Tests  1 failed | 48 passed (49)
```

Sanity pass, same ambient exports, unmutated: **49/49 green**. So the red is the mutation and not the export. The blank-don't-delete reasoning holds up under the condition it was written for.

**EXECUTED — `#289` is a real coverage gain, not bookkeeping.** I planted a false claim (`This server has seventeen mutating routes today.`) into `docs/prds/prd-43-the-claim-is-a-test.md` and ran `route-class-law` on two trees:

| tree | result |
|---|---|
| pre-297 `main` (exclusion in place) | **94/94 green — the exclusion hides it** |
| with this PR | **red**: `no swept file states a recognised route-count claim that no CLAIMS row declares` |

That is the before/after the issue argued for, executed rather than reasoned.

**EXECUTED — re-derived the factual claims rather than taking them:**

| claim | derived | ✓ |
|---|---|---|
| 53 tracked `docs/prds` md files | `git ls-files` → 53 | ✓ |
| 40 tracked `docs/review` md files | `git ls-files` → 40 | ✓ |
| twelve `CLAIMS` patterns | 12 `pattern:` rows | ✓ |

**Fence: exact.** Three files, all inside their issues' declared fences, no widening. `#288` and `#289` are both prd43 wave 10 — one PR, one (milestone, wave).

### Gates I ran, against the tree that will actually exist

Because #297 and #298 are in flight together and neither CI run has seen the other, I merged them and ran the combined tree — this is the interaction that matters, since **this PR retires the `docs/prds/` exclusion while #298 adds prose to a `docs/prds/` file**:

```
npm run typecheck                       -> 0
npm run lint                            -> 0   (897 files)
VITEST_MAX_WORKERS=6 npm test           -> 7635 passed, 3 expected fail, 6 skipped
doc-citation + route-class + runbook-delivery (merged tree) -> 155/155
```

The interaction is clean: #298's new PRD prose is now genuinely swept and passes.

### One thing I did NOT run, and one failure I am not attributing to you

I could not run `lane-precommit.sh`, `verify-ledger.sh` or `rereview-budget.sh` — correctly not shipped by this repo, so I verified the ledger claims by re-deriving from git instead.

The merged-tree suite had **one** failure: `session-recorder.test.ts > holds the ceiling in a session reached by rotation`. **Not yours** — that file carries a `// @gate-timing` marker, so AGENTS.md has it run alone and serially precisely because a wall-clock assertion under load means nothing, and my run was at 6 workers. **EXECUTED: green 3/3 in isolation on the merged tree**, and neither PR touches `session-recorder`. Flagging it so you know I saw it and why I am not asking you to do anything about it.

*(An earlier full-suite run of mine showed four failures. That run was contaminated — I mutated the working tree while it was in flight. Discarded and re-run clean; the numbers above are from the clean run.)*

### Your call, offered not asserted — one nit, no change requested

**REASONED, not executed.** In `a missing landing ref refuses to certify the pin…`:

```ts
expect(() => pinResolves(landedAncestorSha())).not.toThrow() // the real LANDING_REF is present in this clone
```

By the time this line runs, `pinVerdictCache` already holds `origin/main <sha>` — `a pin that does not resolve exempts nothing…` populated it earlier with the same call. The cache lookup returns before `refResolves(landingRef)` is reached, so this assertion is cache-served and does not actually exercise the default-ref path it reads as covering. Its failure mode is benign and a dozen other tests would redden if `LANDING_REF` broke, so this is a nit and not a finding — but given the file's own standard about assertions that cannot fail for the reason they claim, you may want it to clear the cache first.

Approving on the evidence above.

Issue #302, opened 2026-09-06
prd17 w7: one rule governs what counts as degraded and what is preserved


gabriel-canaan:
> **Absorbed #300 on 2026-09-07.** They are one change, not two — see
> *The degraded predicate* below. #300 is closed pointing here.

## Fence (may touch ONLY)
- `packages/server/src/server/poll-loop.ts`
- `packages/server/src/server/poll-loop.test.ts`
- `packages/server/src/server/summons.test.ts`

## Blocked by
**#278.** Related to #300 — both are about the preserve/raise asymmetry — and
they may be one change. Whoever picks these up should read both and say whether
they are.

## Why
A review seat's answer to #278's own review question — *do the preserve side and
the raise side follow from the same principle?* — was **no**, and it named the
concrete case: a lane marked `parked` in the manifest.

Its finding, EXECUTED and reproduced identically against `61e7178`, so
**pre-existing, not introduced by #278**:

> `CONTROL (nothing parked): frozen raises at t1 and never chatters.`
> Pre-existing — identical at `61e7178`. The repair does not introduce it, but
> it writes the false premise into the file as a load-bearing comment.

Two things to fix, and the second is the one that keeps it invisible:

1. The asymmetry itself — the conditions under which a point is PRESERVED and
   those under which it is RAISED do not follow from one rule.
2. **No fixture anywhere sets `parked`** — including #278's own
   `wireManifestFor`, which drops the field. So the whole `parked` axis is
   unexercised by every test that drives the raiser, and a comment in the file
   now asserts behaviour for it.

### The degraded predicate — absorbed from #300

#278 stopped a TIMED-OUT manifest read from clearing open `off-fence`
summonses. The reasoning is right and worth keeping: a manifest you could not
read is not evidence the fence stopped applying. **The guard is keyed on the
wrong condition** — it covers the read that did not SETTLE, and not the read
that settled and reported itself unavailable.

`readLanesManifest()` returns `{ available: false, reason: … }` for non-ENOENT
read errors (`api/lanes.ts`). So an `EACCES` on the manifest is "settled", skips
the preserve path, and clears every open `off-fence` point.

EXECUTED by a review seat: with a valid manifest opening
`45-ledger-subrows:off-fence`, returning
`{ available: false, reason: 'EACCES: permission denied' }` emitted a
`summons.cleared`; restoring the same manifest emitted a second raise. CONTROL:
the available manifest raised the real point, and restoring it re-raised exactly
that point.

[...]

Issue #301, opened 2026-09-06
prd17 w7: a recovery is recorded only for an error the log contains


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/server/poll-loop.ts`
- `packages/server/src/server/poll-loop.test.ts`

## Blocked by
**#278**, which adds the recovery path this corrects.

## Why
#278 closed a real gap: a timed-out manifest read emitted a `collector.error`
named `summons` that nothing could ever clear, because `collector.recovered` is
emitted only by `withResilience`/`withResumeReconciliation`, which wrap
`loadCollectors()` output — and the raiser is not a collector and is not
wrapped. The fix counts timeouts and emits `collector.recovered` on the next
settling read.

The counter increments on the TIMEOUT, not on the alarm having been RECORDED.
`recordAlarm` is the one publish-without-append path in the recorder and it
reports `{ appended: false }` on a failed write. So a timeout whose alarm could
not be appended still arms the recovery, and a later healthy read then persists
a `collector.recovered` with **no persisted preceding `collector.error`**.

EXECUTED by a review seat: the timeout attempted exactly one alarm, the recorder
reported `appended: false`, and the persisted stream contained **zero**
`collector.error` events followed by **one** `collector.recovered`, folding
`summons` to healthy. Live subscribers may have seen the transient alarm;
replayed history cannot justify its recovery.

This is prd-17's own subject matter. A record that shows a recovery for
something that never errored is a record that cannot be replayed into the truth
— the failure ruling 3's integrity laws exist to abolish.

Unlike #299 and #300 this is NOT pre-existing: the recovery path is new in
#278.

## Definition of done
- The recovery is armed by the alarm having been APPENDED, not by the timeout
  having occurred. `recordAlarm` already returns `{ appended }` — read it.
- A failed alarm append leaves the counter untouched, so no recovery is owed.
- A test asserts the persisted stream, not the emitted one: fold the log and
  assert no `collector.recovered` without a preceding `collector.error` for the
  same name.
- `npm run typecheck`, `npm run lint`, `VITEST_MAX_WORKERS=6 npm test` pass.

## Sibling case
The reverse: an alarm that IS appended while the recovery's own append fails
leaves `summons` in error forever — the exact state #278 set out to make
recoverable. Both halves of the pair have a failing-append case and the fix must
name what happens in each.

## What mutation would this test survive?
Assert on events the loop EMITTED rather than on the persisted log. Both are
[...]

Issue #300, opened 2026-09-06
prd17 w7: an unreadable manifest is treated as an absent one (consolidated into #302)


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/server/poll-loop.ts`
- `packages/server/src/server/poll-loop.test.ts`

## Blocked by
**#278**, which introduces the preserve behaviour this corrects.

## Why
#278 stopped a TIMED-OUT manifest read from clearing open `off-fence` summonses.
The reasoning is right and worth keeping: a manifest you could not read is not
evidence the fence stopped applying.

The guard is keyed on the wrong condition. It covers the read that did not
SETTLE; it does not cover the read that settled and reported itself
**unavailable**.

`readLanesManifest()` returns `{ available: false, reason: … }` for non-ENOENT
read errors — `api/lanes.ts`. So an `EACCES` on the manifest is "settled", skips
the preserve path, and clears every open `off-fence` point.

EXECUTED by a review seat: with a valid manifest opening
`45-ledger-subrows:off-fence`, returning `{ available: false, reason: 'EACCES:
permission denied' }` emitted a `summons.cleared`; restoring the same manifest
emitted a second raise. CONTROL: the available manifest raised the real point,
and restoring it re-raised exactly that point.

That cleared→raised cycle is chattering, on the family whose stated purpose
(`events/summons.ts`) is making flood and chattering computable.

**"Settled" does not mean "the manifest is definitely absent."**

## Definition of done
- The preserve decision is keyed on whether the manifest is KNOWN ABSENT versus
  UNKNOWN, not on whether the read settled. `ENOENT` is knowledge; `EACCES`,
  a parse failure and a timeout are all absence of knowledge.
- `readLanesManifest`'s own `reason` is what distinguishes them — read it rather
  than inferring from the shape.
- A test drives each of: absent (ENOENT), unreadable (EACCES), unparseable, and
  timed out — and asserts which of the four may clear an open `off-fence` point.
- `npm run typecheck`, `npm run lint`, `VITEST_MAX_WORKERS=6 npm test` pass.

## Sibling case
The RAISE side. #278's review asked whether preserve and raise follow from the
same principle and the answer was no. If an unknown manifest may not CLEAR an
off-fence point, it must not RAISE one either — raising on ignorance is the same
error facing the other way. Settle both in one edit and say so.

## What mutation would this test survive?
Key the guard on `available === false` alone. The timeout case passes, ENOENT
[...]

Issue #299, opened 2026-09-06
prd17 w7: a rotation between two appends cannot leak stale summons clears


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/server/poll-loop.ts`
- `packages/server/src/server/poll-loop.test.ts`

## Blocked by
**#278.** This narrows a window #278 already narrowed; it must land first or the
two edits collide in the same function.

## Why
#278 moved `recorder.sessionId` and `recorder.foldSoFar()` to AFTER the awaited
manifest read, closing a window where a rotation landing mid-await made the tick
diff old summons points against a new empty fold and emit phantom
`summons.cleared` into the new session.

That closed the read window. **It did not close the write window.** The diff is
computed, then its events are appended one at a time, and a session switch
between two appends still lets the remaining old-session clears land in the new
session.

EXECUTED by a review seat against the real `SessionRecorder`: three open
summonses, a calm next fold, `openSession()` fired during the FIRST clear's
append. The first append was correctly discarded from the new buffer; **the
remaining two stale clears appeared in the new session.**

A second seat checked the same probe against `61e7178` and reproduced it, so
this is **pre-existing and strictly narrower after #278** — not a regression.
That is why it is a follow-up rather than a fourth review round on #278.

Production rotation is the reachable path: `api/rotate.ts` opens the recorder
before it can call `pollLoop.reset()`.

## Definition of done
- A session switch at ANY point in `raiseSummons()` — including between two
  appends of one diff — cannot put an old session's `summons.cleared` into the
  new session's log.
- The fix is structural, not another narrowing. A seat named the two shapes that
  actually close it: **re-check the session identity before each record**, or
  **make the batch atomic** with respect to the session boundary. Pick one, and
  say on this issue why the other was rejected.
- The test drives the switch mid-batch, not before the tick — the existing test
  covers rotation between ticks and passes at either behaviour.
- `npm run typecheck`, `npm run lint`, `VITEST_MAX_WORKERS=6 npm test` pass.

## Sibling case
The raise side. Everything above is about clears, because a clear is what a
stale `previous` produces — but a raise computed against the old fold and
appended after the switch is the same defect wearing the other hat. Whatever
guard is chosen must cover both, and the test must show it.

## What mutation would this test survive?
Re-check the session only ONCE, before the first append. A test that switches
before the batch passes; only one that switches BETWEEN appends catches it.

**Found by:** both review seats, #278 round 3, EXECUTED with controls against
[...]

PR #303, opened 2026-09-06
prd17 w3: a summons the instrument raised says so in the record

### LINKED ISSUE #278: prd17 w3: a summons the instrument raised says so in the record
## Fence (may touch ONLY)
- `packages/server/src/server/summons.ts`
- `packages/server/src/server/summons.test.ts`
- `docs/adr/`

The two summons files are new. `docs/adr/` is claimed for the ADR ruling 5 owes.

- `packages/server/src/server/collector-loader.test.ts` (packages/server/src/server/collector-loader.test.ts only, and only the inline fake SessionRecorder's shape — no assertion, no collector, no production file)

## Blocked by
Nothing. Wave 0 was discharged by prd-17 ruling 5 on 2026-09-05, and #217's
claim on the ADR directory cleared when it merged on 2026-09-04.

## Why
prd-17 ruling 5, ruled in session 2026-09-05: **the instrument raises its own
summons, server-side, on the poll loop's tick.** The record must be complete
whether or not anyone is watching — a summons that exists only while a browser
is open is an observer-dependent record, which is prd1's founding insight
recurring one level down.

Read ruling 5 before starting; it names the seam and the reason, and it records
what was rejected so those do not get re-proposed.

Nothing is duplicated. buildFleet is core's and the server already calls it
(packages/server/src/api/lab.ts); the lane manifest is already read server-side
(readLanesManifest in packages/server/src/api/lanes.ts). What is new is a clock
and an edge, and poll-loop.ts already has both halves: it ticks on an interval
with an injectable `now`, holds the recorder, and persists state through a
SnapshotStore.

One shape trap: core's LaneManifest is a record keyed by handle, while
api/lanes.ts's LanesManifest is a versioned array. parseLaneManifest in
packages/core/src/fleet/fences.ts is the reader this wants — do not hand it the
route's shape.

## Definition of done
- An ADR in docs/adr/ records what a summons raiser is, **why it is not a
  collector**, and where the edge state lives. The collector contract hands a
  collector repoPath, now, exec, nextId and emit — never the folded state — so a
  state-reading judge cannot be a collector without widening ADR-0004. Read
  docs/adr/README.md first; that log is append-only.
- The edge-trigger is a **pure function** in its own module: last tick's rungs
  plus this tick's fleet in, summons.raised and summons.cleared out. Put the
  diff inside the tick and there is nothing to test without a running loop.
- A raise fires once, not once per tick, and a clear fires once. The pair is
  keyed on (lane, kind), which is what ruling 1's schema note says a later fold
  would join them on.
- The edge state survives a restart without re-firing every open summons.
  poll-loop.ts's SnapshotStore is where per-collector state already survives one.
- The first entries of the summons `kind` vocabulary are chosen and written
  down. Ruling 1 left `kind` an open string on purpose; this issue picks the
  first members and says why, it does not close the enum.
- The gate's own behaviour and every existing poll-loop guarantee are unchanged:
  the per-exec ceiling, the per-collector watchdog budget, and the session
  boundary reset all still hold.

## Sibling case
Raise and clear are two edges, and the clear is the one that gets missed. A
condition that ends while the process was down, or ends between two ticks that
both saw it, must still clear exactly once. The reducer's own arm records that a
clear with no preceding raise is a real state of the world and must fold — so
this must never emit a clear that assumes its raise was seen.

Second: the ladder needs a clock, and the poll loop's `now` is injectable
precisely so tests are deterministic. Reaching for the wall clock inside the
pure module is the miss, and it is the one that makes the tests flaky rather
than red.

Make the raise fire on every tick instead of on the edge. A test that ticks once
and asserts one event passes. The assertion has to tick repeatedly with the
condition unchanged and assert the count stays at one.

Second: drop the snapshot persistence and keep the in-memory state. Every test
in one process passes; only a test that rebuilds the loop from a stored snapshot
catches it.

**Model:** sonnet. **Wave:** 3. **Milestone:** prd17.

gabriel-canaan:
Closes #278

prd-17 ruling 5: **the instrument raises its own summons, server-side, on the
poll loop's tick.** The record must be complete whether or not anyone is
watching — a summons that exists only while a browser is open is an
observer-dependent record, which is prd1's founding insight recurring one level
down.

`summons.ts` is a pure edge-trigger: last tick's open `(lane, kind)` points plus
this tick's judged conditions in, `summons.raised`/`summons.cleared` out.
`poll-loop.ts` feeds it from `buildFleet(recorder.foldSoFar(), …)` — the same
fold every other route already reads. **Nothing is duplicated:** `buildFleet` and
`readLanesManifest`/`parseLaneManifest` are called where they already were, and
`isSummonsKind` derives which pathologies raise from core's own
`PATHOLOGY_RANK` rather than a second hand-maintained list.

`docs/adr/0038` records what a summons raiser is, **why it is not a collector**
(the contract hands a collector `repoPath`/`now`/`exec`/`nextId`/`emit`, never
the folded state, so a state-reading judge cannot be one without widening
ADR-0004), and where the edge state lives.

## What is the sibling case?

**Raise and clear are two edges, and the clear is the one that gets missed.**
`diffSummons`'s `cleared` array is built from `previous` alone — every
`(lane, kind)` point absent from this tick's judgement clears, with no check that
*this process* emitted the matching raise. That is deliberate: the reducer's own
arm records that a clear with no preceding raise is a real state of the world.
A test hands `diffSummons` a `previous` point it never produced — standing in for
a snapshot loaded after a restart, or a condition whose whole lifetime fell
between two ticks — and confirms it still clears exactly once.

**The session boundary.** `rotateSession()` empties the fold via
`openSession()` *before* the route calls `pollLoop.reset()`, so a tick landing in
that gap would diff old points against a new empty fold. `summonsSessionId`
records which session's fold the state was diffed against, and the reads happen
**after** the awaited manifest read, not before it.

Certified per fix, none reusing an earlier proof:

| mutation | result |
|---|---|
| raise fires every tick instead of on the edge | 4 red, at both the pure and poll-loop levels |
| drop the snapshot persistence | restart test red |
| drop the hydrate | 2 red |
| `cleared` always `[]` | 6 red, incl. the sibling-case test |
| `isSummonsKind` always true | 2 red |
| `SUMMONS_SNAPSHOT_KEY` set to `'git'` (a real loaded collector) | red — the law asserts the **real registry**, not a restated list |
| move the session guard back before the await | 3 phantom clears |
| disable off-fence preservation | clears an open point |
| retain `manifestInFlight` across `reset()` | blocks recovery |
| disable the recovery emission | leaves the collector in error |

## Verification — three rounds, six seat passes

| round | verdicts | findings |
|---|---|---|
| initial | Opus merge / Codex block | 4, **all in lane code** |
| fix re-review 1 | both block | 5, all in round 1's repair |
| fix re-review 3 | **Opus safe to merge** / Codex block | 4 |

Round 3 ran on an **explicit operator override** of a CHURN refusal — two rounds
of findings in one file. The override is on the record with its reason: the
refusal exists to stop a session chasing successive *spellings*, and round 2
instead changed the **shape** of control flow around an `await` in the
instrument's heartbeat, where both prior rounds had found real defects.

Round 3's split is resolved by evidence, not by preference: **Opus checked out
`61e7178`'s `poll-loop.ts` and reproduced Codex's remaining findings against
it** — so they are pre-existing or narrowed, not introduced. Its words on the
session window: *"strictly narrower than at `61e7178`, where the whole manifest
read was also inside the window."* Merging strictly improves the tree.

**Four follow-ups filed with their evidence rather than chased:** #299 (a
rotation between two appends still leaks stale clears), #300 (an unreadable
manifest treated as an absent one), #301 (a recovery recorded for an error the
log never contains — the one item genuinely introduced here), #302 (the `parked`
axis unexercised). All four are fenced on `poll-loop.ts` and blocked on this PR.

**Three commits carry FORCED ledger rows, and the reason is an orchestrator
error, not a missing review.** Seats ran on every one; the runs used `--out` to
put reports in the session scratchpad instead of the default `.git` location, so
the ledger cannot see their round manifests. The reports exist and the verdicts
are quoted above.

A fence widening on `collector-loader.test.ts` is recorded on the issue in two
parts — first for the fake `SessionRecorder`'s shape, then amended to permit one
assertion enforcing the reserved snapshot key against the real registry.

Gates: `npm run typecheck` clean, `npm run lint` 899 files clean,
`VITEST_MAX_WORKERS=6 npm test` → 415 files, 7665 passed, 3 expected fail,
2 skipped.

---

## Declared review gap

> All three commits carry FORCED ledger rows. Seats DID run on each — six passes across three rounds, verdicts quoted in the body — but the runs used --out to write reports into the session scratchpad rather than the default .git location, so the ledger cannot see their round manifests. That is an error in how the runs were invoked, not a review that did not happen.

```
  forced  7f1beea  fix(server): raiseSummons reads sessionId/foldSoFar AFTER its await, not before (review of #278, round 2)
          ^ Seats DID run on this sha — two of them — but the ledger cannot see the round manifest because the runs used --out to put reports in the session scratchpad instead of the default .git location. That is an orchestrator error in how the runs were invoked, not a review that did not happen; the reports are at scratchpad/w3-verify/. Recorded forced so the gap is visible rather than implied. Fix re-review round 3, operator-authorised past a CHURN refusal: Opus SAFE TO MERGE, Codex not-safe. Opus checked out 61e7178's poll-loop.ts and reproduced Codex's gaps against it — pre-existing or narrowed, not introduced. All four certified mutations reproduce, killing exactly one test each; no pre-existing guarantee moved. Remaining items filed as #299/#300/#301/#302.
  forced  b0bd28a  fix(server): the summons raiser cannot stall the loop or misfire across a rotation (review of #278, round 1)
          ^ Seats DID run on this sha — two of them — but the ledger cannot see the round manifest because the runs used --out to put reports in the session scratchpad instead of the default .git location. That is an orchestrator error in how the runs were invoked, not a review that did not happen; the reports are at scratchpad/w3-verify/. Recorded forced so the gap is visible rather than implied. Fix re-review round 1: both seats NOT SAFE, 5 findings all in the lines the fix added, chiefly the session guard being check-then-use across the new await (3 phantom summons.cleared on rotation). Superseded by 7f1beea.
  forced  8de3c1e  feat(server): the instrument raises its own summons, on the poll loop's tick (#278)
          ^ Seats DID run on this sha — two of them — but the ledger cannot see the round manifest because the runs used --out to put reports in the session scratchpad instead of the default .git location. That is an orchestrator error in how the runs were invoked, not a review that did not happen; the reports are at scratchpad/w3-verify/. Recorded forced so the gap is visible rather than implied. Initial verify: SPLIT, Opus ready-to-merge / Codex not-ready, 4 findings ALL in lane code — an unbounded manifest read that permanently stalled the tick stream, an unnamespaced snapshot key, and two limbs no test touched.
──
3 commit(s) carry a FORCED row, which is a stated reason, not a review.
```


ciaran-slow (approved):
Reviewed commit-by-commit in a fresh clone at `786f86b`, on node v22.23.2. No changes requested; nothing pushed to the branch.

This PR was opened `--unverified` with three FORCED ledger rows. **This review is the independent pass those three commits lacked** — every claim below says whether I ran it (`EXECUTED`) or argued it from the code (`REASONED`).

**The tests are not vacuous.** Two mutations of my own, neither on the author's certified list, each killing exactly one test:

| mutation | result |
|---|---|
| `if (diff.raised.length > 0 \|\| diff.cleared.length > 0)` → `if (false)` (persist write dropped, hydrate left intact) | `EXECUTED` — 1 failed / 43 passed. Kills exactly `MUTATION 2 (DoD): a condition still open across a restart does not re-raise` |
| `const next = manifestDegradedThisTick ? [...diffedNext, ...preservedOffFence] : diffedNext` → `const next = diffedNext` (splice kept, carry-forward dropped) | `EXECUTED` — 1 failed / 43 passed. Kills exactly `preserves an open off-fence summons across a timed-out read, and does not re-raise it once the read recovers` |

The second is the one worth naming: it leaves the *clear* suppressed and breaks only the *re-raise*, and the test still caught it — because it asserts `offFenceRaises).toHaveLength(1)` across all three ticks rather than merely asserting no clear. That is the anti-chattering assertion doing real work.

`poll-loop.test.ts:1277` (`without a snapshotStore, a restart forgets the edge-state and re-raises`) is the control that stops MUTATION 2 being a tautology. It is the right test and it is present.

**Three claims I checked rather than took on trust, all of which hold:**

1. *"A degraded tick cannot re-raise a preserved off-fence point."* `REASONED`, from source: `manifestDegradedThisTick` implies `manifest === null` → `buildFleet.ts:289` gives `fence = undefined` → `:359` `fenced: false` → `diagnose.ts:336` `detectOffFence` returns `null` on `!lane.fenced`. So off-fence can never appear in `conditions` on a degraded tick, `preservedOffFence` and `diffedNext` are disjoint, and `next` cannot acquire a duplicate. `buildFleet.ts:123` states this as ruling 19 independently.

2. *"An existing poll-loop guarantee moved."* It did not. The `raceBudget` generalization is pure type-parameterization — same `setTimeout(TIMED_OUT, budgetMs)`, same `timer.unref()`, same resolve/reject arms — and the collector call site passes the identical `tickBudgetMs` it previously closed over. `REASONED`, from the diff.

3. *A latched `summons` error surviving a `reset()`.* I expected one: `reset()` zeroes `summonsManifestFailures`, so a recovery armed before the reset is disarmed while the `collector.error` stays folded. Not reachable — both callers open a new session first (`api/rotate.ts:95`, `api/retarget.ts:269` both build the new store from `*.opened.sessionId`), so the error is in the closed session's log and the new fold starts clean. `REASONED`.

`setManifestReadOverride` is a `vi.hoisted()` mock inside the test file, not a production seam.

## Gates

- `EXECUTED` `gh pr checks 303` — **10/10 legs green across all three expected jobs**: `build-test-boot` ×3, `pack-smoke` ×6, `windows-suite` ×1. Counted against the checklist rather than read off the bar.
- `EXECUTED` `npm run typecheck` clean; `npm run lint` clean, 899 files.
- `EXECUTED` `VITEST_MAX_WORKERS=6 npm test` → **414/415 files, 7660 passed, 3 expected fail**. One failure: `recorder/session-recorder.test.ts > holds the ceiling in a session reached by rotation`, timing out at 5000ms. **Not this PR's**: the file is untouched by the diff (`git diff --name-only realmain..pr303` does not list it), it passes **57/57 in isolation**, and both ubuntu CI legs pass it. Known load-sensitive timing flake at 6 workers.
- `EXECUTED` base is `main` and `origin/main` is already an ancestor of the head — no integration-branch trap. Both merge commits have an empty combined diff, so neither is an evil merge.
- Fence: 7 files, all inside #278's declared fence including both recorded widenings on `collector-loader.test.ts`. Both were recorded on the issue **before** the change, as AGENTS.md requires.

**What I did not run:** `scripts/gate.sh` (the operator's step, and it lands rather than checks), and the load-batch concurrency probe. Neither is mine to run here.

The four follow-ups (#299–#302) are unusually good: each carries EXECUTED evidence, and three of the four are shown pre-existing by reproduction against `61e7178`. #301 is correctly flagged as the one genuinely introduced here. Filing them beat chasing them.

One grooming note: all four fence `packages/server/src/server/poll-loop.ts` and `poll-loop.test.ts`, so they mutually overlap and cannot be dispatched as one wave — `fence-lint.sh 299 300 301 302` would hard-fail on the overlap. #302 already asks whether it and #300 are one change; that question probably wants answering across all four before any of them is boarded.

## Verdict

Approved. I found nothing to fix, so there is no reviewer commit on this branch. Landing stays the operator's step.

PR #298, opened 2026-09-06
prd-43 wave 9: the runbook states what a PR requires before it is opened

gabriel-canaan:
prd-43 wave 9 — the runbook states what a PR requires before it is opened, plus the
Sequencing corrections and rulings that reviewing it produced.

Closes #285

## Commits

| commit | what becomes true |
|---|---|
| `022abef` | the wave-8 paragraph counts the sites #232 found, not the rows it left |
| `ae0368f` | this PRD is read by the suite and enforced by none of it |
| `abe4d83` | a PR states its two preconditions where the runbook is read (#285) |
| `7257014` | wave 10 is declared, and the superseded wave-7 paragraph says so |
| `dcdf06f` | five claims in the wave-9 commits were false, and the marker swallowed its own ruling |
| `e86225b` | wave 10 is built and in review, not waiting to be built |
| `fd48918` | the route-count exclusion is already vacuous, and this paragraph said otherwise |
| `cf96267` | rulings 6, 7 and 8 — the three open questions are ruled, by what shipped |
| `3b9e072` | counts ship their derivation, and findings are listed rather than totalled |

`grep -c '/verify' AGENTS.md` returned 0 before this. The rule that a PR needs an
independent review pass over every commit, and the operator's word for that specific
PR, lived only in a project memory and in a skill that loads conditionally — so a
session that never invoked the skill never met the rule. The new section states both
conditions, names `--strict` and why, and ends by saying it is NOT the enforcement: a
PreToolUse hook is, and reading a paragraph is not compliance.

## What is the sibling case?

The other standing constraints that lived only in prose. Checked rather than assumed:
PR authorisation has a hook, verify-before-PR has a ledger that hook reads, and lane
permissions are enforced by a preflight that refuses a wrapper skipping them. Scope was
the one still being restated by hand; it now has a skill and a hook of its own, both
outside this repo.

Within this branch the sibling is the PRD's own Sequencing, and it is why five of these
nine commits exist. A Sequencing section describes work that changes underneath it, so
its TENSE is load-bearing and nothing checks it — waves 5, 6, 7 and 10 all described
landed work as pending at some point during this branch's life.

## What mutation would this test survive?

**None, and that is the finding rather than an excuse.** `docs/prds/` is excluded from
every enforcement sweep by ruling 8 — which this branch also writes down — and
`AGENTS.md`'s citation law checks numbers-as-citations, not numbers-as-counts. Every
defect in these nine commits was caught by a human or a review seat. No gate could have
caught any of them.

[...]

PR #304, opened 2026-09-06
prd27 w4: a lapsed declaration reads as lapsed, and configured-but-silent says so

### LINKED ISSUE #218: prd27 w4: a lapsed declaration reads as lapsed, and configured-but-silent says so
## Fence (may touch ONLY)
- `packages/core/src/collector.ts` (`deriveRung`: L2 told apart from L4)
- `packages/core/src/collector.test.ts`
- `packages/core/src/selectors/lapse.ts` (new) and `lapse.test.ts` (new)
- `packages/core/src/selectors/index.ts`
- `packages/server/src/collectors/sessionlog/collector.ts` (the `attention` capability strings only)
- `packages/server/src/collectors/sessionlog/collector.test.ts`
- `packages/server/src/collectors/beacon/collector.ts` (the manifest only)
- `packages/server/src/collectors/beacon/collector.test.ts`
- `docs/design-notes/beacon-lapse-interval.md` (new)

## Blocked by
**#283** (wave 3) — there is no `declared` reading to lapse until the fleet believes a beacon.
**#282** (wave 2) — the interval is **measured** from real hook beacons, so `rhizomorph env --hooks claude` must have run in real sessions long enough to observe the cadence of `working` beacons during an active turn. The design note cites that capture.

Sibling lanes: nothing in `packages/core/src/reduce.ts`, `fleet/buildFleet.ts`, `api/**` or `packages/web/**` is touched here — #283 owns the fold and the voice; this issue adds the third reading and the rung.

**Regroomed 2026-09-05 (w2 → w4).** The original body bundled the fold, the fleet's belief, the disagreement voice, the emitter and the lapse into one issue with a fence that named none of `packages/core`, `packages/web` or `api/meta.ts`. The PRD's own sequencing puts the lapse last, after something real exists to measure. #281 (the witness) and #282 (the emitter) are wave 2; #283 (the fold, the belief, the disagreement) is wave 3; this is wave 4. The original text is kept below.

## Why

prd-27 ruling 6 and the 2026-08-24 amendment, verbatim:

> A previously live beacon that goes quiet becomes a first-class lapsed state with its own voice ("declared attention lapsed Nm ago; reading turn shape"). The interval is not guessed here: the implementation wave owes a design note with a measured value before the state ships.

And ruling 3's carve-out, closed by the same amendment: **configured-but-silent reads `partial`, with the reason said.** "Configured and quiet" differs from "never offered", and the reading names what would strengthen it.

## Definition of done

- Three readings, three voices, each a test: **never declared** (no beacon for the lane, ever — attention stays whatever the other organs say, no new word); **configured but silent** (the beacon directory exists and holds a writer's file, but no attention beacon for this lane has arrived — `attention: partial`, reason `hooks configured; no beacon for this lane yet`, remedy naming the first hook that would speak); **lapsed** (a lane's newest attention beacon is older than `BEACON_LAPSE_MS` while the lane is still `present` — the condition's why reads `declared attention lapsed <span> ago; reading turn shape`, and the fleet falls back to the inferred reading with `inferred: true`).
- `packages/core/src/selectors/lapse.ts` exports `BEACON_LAPSE_MS` and a pure `declarationStatus(declared, now, lastWorkTs) → 'live' | 'lapsed'`; `diagnose.ts` reads it. Tests assert both sides of the boundary — `BEACON_LAPSE_MS - 1` live, `BEACON_LAPSE_MS + 1` lapsed — **using the exported constant, never a locally typed literal**, and a third test asserts the constant equals the value the design note states, read from the note's text, so the code and the note cannot drift.
- `docs/design-notes/beacon-lapse-interval.md` records the measured cadence of `working` beacons in real Claude Code sessions with #282's hooks installed (how many sessions, how long, the longest gap between beacons during an active turn, on which machine class), and derives the interval from it with the margin stated. A constant with no note is what the amendment exists to prevent.
- `deriveRung` tells **L2** (attention `provided` by the beacon organ) from **L4** (attention `provided` by tmux/workmux): a fleet whose only declaring witness is `beacon` reports L2; `doctor` and `/api/meta` agree, and `rungInfo('L2')` is now reachable by a real collector (the doc comment in `collector.ts` that says it is not is corrected).
- `SESSIONLOG_CAPABILITIES.attention` tells the truth about which case the reader is in: the `remedy` no longer says *would declare it* (a beacon exists) — it names `rhizomorph env <lane> --hooks claude`. `BEACON_CAPABILITIES.attention` moves off the static `absent` to `partial` with the configured-but-silent reason, since #283 now reads beacons per lane.
- `doctor` prints the three readings per lane and the rung. `CHANGELOG.md` `[Unreleased]` › `Added`.

## Sibling case

"Never declared", "declared then lapsed" and "configured but never spoke" are three states that a boolean collapses into two. The sibling miss here is a lapse implementation that renders a never-configured harness as lapsed — which is precisely the false summons this PRD was written to stop, in a new costume. Its mirror: a lapse that fires while the lane is not `present` (the worktree is gone) — a landed lane has not lapsed, it has finished.

Set the lapse interval to a different value and re-run. A test that passes at any interval — the `TTL - 1` / `TTL + 1` shape AGENTS.md names — is the failure mode here, because the interval is the one thing this wave measured. Assert against the boundary from both sides, using the note's value rather than a locally-typed literal; the note-equals-constant test must redden when either side changes alone.

---

### Fence (may touch ONLY)
- `packages/server/src/collectors/sessionlog/`
- `packages/server/src/collectors/beacon/`
- `docs/design-notes/`

### Blocked by
Wave 1 — the beacon collector and its ADR. There is no declaration to lapse, and nothing to disagree with, until beacons arrive.

Sibling lanes: #202 owns `docs/adr/` — this wave writes a **design note**, not an ADR, because what it owes is a measured value rather than a structural decision.

### Definition of done (original)
- A previously-live beacon that stops arriving becomes a **lapsed** state, distinct from "never declared" and distinct from "declared and quiet". Three readings, three voices.
- **Configured-but-silent reads `partial`, with the reason said.**
- A design note under `docs/design-notes/` records the lapse interval **with the measurement behind it**.
- When a beacon declares one attention state and the turn-shape machine infers another, the disagreement is voiced rather than resolved silently. → now #283.
- The `sessionlog` collector's `attention` capability moves off `partial` **only** where a beacon actually declares it.

ciaran-slow:
Closes #218

prd-27 wave 4, and the last wave of prd27 — waves 1–3 (#217, #281, #282, #283, #290) are on `main`. One issue, so one commit; the wave PR is still the unit of queueing.

## What becomes true

A hook beacon that goes quiet while its lane keeps working **stops being believed**. Until now a declaration stood forever: a harness that crashed, or whose hooks were uninstalled mid-session, left the fleet asserting `declares working` indefinitely with nothing to contradict it.

- **A declaration lapses.** After `BEACON_LAPSE_MS` a `working` or `stopped` declaration is contradicted by age alone; a `waiting` one lapses **only once work follows it**, because a lane genuinely waiting on a human is silent for as long as the human takes — doubting that would be the false summons in reverse. A lapsed declaration has no precedence in `detectWaiting`: every ruling-4 clause sees `null`, the inference that stands takes over, and it says why — ` · declared attention lapsed <span> ago; reading turn shape`, on all three arms (pane-stillness, transcript-shape, and the workmux arm's `olderBeacon` slot). The STATE card speaks the same clause.
- **Configured-but-silent says so.** `attentionReading` gives one lane one of four readings — `never-declared`, `configured-silent`, `live`, `lapsed` — and `rhizomorph doctor` prints a line per present lane off the same `buildFleet` the dashboard uses, so doctor and the STATE column cannot disagree. A lapsed lane `warn`s; nothing here can `fail`.
- **L2 is reachable for the first time.** `attention` now carries a `witness` on the capability manifest (`'rig' | 'beacon'`), so `deriveRung` tells *who* declared, not merely that someone did: beacon-provided is **L2**, rig-provided is **L4**, and a tie goes to the rig. A configured-but-silent beacon — `partial`, no telemetry — can **no longer read as L3**, which is precisely the false PTY rung ADR-0036 recorded as the reason #217 had to leave the manifest `absent`. Recorded as [ADR-0039](docs/adr/0039-attention-names-its-witness-on-the-manifest.md).
- **Presence is the caller's.** A landed lane has not lapsed, it has finished — it gets no line at all.

## The measurement — the constant is measured, not chosen

prd-27 ruling 6 and the 2026-08-24 amendment: *the mechanism is ruled, the number is measured.* The derivation rule was fixed in the plan **before** the measurement, so the result could not be steered.

| | |
|---|---|
| sessions | **3**, five prompts each = **15 turns**, 2026-09-07, manual-approval mode confirmed on the pane before every session |
| beacons | **77** (60 `working`, 2 `waiting`, 15 `stopped`); 60 counted intra-turn gaps |
| **longest intra-turn gap `G`** | **57 s** — session 1, turn 2, `UserPromptSubmit → Stop`: the no-tool 900-word generation turn, which is the turn the interval is really about (one beacon at the prompt, none until `Stop`) |
| median gap | **5 s** |
| machine | Apple Silicon Mac mini, 8 GB (the operator's box) |
| Claude Code | `2.1.261` for session 1; the CLI self-updated mid-capture, so sessions 2 and 3 ran on `2.1.263`. Both are named in the note, and **the longest gap came from session 1, on the older build.** |
| **derivation** | `max(3 × 57 000, 2 × TURN_SETTLE_MS = 150 000)` = 171 000 → ceil to the minute → **`BEACON_LAPSE_MS = 180000`** |

A gap whose first line is a `Notification` is excluded from the count by construction — that gap is the human's latency to a permission dialog, which is exactly the declared `waiting` case and is unbounded by design. `G` is well inside the plan's 5-minute stop condition.

The 3× margin covers a turn twice as slow as the slowest measured before the declaration is doubted; the 2×`TURN_SETTLE_MS` floor ties the lapse to the transcript organ's own settle window, so a beacon is never doubted sooner than the organ doubts a turn.

**The verifier recomputed `G` independently** from the note's raw capture block alone, applying the plan's turn/gap rules rather than re-reading the builder's arithmetic: 77 lines, 15 turns, 60 gaps, `G = 57 000 ms`, median 5 000 ms — every one of the 15 per-turn rows matched. The 77-line block is byte-identical to the capture file. `TURN_SETTLE_MS = 75_000` confirmed at source.

The note (`docs/design-notes/beacon-lapse-interval.md`) is the source of truth for the value, and a law in `beacon/collector.test.ts` reads the number back out of the note's prose — so changing the constant alone reddens it, and changing the note alone reddens it from the other side. Neither can be quietly retuned. The raw capture was grep-audited for `/Users/`, `/home/`, `C:\`, the temp roots and the hostname before it went in: zero matches (the emitter's line shape carries no path by construction, and the lane `2-core` is synthetic).

## What is the sibling case?

The issue names the pair, and both are tested from both directions:

- **A never-configured harness rendered as lapsed.** It cannot reach the arithmetic: `attentionReading` returns `never-declared` from an empty fold before any age is computed, and `declarationStatus(null, …)` is `null`, never `'lapsed'`. Doctor prints "never declared" with an `ok`.
- **A lapse that fires on a non-`present` lane.** `declarationStatus` deliberately does not read `present`; `detectWaiting`'s existing exemptions and `checkDeclaredAttention`'s `lane.present` filter decide it, so the landed-lane case goes through the guard #283's tests already pin.

**The sibling the builder had to go looking for is a third one**, and it is why §4.1 names three call sites rather than two: the lapsed clause is appended in **three** places in `detectWaiting`. The transcript-shape and pane-stillness arms are the obvious pair — #296's review already caught a clause added to one and not the other — and the workmux arm's `olderBeacon` slot is the one a reader skims past. All three are pinned byte-exact and independently: dropping the clause from any single arm reddens exactly one test.

On the manifest side, the sibling risk is structural: `beaconCapabilitiesFor` is the **only** constructor permitted to build a beacon `provided`, because an optional `witness` forgotten anywhere else reads as **L4 — the top rung — silently**. ADR-0039 records that as its named bad consequence, and the test pins the field with `toEqual`, not a level check.

**The honest answer first: a `BEACON_LAPSE_MS ± 1` boundary pair survives any interval whatsoever.** That is AGENTS.md's own named failure shape, and since this issue exists to introduce exactly such a constant it is defended twice — every boundary case imports the constant instead of typing a literal (so they all move together; a grep of all nine touched test files finds no typed `180_000`), and the *value* is held to the measurement by the note law.

The verifier applied **21 real source mutations**, ran the nine touched test files (374 tests) against each, and restored the tree between runs. Two results are worth surfacing rather than burying:

- **M8 — removing `mergeCapabilities`' tie-break survived `meta.test.ts` and `doctor.test.ts`.** Both "L4 once workmux is active" assertions stayed green, because the route and the CLI happen to list the beacon *last*, so first-seen-wins keeps the rig there anyway. The unit tests in `collector.test.ts` are what actually pin the rule — and they run it in **both input orders**, which is the point, since the rule being replaced was "first seen wins" and an order-sensitive fix passes in exactly one order. The build report's "all in both input orders" holds for the unit tests, not for the route and CLI; recorded here rather than left as an overclaim.
- **M9 — beacon `partial` reading L3 again survived `meta.test.ts`**, as expected on a box where the transcript organ is present; `collector.test.ts` and the beacon manifest test are what close it.

No mutation survived everything, and no test in the diff is decorative. Every other mutation reddened the test the plan predicted: ignoring `lastWorkTs`, never lapsing `working`, flipping the boundary to `<`, swapping never-declared with configured-silent, `deriveRung` ignoring `witness`, `beaconCapabilitiesFor` forgetting it, dropping the `present` filter, the doctor `catch` rethrowing, and each of the three arms in turn.

The 3^6 capability-exhaustiveness law now runs **twice** — once unchanged (no `witness`, so the pre-existing law is untouched) and once with `attention` signed `beacon` across all 729 combinations, asserting totality and that `provided` always reads L2.

## Deviations from the plan

Three, all recorded on the issue before or at build time, all judged and accepted by the verifier:

1. **A `vi.mock` seam for the unreadable-session doctor case.** The plan's §5.3 seeds "a session file of garbage bytes" and expects a `warn`. It cannot work as written: ADR-0011 says a recording never rots, so `readSessionLog` swallows *every* file-level failure — bad bytes, an unopenable path, a vanished file — and returns an empty session. (Confirmed by seeding both garbage bytes and a directory at the session path; both read `ok`.) The guard is still real, because `reduceAll`/`buildFleet` can throw, so the fault is injected at the read seam with the same `vi.mock` passthrough shape `beacon/collector.test.ts` already uses. Mutation M17 (rethrow inside the `catch`) reddens exactly that test, so **the mock does exercise the guard** rather than paper over it, and a case pinning the honest lenient ADR-0011 outcome was added alongside.
2. **Built on opus, groomed as sonnet.** Recorded in plan §0 before the build, with the reason: this issue changes rung semantics, adds a doctor check that folds a session, and derives a constant from a measurement. A deviation from the grooming, not from the plan.
3. **`S-Tab` → `BTab`** in the measurement procedure. `tmux send-keys … S-Tab` is *accepted and silently does nothing*, leaving auto mode on — which is exactly the failure `CAPTURE.md` warns about, with no error to notice. `BTab` is the working key name. A tmux key name, not a design change; the mode line was read off the pane before every session. Follow-up below.

## Fence widenings

The plan declared five before the change; the build declared a sixth, forced, before making it.

1. `packages/core/src/selectors/condition.ts` + `condition.test.ts` — the DoD requires the card to speak the lapse, and `declaredClause` is where a card names a declaration.
2. `packages/server/src/api/meta.ts` + `meta.test.ts` — the DoD requires `doctor` and `/api/meta` to agree on L2; `/api/meta` builds its ladder from static manifests, so the live attention has to be applied there.
3. `packages/server/src/collectors/beacon/index.ts` — one export line for the helper `meta.ts` and `doctor.ts` share.
4. `docs/adr/0039-*.md` (new) + a row in `docs/adr/README.md`.
5. `CHANGELOG.md`.
6. **`packages/core/src/fleet/buildFleet.test.ts` — unplanned and forced.** Plan §8 says explicitly *"not in the fence — do not add to it"*, and no case was added. What could not be avoided: its existing `(b3)` test declares `working` at `NOW - 300_000`, which the newly measured interval **lapses** — so `staleDeclaredWorking` correctly stops speaking and the byte-exact assertion goes red. One declaration was re-timed to `NOW - 170_000` and its expected span to `2m50s`. The test's claim is unchanged; only its clock is.

**The verifier judged the sixth widening explicitly and accepted it**, on two checks: (1) it still tests clause b3 — the lane's last work sits at `NOW - 160_000`, so `170_000` remains *older than the last work* and is now also *younger than the lapse*, which is exactly the window `staleDeclaredWorking` speaks in; leaving `300_000` would have been a red test, not a preserved one. (2) The byte-exact string is honestly re-derived — `formatSpan(170_000)` is `2m50s` at source, not guessed. One noted fragility, carried in comments beside all three re-timings: `170_000` is only "younger than the lapse" because the constant is 180 000, so a future re-measurement that *lowered* the interval would redden these three and need re-timing again.

`git diff --name-only origin/main...HEAD` is 23 files = the issue's 13 fence paths + these six widenings. Nothing under `packages/web/**`; no `reduce.ts`, `state.ts`, `fleet/buildFleet.ts`, `api/doctor.ts`; no era snapshot (`SessionState` is untouched, so **no era-1 re-bless**); no lockfile change.

## Verification

**Verdict: ready to merge.** Verified 2026-09-07 on a fresh worktree from `origin/218-lapsed-declaration` at `a2ea67c`, by a **different model** (`fable`; the commit's own trailer is `Claude opus`), against base `origin/main` at `29cde14`. **No must-fix finding**, and no fix commit was needed — the commit on this branch is the builder's, unamended.

The verifier ran every gate independently, walked the Definition of done criterion by criterion (all met), ran the 21 mutations above, and did a sibling hunt across every reader of `.declared` outside the diff (`buildFleet.ts` sets the raw fact and lets consumers decide; the web's `beaconFlow` pill counts occurrences rather than making a per-lane truth claim; no consumer renders a lapsed declaration as current). Architecture checks: no `node:*` in `packages/core` (the note law lives in `server` for exactly that reason), no write into the watched repo, no new route, no CORS or schema change.

One nuance the verifier recorded as REASONED and not a defect: the DoD's exact string `hooks configured; no beacon for this lane yet` is carried by the **per-lane** reading that doctor prints, while the organ's **session-wide** manifest carries a longer sentence containing `ruling 3`. That split is deliberate (plan §1.5, §4.3) — the manifest is not per lane, and `meta.test.ts` already asserts `ruling 3` on it.

## Post-verify fixes

The first push of this branch (`f703294`, lane `a2ea67c`) went **9/10 green**, with `windows-suite` red on exactly one UNEXPECTED FAILURE: the new note law's `carries a measurement, not prose …` case, failing `no ```jsonl raw-capture block in the note`.

The note was correct; the reader was not. `docs/design-notes/` carries no `eol=lf` pin (unlike `fixtures/**`), so the note arrives on a Windows checkout with **CRLF** line endings — and the block anchor `/```jsonl\n([\s\S]*?)```/` matches a literal `\n`, so it never fired. The `BEACON_LAPSE_MS` anchor has no newline in it and matched either way, which is why only one of the two cases went red.

One line, in the law's own `describe`:

```
-    const NOTE = readFileSync(NOTE_PATH, 'utf8')
+    // The note is prose under docs/, not a pinned fixture: a CRLF checkout (the windows-suite leg) must read the same bytes-as-text this law was written against.
+    const NOTE = readFileSync(NOTE_PATH, 'utf8').replace(/\r\n/g, '\n')
```

Proven rather than assumed: the note's text written out with CRLF endings matches `/BEACON_LAPSE_MS = (\d+)/` **before** normalisation (→ `180000`) and **fails** the jsonl-block regex; after `.replace(/\r\n/g, '\n')` both match, the block yielding its 77 non-empty lines. This is the `line-endings` cause class `.windows-known-failures` documents, but it is a fixable defect in a brand-new law rather than an entry for that list, so nothing was added there.

Amended into the single lane commit — lane `a2ea67c` → **`7f4ce6b`**, wave branch `f703294` → **`fa4a37c`** — with the reason written into the commit body under `Post-verify fix (windows-suite):`. **The `fable` verify pass above predates this change**: it read `a2ea67c`, and this one-line test-reader fix is the only difference between what it read and what is on the branch now (`git diff a2ea67c 7f4ce6b` is the hunk shown above and nothing else). Gates re-run on the amended lane commit — typecheck, lint, build, and `VITEST_MAX_WORKERS=6 npm test` (same single pre-existing `session-recorder` timeout, green run alone) — and typecheck, lint and the beacon collector file (42 passed) re-run on the re-assembled wave branch.

### ADR renumbered `0038` → `0039`

While this PR sat unmergeable, [#303](https://github.com/launchpad-26/rhizomorph/pull/303) landed on `main` (`45c61e3`) carrying a **different** `docs/adr/0038-a-summons-raiser-judges-the-fold-on-the-tick.md` and its own README row. Two records claimed `0038`, so #304 could not merge and never got CI. `docs/adr/README.md` §Conventions settles which one moves without a judgement call — numbers are **sequential and never reused**, and the log is append-only — so the record that landed first keeps `0038` and this one takes the next free number, `0039` (confirmed free: a tree listing of `docs/adr/` on `origin/main` ended at 0038).

The rename is mechanical and nothing else in the change moves: the record is renamed, its H1 `# 0038.` → `# 0039.`, its README row, the CHANGELOG link text and path, and the ten `ADR-0038` code citations (`packages/core/src/collector.ts` ×4, `collector.test.ts` ×2, `collectors/beacon/collector.ts`, `api/meta.ts`, `api/meta.test.ts` ×2). Diffed old lane commit against new, that is **13 lines across 8 files, all of them the number**; #303's own `0038` references were not touched.

Amended into the single lane commit — lane `7f4ce6b` → **`de87fb1`**, wave branch `fa4a37c` → **`98ce4d6`** — re-assembled by cherry-picking onto current `origin/main` (`45c61e3`). The **one** conflict was the `docs/adr/README.md` table tail, which is the collision itself: both sides append a row after `0037`. Resolved to `0037`, `0038` (#303's row, verified byte-identical to `origin/main`), `0039` (ours), so the table's diff against `main` is **one added line**. Patch-vs-patch the lane and the wave still carry the same change: each commit's own diff has **byte-identical added/removed lines** (1,529 of them) and they differ only in the README hunk's context, which is exactly the `0038` row `main` has and the lane branch does not.

Gates re-run, EXECUTED, node v22.23.2, dependencies installed first — on the amended **lane** commit: typecheck, lint, build, the four targeted files (151 passed, `doc-citation-law.test.ts` among them), and `VITEST_MAX_WORKERS=6 npm test` → **414/415 files, 7681 passed**, the single red being the same pre-existing `session-recorder.test.ts › holds the ceiling in a session reached by rotation` 5 s timeout (57 passed run alone, and red on untouched `origin/main` too). On the **assembled** branch: typecheck, lint, and `doc-citation-law` + `beacon/collector` + `api/meta` (116 passed).

## How the wave was assembled

Per the open-wave-pr skill: never opened from the lane branch.

```
git fetch origin
git checkout -b prd27-w4-lapsed-declaration origin/main      # 45c61e3
git cherry-pick de87fb1c42ffdf123778cbc9d9ac8b2a3a717ae3     # one conflict: the docs/adr/README.md table tail
```

Cherry-picked onto a `main` that had moved three merges past the lane's verified base (`f479a9a` → `d7601ec`, `d804b5a`, `45c61e3`), giving `98ce4d6` — one conflict, the ADR README table tail, resolved as described under Post-verify fixes. Because `origin/main` advanced, a raw diff of `de87fb1` against `HEAD` is *not* the right equivalence check — it would report the unrelated `main` commits. The check that means something is patch-vs-patch: each commit's own diff — lane against its parent, wave tip against its parent — carries **byte-identical added and removed lines** (differing only in the README hunk's context; see Post-verify fixes), so the wave branch carries the lane's reviewed change exactly, and the 23-file footprint matches the verify report.

### Gates on the assembled branch — all EXECUTED, node v22.23.2, `npm install` first

The bundle is the first run where the wave's commits sit on current `main`, so this is the run that can catch cross-lane breakage.

| gate | command | result |
|---|---|---|
| typecheck | `npm run typecheck` | **pass** — contract, core, server, web |
| lint | `npm run lint` | **pass** — 899 files, no fixes applied |
| build | `npm run build` | **pass** — `dist/main.js` 864.5 kB |
| test | `VITEST_MAX_WORKERS=6 npm test` | **7688 passed**, 1 failed, 3 expected-fail, 6 skipped (7698); 415 files; 261.47 s — the single failure is the pre-existing one below |
| boot smoke | `npm start -- --port 0` | **pass** — see below |

Boot smoke, on a throwaway `RHIZOMORPH_DATA_DIR` (removed afterwards): `GET /` → **200**, `text/html`, 1628 bytes. `GET /api/meta` bare → **401**, `missing or invalid x-rhizomorph-capability header`. With the page's capability token (read from the served `<meta name="rhizomorph-capability">`) → **200**, and:

- `rung: "L1"`
- `capabilities.beacon.attention` = `{ level: 'partial', witness: 'beacon', reason: '…configured-but-silent reads partial (prd-27 ruling 3)…', remedy: '…rhizomorph env <lane> --hooks claude…' }`
- `capabilities.sessionlog.attention.remedy` names `--hooks claude` and no longer says "would declare"

The rung reads **L1** here rather than the `L4` the build and verify boxes reported, because this worktree has no tmux/workmux rig attached — and that difference is worth stating, because it exercises the guard this issue exists for from the other side: a **configured-but-silent beacon on a machine with no rig reads L1, not L3.** That is exactly the false PTY rung ADR-0036 named, closed by type.

### Pre-existing, not this wave's

The one red in the full run is `packages/server/src/recorder/session-recorder.test.ts › SessionRecorder — the window has a stated ceiling (prd-44 ruling 4, #37) › holds the ceiling in a session reached by rotation, not only in the first one`, which times out at the 5000 ms default under the 6-worker suite.

It is not this wave's, and that was established by execution three independent times rather than asserted:

- **The builder** stashed the entire change (`git stash push -u`) and ran the full suite against the untouched `origin/main` tree: the same single test fails identically there.
- **The verifier** confirmed it independently, and checked the coupling: `git diff --name-only origin/main...HEAD` contains nothing under `packages/server/src/recorder/`, `session-recorder.ts` imports only `initialSessionState, reduce, reduceAll` from core, and this diff touches none of them — so no code path the test exercises changed. The file was last touched on `main` at `05e47ed`.
- **This assembly run**, on the bundled branch: it was the **only** failure in 415 files, and run alone it is **57/57 passed in 10.46 s** (`npx vitest run packages/server/src/recorder/session-recorder.test.ts`).

It is a latent load-sensitive flake: the case seeds 75 000+ events at the real ceiling and sits at roughly 83% of the default timeout even on a quiet box, so it tips over under six workers. It belongs to a fence that owns `recorder/`, and it will surface again in `scripts/gate.sh`'s load pass — the honest fix is a raised per-test timeout, since the file's own header explicitly declines the `// @gate-timing` marker.

## Follow-ups

None blocks this wave; each has a home and a named fence.

1. **`GET /api/doctor` can never report L2.** The route still calls `checkEnrichmentLadder(checks, repoPath)` with two arguments, so its beacon contributor is the static `partial` regardless of the fold. The third parameter was made optional precisely so the route kept compiling, and `api/**` is fenced out of #218 by name. Concrete divergence: on a machine with a recorded beacon and no tmux/workmux, the CLI says `L2 — beacon (declared attention)` while the route's ladder says L0/L1. The issue's clause — "`doctor` and `/api/meta` agree" — does hold; it is the *third* surface that lags. One-line fix at the call site (pass the recorder's fold) plus the attention checks, in a fence that owns `api/doctor.ts`.
2. **`docs/architecture.md`'s decisions log has no line for ADR-0039** — nor for **ADR-0036 or ADR-0037**, which are cited only in the body. The build skill asks for a decisions-log line when an ADR changes the current architecture; the plan enumerated its widenings exhaustively and `architecture.md` was not among them, so touching it would have been an undeclared seventh widening. Worth doing for the three records together.
3. **`packages/server/src/collectors/beacon/fixtures/CAPTURE.md` should say `BTab`, not `S-Tab`.** The recipe's current key name is accepted by tmux and silently does nothing, leaving auto mode on — the exact failure the document warns about, with no error to notice. The design note carries the correction so it is not lost, but the recipe is where a future capture will read it.
4. **The `session-recorder.test.ts` rotation-ceiling timeout** above — a raised per-test timeout, in a fence that owns `recorder/`.
5. **Carried from wave 3: the STATE-hover screenshot capture** still wants doing at a `main` HEAD.

## A precondition I could not check

`AGENTS.md` §"Opening a PR" (added on `main` in #298, after this lane was built) requires the range to be confirmed against the verify ledger via `~/.claude/skills/verify/scripts/verify-ledger.sh check --strict`. **That tooling is operator-local and is not present in this worktree** — the section says so itself — and no `pr-approval-guard` hook fired here. So I am recording the gap in the PR body where a reviewer sees it, which is what the section prescribes as the honest escape, rather than reporting a check I did not run.

What *is* true: an independent verify pass on a different model (`fable`) is posted in full on #218, with a `ready to merge` verdict, no must-fix findings, and 21 executed mutations. What I cannot show is the ledger row recording it. Worth an operator confirmation before merge.


gabriel-canaan:
## Review — reads as ready to land

Reviewed on a fresh worktree from `origin/prd27-w4-lapsed-declaration` at `98ce4d6`, node v22.22.2, `npm install` first. `origin/main` merges in as **"Already up to date"** — the cherry-pick onto `45c61e3` means this branch already *is* current `main` plus the one commit, so everything below was exercised against the tree that will land, not an older one.

**The measurement reproduces exactly.** This is the load-bearing claim of the whole wave, so I recomputed `G` from the note's raw capture block alone — extracting the ```` ```jsonl ```` block, applying the note's own stated rules (a turn is `UserPromptSubmit` → next `Stop`; an intra-turn gap is between consecutive lines inside one turn, excluding any gap whose *first* line is a `Notification`) rather than re-reading the arithmetic. **EXECUTED:**

```
lines: 77  turns: 15  unclosed: 0
kinds: {"working":60,"waiting":2,"stopped":15}
gaps counted: 60   G = 57000 ms   median = 5000 ms
derivation: max(3*G, 2*TURN_SETTLE)= 171000   ceil-to-minute = 180000
```

All **15 per-turn rows** matched the note's table byte for byte — length, beacon count, longest gap *and* the hook pair named in parentheses. `TURN_SETTLE_MS = 75_000` confirmed at `packages/server/src/collectors/sessionlog/lane-state.ts:68`. The derivation lands on 180 000 independently.

**Gates, all EXECUTED in a clean worktree:**

| gate | result |
|---|---|
| `npm run lint` | pass |
| `VITEST_MAX_WORKERS=6 npm test` | **416/416 files, 7718 passed**, 3 expected-fail, 2 skipped, **0 failed** |

Worth noting for the author: the `session-recorder.test.ts` rotation-ceiling timeout you reported as the single pre-existing red **did not reproduce here** — full suite, same worker count, zero failures. That is a fourth independent data point for your load-sensitive-flake diagnosis, and it strengthens it rather than contradicting it.

`gh pr checks 304`: **10/10 legs pass** — `build-test-boot` ×3, `pack-smoke` ×6 (including both Windows legs), `windows-suite`. The CRLF fix in the note law is witnessed green on the leg that caught it.

**Four mutations of my own, all caught.** Restore asserted with `git diff HEAD --quiet` after each, not announced:

| # | mutation | result |
|---|---|---|
| A | `BEACON_LAPSE_MS` 180 000 → 240 000 **and** the note's stated value with it, capture left untouched | **CAUGHT** — 1 failed / 373 passed |
| B | drop the `lapsed !== '' ? lapsed :` branch from the workmux arm's `olderBeacon` | **CAUGHT** — 1 failed, exactly the one arm |
| C | `lapsedForMs` returns `now - declared.at`, forgetting the interval | **CAUGHT** — 4 files, 7 tests |
| D | `waiting` lapses on age alone (drop the `lastWorkTs > declared.at` condition) | **CAUGHT** — 2 files, 4 tests |

Mutation A is the one I expected to survive, and it did not — see the first offered note below.

**Fence audit.** `git diff --name-only origin/main...HEAD` is **23 files** = the issue's 14 fence paths + the 9 files of the 6 declared widenings, exactly. Nothing under `packages/web/**`, no `reduce.ts`, `state.ts`, `fleet/buildFleet.ts`, `api/doctor.ts`, no lockfile.

**Sibling hunt, independently.** Grepped every `.declared` reader outside tests across core, server and web. `buildFleet.ts:304` sets the raw fact; `beaconFlow` in `packages/web/src/app/StatusBar.tsx:390` counts occurrences and takes min/max `at` — it makes no per-lane currency claim, so a lapsed beacon having *occurred* is still true there. `declaredClause` and `detectWaiting` are the only two places a declaration is rendered as current, and both now gate on `declarationStatus`. I also checked the claim that `beaconCapabilitiesFor` is the only constructor signing a beacon detail: `witness: 'beacon'` on a `CapabilityDetail` appears in exactly two places, both in `packages/server/src/collectors/beacon/collector.ts` (lines 60 and 78) — every other `witness` hit in the tree is the unrelated `AgentStatusWitness`. **Confirmed.**

**Follow-up 1 is real and correctly scoped.** `packages/server/src/api/doctor.ts:123` does call `checkEnrichmentLadder([...], repoPath)` with two arguments, so the route's beacon contributor stays static and the route cannot report L2. I checked whether this breaks an existing route-vs-CLI parity law: it does not — `packages/server/src/api/doctor.test.ts`'s parity tests compare specific check ids, and one of them exists precisely to assert the route and CLI *deliberately differ* on `telemetry`. Before this PR neither surface could reach L2, so this is a lag, not a regression. Agreed it is follow-up, not blocking.

**ADR renumbering is clean.** `0038` is #303's summons raiser, `0039` is this record; both README rows present and correct; the only surviving `ADR-0038` citation in the tree is in `packages/server/src/server/collector-loader.test.ts`, which is #303's own file and correctly points at #303's record.

### No must-fix findings.

### Two notes — your call, offered not asserted

**1. The constant is pinned in a second place, and the PR body understates it.** The body says a `± 1` boundary pair survives any interval and that the value is held only by the note law. Mutation A says otherwise: retuning the constant *and* the note together — which the note law cannot see — still went red, at `packages/core/src/selectors/condition.test.ts:462`, whose byte-exact assertion reads `declares working 3m00s ago`. `3m00s` is `BEACON_LAPSE_MS` written in another alphabet, so it is a second, independent pin.

That is good news for the law's strength. The reason it is worth a line: the note tells a future reader to *"re-measure and re-derive; do not tune the constant to taste"*, and a re-measurement that changes the interval will redden that assertion for a reason unrelated to its stated claim ("still names the beacon exactly at the interval"). You already carried exactly that warning in comments beside all three re-timings in `buildFleet.test.ts` — this one assertion is the same fragility without the accompanying comment. AGENTS.md's own "a path and its encoding are one edit" applies.

**2. A one-word doc inaccuracy in `packages/core/src/collector.ts`.** `AttentionWitness`'s doc comment says the field is *"Meaningful on `attention` only; the type allows it elsewhere and nothing reads it there."* But `mergeCapabilities`' new tie-break reads `best.witness === 'beacon'` generically, inside the `for (const signal of SIGNALS)` loop — so it reads it on **every** signal, not only `attention`. Nothing signs a non-attention detail today, so no behaviour differs and this is not a defect. It matters only as the sentence a future author would rely on when signing, say, `identity` with a witness, and then quietly changing that signal's merge order. "nothing reads it there" → "only the tie-break reads it, on every signal" would settle it.

Neither is worth re-triggering ten CI legs on a green, mergeable PR — offered for the next commit that touches these files.

### The ledger precondition

Your "A precondition I could not check" section is the right call and I am not going to paper over it: I cannot supply the missing evidence either. `verify-ledger.sh` is operator-local and absent from a fresh clone, so a ledger row recording the `fable` verify pass on `a2ea67c` is not something a reviewer without a configured operator environment can produce. **This carries to the operator, not past them** — AGENTS.md is explicit that a green suite, a clean fence audit and a passing mutation are each *not* condition 1. Everything above validates artefacts; none of it records that the review happened.

What I can attest is my own pass, which is an independent second reading of every commit in the range with the evidence above.

ciaran-slow:
## Independent review pass — the exact range on this PR (`98ce4d6`, lane `de87fb1`)

**This is the independent review pass AGENTS.md §"Opening a PR" condition 1 asks for**, recorded here because the operator-local ledger tooling (`verify-ledger.sh`) does not exist on this machine. The full report is on the issue: https://github.com/launchpad-26/rhizomorph/issues/218#issuecomment-5563813681.

The earlier fable verify on #218 read lane `a2ea67c`. Two fixes were then amended in by opus lanes (`a2ea67c → 7f4ce6b → de87fb1`) with no fresh independent read. This pass — on **fable**, against a commit trailed `Co-Authored-By: Claude opus` — read the whole `origin/main...98ce4d6` diff (23 files, +1431/−98), not only the delta.

### Verdict: **ready to merge.** No must-fix finding.

**The delta since `a2ea67c` is exactly what the PR body says (EXECUTED):** `git diff --stat a2ea67c de87fb1` = 9 files, 15+/14−. Fourteen lines are the `0038 → 0039` renumbering of *this* ADR (rename + H1, README row, CHANGELOG link, nine code citations across `collector.ts`, `collector.test.ts`, `api/meta.ts`, `api/meta.test.ts`, `beacon/collector.ts`); the fifteenth is `readFileSync(NOTE_PATH, 'utf8').replace(/\r\n/g, '\n')` with its comment in `beacon/collector.test.ts`. Nothing else. `git diff a2ea67c de87fb1 -- docs/design-notes packages/core/src/selectors/lapse.ts` is **empty** — `BEACON_LAPSE_MS = 180000` and the note's raw capture are byte-identical to what the prior pass recomputed. No `0038` reference to this ADR survives; #303's `0038-a-summons-raiser…` and `collector-loader.test.ts` are untouched against `main`; the README table diff against `main` is one added row.

**The wave commit is the lane's patch (EXECUTED):** `git diff HEAD^ HEAD` and `git diff de87fb1^ de87fb1` differ only in the README hunk's context (#303's row is context on the wave side); their `+`/`-` lines are **1,529 each, byte-identical**.

**The CRLF fix bites (EXECUTED):** `git check-attr` shows the note at `eol: unspecified` while `fixtures/**` is `eol=lf`, so a Windows checkout does hand the note over as CRLF. Writing the note out with 222 CRLFs and running the law's regexes verbatim: `/BEACON_LAPSE_MS = (\d+)/` matches either way (`180000`); `` /```jsonl\n…/ `` **fails un-normalised** and matches after `.replace` (77 lines) — the exact one-of-two split the windows-suite red showed. The law's assertions are unchanged in strength.

**Gates (EXECUTED, node v22.23.2):** typecheck pass · lint pass (901 files) · build pass · `VITEST_MAX_WORKERS=6 npm test` **416 files: 415 passed, 1 failed; 7713 tests passed, 1 failed, 3 expected-fail, 6 skipped; 256 s.** The one red is the pre-existing `packages/server/src/recorder/session-recorder.test.ts › holds the ceiling in a session reached by rotation` at the 5 000 ms default timeout (5 920 ms under load) — the same single red the two prior runs recorded. This PR touches nothing under `recorder/` (`git diff --name-only origin/main...HEAD | grep recorder/` is empty) · `session-recorder.test.ts` alone pass — **57/57**, 9.73 s · `gh pr checks 304` on `98ce4d6`: all 10 legs green, including **`windows-suite`**.

PR #305, opened 2026-09-06
Ci cost bench split

ciaran-slow:


gabriel-canaan:
## Review — one must-fix, then this is good work

Reviewed at `1239e0a` with `origin/main` merged in (the branch was **behind** main — see the last section). Node v22.22.2, `npm install` first.

**The headline measurement reproduces, and the saving is almost exactly as claimed.** Measured `--maxWorkers=1`, one file at a time, on the same box minutes apart — **EXECUTED**:

| shape | file | tests | test time |
|---|---|---|---|
| before | `streamState.test.ts` (bench included) | 29 | **155.47 s** |
| after | `streamState.bench.test.ts` | 1 | 56.34 s |
| after | `streamState.test.ts` (cheap laws) | 28 | 2.59 s |
| | **after, total** | **29** | **58.93 s** |

155.47 s → 58.93 s, a **96.5 s saving**. Your claim was 141.65 s → 42.22 s, a 99.4 s saving — my box is slower on both sides, but the *saving* lands within 3 s of yours and the test count is conserved exactly (29 = 1 + 28).

**"Every `expect` is byte-identical to before" — CONFIRMED mechanically.** Extracted every `expect(` from the base `streamState.test.ts` and from both post-split files, stripped line numbers, sorted, diffed: **87 vs 87, zero difference.** Nothing added, nothing dropped, nothing reworded.

**`ROUNDS` 3 → 1 is safe, and I checked the margins rather than taking the ratio on trust.** With one sample the median is the sample, so the growth-shape law is what carries the risk. Its assertions are `rows[1].beforeMs > rows[0].beforeMs`, `rows[2].beforeMs > rows[1].beforeMs`, and `afterGrowth < beforeGrowth`. Against the recorded table the monotonicity steps are ~21x and ~33x, and the growth comparison is ~13.6 against ~710 — a **52x margin**. Both directions of noise on the tiny `after` numbers were checked: an unrealistically fast `after[0]` gives ~122, an unrealistically slow one gives ~1.2, and both stay far under 710. A single sample cannot flip any of the three.

**Gates on the merged tree, all EXECUTED:** typecheck pass, lint pass, `VITEST_MAX_WORKERS=6 npm test` → **416/416 files, 7665 passed**, 3 expected-fail, 2 skipped, **0 failed**.

**Every CI figure in commit 2 checks out.** I re-derived them with the commands the paragraph now names, rather than reading the paragraph:

- `CI` on `main`, last 12 successful runs: **8–10 min** (claim: 9–10) ✓
- `Windows suite`, last 10: **12–15 min** (claim: full bar 13–15) ✓
- `Windows suite` first run: **2026-09-03T02:12:32Z** — the claimed join date, exactly ✓
- "it, not `CI`, is what a PR actually waits on": 13 vs 9 ✓
- the arithmetic: 10 min / 21 h = **0.79%**, and 20 min / 21 h = **1.59%** — both as stated ✓

Replacing a stale number with the command that re-derives it is the right shape of fix, and keeping the instruction while re-founding it on the arithmetic rather than the figure is the part I'd have argued for anyway.

**No law pins timing-set membership or count.** I checked the one that could: `gate-honesty-law.test.ts` extracts gate.sh's real discovery block and runs it, but asserts only that *it itself* is not enrolled, plus a non-vacuity check. Adding a file passes it. Verified by running it.

---

### Must-fix — the file is enrolled in the timing set **twice**, and the commit's stated mutation is therefore false

`packages/web/src/app/streamState.bench.test.ts:54` explains the mechanism and quotes the marker literally:

```
 * `scripts/gate.sh` derives its timing set from a `// @gate-timing` marker or a
```

`gate.sh` discovers markers with `grep -rlF '// @gate-timing'` — a **fixed-string content grep**, which does not care that the text sits inside a prose docstring. So the file is enrolled by its name *and* by an accidental marker match.

**This is the exact failure `gate-honesty-law.test.ts`'s own docstring records happening to that file** ("a fixed-string content grep does not care whether the text it matches sits inside a JS string literal — it matched this very file"). The repo has been bitten by this before and wrote a law about it; the law is scoped to the one file, so it does not see this new instance.

The commit message says:

> What it will NOT survive is losing the `.bench.test.ts` name: gate.sh's timing pass then matches one fewer file and its ratchet fires on the drop.

**EXECUTED — that is not what happens.** I renamed the file to `streamState.perf.test.ts`, ran gate.sh's real discovery, and restored (restore asserted with `git diff HEAD --quiet`, not announced):

```
timing-set count after losing the .bench name: 7 (was 7)
```

The ratchet does **not** fire. The safety net the commit claims does not exist — the prose keeps the file enrolled on its own. Today's behaviour is still correct (the file belongs in the set and is in it), so this is not a live bug; what is wrong is that the stated protection is absent, and the failure it would catch is the #209 trap the ratchet exists for.

**The fix is one line** — drop `// ` from inside the backticks. The sentence reads identically to a human and stops matching the grep:

```diff
- * `scripts/gate.sh` derives its timing set from a `// @gate-timing` marker or a
+ * `scripts/gate.sh` derives its timing set from a `@gate-timing` marker or a
```

**Verified, both directions:**

- with the fix, the marker grep matches only the three `scene/` files again, and `TCOUNT` is still **7** (via the name) — so the ratchet floor is unchanged and nothing surprising happens at landing;
- with the fix, renaming away from `.bench.test.ts` now drops `TCOUNT` to **6** and the ratchet fires `timing-regression`, exactly as the commit message claims;
- `gate-honesty-law`, `runbook-delivery-law`, `doc-citation-law`, `streamState.test.ts` and the bench itself: **5 files, 290 passed** with the fix applied.

### Offered, not asserted

`gate-honesty-law.test.ts`'s protective test asserts that *one named file* is not falsely enrolled. AGENTS.md's own lesson from the fixture-hygiene guard applies — *scope a guard by what a file is, not by what it is called*. A general form ("no file is enrolled by both mechanisms", or "every marker match is intentional") would have caught this instance the moment it was written. Worth a follow-up issue rather than this PR.

### Process — this needs the operator, not the author

Four things AGENTS.md asks for that this PR does not have. The *content* is not missing — commit `c61a14c`'s body carries a genuine sibling-case answer (`panels/ledger/perf.test.ts` and `StreamContext.test.tsx`, left alone with measured reasons) and a mutation answer, and they are good. They are just not where the rules put them:

1. **The PR body is completely empty** — no `Closes #`, no sibling case, no mutation question, no fence widening. All of it exists in the commit messages.
2. **No issue exists for this work.** Nothing in the 28 open issues covers it. Every issue must carry a milestone, and whether this is `charter-laws` or a small PRD is a ruling I should not make.
3. **Branch and title are off-convention** — `ci-cost-bench-split` / "Ci cost bench split", rather than `<issue>-<slug>` and `<area>: <what becomes true>`.
4. **The branch was behind `main` when CI ran.** The 10/10 green bar on this page was measured against a tree without #303 on it. I merged `origin/main` in and re-ran everything above green, so the risk is retired — but the green bar itself is not evidence for the tree that will land.

None of (1)–(3) is something a reviewer should paper over by writing the body themselves: doing that would make the PR *look* compliant while the tracker gap — the substantive part — stayed open.

---

**Verdict: request changes**, on the one-line marker fix alone. Everything else here is measured, reproduced and correct, and the two commits are unusually well argued. Once the line lands and the tracker question is settled, this should go straight in.

gabriel-canaan:
## Fixed, not handed back — `8432d6b`

The must-fix above is pushed to this branch as `8432d6b`, per the reviewer-fixes mandate. Fast-forward, no force-push; your two commits are untouched.

**What changed:** one word in the docstring — `` `// @gate-timing` `` → `` `@gate-timing` `` — plus a note beside it saying why it is spelled without its slashes, so the next editor does not helpfully restore them and silently re-enrol the file.

**Both directions re-verified after the commit:**

- the marker grep matches only the three `scene/` files again;
- `TCOUNT` is still **7** via the `.bench.test.ts` name, so the ratchet floor is unchanged and landing sees no surprise;
- renaming away from `.bench.test.ts` now drops `TCOUNT` to **6** and fires `timing-regression` — your commit message's stated mutation is true as written.

**Gates on the pushed commit:** typecheck pass, lint pass, and `gate-honesty-law` + `runbook-delivery-law` + `doc-citation-law` + `streamState.test.ts` + the bench itself — **5 files, 290 passed**.

**What I did not run on this exact commit:** the full suite. I ran it green (416/416 files, 7665 passed) on the pre-fix tree with `origin/main` merged in; this commit changes one comment line in a test docstring on top of that, and the five files that could plausibly read it are the ones above. CI's ten legs are the real check and are running now.

This is a fix to the mechanism your PR is about rather than a typo, so I am **not** approving my own repair — it wants your confirmation, and the tracker questions in the review (the empty body, and that no issue exists for this work) are still open and are the operator's call.

gabriel-canaan:
**CI on `8432d6b`: 10/10 green.** Closing the loop on my note above — the full suite is now witnessed on the fix commit across all three jobs and ten legs, each with a real duration rather than a skip: `build-test-boot` ×3 (4m41s / 8m22s / 8m57s), `pack-smoke` ×6 including both Windows legs, and `windows-suite` at 12m16s.

That is the check I said I had not run locally on this exact commit. Nothing further outstanding from my side; the fix wants your confirmation, and the tracker questions are the operator's.

Issue #307, opened 2026-09-07
prd27 w5: the wave follow-ups land — the PRD and decisions log tell the truth, the doctor route reads L2, beacon is a connection source


ciaran-slow:
## Fence (may touch ONLY)
- `docs/prds/prd-27-the-declared-voice.md` (the `> **Outcome:**` header block only)
- `docs/architecture.md` (the `## Decisions log` list only — three appended entries)
- `packages/server/src/collectors/beacon/fixtures/CAPTURE.md`
- `packages/server/src/cli/doctor.ts` (`checkDeclaredAttention` split into a pure fold-reader plus the file-reading wrapper; nothing else)
- `packages/server/src/cli/doctor.test.ts`
- `packages/server/src/api/doctor.ts`
- `packages/server/src/api/doctor.test.ts`
- `packages/core/src/selectors/connection.ts`
- `packages/core/src/selectors/connection.test.ts`
- `packages/web/src/app/StatusBar.tsx`
- `packages/web/src/app/StatusBar.test.tsx`
- `packages/server/src/api/meta.test.ts` (the connection-sources loop only)
- `packages/server/src/recorder/session-recorder.test.ts` (one per-test timeout only)
- `CHANGELOG.md`

## Blocked by
Nothing — #304 (wave 4) landed at `3f36355`; every item below is a follow-up recorded on the wave-2, wave-3 and wave-4 PRs (#287, #296, #304).

Sibling lanes: **#213** (prd14 w1, Ready, not live) claims `docs/architecture.md` and `CHANGELOG.md` — this issue appends three entries to the decisions log and one changelog line; whichever lands second rebases a list tail. **#220** (prd30 w1, Backlog) owns the native-`title` removal that the STATE-hover screenshot waits on (below). No live issue claims any path here (checked 2026-09-07 against every In-progress/In-review fence).

**Model:** opus. **Wave:** 5 — a cleanup wave: nothing here changes a ruling; every line closes a gap a wave PR named.

## Why

Four waves of prd-27 landed with follow-ups recorded in their PR bodies rather than fixed in place, because each was outside the fence of the issue that found it:

- **The PRD lies about itself.** Its Outcome header still reads *"Two of four Success criteria hold; the hook beacon and the capability-version mismatch voice remain the build."* The hook beacon landed in #282; declared attention, the disagreement voice, the lapse and the L2 rung landed in #283/#290/#218. A reader groomed from that header would re-plan shipped work.
[...]

Issue #306, opened 2026-09-07
prd17 w7: six line citations in gate.sh point at the wrong lines (consolidated into #292)


gabriel-canaan:
Five absolute line citations in `scripts/gate.sh`'s comments point at unrelated
comment lines, and one relative offset is spelled out in words so no regex over
digits can see it. All six predate #274's branch — they are byte-identical at
`d804b5a` — and were surfaced by the round-3 re-review of #274 while checking
the `N lines up` law that commit added.

Verified at `d804b5a`, each cited line read back off the file:

| citing line | cites | what that line actually is |
|---|---|---|
| `:354` | `:41` | a comment line |
| `:359` | `:116` | a comment line |
| `:440` | `:142` | a comment line |
| `:465` | `:96` | a comment line |
| `:472` | `:74-80` | comment lines |

`:13`, `:14` and `:18` are the three that are still correct (`set -uo pipefail`,
the `PATH` export, `cd "$root"`), which is what makes the rot easy to miss: spot
-checking one citation says "fine".

`:440` also carries **"eighteen lines further down"** — the same offset class as
the `110 lines up` #274 removed, but spelled as a word, so the law added in
`982a71a` cannot match it.

## Why this is its own issue and not part of #274

#274's round-3 commit narrowed its law's own claim to the one spelling it
actually checks, and recorded these six in the test's doc comment rather than
sweeping them. Re-deriving five citations means reading what each one *meant* —
which producer or guard it was pointing at — and that is a change with its own
reasoning, in a file another lane may hold.

## Done when

- Each of the five absolute citations either resolves to the line it means, or
  is replaced by a quoted sentence, whichever the surrounding prose needs.
- `"eighteen lines further down"` is re-derived or replaced by a quotation.
- The `N lines up/down` law in `gate-honesty-law.test.ts` is widened to the
  forms it currently misses — `gate.sh:NNN`, `line NNN`, `lines N-M`,
  `N lines further/prior/back`, and the spelled-out numbers — **in the same
  edit** as the fixes above, so it never ships red.
- A mutation per widened form, certified, showing the law reddens on each.

## Fence

- `scripts/gate.sh`
- `packages/server/src/gate-honesty-law.test.ts`

Out of scope: the producer doc comment's own 20 citations, which `982a71a`
re-derived and now holds with a set-equality law.

gabriel-canaan:
[...]

PR #308, opened 2026-09-07
fix(doctor): the wave follow-ups land — the route reads L2, beacon is…

ciaran-slow:
… a connection source, the PRD tells the truth  (#307)

Four waves of prd-27 landed with follow-ups recorded in PR bodies rather than fixed in place, because each was outside the fence of the issue that found it. This closes them; nothing here changes a ruling.

`GET /api/doctor` could not say L2. #218 gave `checkEnrichmentLadder` an optional `declared` parameter so the route would keep compiling, and the route passed nothing — so a beacon-only machine read L4 from the route and L2 from `rhizomorph doctor`, two surfaces answering the same question two ways. `ServerDoctorOptions.foldSoFar` is the seam that closes it: a thunk, read live per call because `ctx.recorder` is re-pointed on retarget (prd-20 ruling 5). The per-lane `attention` readings come with it, phrased by a pure `declaredAttentionChecks(state, now)` extracted from `checkDeclaredAttention` and shared by both surfaces, so neither can word a reading its own way. During replay the route emits `notApplicableDuringReplay('attention')` and never calls the thunk at all.

`beacon` joins `CONNECTION_SOURCES` and `Connection`, its flow folded from `state.declared` — one standing declaration per lane, its `at`. #283 read that flow locally in `StatusBar.tsx` because the constant was outside its fence; that local `beaconFlow` is gone and every pill on the bar now reads `connection[source]`, so the connect page, `/api/meta`'s `connection` block and the Beacon pill can no longer disagree about how many sources exist. `web/src/connect/meta.ts` iterates the constant and gained the source for free, with its tests unedited.

`CAPTURE.md` step 4 named `S-Tab`, which tmux accepts and silently ignores; `BTab` is the key that cycles Claude Code off auto-accept, so the recipe as written yielded a capture with no `Notification` line — exactly the failure #282's law now catches.

`session-recorder.test.ts`'s rotation ceiling law folds 75,003 events and takes 4–10 s alone on an 8 GB machine, past vitest's 5 s default; it carries a `30_000` per-test timeout with the measurements named. It is a count law, not a timing law, and the file's stance (no `@gate-timing`, no `.bench`) is unchanged.

The PRD's Outcome header claimed two of four Success criteria and named the hook beacon as unbuilt; it now states what is true at `3f36355`, naming each wave's sha, and marks ruling 6's remainder ungroomed. The decisions log gains ADR-0036, ADR-0037 and ADR-0039, dated by each ADR's own `Date:` line.

Fence widened twice, both forced by a test that cannot pass as written: `packages/server/src/server/build-app.test.ts` asserts `/api/meta`'s body by exact equality and gains the `beacon` block, and `.citation-prior-tracker` gains `CHANGELOG.md  #307` (with its `violations=` count) because the changelog entry cites an issue above the live tracker maximum, the same row #305 and #306 already carry.

Co-Authored-By: Claude opus


gabriel-canaan:
## Review — one gap, fixed on the branch; nothing else blocking

Reviewed at `70a395a` with `origin/main` merged in (the branch was behind — #304 and #305 landed after it was pushed). Node v22.22.2, `npm install` first.

### Fixed on the branch: the sibling call site — `8144d99`

`registerDoctorRoute` builds the prober **twice** — once at registration, once when a retarget re-points `ctx.repoPath` — and this PR correctly added `foldSoFar` to both. Only the first was pinned.

**EXECUTED.** I dropped `foldSoFar` from the retarget rebuild alone and ran the file:

```
Test Files  1 passed (1)
      Tests  38 passed (38)
```

The whole suite stayed green with the seam gone. What that allowed: a `/api/doctor` that reports L2 and the per-lane attention lines until the first retarget, then silently reverts to the static manifest for the rest of the process's life — the exact defect this PR closes, re-opened in the one path the route exists to survive. The existing `the invalidation hook` test pins `repoPath` through that same rebuild, so the path is covered; it just cannot see a second field added to the same literal.

`8144d99` adds the test. **Certified:** with it, the same mutation reddens **exactly one test** (1 failed | 38 passed) — it kills the test written for it and nothing else. Test-only, no production change; it reuses this describe's own `recorderWith` and `rungIn` helpers rather than pinning a stale ladder string, and mirrors the existing retarget test's shape. Tree restored and re-verified clean after every mutation run.

### A concern I chased and withdrew — recorded because the next reader will have it too

The extraction moved `buildFleet` out of `checkDeclaredAttention`'s `try`: on `main` it sat inside, and it now runs inside the pure `declaredAttentionChecks`, called at `doctor.ts:761` outside the guard. I mutated `buildFleet` to throw and the error propagated straight out of `runDoctor`, killing 15 tests including `session boundary check` ones with nothing to do with attention — while `an unreadable newest session warns instead of throwing` stayed green, because it injects at the read seam that is still inside the try.

That looked like a narrowed guard with a test that could no longer see the narrowing. **It does not hold, and the PR is right:**

- `buildFleet` has **no throw site** — no `throw` and no non-null assertion anywhere in `fleet/buildFleet.ts` or `fleet/diagnose.ts`. My mutation injected a throw into a function that cannot produce one, which is manufacturing a defect rather than finding one.
- #218's rationale said "`reduceAll`/`buildFleet` can throw"; the updated test docstring narrows that to "the read itself and the fold over it", which is **more** accurate, not less.
- The docblock's "Never `fail`" is about the `DoctorCheck` status vocabulary, not about exceptions, so nothing there is contradicted. I misread it first time.

Recording it because the diff genuinely looks like a narrowed guard until you check `buildFleet` for throw sites.

**Gates on the merged tree, EXECUTED:** typecheck pass, lint pass, `VITEST_MAX_WORKERS=6 npm test` → **417/417 files, 7727 passed**, 3 expected-fail, 2 skipped, **0 failed**. `gh pr checks`: **10/10**, `windows-suite` included at 13m59s.

**Fence exact: 16 files = the issue's 14 fence paths + the 9→2 files of both declared widenings.** Nothing outside it.

**The DoD's parity test exists**, and is one of six new route tests: *"the route and /api/meta derive the same rung from the same recorder, on both machines"*. The set also covers L2 beacon-only, L4 with the rig winning the tie, no-beacon, replay-with-thunk-never-called, and no-fold-seam.

**The retarget-safety reasoning is right and the code matches it.** `foldSoFar: () => ctx.recorder.foldSoFar()` reads `ctx.recorder` *inside* the thunk, so a re-pointed recorder is picked up without the prober being rebuilt — and the doc comment says why capturing it would have re-created the staleness the paragraph above it exists to describe.

**Replay never calls the thunk.** `replay || options.foldSoFar === undefined ? null : options.foldSoFar()` short-circuits before the call, and `notApplicableDuringReplay('attention')` is emitted instead. Confirmed in the code and pinned by a test.

**The `.citation-prior-tracker` row is legitimate, which I checked rather than assumed.** `liveMaximum()` derives from merged PR numbers on `origin/main` since the reset date, so at authoring time `#307` genuinely exceeded it and needed a baseline row; the row stays valid against the frozen `recordedMaximum()` of 265. Same shape as the `#304`/`#305`/`#306` rows already there, and the `violations=` count was incremented with it.

**The PRD header's provenance checks out, sha by sha.** All four are on `main` and each merge subject names the exact PR claimed:

| sha | subject |
|---|---|
| `a36d15c` | Merge pull request **#267** from launchpad-26/217-beacon-collector |
| `29cde14` | Merge pull request **#287** from launchpad-26/prd27-w2-declared-witness |
| `f479a9a` | Merge pull request **#296** from launchpad-26/prd27-w3-fleet-believes |
| `3f36355` | Merge pull request **#304** from launchpad-26/prd27-w4-lapsed-declaration |

**The decisions-log dates match each ADR's own `Date:` line exactly** — 2026-09-04 / 09-05 / 09-06 for ADR-0036 / 0037 / 0039. ADR-0038 correctly left alone as prd-17's.

**`CAPTURE.md` fixes both places**, not just the step: line 50's `send-keys … BTab` and the prose at the recipe's foot, each carrying the reason `S-Tab` is accepted and ignored.

**`beaconFlow` is gone from `StatusBar.tsx`** and `CONNECTION_SOURCES` carries the sixth source; `web/src/connect/meta.ts` and its tests are untouched in the diff, which is the proof the widening was additive rather than an assertion about it.

### Offered, not asserted

`session-recorder.test.ts`'s `30_000` is well argued (measurements named, count-not-clock stance restated). Worth knowing it did not reproduce here either way — that law passed in my full run on this branch and in my #304 run, so on this box the headroom is untested. That is corroboration for it being machine-specific, not an objection.

**Verdict: ready once the author confirms `8144d99`.** I am not approving my own addition, and the rest of the change is measured, cited and correct.

gabriel-canaan:
**Superseded by #309 — same change, and that one is better based.**

While reviewing #309 I found it is not a similar PR but the *identical* one: `70a395a` here and `8b1dd22` there share patch-id `d47bb59caf5b`, and a `diff` over the two commit messages returns nothing. #309 is this commit cherry-picked onto a different base.

Two differences, both favouring #309:

- **branch** — this PR was opened from the lane branch `307-wave-follow-ups`; #309 uses `prd27-w5-follow-ups`, the `prd<NN>-w<N>-<slug>` form. AGENTS.md's assembly convention is never to open the wave PR from the lane branch.
- **base** — this PR sits on `3f36355`, **4 commits behind `main`**; #309 sits on `ca80fd3`, which is current `origin/main`.

Both close #307, which is the hazard worth resolving soon: whichever merges second closes nothing, and the issue may already be shut by the first.

My review above stands and applies to both by patch-id equality, and `8144d99` (the retarget fold-seam test) is cherry-picked onto #309 as `bdd7d7f`, re-certified there. So closing this one loses nothing.

Suggest closing in favour of #309 — but that is the operator's call, not mine, and I have not touched either PR's state.

ciaran-slow:
Closing as superseded by #309, per the review on both PRs: this PR was opened from the lane branch `307-wave-follow-ups` and carries the same commit as #309's wave branch (patch-id `d47bb59caf5b`, byte-identical diff and message). #309 is the wave PR the working agreement asks for (one PR = one (milestone, wave)), is based on current `main`, and already carries the review's retarget-seam test as `bdd7d7f` (this branch's `8144d99` is its byte-identical twin). The lane branch is left in place as provenance until #309 merges, as the open-wave-pr stage requires.

PR #309, opened 2026-09-07
prd27 w5: the wave follow-ups land — the doctor route reads L2, beacon is a connection source, the PRD tells the truth

### LINKED ISSUE #307: prd27 w5: the wave follow-ups land — the PRD and decisions log tell the truth, the doctor route reads L2, beacon is a connection source

ciaran-slow:
Closes #307

prd-27 wave 5 is one issue, so this bundle is one commit — the wave's follow-ups,
recorded on #287, #296 and #304 as waves 1–4 landed. Assembled onto `origin/main`
at `ca80fd3`; the lane commit is `70a395a` on `307-wave-follow-ups`.

## What became true

1. **`GET /api/doctor` reads L2 off the running recorder's fold.** `ServerDoctorOptions`
   gains `foldSoFar?: () => SessionState`, read live per call — a thunk, never a captured
   `ctx.recorder`, because retarget re-points it (prd-20 ruling 5). `registerDoctorRoute`
   passes `() => ctx.recorder.foldSoFar()` at both prober sites; `createRouteDoctorProbe`
   carries it through in `baseOptions`. The ladder call becomes
   `checkEnrichmentLadder([impliedTargetPath, ...baseChecks], repoPath, fold?.declared ?? {})`,
   and the route emits per-lane `attention` lines **before** the ladder — the order the CLI
   prints them. Absent seam → no `attention` lines and the static ladder; replay →
   `notApplicableDuringReplay('attention')` and the thunk is never called.

2. **One function phrases the readings.** `declaredAttentionChecks(state: SessionState, now: number): DoctorCheck[]`
   is extracted from `checkDeclaredAttention` in `cli/doctor.ts` — `buildFleet`, the `present`
   filter/sort, the "no present lane" line and the four-arm switch, each `now()` replaced by the
   single `now` the caller supplies. `checkDeclaredAttention` keeps its signature and its two own
   lines (no session; unreadable); its `try/catch` now wraps the read and the fold only, the fleet
   build having moved into the pure function. The route and the CLI can no longer say the same
   reading two ways.

3. **`beacon` is the sixth connection source.** `CONNECTION_SOURCES` and `Connection` gain it;
   `selectConnection` folds `state.declared` — each *standing* record contributes its `at`, count =
   records, not every beacon ever seen (the fold keeps one per lane, #283).

4. **The StatusBar stops deriving its own flow.** `beaconFlow`, its doc block and the
   `Pick<SourceFlow, …>` typing are gone; `SourceKey` is replaced by the imported
   `ConnectionSource`, `SOURCES` is `CONNECTION_SOURCES`, `sourceStatus` takes a full
   `SourceFlow`, and the pill loop reads `connection[source]` for all six. A seventh source
   added to the constant and not to the bar is now a compile error
   (`SOURCE_LABEL: Record<ConnectionSource, string>` stops being exhaustive).

5. **The capture recipe names the key that works.** `CAPTURE.md` step 4 is `BTab` with the reason
   (`S-Tab` is accepted by tmux and silently ignored — #218's measurement), and the prose at the
   recipe's foot agrees. **The ceiling law carries a per-test timeout** — `}, 30_000)` on
   `holds the ceiling in a session reached by rotation…`, with the comment naming the four
   measurements (4.2 s, 9.3 s, 9.7 s, 10.5 s alone; 5.9 s under six workers). No `@gate-timing`,
   no `.bench` rename: the law asserts a COUNT, only the harness clock moved.

6. **The PRD tells the truth.** The `> **Outcome:**` blockquote (and nothing else in that file) now
   states what is true at `3f36355`, naming each wave's merge sha; `docs/architecture.md`'s
   `## Decisions log` gains three entries for ADR-0036, ADR-0037 and ADR-0039, appended after the
   ADR-0031 entry and dated by each ADR's own `Date:` line. ADR-0038 is prd-17's and is untouched.
   **This wave adds no ADR** — `git diff --name-only origin/main...HEAD | grep docs/adr/` is empty.

## What is the sibling case?

The route and the CLI are the same check set recombined, and the defect closed here is that one of
the two gained a reading the other did not. **The sibling is the third reader of the fold:**
`/api/meta`'s ladder already took `beaconCapabilitiesFor(folded.declared)` (#218), so a fix that
gave only the route its fold would have left `/api/meta` as the next surface to disagree.
`the route and /api/meta derive the same rung from the same recorder, on both machines` asserts the
rung token in the route's `ladder` message equals `/api/meta`'s `rung` for the same
`SessionRecorder` — beacon-only (both L2) and workmux (both L4). All three readers now derive the
rung from one fold, and a fourth phrasing has somewhere to fail.

For the connection selector the sibling is `web/src/connect/meta.ts`'s `parseConnection`, which
iterates `CONNECTION_SOURCES`: it gained `beacon` for free, and **the connect page's tests passing
unedited** (`6 files, 229 tests`) is the proof the widening was additive rather than a second place
to keep in step. For the extracted function the sibling is the CLI's own two lines (no session;
unreadable) — the only readings `checkDeclaredAttention` still phrases itself, because neither is a
fact about a fold.

| mutation | the test that reddens | verified |
|---|---|---|
| route passes `{}` to the ladder instead of `fold?.declared ?? {}` | `the route reads L2 off the recorder's fold …` + `the route and /api/meta derive the same rung …` | **EXECUTED** — 2 failed / 36 passed |
| route reads the fold during replay (drop the `replay \|\|` guard) | `replay: attention is not applicable, and the fold thunk is never called` (a `reads` counter asserted `0`, and the thunk throws) | **EXECUTED** — 1 failed / 37 passed |
| `beacon` in the constant but the `state.declared` fold loop removed | `folds declared attention as beacon flow` **and** `the Beacon pill reads waiting until a lane's attention is declared, then live` | **EXECUTED** — the core test and the pill both redden, which is the proof the bar now reads the selector |
| the shared reader says `declares` instead of `declared` | 3 in `cli/doctor.test.ts` (incl. `declaredAttentionChecks phrases the four readings from a fold alone`) + 2 in `api/doctor.test.ts` | **EXECUTED** — 5 failed across both files: one wording change reddens both surfaces, so neither can drift alone |
| every beacon counted rather than the standing declarations | `folds declared attention as beacon flow` — three beacons over two lanes must read `count: 2` / `firstEventTs: 20`, not `3` / `10` | REASONED — not expressible as a mutation of `selectConnection`, which only ever sees `state.declared`; the test pins it from the outside |
| a kind outside the vocabulary counted as flow | `a beacon of a kind outside the vocabulary contributes nothing to beacon flow` | |
| a seventh source added to the constant and not to the bar | `tsc` | |
| route captures `ctx.recorder` instead of reading it in the thunk | **not covered by a test.** Argued from the code, and the reason it is a thunk is in the option's own doc; the retarget-invalidation test above it covers the `repoPath` half only. Named, not claimed. | |
| `beaconFlow` kept in `StatusBar.tsx` beside the widened constant | **not a test** — the StatusBar tests pass either way, which is exactly why the DoD required the identifier to be *gone*. `grep -n "beaconFlow\|stays five" packages/web/src/app/StatusBar.tsx` → no output, exit 1 | **EXECUTED** |
| timeout set back to `5_000` | the full-suite run on the 8 GB machine, quoted below — not unit-testable, and the honest form of that is a measured suite result rather than an assertion | |

**Survives, and known:** the PRD's Outcome header and the three decisions-log entries are prose,
and no test holds them to the tree. They name shas (`a36d15c`, `29cde14`, `f479a9a`, `3f36355` —
all merge commits on `main`, resolved by the verifier against `gh pr view` mergeCommit oids) and
issue numbers so a reviewer can check them against the history rather than against a claim.

## Deviations from the plan

1. **The L2/L4 route legs use a hoisted exec flag rather than a per-call stub.**
   `registerDoctorRoute` has no `exec` seam (`ServerContext` is outside the fence, deliberately,
   per that file's own comment), so the existing module-level `vi.mock('../server/exec.js')` gained
   `execState.workmuxMissing` — a `vi.hoisted` flag defaulting to `false`, so every pre-existing
   test in that file sees the identical healthy stub, and `teardown()` resets it. This is how the
   DoD's "disables workmux" is reached through the real HTTP route rather than only through
   `runServerDoctor`. Verifier's judgement: sound — `isMissingBinary` is
   `failed && errorMessage !== undefined`, so the ENOENT shape the flag returns is what a real
   missing binary produces.
2. **One extra fixture knob, `workmuxDisabled`, on the parity test.** The rung-parity test needs the
   beacon-only machine described to *both* readers, because they learn about an absent workmux
   through different honest seams: the route probes the binary (`checkOptionalTool`), `/api/meta`
   reads the fold's `collector.disabled`. So the L2 fixture records a `collector.disabled` for
   workmux as well as flipping the exec flag. Setting only one of the two is a state where the two
   readers legitimately disagree and neither is wrong about what it measured — the test comment
   says so.
3. **One comment in `cli/doctor.test.ts` corrected.** The "unreadable newest session" test's
   docblock said the guard covers "the fold and the fleet built on it"; the fleet build moved into
   the pure reader, so the sentence now names the read and the fold. **No assertion changed.** One
   consequence worth naming (verifier, a suspicion rather than a finding): a throw inside
   `buildFleet` would now escape `checkDeclaredAttention` rather than degrade to the warn line —
   `buildFleet` is a pure selector over a fold `reduceAll` just produced, and no input was found
   that makes it throw.
4. **Verify-fix `(old-tracker #147)`** — `docs/prds/prd-27-the-declared-voice.md:12` cited a bare
   `#147`; live `#147` exists and is an unrelated prd-44 issue (CLOSED). The PRD means the
   prior-tracker issue, as the DoD's own "old-tracker #223/#147" wording and the PRD body's dated
   `#147, 2026-08-04` show — the header qualified `#223` and not `#147`. No ruling text changed.
   Applied in the same verify-fix: the second `createRouteDoctorProbe({…})` literal in
   `registerDoctorRoute` was indented with its block (whitespace only; biome's formatter is
   `enabled: false`, so no gate reads it).

`StatusBar.test.tsx`: only the comment block above the Beacon-pill tests changed — **no assertion in
that file was edited**, and both Beacon tests pass unedited, which is the proof the widening did not
change what the pill reads. `meta.test.ts`: the connection-sources loop gained `'beacon'` and its
inline body type gained `beacon: SourceFlowForTest` (the loop cannot index the type without it).
The `excludes target-path, web-build and port` id-list test in `api/doctor.test.ts` is
**unchanged**: it calls `runServerDoctor` with no `foldSoFar`, so no `attention` id enters its list
— the plan asked which way that went, and this is it.

## Fence widenings

Both were recorded on the issue's build report before the change, and the verifier proved each was
**forced** by breaking it:

- **`packages/server/src/server/build-app.test.ts`** (one line + a two-line comment). Its
  `GET /api/meta` integration test asserts the whole body by `toEqual`, so a sixth `connection` key
  cannot pass as written. Added
  `beacon: { source: 'beacon', firstEventTs: null, lastEventTs: null, count: 0 }`. *Proof:* deleting
  that line and running the file → `× GET /api/meta reports repo and session info`.
- **`.citation-prior-tracker`** (one row + the count). `doc-citation-law.test.ts` derives the live
  tracker maximum from `origin/main`'s merge subjects, so the changelog's `(#307)` reads as a ceiling
  violation. Added `CHANGELOG.md  #307` in sorted position and bumped `violations=151` → `152`;
  `live-maximum=265` and the measured date are untouched, as that file's header demands, and
  307 > 265 so the row is a genuine violation by that file's own staleness check. Same row
  `CHANGELOG.md  #305` and `#306` already carry. *Proof:* removing the row and restoring `151` →
  `doc-citation-law` red with `[{"file":"CHANGELOG.md","num":307}]`; row present but count left at
  151 → `× the baseline file parses … measured line matches its own entry count`. So the row **and**
  the count bump are each required.

`git diff --name-only origin/main...HEAD` = 16 files = the issue's 14 fenced paths + these two.
`docs/screenshots/**`: 0 files touched.

## The changelog overlap with #216

`scripts/fence-lint.sh 307 216 274 275 235` reports one overlap involving this issue —
`CHANGELOG.md` vs #216 (prd20 w1), because #216 lists the changelog as a fence path. **Ruled at
dispatch, and it is not a fence collision:** the changelog is deliberately unclaimed by every
groomed wave (each issue writes its own `[Unreleased]` entry; the assembler reconciles adjacent
lines — `AGENTS.md`'s standing convention), so whichever lands second merges one adjacent entry and
**both entries are kept in full**. No code path overlaps. The run's other overlap
(`route-class-law.test.ts`) is #216 vs #275 and does not involve this issue.
`scripts/fence-lint.sh 307` alone: PASSED.

## Independent review of this range

- **The range was read by an independent review pass.** Builder model **opus** (commit trailer
  `Co-Authored-By: Claude opus`); verifier **fable** (`Co-Authored-By: Claude fable`) — a different
  model, as the verify skill requires. Verdict **ready to merge**, no code finding, no law weakened,
  after one two-word prose fix the verifier applied itself as a verify-fix. The pass ran all four
  gates plus the boot smoke on its own checkout and certified five mutations (the EXECUTED rows
  above), each restored byte-identical with `git status` clean afterwards.
  - verify report: https://github.com/launchpad-26/rhizomorph/issues/307#issuecomment-5564622522
  - verify-fix: https://github.com/launchpad-26/rhizomorph/issues/307#issuecomment-5564670442
  - build report: https://github.com/launchpad-26/rhizomorph/issues/307#issuecomment-5564517744
  - plan: https://github.com/launchpad-26/rhizomorph/issues/307#issuecomment-5564286094
- **No ledger row can be recorded.** The operator-local tooling `AGENTS.md` names —
  `~/.claude/skills/verify/scripts/verify-ledger.sh`, and with it `pr-open.sh` — **does not exist on
  this machine** (`ls` → No such file or directory; the whole `~/.claude/skills/verify/` directory is
  absent). It is not shipped by this repository, so `check --strict --range` could not be run and
  no `verify` seat row exists to check.
- **This section is the honest escape `AGENTS.md` prescribes** for exactly that gap: the review
  happened and is linked above, the ledger evidence does not exist, and the gap is written into the
  PR body where a reviewer sees it rather than inferred from a green suite. A green suite is not
  condition 1, and this PR does not claim it as one.

## Not in this wave

- **STATE-hover screenshots** wait on #220 (prd30 w1) removing the native `title` attribute —
  `docs/screenshots/**` and `packages/web/src/panels/fleet/**` are untouched here, as #307's own
  issue says.
- **The identity join, `dissent` on `Lane`, and the two lapse-boundary notes** want rulings before
  anyone builds them.
- **Ruling 6's version-mismatch voice (old-tracker #223) and authoritative cost coverage
  (old-tracker #147)** are **not yet groomed** — the PRD's Outcome header now says so in those
  words.
- **`docs/architecture.md:270`** pins `api/doctor.ts:221-222` for "the route passes
  `replay: ctx.readOnly`". It was **already stale on `origin/main`** (those lines are a docblock
  there) and this diff moves it further. Pre-existing and outside this fence (decisions log only) —
  it belongs with whichever issue next touches that narrative, in the "named rather than cited by
  line" idiom the paragraph above it already uses. Same class as #306.
- **`StatusBar.test.tsx:327`** says "over all five sources" — a historical description of #251's fix
  at the time, harmless, left as it stands.

## How the wave was assembled

**Two pre-checks first, both learned on #304:**

1. **No ADR number can collide.** `git ls-tree --name-only origin/main docs/adr/ | tail -3` →
   `0038-…`, `0039-…`, `README.md`; and the assembled diff adds nothing under `docs/adr/`
   (grep over `git diff --name-only origin/main...HEAD` → empty). This wave writes *about* ADRs
   0036/0037/0039 in the decisions log; it adds none.
2. **`main` moved, and not into this wave's files.** The lane branched at `3f36355` (PR #304).
   Since then `main` gained **four commits — PR #305, `ci-cost-bench-split`**: `ca80fd3` (merge),
   `8432d6b` (`fix(web): the bench is enrolled in the timing set by its name alone`), `1239e0a`
   (`docs(runbook): CI's stated duration is the measured one`), `c61a14c`
   (`perf(web): the #183 bench stops being the suite's most expensive file`). Those five files —
   `AGENTS.md`, `packages/web/src/app/StreamContext.test.tsx`,
   `packages/web/src/app/streamState.bench.test.ts`, `packages/web/src/app/streamState.test.ts`,
   `packages/web/src/hooks/useEventStream.ts` — intersect this wave's 16 files in **nothing**. This
   matches what the verifier saw; `main` did not move again in between.

**Assembly:** `git checkout -b prd27-w5-follow-ups origin/main` (at `ca80fd3`) then
`git cherry-pick 70a395a`. **No conflict at all** — not even the `CHANGELOG.md` one the convention
allows for, since #216 has not landed its entry yet. One commit, `8b1dd22`, preserving
one-commit-per-issue and both trailers.

**Patch identity confirmed:** `git diff 70a395a^ 70a395a` vs `git diff HEAD^ HEAD`, compared on
added/removed lines with the `+++`/`---` file headers stripped — **707 lines each, byte-identical**
(`diff` exit 0). `16 files changed, 587 insertions(+), 120 deletions(-)`, the same as the lane's.

**Gate on the assembled branch** (node v22.23.2, after `npm install`; `scripts/gate.sh` was **not**
run — running it *is* the landing, and that is the operator's step):

| gate | command | result |
|---|---|---|
| test | `VITEST_MAX_WORKERS=6 npm test` | **`Test Files 417 passed (417)` · `Tests 7723 passed \| 3 expected fail \| 6 skipped (7732)` · 274.31 s · exit 0 — FULLY GREEN, the ceiling law included** |
| typecheck | `npm run typecheck` | exit 0, all workspaces |
| lint | `npm run lint` | `Checked 902 files in 1410ms. No fixes applied.` exit 0 |
| build | `npm run build` | exit 0 (server esbuild, vite web, app — `dist/main.js` 864.5kb, `dist/preload.cjs` 1.3kb) |
| boot smoke | `npm start -- --port 0` | server up on an ephemeral port; **`GET /` 200 `text/html`**; **`GET /api/doctor`** (page capability token) 200 → **28 `attention:<lane>` lines** (`never declared — attention read from the other organs (rung below); hooks: …`) **ordered before** the single `ladder` line (`this repo sits at L3 — PTY wrapper — next: …`); **`GET /api/meta`** 200 → `connection` carries the **six sources in `CONNECTION_SOURCES` order with `beacon` sixth and last of the sources** — `git,tmux,workmux,sessionlog,otel,beacon` — followed by the two non-source keys `uninstrumentedSessions,refusals`; `connection.beacon = {source:'beacon', firstEventTs:null, lastEventTs:null, count:0}`. Without the token both routes 401. |

The full-suite line is the point of item 5: this is the 8 GB Apple Silicon machine the DoD names,
and the suite is **fully green on it** — the single red of prd-27 waves 3 and 4 is gone. The file
count is 417 rather than the lane's 416 because `main` gained `streamState.bench.test.ts` in PR #305.

**One pre-existing seam difference the smoke shows, and it is not this wave's:** `/api/meta` reported
`rung: L1` while `/api/doctor`'s ladder reported `L3` on the same boot. The two readers learn about
workmux through different honest seams — the route probes the binary, `/api/meta` reads the fold's
`collector.disabled`, and nothing had declared a disabled collector on this fresh boot. It is
documented in `api/meta.ts:130–131`, the verifier saw the same thing, and deviation 2 above is the
test-side answer to it.

Lane branch `307-wave-follow-ups` and its worktree are left in place as the commit's provenance.
Nothing was merged; the merge is manual, on GitHub, after review.


gabriel-canaan:
## Review — this is #308 again, and it is the one to keep

**#309 and #308 are the same change.** Not similar — identical:

```
8b1dd22  (this PR)  patch-id d47bb59caf5b
70a395a  (PR #308)  patch-id d47bb59caf5b
```

Byte-identical diffs, and `diff` over the two commit messages returns nothing either. #309 is `70a395a` cherry-picked onto a different base. Both carry `(#307)` and both would close the same issue.

**This one is the better-formed of the two**, on both counts the runbook cares about:

| | #308 | #309 |
|---|---|---|
| branch | `307-wave-follow-ups` (the **lane** branch) | `prd27-w5-follow-ups` — `prd<NN>-w<N>-<slug>` ✓ |
| base | `3f36355`, **4 commits behind `main`** | `ca80fd3` — **current `origin/main`** ✓ |

AGENTS.md's assembly convention is *never open the wave PR from the lane branch*, which is what #308 did; and a branch that is behind is not the tree that lands. So my recommendation is: **land #309, close #308** with a note pointing here. Two open PRs closing one issue is its own hazard — whichever merges second closes nothing, and the issue may already be shut by the first.

### The content review is on #308 and carries over verbatim

Posted at https://github.com/launchpad-26/rhizomorph/pull/308#issuecomment-5564795704. I am not re-deriving it here, and the patch-id equality above is the licence for that — it is the same bytes. In summary it verified: the fence exact at 16 files; the DoD's route↔meta parity test present among six new route tests; the retarget-safe thunk reading `ctx.recorder` inside the closure; replay never calling the thunk; the `.citation-prior-tracker` row legitimate against `liveMaximum()`; all four shas the PRD header cites on `main` with merge subjects naming exactly the claimed PRs; and the three decisions-log dates matching each ADR's own `Date:` line. It also records a concern I chased and **withdrew** — the `buildFleet` guard scope — because `buildFleet` has no throw site, so the narrowing is correct and my first reading of it was wrong.

### The one gap, re-verified here and fixed — `bdd7d7f`

I did re-run the finding against **this** branch's own tree rather than inheriting it. `registerDoctorRoute` builds the prober twice — at registration, and again when a retarget re-points `ctx.repoPath` — and this PR adds `foldSoFar` to both. Only the first was pinned.

**EXECUTED on `8b1dd22`:** dropping `foldSoFar` from the retarget rebuild alone left `api/doctor.test.ts` at **38/38 passed**. The gap allows a `/api/doctor` that reports L2 and the per-lane lines until the first retarget, then silently reverts to the static manifest for the life of the process — the defect this PR closes, re-opened in the one path the route exists to survive.

`bdd7d7f` cherry-picks the test onto this branch. **Certified here:** the same mutation now reddens **exactly one test** — `1 failed | 38 passed` — killing the test written for it and nothing else. Tree restored and verified clean after every run. Test-only; no production code changed.

### Gates on this branch, with the test — all EXECUTED

| gate | result |
|---|---|
| `npm run lint` | pass |
| `VITEST_MAX_WORKERS=6 npm test` | **7728 passed**, 3 expected-fail, 2 skipped, **0 failed** |

For what it is worth, #308's CI came back **10/10 green** on the identical change plus the identical test, `windows-suite` included.

**Verdict: ready once the author confirms `bdd7d7f`** — I am not approving my own addition — and once one of the two PRs is closed. Everything else here was measured on #308 and re-measured on this tree.

PR #310, opened 2026-09-07
prd17 w5: the gate's verdict reaches the beacon record

### LINKED ISSUE #274: prd17 w5: the gate's verdict reaches the beacon directory

gabriel-canaan:
Closes #274

prd-17 wave 5. The landing gate already printed a `v1` verdict line; this makes
that same line reach `gate.jsonl` in the beacon directory, so a landing shows up
in the record rather than only in the operator's terminal.

## What is here

Four commits. The first is the feature; the other three are review repairs, each
answering a round of independent seats.

| | |
|---|---|
| `eb7009d` | the verdict line also reaches `gate.jsonl` in the beacon directory |
| `c424bc8` | the write cannot hang, silently drop a special-char repo, or corrupt the next landing |
| `982a71a` | the append flushes inside its lock, the tsx pin closes the launcher class, no comment cites by offset |
| `0b07ea8` | the producer citations are held by the file, the pins match statements not mentions, the offset law claims only what it checks |

## What the review rounds actually found

Four rounds ran. **Every finding after round 1 was in a repair written during
review, not in the built feature** — which is worth knowing before reading the
diff, because it is where the risk in this PR sits.

- **Round 1** (2 seats): `npx --no-install tsx` falls back to an unbounded
  registry lookup — MEASURED 70 s against a connection-refused registry and
  still running past five minutes against a black-holed one, inside the
  function whose own comment says *"A landing may fail; it may not hang."* Plus
  a torn append fusing the next landing's JSON onto a partial prefix and losing
  **both** verdicts, and `import()` on a bare path failing silently on `%`,
  `#`, `?`, LF and CR repo paths.
- **Round 2** (1 usable seat): the round-1 repair existed **only inside a merge
  commit's conflict resolution**, so `scripts/gate.sh`'s own first step — a
  plain rebase — deleted the fix *and its tests together*, leaving the suite
  green, the fence audit green and `merge-base --is-ancestor` green while all
  three round-1 defects returned to `main`. The branch was flattened in
  response; it is now four ordinary commits and the tree is byte-identical to
  the pre-flatten head.
- **Round 3** (1 usable seat): the anti-`npx` guard pinned a *token*, not the
  property — `npm exec --offline --yes -- tsx` slipped it and hung identically.
  The append flushed at `close()`, i.e. **after** `LOCK_UN`. Four comments still
  justified the mechanism round 2 had removed.
- **Round 4** (1 usable seat): the round-3 commit **staled 17 of the 20 producer
  citations** while its own body claimed the sibling sweep had found none left —
  the third recurrence of that one class. Its new pins were also weaker than
  claimed: a trailing `#` comment inside the embedded Python defeated the flush
  order pin, and a launcher that *prefixed* the resolved path slipped the tsx
  pin.

## What is the sibling case?

Named per round in each commit body. The one worth repeating here: the producer
doc comment's 20 line citations had staled three times because the *counts* were
pinned and the *citations* never were. They are now extracted from the test
file's own source between markers and compared **as a set** against the
producers found in `scripts/gate.sh`, so a line added anywhere above a cited
producer reddens the build in the commit that adds it. The two marker guards are
load-bearing: without them, deleting a marker makes the law `expect([]).toEqual([])`
and it passes having checked nothing.

**Not covered, deliberately:** the other line-number citations in that file
(`:96`, `:82`, `:24`, `:57`, `:74/:77`, `:100`) and six pre-existing wrong
citations in `scripts/gate.sh` itself. Filed as #306 rather than swept in —
widening the law without fixing them would ship it red.

Seven, each applied to the real committed files, run, reverted, and certified.
Three of them were **green** before the final commit and are the reason it
exists:

| | mutation | |
|---|---|---|
| M1 | resolved binary → `npm exec --offline --yes -- tsx` | RED |
| M2 | delete `f.flush()` | RED |
| M2b | move `f.flush()` after the unlock | RED |
| M3 | reintroduce an `N lines up` citation | RED |
| M4 | delete `f.flush()`, leave a trailing comment naming it | RED *(was green)* |
| M5 | `npx --no-install "<resolved path>" -e` | RED *(was green)* |
| M6 | insert one line above a cited producer | RED *(was green)* |

M6 models what actually went wrong three times — an unrelated edit shifting
lines under a paragraph nobody re-read — and nothing in the suite could see it
until now.

## Fence widening

`packages/server/src/gate-honesty-law.test.ts` was added to #274's fence during
round 2 and recorded on the issue at the time, via `fence-widen.sh`. No further
widening in rounds 3 or 4: every change since is inside `scripts/gate.sh` and
that file.

## Gates

Node 22.22.2, re-derived on `0b07ea8`:

- `npm run lint` — 897 files, no fixes applied
- `VITEST_MAX_WORKERS=6 npm test` — **414 files, 7653 passed**, 3 expected fail, 2 skipped
- `packages/server/src/gate-honesty-law.test.ts` alone — 214 passed, 1 pre-existing skip

## What a reviewer should look at hardest

`0b07ea8` — it is the one commit in this PR that **no independent seat has
read**, for the reason recorded below. It is test-only, and its own claims are
the ones most worth checking: that the citation law cannot pass vacuously, and
that the narrowed offset law's name now matches what it checks.

---

## Declared review gap

> 0b07ea8 has no independent review record: it answers round 4's findings and was written after the last seat pass, and rereview-budget.sh refuses a further round (cap 1 — every finding that round was in a test file). The other three commits are recorded in the ledger with their seat counts and verdicts. Two of the four rounds ran with only one usable seat: round 2's Codex seat died when the 5h rolling window hit 100%, and round 4's hit a host SIGTRAP on every command and correctly refused to return a verdict rather than fabricate one.

```
  UNVERIFIED  0b07ea8  test(gate-honesty-law): the producer citations are held by the file, the pins match statements not mentions, the offset law claims only what it checks (review of #274, round 4)
  ok      982a71a  fix(gate): the beacon append flushes inside its lock, the tsx pin closes the launcher class, no comment cites by offset (review of #274, round 3)
  ok      c424bc8  fix(gate): the beacon write cannot hang, silently drop a special-char repo, or corrupt the next landing (review of #274, round 2)
  ok      eb7009d  feat(gate): the gate's verdict line also reaches gate.jsonl in the beacon directory (#274)
──
Run the seats, then:  verify-ledger.sh record --sha <sha> --seats <n> --verdict <text>
Or record why review is not needed:  verify-ledger.sh record --sha <sha> --force "<why>"
A forced row is visible here and in any PR body that cites this check.
```


ciaran-slow (approved):
Reviewed commit-by-commit in a clean clone of the PR head (node v22.23.2, macOS). One finding, fixed in `27e695e` on this branch; everything else verified in your favour.

## What verified — EXECUTED, in a clean clone

- `npm run typecheck` clean across all workspaces; `npm run lint` 897 files, no fixes applied — matching your figure exactly.
- `VITEST_MAX_WORKERS=6 npm test`: **414/414 files, 7649 passed, 3 expected fail, 6 skipped.**
- `gate-honesty-law.test.ts` alone: 215 passed.
- All ten CI legs executed and passed — `gh pr checks 310` shows `build-test-boot` ×3, `pack-smoke` ×6 including both Windows legs, and `windows-suite`. Checked against the three-job/ten-leg list in AGENTS.md rather than the green bar.
- Fence is clean: the diff is exactly `scripts/gate.sh` + `packages/server/src/gate-honesty-law.test.ts`, and the second is recorded on #274 with a widening comment dated before the change. Nothing else is touched in any of the four commits.

**The feature reaches a real reader, end-to-end** — the thing most worth checking, and it holds:

- ADR-0036 specifies `<writer>.jsonl` per writer; you write `gate.jsonl` and the line carries `"writer":"gate"`. Consistent, and `collector.ts` tails every `*.jsonl` (`BEACON_FILE_SUFFIX`).
- **The DoD's fresh-install case genuinely works.** `createBeaconCollector`'s `poll` re-runs `readdir` on the beacon directory *every tick* and treats ENOENT as the ordinary empty state, so a `gate.jsonl` created by a first landing after your `os.makedirs` is discovered on the next tick rather than missed by a one-shot enumeration. Worth stating explicitly because a collector that enumerated once at startup would have made the whole mkdir requirement pointless, and nothing in the PR asserts this end of the contract.
- The sibling case holds as claimed: the write sits inside `emit_gate_verdict`, reached from `fail()`, from `emit_gate_verdict clean` and from `emit_gate_verdict push-failed`, so holds and merges both record. The `GATE_OUTFILE=""` at the end also makes a second call a no-op return, so the `PUSHED` branch cannot double-emit.
- I checked `push_or_warn`'s `git push origin main 2>&1 | tail -1 || …` for the pipeline-status trap that would make `push-failed` dead code and silently record a local-only landing as `clean`. It is not a defect — `set -uo pipefail` at `gate.sh:13` makes the pipeline carry git's status. Reporting the check because a reader of that function will wonder, and it is outside this PR's range anyway (`3868be0`).

## One finding — the round-4 tsx pin (fixed in `27e695e`)

`EXECUTED.` The test named *"the ONLY program this machinery invokes is the resolved tsx binary — **a launcher of any spelling fails here**"* stays green on a launcher.

Round 4's finding 3 replaced the suffix pin with an anchor at the start of the invocation line, on the stated property that the pin "must hold the whole command word". An anchor is still a *position* pin, and a `\` continuation puts the command word on a different physical line from the `-e` it anchors to. I extended the continuation already at `gate.sh:244` rather than inserting a line — so the line count is unchanged and `0b07ea8`'s producer-citation law could not mask the result:

```
beacon_dir=$(REPO_PATH="$root" MODULE_PATH="...paths.ts" npx --no-install \
  "$root/node_modules/.bin/tsx" -e '
```

- before `27e695e`: **215/215 GREEN**, with `npx` genuinely in front of the binary
- after: **RED**, 1 failed | 214 passed, the message naming the command it found

**Exposure is nil, and the claim was the defect** — the same shape you recorded for round 4's own finding 3, so I re-measured rather than assuming it carried over. From a scratch directory with no `node_modules` above it, against a connection-refused registry (`npm_config_registry=http://127.0.0.1:9/`): bare binary `rc=0 387ms`, `npx --no-install <resolved path>` `rc=0 538ms`, `npm exec -- <resolved path>` `rc=0 343ms`. Handed an absolute path, neither launcher consults the registry, so no prefixed form reaches the unbounded lookup round 2 measured at 70s. Nothing hangs; a test whose own name asserts more than it checks is what is repaired.

The fix stops asking *where* the program sits, since that is the third spelling of one defect (round 3: the token adjacent to `-e`; round 4: the start of its physical line; this: any earlier physical line). Continuations are joined into one logical line, the `\w+=$(` opener and every leading `VAR=value` environment assignment are stripped — neither is a command word — and what remains *begins* with the program that runs. `toHaveLength(1)` is kept as the anti-vacuity guard. Verified there is exactly one `\`-continuation in the whole `VERDICT_MACHINERY` slice, so the join cannot fuse unrelated statements, and the `\` in the embedded Python's `b"\n"` is followed by `n`, not a newline.

No fence widening: `gate-honesty-law.test.ts` is already in #274's recorded fence, and `scripts/gate.sh` is untouched by my commit — the mutation was reverted and `git diff` against `0b07ea8` is empty for both files.

- **The `tsx` invocation is the one unbounded step left in a function whose own law is "A landing may fail; it may not hang."** The tee drain has 10s, the flock has 5s, and round 2 removed npx's registry lookup — but `tsx` itself has no ceiling. `REASONED`, not executed: I could not construct a case where it hangs rather than failing fast, and the `install-broken` path you name would give ERR_MODULE_NOT_FOUND, not a stall. Raising it only because the surrounding bound is explicit and this one is not.
- The order pin on the flush matches `f.flush()` as a trimmed statement, so a flush wrapped in an unreachable guard would still satisfy write < flush < unlock. `REASONED`; not a realistic edit, and your doc comment already scopes the guarantee to removal and reordering. Not worth a round.

## What I did not run

- **I did not run `scripts/gate.sh`.** Landing is the operator's step, and running it *is* the landing.
- No load-batch probe and no serial `@gate-timing` pass — that is the landing gate's `[load-batches]` argument, not a reviewer's.
- I did not reproduce round 2's 70s / past-five-minutes black-holed-registry measurement. I measured only the prefix forms above, against a connection-refused registry, which is what my own finding turns on.
- My skip count differs from yours — 6 skipped / 7649 passed here against your 2 / 7653. Same 414 files and same 3 expected fails, so I read this as platform or node-version dependent (macOS, v22.23.2) rather than anything this PR did. Flagging it as a discrepancy I did not chase, not as a finding.

Approving on the strength of the above. **I am not merging it** — `27e695e` is my own commit on this branch and wants a second pair of eyes, and landing here is `scripts/gate.sh` in the operator's hands, with `[load-batches]` since this range touches tests.

gabriel-canaan:
**Declared review gap now closed — updating the body's claim rather than leaving it stale.**

The body says `0b07ea8` has no independent review record. That was true when the PR was opened; it is not true now, and both commits are recorded without a forced row:

- `0b07ea8` — **single reviewer: Ciaran Slow.** Read commit-by-commit here, and the review found a real defect in this very commit's tsx pin: the anchor it introduced is a POSITION pin, and a backslash continuation puts the command word on a different physical line from the `-e` it anchors to.
- `27e695e` — the reviewer's own repair, so it needed a second pair of eyes. Verified by **re-running mutation M7**, not by reading it. The launcher prefixes the resolved path on the *existing* continuation line, so no line is added and the producer-citation law cannot mask the result (line count identical, checked).

  | | M7 applied | |
  |---|---|---|
  | CONTROL — test file at `0b07ea8` | same mutation | **214 passed, GREEN** — the launcher runs unseen |
  | PROBE — test file at `27e695e` | same mutation | **RED**, naming the command it found |

  The exposure measurement is re-read and accepted: handed an absolute path, neither `npx` nor `npm exec` consults the registry, so what was repaired is a claim, not a reachable hang.

`verify-ledger.sh check --strict` now reports *every commit in range has a review record*.

Worth recording for whoever reads this later: this is the same defect three iterations deep — token pin, then suffix pin, then position pin, now a logical-line pin. Each fix was narrower than the property its own name asserted, and two of the three were mine.

Issue #311, opened 2026-09-07
drawer law: the file-count comment and its assertion say the same thing


gabriel-canaan:
## Why

prd-43 ruling 2 — *a count stated in prose is derived from the thing it counts* — with
ruling 1 establishing that a code **comment** is prose for that rule's purposes. #235
fixed the two `lab/` laws that stated an exact count in a comment and then asserted a
lower bound. A third law does the same thing, in another package, and #235's fence
could not reach it.

`packages/web/src/drawer/readonly.test.ts`, at its coverage assertion:

> 11 real files after #562 (8 in `drawer/`, 3 in `why/`): the peek retired `Tabs.tsx`
> and `Trace.tsx` outright, so this pin comes DOWN by two — with the count still
> **exact-ish rather than a loose lower bound**, so a surface quietly dropping out of
> the walk fails loudly here too.

The assertion beneath it is `expect(sourceFiles().length).toBeGreaterThanOrEqual(11)`.

"exact-ish" is the tell: the author knew the assertion was weaker than the sentence.

## What is actually wrong, stated narrowly

The comment claims the pin is exact so that "a surface quietly dropping out of the walk
fails loudly here too". **A file silently ADDED is not caught**, which is the half a
floor cannot do. EXECUTED at `e5895a4`:

```
control                                    15 passed
+ packages/web/src/drawer/zz-probe.ts      15 passed   <- a 12th file, invisible
- packages/web/src/why/NearestEntry.tsx     1 failed   (expected 10 to be >= 11)
```

So the floor is tight downward and blind upward, exactly as #235 found for the `lab/`
pair. The counts themselves are correct today: `find` over the two directories gives
8 + 3 = 11, matching the comment.

## Definition of done

- The comment and the assertion say the same thing. Either the assertion becomes exact
  or the comment stops claiming exactness — not one file each way.
- The counts are re-derived by running the law, not by copying the numbers above.
- The law stays green on landing. It is not disabled, skipped, or given an allowlist.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` pass.

## What mutation would this test survive?

The one this issue is filed on: **add any source file under `drawer/` or `why/`**. Green
today; must redden after.

**Read #235's landed commit before choosing a shape.** It went through three review
rounds and each repair opened the next hole, all in the same helper:

1. swapping `toBeGreaterThanOrEqual` for `toBe` — survives a *compensated* shrink, where
[...]

PR #312, opened 2026-09-07
prd43 w8: the outbound-call table and the file-count laws state their real counts

### LINKED ISSUE #235: prd43 w8: a law's file-count comment and its assertion say the same thing

### LINKED ISSUE #275: prd43 w8: every row in the outbound-call table states its real expected count

gabriel-canaan:
Closes #235
Closes #275

prd-43 wave 8 — the milestone's last wave. Two lane commits and two paper commits.

## What becomes true

- **#275** — every row in the outbound-call table renders its real expected count. The `describe.each` title was `'counts %s exactly %i time(s)'` over `[label, source, expected]`, and vitest fills specifiers positionally, so `%i` took `source` (a string) and coerced to `NaN`. The `expected` count — the one number the title exists to state — was never read.
- **#235** — the two `lab/` file-count laws assert what their own comments claim. Both stated an exact count in a comment, insisted in that comment that it was *not* a loose lower bound, and then asserted `toBeGreaterThanOrEqual`.

## Fence widening: none. But the PRD is in no issue's fence

`docs/prds/prd-43-the-claim-is-a-test.md` is touched by the two paper commits and is claimed by neither issue. That is deliberate and is recorded here rather than as a widening, the way PR #298 did it: the PRD is the plan of record for the wave being closed, and amending it is part of closing the wave, not part of either issue.

The other three files are exactly the declared fences. `scripts/fence-lint.sh 235 275` → PASSED, overlaps: none.

## What is the sibling case?

**#275:** the class is `.each` titles whose specifiers do not match the row tuple. A review seat swept **7,649 rendered check lines** across the whole suite — zero titles carry `NaN`, `[object Object]`, or an unconsumed specifier. The one near-miss found (`scene/palette.test.ts:478`, where `%s` binds the *actual* rather than the *expected* column) is out of fence and left alone.

**#235:** the two laws are each other's sibling — same defect, introduced together, docstrings referencing one another. A **third** instance exists at `packages/web/src/drawer/readonly.test.ts:180` (same shape, hedged in its own comment as "exact-ish"). It is in another package, outside this fence, and was **filed as #311 and fixed on its own branch** rather than widened into here.

Every one below was executed and restored; the working tree was left clean.

**#275** — revert control renders **exactly 53** `exactly NaN time(s)` titles; the fix renders **0**. Mutating a row's expected count 1→2 fails at the scalar equality.

**#235** — the issue named a compensated shrink as the mutation worth catching: delete two files from `compare/`, add two at `lab/`'s root, total still 17. A matcher-only fix survives it. This commit asserts **per-directory** counts, so it reddens naming the directory:

```
-   "compare": 7,
+   "compare": 5,
```

The accumulator is `Object.create(null)`, not `{}` — a plain literal inherits `__proto__` as an accessor, so `counts['__proto__'] = n` assigns through the setter and creates no own key, leaving a real file under `lab/__proto__/` invisible. Reddens now, naming `__proto__: 1`.

## Review

Four rounds. Rounds 1–3 each found a real defect **introduced by the previous round's repair** — a matcher-only fix, then a `(root)` sentinel a directory can collide with, then the `{}` accumulator. Round 4 found nothing, from two seats.

Round 4's seats converged independently on why the fix is exactly the right width: `__proto__` is the only property on `Object.prototype` implemented as an accessor rather than a data property, so every other worrying name (`constructor`, `hasOwnProperty`, `toString`, `valueOf`, `length`, numeric-like, NFC/NFD variants) already shadows correctly on a plain object. One seat enumerated 15 key forms as a grammar table, all EXECUTED.

`verify-ledger.sh check --strict` — every commit in range has a review record, no forced rows.

**Two findings are recorded here and deliberately not acted on** — neither filed nor fixed, and this is the whole record of them. Both are REASONED, both concern the PRD's shipped state-derivation command, and neither has a live instance: it lacks an explicit `--repo`, which matters only in a clone whose remote is a fork; and `--limit 200` truncates silently at a 201st issue, against a milestone holding 34.

They are left because acting on them costs more than they are worth. Amending the paper commit invalidates its ledger row, which needs a review round, which the round budget refuses on churn — for two shell flags on a documentation command, in the PR that closes this PRD. This milestone grew from 9 issues in 5 waves to 34 in 10 precisely by treating every real-but-small finding as work; two more would be that pattern one last time. If a fork clone or a 201st issue ever appears, this paragraph is the pointer.

## Gates

`typecheck` clean · `biome lint` 902 files clean · suite **417 files / 7741 passed | 3 expected fail | 2 skipped**, on this head, level with `origin/main`.


ciaran-slow (approved):
**Approving.** Both fixes are real, and both mutations the PR claims were re-run here rather than taken on trust.

### #235 — the compensated shrink, certified

The interesting claim is that the grouped assertion catches a defect the old one could not. Both halves were run (EXECUTED):

Mutation — `git mv` two files from `lab/compare/` to `lab/`'s root, so the total stays 17:

```
AssertionError: expected { '': 7, branching: 2, …(2) } to deeply equal { '': 5, branching: 2, …(2) }
    "branching": 2,
-   "compare": 7,
+   "compare": 5,
    "launch": 3,
```

Red, and it **names the directory that moved** — which is the whole point.

Then the control that matters: I restored the pre-fix assertion (`expect(sourceFiles().length).toBeGreaterThanOrEqual(17)`) with the two files still moved. **10 passed (10).** The old law is green through the exact defect the issue was filed about. That is the certification, and it is the difference between a matcher change and a real fix.

`Object.create(null)` is the right call and the docstring's reasoning about the `(root)` sentinel is correct — parentheses are legal in a POSIX path. Worth noting `toEqual` ignores the prototype difference (only `toStrictEqual` would care), so the null-prototype accumulator costs nothing at the assertion site.

### #275 — the title states the count

`%s (%s) %i` against `[label, source, expected]` is positionally right. On the fix, the rendered titles carry **0** instances of `exactly NaN time(s)` and **53** real counts — 17 × `exactly 0 time(s)` and 36 × `exactly 1 time(s)`. That 53 matches the PR's stated control figure exactly. (EXECUTED)

### Paper

The Sequencing note is the right level to write at — a section that cannot be kept current by hand saying so and handing over the command beats a seventh correction. The `tr '\n' ' '` normalisation before `grep -o` is a genuine catch; a line-oriented count really does under-report a phrase wrapped across a break.

The wave 5 / wave 8 merge is justified on this PRD's own definition of a wave, and `scripts/fence-lint.sh 235 275` reporting no overlap is the derivation rather than the assertion. Both issues are milestoned `prd43` and both titles read `w8`, so the board agrees with the paper.

The PRD-in-no-fence call is recorded in the body before the fact, the way #298 did it. That is the correct handling.

**No findings.**

---

### How this was checked

gabriel-canaan:
Landed on `main` as `f4820552` (#275), `b5de19ac` (#235), `10c0d141` and `bb39d193` (the two paper commits), merged via `scripts/gate.sh` with 4 load batches.

**This PR reads CLOSED rather than merged, and that understates what happened.** The gate merged cleanly and its push then hit a non-fast-forward — `origin/main` had moved 23 commits (prd-52 and prd-53 landed in the interval). Per the runbook that is the after-the-merge case, so `main` was rebased onto the new base and pushed directly. That reminted every sha, so GitHub could not match this branch's head to `main` by ancestry and never marked the PR merged; `Closes #235` / `Closes #275` therefore never fired, and both issues were closed by hand against their commits on `main`.

The content is provably the same work: the pre-rebase tip and the landed tip share patch-id `b5b9881d93ba`.

Merging this PR from the GitHub UI was considered and rejected — it would have pulled the pre-rebase lineage into `main`'s ancestry alongside the rebased one, putting all four commit messages in the history twice under different shas.

prd-43 closes at 34/34.

PR #321, opened 2026-09-07
prd52: the world composes — the scene holds several colonies, and the size it holds is a law

### LINKED ISSUE #313: prd52 w1: the scene lays out a world of colonies, and one colony is unchanged
## Fence (may touch ONLY)
- `packages/web/src/scene/geometry/world.ts`
- `packages/web/src/scene/geometry.ts`
- `packages/web/src/scene/world.test.ts`

## Blocked by
Nothing. This is the Keystone: additive, zero-claimant, and every later wave consumes it.
Sibling lanes hold other surfaces: nothing else in prd-52 may enter `scene/geometry/world.ts`.

## Definition of done
prd-52 ruling 2. `layoutWorld(sources, options)` sits above `layoutScene` and returns a `WorldGeometry` of colony-local geometries, each carrying the `origin` that places it.

- One `ColonyGeometry` per source, in the order the caller gave them, each carrying the caller's own id.
- **At one colony the geometry is byte-identical to `layoutScene(fleet, options)`** — asserted with `toStrictEqual` against the shipped `fleet20` AND `pathology` fixtures, not a hand-built one.
- The single colony's `origin` is exactly `{ x: 0, y: 0 }` — `toStrictEqual`, not `toBeCloseTo`; a colony 0.0001px off the origin has had a translation applied to it.
- Colony boxes stay inside the world and no two share an origin.
- `threadCount` is the sum across colonies, proven against two colonies of genuinely different sizes.
- The column count minimises cell distortion measured **symmetrically**, so a linear `|aspect - 1|` score fails the law rather than passing it.
- The colony id is the caller's to supply and is not derived from `RootMass` — `repoName` collides exactly in the several-people-one-repository case this layer exists for.

### LINKED ISSUE #314: prd52 w2: the display list is composed across colonies, and each colony pays its fixed costs once
## Fence (may touch ONLY)
- `packages/web/src/scene/marks/index.ts`
- `packages/web/src/scene/worldMarks.test.ts` (new file)

## Blocked by
#313.
Sibling lanes hold other surfaces: the cache chain holds `scene/geometry/layout.ts` and `scene/ribbon.ts` this wave; `scene/marks/thread.ts` belongs to the detail-substitution lane in a later wave; `scene/marks.test.ts` is that lane's and not this one's — put this wave's laws in a new file. Do not enter `scene/gl/`.

## Definition of done
prd-52 ruling 1. `worldMarks` composes one display list for the whole world.

- Each colony's marks are placed by its `origin`; nothing is removed, per `marks/index.ts`'s own law that density is managed by thinness, stillness and depth layering, never by removal.
- Depth order holds **across** colonies, not only within one — `byDepth`'s finished-behind-living rule extends to the world.
- **Each colony pays its fixed costs once.** `ambient-world`, `root`, `dissolve` and `ambient-screen` are emitted once per colony — never once per world, never once per thread. A count assertion against three colonies, and a mutation that emits one per world turns it red.
- At one colony the composed list equals `sceneMarks`' list.
- Every colony's threads reach the list; a world of N colonies emits every thread it was given.

### LINKED ISSUE #315: prd52 w2: a growing thread answers from cache at both links
## Fence (may touch ONLY)
- `packages/web/src/scene/geometry/layout.ts`
- `packages/web/src/scene/ribbon.ts`
- `packages/web/src/scene/livingCache.test.ts`

## Blocked by
#313.
Sibling lanes hold other surfaces: `scene/marks/index.ts` belongs to the composition lane this wave. `scene/geometry.test.ts` is NOT in this fence and must stay green untouched — that is the proof geometry output did not move.
**Model:** opus. The invalidation argument is the whole difficulty; a wrong key is a stale picture, not a failing test.
**Wave:** 2.

## Definition of done
prd-52 ruling 4 — repaired at both links, or at neither.

- The **pre-truncation** spine is cached keyed *without* `growth`, and re-truncated per frame. `layout.ts` already builds `full` then truncates; this caches in front of that seam.
- The outline cache in `ribbon.ts` is re-keyed off raw array identity, so a lane whose growth advanced since the previous frame **hits** it. Today the fresh truncated array misses by construction.
- **A hit/miss count assertion, never a wall clock** — the file's standing discipline.
- Beware the scope of "growing": `GROWTH.thickenTau = 30_000` keeps `thicken` moving while a lane does *work*, and `sizeFrac` feeds `rim` through `lifecycleFrac`, so the living key churns for most of a busy fleet rather than a few new arrivals. The cached spine must not be silently invalidated by `rim` movement.
- `livingCache.test.ts`'s existing identity, invalidation and byte-safety contracts still hold.
- **`geometry.test.ts` stays green with no edits** — geometry output is unchanged.
- A mutation reverting either link alone leaves the cache-hit law red.

### LINKED ISSUE #316: prd52 w3: the camera frames the whole world
## Fence (may touch ONLY)
- `packages/web/src/scene/camera.ts`
- `packages/web/src/scene/camera.test.ts`

## Blocked by
#313, #314.
Sibling lanes hold other surfaces: the selection lane holds `fleet/selection.tsx` this wave. `scene/view/useCamera.ts` and `scene/view/useFrameLoop.ts` are the view-wiring lane's, in a later wave — this lane changes the pure function only and wires nothing.

## Definition of done
- `contentBounds` gains a world form that unions N colonies' bounds, each offset by its `origin`.
- **At one colony the world bounds equal today's single-geometry bounds exactly.**
- An empty world, and a world whose colonies are all empty, both answer without throwing — today's empty-scene fallback still applies.
- `fitCamera` over the world's bounds frames every colony: no colony's threads fall outside the fitted box.
- Existing camera laws stay green; this is additive.

### LINKED ISSUE #317: prd52 w3: a selection names its colony
## Fence (may touch ONLY)
- `packages/web/src/fleet/selection.tsx`
- `packages/web/src/fleet/selection.test.tsx`

## Blocked by
#313.
Sibling lanes hold other surfaces: the camera lane holds `scene/camera.ts` this wave. **The eleven `MAIN_SELECTION` consumers across `drawer/`, `lane-page/`, `scene/index.tsx`, `scene/view/` belong to the view-wiring lane in a later wave — do not edit them.** This lane is additive: the bare `MAIN_SELECTION` keeps working and is retired later, by that lane.

## Definition of done
prd-52's law 6.

- A selection id carries its colony, so **no id resolves to two lanes in a world**.
- Two colonies holding a same-named lane produce different selection ids — the assertion that would fail today, since `MAIN_SELECTION` is one hardcoded id and N colonies are N masses.
- A colony id and a lane id round-trip: both are recoverable from the composed id.
- **The single-colony behaviour is unchanged and no consumer is edited in this issue.** Additive only.

### LINKED ISSUE #318: prd52 w4: the scene thins its material as the world grows, and the encoded channels do not move
## Fence (may touch ONLY)
- `packages/web/src/scene/marks/thread.ts`
- `packages/web/src/scene/geometry/scale.ts`
- `packages/web/src/scene/marks.test.ts`

## Blocked by
#313, #314.
Sibling lanes hold other surfaces: `scene/marks/index.ts` landed in wave 2 and is read, not reopened. Do not enter `scene/palette.ts` — this lane reads tokens and never writes them.
**Landing note:** `marks.test.ts` carries `// @gate-timing`, so the landing needs `scripts/gate.sh <handle> <fence> <load-batches>` with a non-zero third argument.

## Definition of done
prd-52 ruling 5 — substitutes, never hides.

- A detail level is derived from **world thread count**, following `geometry/scale.ts`'s own `LABELS_ALL_MAX` precedent. **Never from measured frame time** — that would move a setting the operator chose, which D36 rules is theirs.
- Above the threshold the bloom and underglow passes thin, reusing the shipped quality gate already at `marks/thread.ts`. Below it the picture is unchanged.
- **Locked channels do not move.** Thread hue, thread brightness, radial position and encoded width are identical at every detail level, and a mutation that varies any one of them turns the law red.
- Nothing is removed: every thread still emits its core ribbon at every level.

### LINKED ISSUE #319: prd52 w5: the view renders through the world path
## Fence (may touch ONLY)
- `packages/web/src/scene/view/useCamera.ts`
- `packages/web/src/scene/view/hitTest.ts`
- `packages/web/src/scene/view/hitTest.test.ts`
- `packages/web/src/scene/index.tsx`
- `packages/web/src/drawer/attach.ts`
- `packages/web/src/drawer/index.tsx`
- `packages/web/src/drawer/index.test.tsx`
- `packages/web/src/lane-page/LanePage.tsx`

## Blocked by
#313, #314, #316, #317. Split out of wave 4 deliberately: this is the largest fence in the PRD, nine files across five directories, and it is the one lane worth reviewing alone.
Sibling lanes hold other surfaces: nothing runs beside this lane. `scene/marks/thread.ts` and `scene/geometry/scale.ts` landed in wave 4 and are read, not reopened.
**Model:** opus. **Wave:** 5.
**Landing note:** `SceneView.test.tsx` carries `// @gate-timing`; the landing needs a non-zero `load-batches`.

## Definition of done
- `useFrameLoop` builds its frame through `layoutWorld`/`worldMarks` and frames it with the world camera, rather than through the single-colony path.
- **With one colony the rendered frame is unchanged** — `SceneView.test.tsx`'s existing laws stay green with no assertion edits. That is ruling 2 reaching the running app.
- Hit-testing resolves a (colony, lane) pair, and the drawer and lane page open the right lane in a world of more than one colony.
- The bare `MAIN_SELECTION` id is retired in favour of the namespaced form from #317; all eleven consumers move together.
- The app still renders one colony in production: the multi-colony fixture is test-only and does not join `StreamSource` (operator ruling, 2026-09-07).

### LINKED ISSUE #320: prd52 w6: the model floor is measured on the composed world
## Fence (may touch ONLY)

## Blocked by
#313, #314, #315, #316, #318, #319. Last on purpose: it measures what every earlier wave actually built, and a sweep that runs earlier re-lays ground the others are about to dig.
Sibling lanes hold other surfaces: nothing runs beside this lane.
**Landing note:** `perf.test.ts` carries `// @gate-timing`; the landing needs a non-zero `load-batches` or the serial pass never runs and the numbers are untrustworthy.

## Definition of done
prd-52 ruling 3 — the size is restated by measurement, never by promise.

- The model-floor cells run **through `layoutWorld`**, at 1x30, 3x30, 1x180 and 3x60, replacing the sum-of-three-independent-layouts proxy.
- Reported **by builder** — lift the existing `time`/`perThread` closures to file scope rather than writing new ones.
- A **growing-versus-pinned** arm reports what wave 2's cache chain bought.
- **Timings reported, never asserted.** Every law here is a count or a shape; no `expect` bounds a millisecond.
- The comment block at `packages/web/src/scene/perf.test.ts:1555-1572` gains one line naming #190 and prd-49 — that issue's own deferred request, which asks for it in the same change as a re-measurement and not before.
- **The composed 3x30 reading is recorded**, so prd-49's clause-2 threshold — suspended by the 2026-09-07 amendment because it was calibrated against the summed number — can be re-derived from it.

KelliherL:
Closes #313
Closes #314
Closes #315
Closes #316
Closes #317
Closes #318
Closes #319
Closes #320

## What this is

prd-52 — *the world composes* — blessed 2026-09-07, all six waves in one branch, **by operator ruling, as one batch for review.** This bundles across waves, which AGENTS.md's working agreement forbids, and the operator took that call knowingly with the multiplayer conversion landing tomorrow. The mitigation is the one the agreement itself names: **the commit is the unit of review.** One commit per issue, in wave order, each self-contained with its rationale, its laws and its mutations. Read it commit by commit.

The scene renders exactly one colony and the conversion arrives on top of it. prd-33 ruling 7 promised *"no redesign is required"* when the team layer landed; audited against the tree, that held for the architecture and failed for the implementation — placement, a union camera fit, cross-colony depth order and namespaced selection were never written. This branch writes them, and pins the laws that make a scale regression fail a test instead of being discovered in a session.

## Read in this order

| commit | issue | the fact it makes true |
|---|---|---|
| `3dfcd0a` | — | the PRD: six rulings, blessed at `db61452`, ruling 7 at `b18d93c` |
| `90a1234` | — | prd-49's tripwire says which cell it watches now that the scene composes (posted to #190) |
| `2673964` | #313 | the scene lays out a world of colonies, and one colony is unchanged |
| `69a51ed` | #314 | the display list is composed across colonies; each colony pays its fixed costs once |
| `28d6df3` | #316 | the camera frames the whole world |
| `f27937b` | #317 | a selection names its colony |
| `8685dc9` | #315 | a growing thread reuses the curve growth cannot change |
| `813ae9c` | #318 | material thins as the world grows; the encoded channels do not move |
| `353713a` | #319 | the view renders through the world path |
| `a63a9ca` `1c220c1` | #320 | the model floor is measured on the composed world |
| `7dd06a5` `51a4234` `a6cf44b` | — | placement ruled: one surface, stable arrival order, fixed ring slots |
| `b18d93c` | — | ruling 7: a departed colony stays as landed mass |

**`2673964` has no `(#313)` in its subject.** It predates its own issue — the milestone could not exist before the blessing — and it now sits behind three later commits. Carried here as `Closes #313` rather than rewriting history.

## What is the sibling case?

- **#313 / placement.** The obvious case is N=1; the sibling is *every other colony*. The first cut shrank each colony's box to fit N, which changed a teammate's `rootRadius`/`rx`/`ry` — so their work rendered at a different scale from the one they were looking at. Now every colony is laid out at the full box and only the origin differs; the byte-identity law holds for all of them, not just the solo case.
- **#314.** The sibling of "translate the marks" is *the next mark kind somebody adds*. A walker over the display list needs a case per kind and every later kind is a silent miss. The origin goes into `layoutScene`'s centre instead — one term, cannot rot.
- **#316.** `contentBounds` of one colony vs the union: the sibling is the *empty* world, which returns `null` rather than a box at the origin the camera would fly to.
- **#317.** The bare `MAIN_SELECTION` vs the namespaced form: the sibling is a lane id that itself contains the separator (`feat/thing`), so the split is on the **first** `/`, and a colony id containing one is refused rather than mis-parsed later.
- **#315.** The living cache vs the base cache: they are complementary, not alternatives. At rest the living cache answers and the base is never consulted; only a lane whose sole change is growth reaches the second.
- **#318.** `thread.ts`'s three gates were rerouted through `drawnQuality`; the sibling is `ambient.ts`'s two gates, which are prd-33's material and are **deliberately left** on the raw dial — stated in the commit, not missed.
- **Placement.** "Joining moves nobody" had a sibling I did not see until the law caught it: *leaving*. Removal from a slotted ring shifts every colony after the gap. Ruling 7 closes it — a departed colony keeps its slot as landed mass.

**Twenty-five mutations run across the branch.** The ones that survived, and why each is stated rather than hidden:

- **M4** (#313, aspect scoring) — *survived on first write.* The test asserted a bound both scorings met. Replaced with a symmetric-distortion property over five boxes; the linear scorer now fails at 6 colonies in 480×500.
- **M17, M18** (#319) — *survived, correctly.* Routing the frame loop back through the single-colony path, or fitting the camera to the colony rather than the world, changes nothing observable at N=1 — because ruling 2 says the two must be byte-identical. Any law claiming to catch that revert would be one that cannot fail for the reason it claims. Guard is #320 measuring through `layoutWorld`, and this review.
- **M20** (placement) — *survived.* Folding `{0,0}` into the first colony's options is arithmetically identical to passing them untouched; that branch is defensive, not load-bearing.
- **M25** (ruling 7) — *inconclusive on first try*: the mutant did not typecheck. Redone with one that does; all three ruling-7 laws go red.

Everything else went red as intended, including the two that found real bugs:

- **#314's origin-shift law found the spine caches keyed on `width x height`** — a second colony sharing a box size was served the first colony's cached spines and would have drawn on top of it. Invisible to every existing test.
- **The moves-nobody law caught the ring re-spacing by count** (`51a4234` landed with it red; `a6cf44b` fixes it with eight fixed 45° slots). That commit is left in history so the law is seen doing its job.

## Fence widenings — recorded on the issue before the change

- **#314** → `geometry/types.ts`, `geometry/layout.ts`, `geometry/world.ts` (the origin lands in `centre`).
- **#318** → `marks/frame.ts`, `marks/index.ts` (the world's thread count travels on the frame; `geometry.threads.length` is one colony's and would thin at N× the ruled size).
- **#319** → *narrowed* from ten paths to four; the `MAIN_SELECTION` retirement across eleven consumers is moved to unfiled work, since #317 made the namespaced form additive on purpose and retiring the bare id changes nothing while the app renders one colony.
- **#315** → *scope amended* to link 1 only. The outline cache's miss for a growing lane is real work, not a stale key; the 15 ms figure that made it look broken is the harness's pessimal construction.

## Measured, on the composed path

`perf.test.ts`'s model floor now runs `layoutWorld` + `worldMarks` instead of summing three independent renders. One loaded box, one run — ratios are the claim:

| cell | threads | of budget |
|---|---|---|
| 30×1 | 30 | 34.9% |
| 30×3 | 90 | 76.6% |
| 60×3 | 180 | 159.4% |
| **180×1** | **180** | **126.5%** |

**One colony is cheaper than three at the same thread count** — the per-colony fixed costs dominate. The cost of multiplayer is the number of people. And the growing-vs-pinned arm reads **1.12×**, which brackets #315's win at twelve percent: the renderer spike called it "the cheapest large win available"; against a measurement it is cheap but not large.

## ⚠ No verify ledger backs this range

AGENTS.md's PR-opening section requires every commit be recorded in an operator-local verify ledger, and says a green suite is not that. **This machine does not have the kit** — `verify-ledger.sh`, `lane-precommit.sh` and `pr-open.sh` are not shipped by the repo, are not in WSL or Windows `~/.claude/skills/`, and are not in `KelliherL/RhizoReview`. Nobody in the cohort could point to it. The runbook's own honest escape, `pr-open.sh --unverified "<why>"`, is part of the same missing kit — so this paragraph is that escape, written by hand.

What *does* back the range: the full suite green under `--maxWorkers` (418 files, 7757 tests, 3 expected-fail, 2 skipped), typecheck 0, lint clean on 904 files, and the 25 mutations above. That is artefact evidence, and the runbook is right that it is not the same as an independent read. **Reviewer: this is the read.**

## Landing

- Touches three `// @gate-timing` files (`marks.test.ts`, `SceneView.test.tsx`, `perf.test.ts`) → the landing needs `scripts/gate.sh <handle> <fence> <load-batches>` with a **non-zero third argument**, or the serial timing pass never runs.
- prd-49 amended and posted to #190: clause 2 follows the composed cell with its ~60% threshold **suspended** until re-derived from #320's reading; clause 1 is not tripped by a synthetic fixture. The multi-colony fixture is **test-only** by operator ruling and does not join `StreamSource`.
- What this branch does **not** do: open the frame loop to N fleets (it composes exactly one; the seam is the world's now), retire the bare `MAIN_SELECTION`, build any data path, or promise 60 fps at 180 threads. prd-33 ruling 13 stands: raising the ceiling is a measurement, not a decision.


ciaran-slow (commented):
Read commit by commit, in the order the body sets out. **One finding, and it is fixable in a line** — everything else verified in your favour, including the claims about your own history.

## What verified

**Your account of `51a4234` → `a6cf44b` is exactly true.** I checked out both:

- at `51a4234`: `× moves NOBODY when a colony joins…` — **1 failed | 12 passed**
- at `a6cf44b`: **15 passed (15)**

The red commit really is left in history with the law visibly doing its job. That is worth more than a clean bisect. (EXECUTED)

**The moves-nobody law bites on the real defect.** I re-introduced ring re-spacing by live count — `ringOrigin(i-1, w, h, sources.length-1)` with the angle divided by `count` — and got `× moves NOBODY when a colony joins` plus `× keeps every colony's content clear of every other's, across two full rings`. Note my first attempt at this mutant was **inert** (I added the parameter but left the default, so behaviour was unchanged and 18/18 passed); the law only reddens once the caller actually passes the count, which is the correct sensitivity. (EXECUTED)

**All three ruling-2 laws can fail.** Translating the lone colony 0.5px reddens `places the only colony at the world origin exactly, not near it` and nothing else — correctly, since the geometry is untouched. Translating it *and* routing it through the composed path reddens all three, including both byte-identity assertions. The origin law and the byte-identity laws are complementary rather than redundant, and each fails for the reason it claims. (EXECUTED)

**Your disclosed M20 reproduces exactly as described.** Folding `{0,0}` into the first colony's options is arithmetically identical and survives — 18/18. Stating that rather than hiding it is the right call, and I agree the branch is defensive, not load-bearing.

**The stale base is not a problem.** This branch was 13 commits behind. Merged onto `7b3f3312` alone: typecheck 0, **419 files, 7784 passed, 3 expected fail, 6 skipped, exit 0**. (EXECUTED)

`RING_SPACING = 1.5` clears the √2 worst case with margin, and having `world.test.ts` prove disjointness empirically rather than trusting the arithmetic is the right choice. `worldBounds` unioning content rather than cells is correct for the same reason `contentBounds` is not the viewport box, and it returns a copy at N=1 so `toStrictEqual` works without an identity claim.

---

## Finding — `f27937b` (#317): a bare lane id containing `/` now reads as the root mass

**EXECUTED, and it is a regression.**

[`selection.tsx:67`](packages/web/src/fleet/selection.tsx#L67) changed:

```ts
-  return selectedId === MAIN_SELECTION
+  return withinColony(selectedId) === MAIN_SELECTION
```

`withinColony` splits on the **first** `/`. So for a *bare* — unnamespaced, single-colony — selection of a lane on branch `feat/main`, it returns `'main'`, and `isMainSelected` answers **true** for a worker lane.

```ts
expect(isMainSelected('feat/main')).toBe(false)
// AssertionError: expected true to be false
```

Same probe on `origin/main` (`7b3f3312`): **3 passed (3)**. On this branch: **1 failed | 2 passed**. Before this commit `'feat/main' === 'main'` was plainly false.

**It is reachable on this commit's own premise.** The docstring immediately above says a lane id is a branch-derived slug and that `feat/thing` "is nothing unusual", and the fixtures use `feat/thing` and `al/pha`. The app renders one colony today, so the bare form is the live path — which is exactly the case that regresses. Two consumers render the wrong surface: [`drawer/index.tsx:74`](packages/web/src/drawer/index.tsx#L74) and [`panels/trace/index.tsx:61`](packages/web/src/panels/trace/index.tsx#L61).

**Why the tests miss it.** `selection.test.tsx` covers `colonySelection('alpha', 'feat/thing')` splitting correctly, and covers `isMainSelected` on the bare root, the namespaced root and a namespaced lane. What it never covers is a bare *lane* id that itself contains the separator. This is the sibling shape the PR body is otherwise so good at: you handled a slashed lane id in the **split** and not in the **membership test**.

**I have not pushed a fix.** The minimal one I would reach for is `COLONY_SEP = ':'` — git forbids `:` in a branch name, so a branch-derived lane id can never contain one, the first-split logic is unchanged, and the `colonyId.includes(SEP)` guard gets stronger. But you wrote a deliberate paragraph choosing `/` and reasoning about the first-split, so overriding that on your branch is your call and not mine. Happy to push whichever way you want it, or to leave it to you.

---

## Two process notes

**The cross-wave bundle.** The body says the operator took this call knowingly. I cannot verify a ruling from the PR body alone — it is asserted here rather than recorded on an issue. Worth a line on one of #313–#320 so it survives this PR.

**Landing.** You already flagged it and it is right: three `// @gate-timing` files are touched, so `scripts/gate.sh <handle> <fence> <load-batches>` needs a **non-zero third argument** or the serial timing pass never runs.

**Your ⚠ section is accurate** — the operator kit genuinely is not shipped by this repo, and writing the `--unverified` escape by hand was the honest move. This review is the independent read you asked for.

---

### How this was checked

ciaran-slow:
Pushed the fix as `9d70136d` — **`fix(fleet): a bare lane id keeps its slashes  (review of #321)`**, on top of `b18d93c2` as a fast-forward. Nothing was rewritten and nothing force-pushed.

### What it does

`COLONY_SEP` becomes `':'`. The ambiguity was in the separator rather than in the split: with `/` between them, `feat/main` is either a lone lane on branch `feat/main` or lane `main` inside a colony called `feat`, and nothing in the string decides which. `git check-ref-format` forbids `:` in a refname outright, so no branch-derived lane id can carry one, and the bare form is unambiguous by construction rather than by care.

I rewrote the `COLONY_SEP` docblock rather than leaving it: it argued *for* `/` and for the first-split at some length, so keeping it would have left the comment contradicting the code — which is the defect class #312 is closing one PR over.

**The sibling.** The colony half was guarded from the first write; the selection half was not. `colonySelection` now refuses the separator in **both**, and checks the selection *before* the bare-form early return, since a bare id holding one is mis-split by `colonyOf`/`withinColony` later with no colony ever having been named. Same shape as the original defect — a rule applied to one of two symmetric halves.

### Evidence (EXECUTED, on this branch)

- Reverting `COLONY_SEP` to `/` and leaving the tests alone reddens **all four** new/changed assertions — including `does NOT read a bare lane id as a namespaced one` and the both-halves refusal. None of them can pass for the wrong reason.
- With the fix: `selection.test.tsx` **34 passed (34)**.
- Full gate on the branch: `npm run typecheck` **0**, `npm run lint` **0**, `VITEST_MAX_WORKERS=6 npm test` → **418 files, 7763 passed, 3 expected fail, 6 skipped, exit 0**.

### Why this changes no rendered output

Nothing outside `selection.tsx` reads or builds a namespaced selection yet — #319 was narrowed and the `MAIN_SELECTION` retirement deferred — so no call site depends on the wire format. Verified by sweeping the tree for `colonySelection|colonyOf|withinColony`: no hits outside `selection.tsx` and its test.

**Fence:** both files are already #317's. No widening.

### Over to you

Per the review ladder this is a substantive fix, so it is **blocked on your confirmation, not merged**. If you would rather keep `/` and solve the ambiguity another way — always namespacing, or having `isMainSelected` take the colony context explicitly — say so and I will drop `9d70136d`; the design call is yours and I have no attachment to mine beyond that it is the smallest one I could find. Landing is the operator's step either way, and still needs a non-zero third argument to `gate.sh` for the three `// @gate-timing` files.

Issue #324, opened 2026-09-07
prd53 w2: a measured outcome is typed and travels per run over a gated route, and a launch above the ceiling refuses by name and override


KelliherL:
## Fence (may touch ONLY)
- `packages/server/src/api/lab.ts`
- `packages/server/src/api/lab.test.ts`
- `packages/server/src/api/index.ts`
- `packages/server/src/api/route-class-law.test.ts`
- `packages/web/src/lab/adapters.ts`
- `packages/web/src/lab/adapters.test.ts`
- `packages/web/src/lab/types.ts`
- `packages/contract/src/lab-measure.contract.test.ts` (new file)
- `packages/contract/src/contract-coverage-law.test.ts`
- `docs/design-notes/lab-launch-ceiling-arms-runs.md` (new file)
- `packages/core/src/events/lab.ts`, `packages/core/src/events/lab.test.ts`, `packages/core/src/events/events.test.ts`, `packages/core/src/fixtures.ts`, `packages/core/src/fixtures.test.ts` (widened 2026-09-08 — the `fork.measured` event and `ceilingOverride` on `fork.dispatched`; see the widening comment)
- `packages/core/src/state.ts`, `packages/core/src/reduce.ts`, `packages/core/src/reduce.test.ts` (widened 2026-09-08 — the fold for `fork.measured`)
- `packages/core/src/eras/era-1/session-state.snapshot.json`, `packages/core/src/eras/CAPTURE.md` (widened 2026-09-08 — the golden fold's new keys, blessed and noted)
- `packages/server/src/lab/fork.ts`, `packages/server/src/lab/fork.test.ts` (widened 2026-09-08 — the override reaches the event; the engine's own ceiling)
- `packages/server/src/cli/lab-fork.ts`, `packages/server/src/cli/lab-fork.test.ts`, `packages/server/src/cli/lab-compare.ts`, `packages/server/src/cli/lab-compare.test.ts`, `packages/server/src/cli/index.ts` (widened 2026-09-08 — `--ceiling-override`, `--json`)
- `packages/web/src/lab/measure.ts` (new), `packages/web/src/replay/mutating-calls-law.test.ts` (widened 2026-09-08 — the one declared web module the contract law requires)

## Blocked by
#322.
[...]

Issue #323, opened 2026-09-07
prd53 w2: one floor and one confound clause live in core, and both surfaces read them


KelliherL:
## Fence (may touch ONLY)
- `packages/core/src/lab/` (new directory, all files)
- `packages/server/src/lab/compare.ts`
- `packages/server/src/lab/compare.test.ts`
- `packages/web/src/lab/compare/summarise.ts`
- `packages/web/src/lab/compare/summarise.test.ts`
- `packages/core/src/index.ts` (widened 2026-09-08 — one export line so the module is reachable as `@rhizomorph/core`)

## Blocked by
#322.
Sibling lanes hold other surfaces: the launch-path lane holds `packages/server/src/api/lab.ts`, `packages/web/src/lab/adapters.ts` and `packages/web/src/lab/types.ts` this wave — **the web-side deletion of `isCleanlyControlled` from `types.ts` is that lane's, not this one's; this lane adds the core copy and points `summarise.ts` at it.** `packages/web/src/lab/LabPage.tsx` is wave 3's.
**Model:** sonnet. **Wave:** 2.

## Definition of done
prd-53 ruling 2 — one floor and one confound clause, in core, and every surface reads the same function.

- A pure, zod-only law module in `packages/core/src/lab/` (ADR-0003-eligible) hosts: the summary floor over **completed runs per arm**; the cross-arm-claim floor over **arms**; the confound predicate (today only in `web/src/lab/types.ts` as `isCleanlyControlled`); and the counterfactual clause — what actually happened is one observation, not a distribution.
- `server/src/lab/compare.ts` and `web/src/lab/compare/summarise.ts` both import the one copy. `renderComparison`'s CLI table gains the confound voice: arms differing in model *and* brief print *"a difference cannot be attributed to either"*.
- **Recorded at review 2026-09-08, NOT before the change** (review of PR #336). The paths below were touched by this issue's commit and were missing from this fence. They are reconciliations the commit message names; the record, not the code, was short:
  - `packages/server/src/cli/index.test.ts` — the compare table's new `run` column moves one asserted header string
- Extends prd-12 ruling 4; never loosens it.
- **Laws (counts):** the floor has exactly one implementation — a grep law that both consumers import it; the confound copy appears in the CLI table for two-dimension arms; `compare.test.ts` and `summarise.test.ts` assert the *same* verdict for the same fixture.
[...]

PR #331, opened 2026-09-07
docs(prd): prd-53 — the lab: one experiment, r runs, a measured outcome, and documents that say what the code does

KelliherL:
**prd-53 — the lab**, `Kind: specifying`, **BLESSED** by Lachlan Kelliher in session 2026-09-07 (status line stamped). Milestone `prd53` = #23; its nine issues are #322–#330, fence-linted per wave.

## What this PR is
- `docs/prds/prd-53-the-lab.md` — the implementation PRD for the lab's Stage 1. Ten rulings, each naming the prd-28 ruling it carries or supersedes (prd-28's paper died in the 2026-08-19 deletion; its numbers are never reused). Keystone: **an experiment is one forkId and an arm holds r runs** — today `dispatchFork` mints one `fork.dispatched` per arm (`packages/server/src/lab/fork.ts:302`) and the web launch fragments n arms into n single-arm forks (`packages/server/src/api/lab.ts:712`), so no arm can ever reach the three completed runs `summarise.ts:36` requires and `ComparisonSurface` never renders against the real server. Every Evidence bullet cites a live path on today's tree.
- `docs/vision-the-lab.md` — **restored as recovered history** from the pre-deletion clone (`19adcda8`, 2026-08-13), under a header stating its source, its date and what has changed since. Absent from the live repo's entire history until now.

## What it does not do
No code. `docs/design-notes/lab-launch-ceilings.md` is not touched — it is prd-50's while #237/#238 are open. The six PRDs that fence `packages/web/src/lab/` as "prd-28's territory" are not edited here; each gets a dated note in wave 5 (#330).

## Laws run
`doc-citation-law` and `corpus-eol-law`: 57/57 — the restored vision doc's one glob (`packages/*/src`) is de-backticked so the citation regex does not read it as a path.

PR #332, opened 2026-09-07
prd53 w1: an experiment is one forkId, and an arm holds r runs addressable from outside dispatchFork

### LINKED ISSUE #322: prd53 w1: an experiment is one forkId, and an arm holds r runs addressable from outside dispatchFork
## Fence (may touch ONLY)
- `packages/core/src/events/lab.ts`
- `packages/core/src/events/lab.test.ts`
- `packages/core/src/state.ts`
- `packages/core/src/reduce.test.ts` (widened 2026-09-07 — the fold's own tests; one `describe` added)
- `packages/core/src/eras/era-1/session-state.snapshot.json` (widened 2026-09-07 — the golden fold moves by one key, `byArm`; blessed per CAPTURE.md)
- `packages/core/src/eras/CAPTURE.md` (widened 2026-09-07 — the move's note)
- `packages/server/src/lab/fork.ts`
- `packages/server/src/lab/fork.test.ts`
- `packages/server/src/lab/paths.ts`
- `packages/server/src/lab/paths.test.ts`
- `packages/server/src/cli/lab-fork.ts`
- `packages/server/src/cli/lab-fork.test.ts`
- `packages/server/src/cli/index.ts`

## Blocked by
Nothing. This is the Keystone — everything in prd-53 consumes it, and nothing above it is honest until it lands.
Sibling lanes hold other surfaces: none run beside this one. `packages/server/src/lab/namespace-law.test.ts` is NOT in this fence and must pass **unchanged** — it is the law this wave must not disturb. `packages/server/src/lab/compare.ts` and `packages/web/src/lab/**` are wave 2's and 3's.
**Model:** opus. **Wave:** 1.

## Definition of done
prd-53 ruling 1 — the experiment is one fork, and an arm holds r runs.

- `forkDispatchedPayloadSchema` gains `run: z.number().int().positive().optional()`, read everywhere as `run ?? 1`. The codebase's own additive convention (`beacon.ts`, `git.ts`, `system.ts`) — **not** `upcast()`, which is identity today and exists for reshaping.
- `ForkState` gains a `byArm` index keyed `${forkId}#${arm}` beside `byFork` and `byLane` (`reduce.ts`'s `appendIndexed`). The comment at `state.ts` *"One arm per handle in practice"* dies with the fact.
- Worktree paths and lane handles become run-aware and **elide `-run-1`**, so every existing single-run fixture path still holds; the new grammar appears only where `run > 1`.
- `lab fork` gains `--fork-id` (exposing the already-injectable `DispatchForkOptions.forkId`, `fork.ts:66-67`) and `--arm-number` (valid with `--arms 1`). `launchExperiment` mints one `forkId` and passes both on each per-arm `runCli` call. `ARM_LINE_RE` (`api/lab.ts:624`, which hardcodes `arm 1`) parses the real arm number.
- **Law 8 — old-behaviour tests rewritten in the same commit, not merely updated:** the `api/lab.test.ts` fixture (~lines 140-166) where two runs share one `worktreePath` — *illegal* once runs are real, not stale; the `parseSingleArmForkStdout` tests (~245-283); literal `fork-N-arm-N` paths in `paths.test.ts` and `fork.test.ts` (survive via elision, gain `run > 1` cases).
- **Laws (counts, never clocks):** after a simulated multi-call web launch, `byFork[forkId].length === arms × runsPerArm` — one forkId per launch however many `runCli` calls; no two dispatch records share a `worktreePath`.
- **Mutations to run and show red:** put `arm` back where `run` goes (worktree-uniqueness law); drop `--fork-id` from the web launch (one-forkId law).

KelliherL:
**prd-53 wave 1 — the keystone.** One issue, one commit. Closes #322.

Before this PR the lab could not hold an experiment: `dispatchFork` minted one `fork.dispatched` per arm with no run dimension (`fork.ts:302`), and the web launch split an n-arm experiment into n single-arm forks with n fork ids (`api/lab.ts:712`). No arm could ever reach the three completed runs `summarise.ts:36` requires, so `ComparisonSurface` never rendered against a real server — the user guide said so; the status lines did not. After this PR one launch is one `forkId`, an arm holds r runs, and every run has its own worktree and handle.

## What changes, by layer
| layer | before | after |
|---|---|---|
| event | `fork.dispatched {forkId, arm, …}` | `+ run?` — optional, read everywhere as `run ?? 1` (additive, like `beacon.ts`/`git.ts`; not `upcast()`) |
| fold | `byFork`, `byLane` | `+ byArm` keyed `armKey(forkId, arm)`; `ForkDispatchRecord.run` |
| paths | `<forkId>-arm-<n>` | `armLeaf(forkId, arm, run)` — **`-run-1` elided**, so every existing path/handle/fixture holds; `-run-<r>` only for r > 1 |
| engine | `arms` | `+ runs`, `+ armNumber` (first arm this call dispatches); arm-major, run-minor loop |
| CLI | `--arms` | `+ --runs <r>`, `+ --fork-id <id>`, `+ --arm-number <k>` (refused unless `--arms 1`) |
| web launch | n calls, n fork ids, `ARM_LINE_RE = /arm 1/` | mints ONE forkId, passes `--fork-id`/`--arm-number`/`--runs` on every call; `parseForkStdout` reads the real arm number and every run; a CLI that answers with another forkId is a named failure |
| ceiling | `arms ≤ MAX_ARMS` | `arms × runs ≤ MAX_ARMS` — every run is a spending lane (prd41 ruling 4). prd53 ruling 6 (w2b, #324) makes it configurable and names the override |
| DTO | `LabRunDTO` | `+ run`; `LaunchResult.forkId`, `LaunchedArmResult.runs[]` — additive, the web is untouched |

## Laws (counts and shapes, never clocks) — and the mutation each survived-then-failed
Every law was shown **red** by a deliberate break, then the break reverted byte-for-byte (checksummed):

| law | where | mutation | result |
|---|---|---|---|
| arms × runs realities under ONE forkId; no two share a worktree or handle | `fork.test.ts` | `armLeaf` forgets the run (arm back where run goes) | **red** — 3 tests across fork/paths/api |
| one launch = one experiment: `byFork[forkId].length === arms × runs`, read from the **log the launch wrote**, not the result | `api/lab.test.ts` | launch stops passing `--fork-id` | **red** |
| the r runs of one arm index under one `armKey` | `reduce.test.ts` | fold never appends to `byArm` | **red** — 2 tests |
| a record with no `run` reads as run 1 | `reduce.test.ts`, `api/lab.test.ts` | `run ?? 0` | **red** |

Plus: schema accepts `run`, refuses 0/negative/fractional, and a pre-prd53 payload parses unchanged with `run` undefined (`events/lab.test.ts`); `--arm-number` without `--arms 1` is refused by name (`lab-fork.test.ts`); `armLeaf` run 1 is byte-identical to the old spelling and runs 1–3 give three distinct leaves (`paths.test.ts`).

## What is the sibling case?
Two dispatch paths reach the engine and both are exercised: the **operator's CLI** (`lab fork --arms 2 --runs 2` → four realities, one forkId — `fork.test.ts`) and the **web launch** (`--arms 1 --fork-id X --arm-number k --runs r`, one call per arm because each arm carries its own treatment — the one-forkId law runs the real CLI through `runCli` and folds the real log). The other sibling is **time**: records written before this PR carry no `run` and are read as run 1 at every layer (schema, fold, DTO); records written after always carry it.

## Old-behaviour tests — rewritten in this commit, not patched (issue-groom law 8)
- `api/lab.test.ts` two runs of one arm sharing one `worktreePath` → now run 1 / run 2 with distinct leaves, and the test asserts the run numbers and the distinct worktrees. The old shape is **illegal by construction** now (`armLeaf`), which is why the fixture could not simply keep its data.
- `parseSingleArmForkStdout` tests → `parseForkStdout`: the real arm number (`arm 2`), a multi-run block with the launcher's extra session line, and a header with no run block → `null`.
- `dispatches free-form arms independently … never sharing … a fork id` asserted **three fork ids for three arms** — the fragmentation, pinned as if it were a law. Now: three worktrees, one `forkId`.

## What this PR does not do
- No floor/confound work (R2, #323), no outcome or `/api/lab/measure` (R3, #324), no ceiling override (R6, #324), no web changes at all — `packages/web/src/lab/` reads the additive fields when its waves come.
- `lab/namespace-law.test.ts` **unchanged** (the law w1 and w2 must not disturb). `.windows-known-failures` gains no entry.
- Not run in a browser; the console's `LaunchPanel` sends no `runs` yet, so a real launch today is `runs = 1` and byte-identical in behaviour to before, except that its arms now share a forkId.

## The golden era corpus moved by one line — deliberately
`eras.test.ts` pins era-1's fold **byte for byte**, and `initialForkState()` now folds to one more empty key. The snapshot was re-blessed with `CAPTURE.md`'s own command (never `vitest -u`), the move checked by diff — **exactly one added line, `"byArm": {}`**, no other key — and the note added in the file's established form: the reducer change is the `fork.dispatched` arm; the new fold is the correct meaning of the old log because era-1 predates the laboratory and now says so under one more name.

## Fence
As filed on #322, plus two widenings each recorded on the issue **before** the change: `packages/core/src/reduce.test.ts` (the fold's own tests; one `describe` added) and `packages/core/src/eras/era-1/session-state.snapshot.json` + `packages/core/src/eras/CAPTURE.md` (the golden fold's one-key move and its note). `scripts/fence-lint.sh 322`: 21 paths, PASSED. 16 files changed — the fence and nothing outside it.

## Gate
`npm run typecheck` · `npm run lint` · `VITEST_MAX_WORKERS=6 npm test` — typecheck **0 errors** · biome **Checked 902 files. No fixes applied.** · suite **Test Files 417 passed (417) · Tests 7756 passed | 3 expected fail | 2 skipped (7761)** — run on this exact tree, staged, before the commit that carries it (the 3 expected-fail and 2 skipped are the baseline's, unchanged)

Two laws taught me something on the way and are worth a line each: `corpus-eol-law` re-hashes the pinned era files against their **index** blobs, so a pre-commit run with the re-blessed snapshot unstaged reads as a violation — the gate above was run on the committed tree. And `doc-citation-law`'s ceiling derives the tracker's live maximum from **merged PR numbers on main**, so a fresh issue number cannot enter the corpus until something above it lands — the `CAPTURE.md` note therefore cites `prd-53 ruling 1`, not `#322`. Both laws are right; neither was changed.


ciaran-slow (approved):
**Approving.** The keystone holds, and the laws bite hard. Every mutation in the PR table was re-run here rather than taken on trust.

### The mutations, re-derived (EXECUTED)

| mutation | result |
|---|---|
| `armLeaf` swaps arm and run | **red** — `the second run onward carries its number…` and `restores arms × runs realities under ONE forkId…` |
| `run: p.run ?? 0` | **red** — `reads a record written before runs existed as run 1` |
| the launch stops passing `--fork-id` | **red — 8 tests**, including `one launch is ONE experiment: every arm and every run folds under the forkId the launch minted` |

That third one is the keystone and it is well guarded. The `parsed.forkId !== forkId` check is doing real work: with `--fork-id` removed, the failure surfaces as a named, explained error rather than as arms quietly filed under ids the launch cannot account for. Reading the count back from the log the launch wrote, rather than from the result the launch returned, is what makes that law worth having.

### The design calls that are right

**`-run-1` elided** is the decision that makes this additive rather than a migration — every existing worktree, handle and fixture keeps its spelling, and the run grammar only appears where nothing existed before to rename. Having `armLaneHandle` delegate to `armLeaf` so handle and worktree leaf cannot drift is the same instinct applied twice.

**Arm-major, run-minor** with the stated consequence — a partial dispatch leaves whole arms behind it, never half of two — is the right loop order for something that spends money, and it matches how the failure is reported.

**`arms × runs ≤ MAX_ARMS`.** Reading prd-41 ruling 4's ceiling as bounding spending lanes rather than arms is the correct reading, and the error message states the arithmetic it refused.

On `armKey`'s collision argument: the justification given is that a fork id never holds `#`, but the property is actually stronger than that. Since `arm` is always a plain integer, splitting at the **last** `#` recovers `(forkId, arm)` uniquely whatever the fork id contains — so the key is injective regardless. Nothing to change; the guarantee is just sturdier than the comment claims.

### One hardening note — not a blocker, and not a vulnerability

`--fork-id` is validated only as non-empty ([`cli/lab-fork.ts`](packages/server/src/cli/lab-fork.ts)), and the value then flows into `path.join` via `armLeaf`. The shape (EXECUTED, on the extracted `armLeaf`/`armWorktreePath` logic):

```
normal : /data/lab/worktrees/fork-abc-arm-1
escape : /tmp/pwn-arm-1        // forkId = '../../../../tmp/pwn'
```

**This is not remotely reachable, and I checked before writing it down:** `parseLaunchRequestBody` destructures only `lane`, `checkpointId`, `arms`, `runs`, and `launchExperiment` mints its own `fork-${randomUUID()}`. So no HTTP client can supply a fork id — the only route in is the operator's own shell, where they can already write wherever they like. Filed as hardening you may want (a `fork-`-prefix or `[A-Za-z0-9-]` guard alongside the non-empty check), not as something holding this PR.

### One observation, offered not asserted

`RUN_BLOCK_RE`'s `(?:.*\n)*?` between the worktree line and the launch line will span into the *next* run's block if a launch line is ever missing, pairing one arm's header with another's launch state. It is unreachable today — `cli/index.ts` prints the `launch` line unconditionally per arm — so this is a note about a coupling, not a defect. Your call whether it wants an anchor.

**No blocking findings.**

**Ordering:** #331 carries the prd-53 PRD this commit cites throughout. Nothing binds them mechanically (`doc-citation-law` is 49/49 green with this PR merged and no prd-53 document present — EXECUTED), but landing #331 first keeps the citations honest.

---

### How this was checked

Issue #334, opened 2026-09-07
prd30 w3: every surface that discloses is proven per file, not per directory


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/app/title-residue-law.test.ts`
- `packages/web/src/lane-page/RunView.test.tsx`
- `packages/web/src/lane-page/LanePage.test.tsx`

## Why

#220's second law pairs every directory that renders `<Disclosure>` with a test
in that directory calling `discloseText`, which opens the card by mouse and by
keyboard and throws unless the two markups match. It is **directory-granular**,
and the re-verify pass proved that granularity has a hole rather than arguing
it — EXECUTED, on `777d9c2b`:

> deleted `RunOutcome.tsx`'s `<Disclosure>` wrapper entirely and the full
> `lane-page/` + residue-law suite (60 tests) stayed green

`packages/web/src/lane-page/` holds **four** files rendering the card —
`PageHeader.tsx`, `SpendDetail.tsx`, `RunSpine.tsx`, `RunOutcome.tsx` — and
#220 covered two. The law reads the directory as proven and cannot see the
other two.

The law's own docstring argues for directory granularity: a test file's name
does not reliably say what it renders (`PageHeader.tsx` is covered by
`LanePage.test.tsx`), so a per-file rule needs a hand-maintained map, which is
a second place to forget. That argument is real but it was written by the
author of the law to justify the author's own choice, and this is the case it
loses to.

**This is not #220's regression.** `RunSpine.tsx` and `RunOutcome.tsx` adopted
`Disclosure` before #220 and were already uncovered on `main`; #220 neither
introduced the gap nor widened it. What #220 introduced is a law that reports
the directory as proven, and a law that reports success over an unproven
surface is the failure mode AGENTS.md already records for `cmd_orphans`.

## Definition of done

- `RunSpine.tsx` and `RunOutcome.tsx` each have a test that opens their card
  through `discloseText`, asserting a specific string the surface derives —
  not merely that a card rendered.
- The law is decided, in the same commit, one way or the other: either
  tightened to per-file with the map made cheap to maintain (a new adopter
  fails until it names its test), **or** left directory-granular with the hole
  named in its docstring and this issue cited. Both are defensible; leaving the
  docstring's current argument standing unqualified is not, because the case it
  loses to is now known.
- Whichever is chosen, **prove it with a mutation**: delete a `<Disclosure>`
  from a covered surface and show the suite going red. The re-verify's method
[...]

PR #335, opened 2026-09-07
prd30 w1: every mark explains itself on hover and on focus

### LINKED ISSUE #220: prd30 w1: no surface explains itself with a native title attribute

ciaran-slow:
Closes #220

prd-30 wave 1, one issue, one commit. The last **50** native `title=`
attributes in the instrument's own surfaces are gone, and the shared disclosure
card is behind every mark instead.

A native tooltip fails prd-04 ruling 1's layman bar four ways at once: OS-delayed
by about a second, invisible to a keyboard, absent on touch, and usually one word
long. Charter §6 is binding — whatever hover discloses, focus discloses — so each
of these was a promise the instrument was not keeping.

**37 became cards, 9 became `aria-label`, 4 use a new inline trigger.** The nine
were not conditions: four duplicated an `aria-label` already present on an
already-focusable element, and five were sentences naming what a control *does*,
which is an accessible name. The issue's own DoD forbids re-spelling a one-word
tooltip into a card-shaped one.

## Why this is one commit and not fifty

Four components — `Vital` (×12), `Figure` (×5), `Cell` (×3), `Missing` (×2) —
each funnelled a dozen `title` props into **one** DOM attribute. Converting those
four carried 22 marks. The rest went through their format modules, so
`*Title(): string` became `*Disclosure(): DisclosureContent` and the surface
stopped flattening a structure it had already been handed: `stateTitle` was
literally `disclosureLines(selectLaneCondition(…))` glued back into one string,
and is deleted.

## The correction the issue body needs

**#220 says "97 native `title=`". That number is a bare grep and it overcounts.**
In the fenced directories there were 77 raw hits, of which **50** were DOM
attributes; 26 were props named `title` on local components, and two of those
(`<PanelFrame title="Fleet">`) render a visible heading and have nothing to do
with tooltips. The residue law below resolves the element each `title=` actually
sits on for exactly this reason — a law that repeated the miscount would fail on
a legitimate heading prop, and the first person it lied to would disable it.

## Every card carries a real age, sourced and not invented

`disclosureLines` refuses a missing fact and refuses a non-finite or negative
elapsed time, so all 37 needed a genuine observation. Each is taken from what the
surface already measured — `lane.ageMs`, `chip.waitMs`, `now - row.lastTs`,
`now - touch.lastTouchedAt`. Where a reading is re-derived from the fold on every
tick and has no "since" of its own, it reports `elapsedMs: 0`, which is core's
own documented register for a continuously-true fact
(`packages/core/src/selectors/condition.ts`) and reads as "confirmed just now"
rather than as a fabricated span.

## What is the sibling case?

**The four funnels.** Converting `<Vital>`'s twelve call sites and missing the one
`<div title={title}>` they render through would have left every vital on a native
tooltip while the diff looked complete. All four are converted, and the residue
law proves it rather than my counting.

**A card around a cell that contains a control.** Hit in `ledger/index.tsx`, where
the tokens cell holds the exemplar jump button — a trigger around the cell would
nest a button in a button. The card wraps the figure; the button gets its own
inline-mode card. `collisions/`, `attention/`, `trace/`, `why/` and
`drawer/Conversation` share the shape and all use the inline trigger.

**`theme/kind.test.tsx`'s `HAND_SPELLED_TAGS`** asserts a count *and* an exact
file set over `why/WhySurface.tsx`, which this commit edits. It did **not** move —
no hand-spelled kind tag was added or retired — so neither half was touched.
Both allowlists the issue named were already empty and stayed empty; **no
allowlist grew.**

Three run, each EXECUTED, each reverted:

1. **Re-added one native `title=`** to a converted fleet cell → the residue law
   goes red naming `panels/fleet/index.tsx:345 <td>`.
2. **Deleted the `?? 0` absence arms** from `fleet/format.ts` → three tests fail
   with `DisclosureError: … is not a measured age (null)`. This is the one that
   matters: it proves the age sourcing is load-bearing, not decorative.
3. **Neutered `requireAge`'s finite check** → four tests red.

**What they still survive**, stated plainly: the *wording* of a why or a remedy.
Several assertions check substance in parts rather than exact strings, because
the sweep genuinely re-phrased some sentences. A card that kept its evidence and
its age but said something subtly wrong would pass.

## Two laws, and the second exists because the first cannot see its own blind spot

`app/title-residue-law.test.ts` holds both.

The **first** finds any native `title=`, resolving the element each one sits on.

The **second** pairs every directory rendering `<Disclosure>` with a test in it
that opens a card. This was **added after the first verify pass**, which caught
the real defect in this PR: a surface can retire its tooltip, adopt the card, and
have nothing ever prove a keyboard user sees what a mouse user sees — while the
first law stays green throughout, because there is no `title=` left to find. That
is what this shipped to review with, on nine surfaces, under a build report
claiming parity was asserted everywhere. One of those surfaces carried a comment
pointing at a test that did not exist, which is worse than the gap: a false
pointer to coverage stops the next reader looking, the same failure AGENTS.md
records for `cmd_orphans`.

Fixed in the amended commit: nine parity tests added, the false comment replaced
with the assertions it claimed, and the second law added so the gap cannot
reopen. Mutation-proven — stripping a directory's parity calls turns it red
naming that directory.

## Known limitation, filed as #334

The second law is **directory-granular**, and the re-verify pass proved that has
a hole rather than arguing it: it deleted `RunOutcome.tsx`'s `<Disclosure>`
wrapper entirely and 60 tests stayed green. `lane-page/` holds four adopting
files and two are covered. **Not this PR's regression** — `RunSpine.tsx` and
`RunOutcome.tsx` adopted the card before #220 and were already uncovered on
`main` — but the law reports the directory as proven, so #334 owns deciding
per-file granularity and covering those two.

## Deviations from the plan

1. **The `~` (inferred) mark kept its own card** instead of folding into the
   STATE card's evidence. `selectLaneCondition` lives in `packages/core`, outside
   this fence; widening into core for one clause was worse than a small card.
   Recorded in `inferredDisclosure`'s docstring with what should happen instead.
2. **`why/NearestEntry.tsx:90` became a card, not an `aria-label`** as planned.
   It is a `<div>`, and an `aria-label` on a non-interactive, non-landmark element
   is ignored by assistive tech — the plan's disposition would have left it
   pointer-only, the exact defect.
3. **Screenshots not regenerated**, checked rather than assumed: Tailwind's
   preflight zeroes button padding, border and font, and the live DOM measures
   every fleet row at 48px with each trigger at `padding: 0 / border: 0`. The
   change is in what hover and focus *do*, which no resting screenshot captures.
4. **Some user-facing wording changed** — one tooltip string is now label +
   why-with-evidence + remedy, and a remedy that was a trailing clause is now a
   named act with its command kept separate so it can be copied. In `CHANGELOG.md`
   under Changed.

## Fence widenings — all recorded before the change

The fence was **narrowed** first, before any code moved (recorded on the issue and
its timeline): `replay/`, `lab/` and `concierge/` came out after `fence-lint`
found them overlapping live lanes #216 and #235, and the overlap was not a
granularity artefact — those lanes' law tests `readFileSync` the very surfaces a
sweep there would edit.

Beyond the narrowed fence this commit touches:

1. **`packages/web/src/disclosure/`** — the `trigger` mode, its tests, and a
   test-only `testing.ts`. prd-30 wave 2's (#221) territory; #221 is Backlog and
   unstarted, so a rebase risk rather than a live collision. **The largest of the
   four.**
2. **`docs/adr/`** — ADR-0040 and its index row (the ADR log law requires it).
3. **`CHANGELOG.md`** — user-visible change.
4. **Test files in the fenced directories** — 19 assertions read `.title` on
   elements that no longer have one.

`scene/`, `theme/theme.css`, `replay/`, `lab/` and `concierge/` are untouched.

## ADR

[ADR-0040](docs/adr/0040-a-mark-that-is-already-a-control-still-discloses.md) —
`Disclosure` gains a `trigger` mode so a mark that is already a control discloses
without nesting a button. Rejected alternatives and their costs are in the record,
including the bad consequences: `role="note"` is a weaker semantic than a button,
and a mode is a place for two paths to drift.

## Gating

On the **assembled** branch, at `origin/main` `1c7d2af5`:

| gate | result |
|---|---|
| `VITEST_MAX_WORKERS=6 npm test` | 421 files / **7840 passed**, 3 expected fail, 6 skipped |
| `npm run lint` | exit 0 |
| `npm run build` | exit 0 |
| boot smoke | served `/` as HTML; `/api/meta` correctly refused without the capability token |

No cross-lane breakage. On the lane the same four were green at 419 files / 7763;
the assembled figure is higher because `main` gained tests from PRs that landed
meanwhile.

Also verified in a real browser rather than only jsdom: loaded fixture 3 in the
built app, focused a fleet-table trigger by keyboard, and read the card back —
with `aria-describedby` pointing at the card's id while open and gone on blur.

## Review provenance — read this before approving

**Two independent verify passes ran, both on Sonnet against an Opus 5 build, and
both are on the issue** ([first](https://github.com/launchpad-26/rhizomorph/issues/220#issuecomment-5575580116),
[re-verify](https://github.com/launchpad-26/rhizomorph/issues/220#issuecomment-5575708346)).
The first returned a blocking finding; it was fixed and answered with the sha.
The second returned *ready to merge*.

**They are not recorded in the verify ledger**, because `verify-ledger.sh` and
`pr-open.sh` are operator-local and are not present on this machine. AGENTS.md's
first condition for opening a PR is the ledger, not a green suite, and this PR
does not satisfy it — so this paragraph is the honest escape `pr-open.sh
--unverified` would have written into the body, put here where a reviewer sees
it. The review happened; the *record* of it does not exist. Treat that as the
gap it is.

One more thing a reviewer should weigh: **the plan, the build, and the fix for
the first verify pass's findings were all written in one context.** The pipeline
wants those cold and separate. The verify passes were genuinely independent; the
authoring was not.


gabriel-canaan (changes requested):
# Review of #335 (rhizo-review) — one substantive finding for you, two fixes pushed

Read commit-by-commit on the branch **with `origin/main` merged in** (it was 9 commits behind), which is a different tree from the one your gating table measured.

- **The rebase preserved the reviewed sweep exactly.** `8ee1b498` is `777d9c2b` rebased onto `1c7d2af5`. I interdiffed the reviewed patch (`e5895a46..dbdfb307`) against this PR's patch (`1c7d2af5..8ee1b498`) across all 53 paths — the same 53 files, byte-identical, but for one hunk (finding 1). Nothing else was lost in the rebase.
- **Post-merge gates**, re-run from a clean install: `npm run typecheck` 0, `npm run lint` 0, `npm run build` 0, `VITEST_MAX_WORKERS=6 npm test` → **423 files / 7885 passed**, one failure, which is main's (finding 4). Your body says 421 / 7840; that tree no longer exists and the higher number is the honest one.
- **Law 1 bites — reproduced independently, not read off your report.** Planted `title="…"` on the `<th>` at `panels/collisions/index.tsx` → red, naming `panels/collisions/index.tsx:129 <th>`, element resolved rather than grepped. Reverted; tree clean vs HEAD.
- **`disclosure/testing.ts` is a genuine parity assertion**, read in full: it opens by mouse, captures, closes, opens by focus against a freshly mounted node, and throws unless `outerHTML` matches. Every call is a charter §6 assertion, as you claim.
- **Your ledger sibling case is right.** `panels/ledger/index.tsx:205-211` — the button-mode card wraps the figure and `ExemplarJumpButton` is its *sibling*, not its child. No nesting.
- **#334 is real**, open, milestone prd30. The known limitation is filed as stated.
- I read all seven `trigger="inline"` sites. Six wrap their own control correctly, five carrying a comment saying why.

## 1. The rebase dropped @KelliherL's review commit — EXECUTED, fixed

`git merge-base --is-ancestor dbdfb307 8ee1b498` exits 1. `panels/attention/AttentionStripView.tsx:229` again reads *"The `title` above still holds the full text."* — the stale comment his review found and fixed. It is the single hunk in the interdiff above: the rebase onto `prd30-w1-open-hand` carried your commit and left his behind.

Cherry-picked as **`c4140a7a`**, his authorship preserved. Nothing else changed.

## 2. The parity law certifies its own directory — EXECUTED, mutation-proven both ways, fixed

`directoriesWithParityTests` credits a directory when some test in it contains the text `discloseText(`. `withoutComments` blanks comments but **not string literals** — and this file carries that text in two of them: the remediation sentence it prints on failure, and the rigged fixture beside the new test. The file lives in `app/`, so `app/` read as proven on the strength of the law's own prose.

- Replaced `discloseText(` with `String(` in `app/Nav.test.tsx` — the only real parity call in `app/`, against `app/Nav.tsx`, which does render `<Disclosure>`. The law stayed **green, 10/10**.
- Removing only this file from the scanned set put `app` back on the unproven list — which isolates the cause to the self-match rather than to the mutation.

That is the failure the second half exists to make impossible, in the one directory that hosts it. It is **not** the granularity limitation in #334; a directory here is self-certifying regardless of granularity.

Fixed in **`9eb45918`**. The exclusion is the rule `shippedFiles` already applies one screen up — *a law test must be able to write the pattern it looks for* — which you applied to law 1 and not to law 2. `THIS_FILE` is derived from `import.meta.url` rather than spelled, so it cannot rot if the law moves beside `one-card-law.test.ts` as its own docstring contemplates, and a second test asserts the derived path is actually in the walk, because an exclusion resolving to a file the walk never yields excludes nothing while reading as fixed.

After the fix: the same mutation goes **red naming `app`**; deleting the `file.name === self` guard turns the new fixture red. Both reverted, tree verified clean vs HEAD.

**This is my fix, so it has had exactly one pair of eyes.** Please confirm it before this merges.

## 3. `trace/TraceTree.tsx:92` nests the inline trigger inside a `<button>` — EXECUTED + REASONED. Yours to answer; I did not fix it

ADR-0040 answers *"a mark that already **is** a control."* This site is the sibling it does not consider: *a mark **inside** a control.* The `<Disclosure trigger="inline">` at `:92` sits within the `<button data-testid="trace-interaction-toggle">` opened at `:79`.

Measured at runtime with the fixture `TraceTree.test.tsx` already uses:

```
trigger is a DESCENDANT of the toggle button: true
trigger: <span role="note" tabindex="0">
descendants of the button carrying tabindex: 1
rows expanded before any click: 0        card open before any click: false
--- after ONE click on the disclosure trigger ---
rows expanded (button's onClick fired): 4
aria-expanded on the toggle: true
card pinned open (disclosure onClick fired): true
```

So one click on the token headline both pins its card and expands the whole interaction — *"two click handlers racing for the same tap"*, which is one of the three harms ADR-0040 names for nested buttons. REASONED, from the HTML content model for `<button>`: "there must be no interactive content descendant **and no descendant with the `tabindex` attribute specified**" — so the focusable `role="note"` span there is invalid markup, the same defect class the mode was introduced to avoid, relocated rather than removed.

Nothing catches it: `TraceTree.test.tsx:50` opens the card by hover/focus and never clicks it.

I have not fixed this because the repair is a design choice in ADR-0040's territory, not a mechanical one — moving the mark out of the toggle changes the row's layout, and suppressing the click inside inline mode changes the mode for the six sites that are correct. It is also the one inline site with no comment explaining the choice, which reads as the one that was not reasoned about.

## 4. `main` is red, and it blocks landing this — not your defect. EXECUTED

`packages/web/src/lab/no-live-fleet-law.test.ts:196` pins `'': 5`; `8f7a60ec` (#324) added `lab/measure.ts`, making it 6. Reproduced on **plain `origin/main` at `20f63b6e`**, with this PR nowhere in the tree, and the last two CI runs on `main` are red. It is the only failure in my post-merge suite. Reported to the operator; out of this PR's scope and untouched here.

In inline mode the trigger span carries `tabIndex={0}`, and focusing the wrapped control **already** opens the card — React's `onFocus` is `focusin`, which bubbles. EXECUTED: focusing only an inner `<button>` gave `card opened by focusing the INNER control alone: true`, with two tab stops in the tree (`span[role=note]`, `button`). So for the five sites that wrap a real control, the wrapper's own tab stop is not what satisfies charter §6 and doubles the tab stops across those rows. It *is* load-bearing at `drawer/Conversation.tsx:226` and `trace/TraceTree.tsx:92`, where the children are not controls. Possibly a note for #221 rather than anything here.

## Not run

The boot smoke, the browser check, and `scripts/gate.sh` (the operator's step). No accessibility checker. CI: all ten legs on `8ee1b498` passed — I read `gh pr checks 335` rather than the badge — but the two pushed commits have not been through CI yet.

ciaran-slow:
## Findings answered (2026-09-08)

Every finding has an answer; the one that was mine to make is fixed at `f7c8f6c2`.

**1. Dropped review commit** — accepted as pushed (`c4140a7a`). Nothing further.

**2. The parity law certified its own directory** — your fix `9eb45918` confirmed by a second pair of eyes, **EXECUTED** on the branch head in a fresh checkout: replacing `discloseText(` with `String(` in `app/Nav.test.tsx` turns the law red naming `app` (`expected [ 'app' ] to deeply equal []`), and the file is green as shipped. The exclusion bites and it is not a no-op. Accepted.

**3. `trace/TraceTree.tsx:92` nested the inline trigger inside the toggle** — **fixed at `f7c8f6c2`**, on the design side you left open. The mark now renders *beside* the toggle in a flex header: the button keeps the arrow and the durations, the token headline stands next to it as inert text and takes the **default** trigger — so it is its own button and the only one on the mark. That is the option that leaves inline mode untouched for the six sites where the control is the mark's child; suppressing the click inside inline mode would have changed the mode for all of them to fix one. `TraceTree.test.tsx` asserts the negative of each harm you measured, against the real tree: the trigger is not a descendant of the toggle; the toggle has no `[tabindex]`, `button` or `role="note"` descendant; a click on the trigger pins the card (`data-open="true"`) and leaves `aria-expanded="false"` with zero rows; the toggle then expands on its own. ADR-0040 is append-only, so it carries a dated note recording the boundary the decision turned out to have (a mark *inside* a control is not a mark that *is* one) and correcting its consequence count to six inline sites.

Gates on the branch at `f7c8f6c2`, Node 22.22.2: typecheck clean · biome 908 files clean · `VITEST_MAX_WORKERS=6 npm test` 421 files / 7843 passed / 3 expected fail / 6 skipped · build clean.

**4. `main` is red** — confirmed **EXECUTED** on plain `origin/main` at `20f63b6e`: `lab/no-live-fleet-law.test.ts` pins the root at `'': 5` and #324's `lab/measure.ts` makes it 6. This is also why this PR's macOS leg went red: CI builds the merge ref, so main's failure lands in every open PR. Out of this PR's scope; a one-line fix-forward branch is prepared for the operator separately.

**Your call item (double tab stop in inline mode)** — agreed it belongs to #221, where the inline trigger's `tabIndex` can be made conditional on whether the wrapped child is itself focusable. Not touched here.

gabriel-canaan (commented):
# Re-review of #335 (rhizo-review, round 2) — all three findings closed, nothing new blocking

Read again on the branch head `f7c8f6c2` **with `origin/main` merged in** (main has moved 9 commits since round 1, including the fix for finding 4). Every claim below is `EXECUTED` unless marked otherwise.

## Round 1's findings, each closed

**1. Dropped review commit** — closed. `c4140a7a` is on the branch, @KelliherL's authorship preserved, and `panels/attention/AttentionStripView.tsx` no longer carries the stale comment.

**2. The parity law certified its own directory** — closed, and closed *properly*. This was my fix (`9eb45918`), so it needed eyes that were not mine, and it got them: you re-ran the mutation independently on the branch head in a fresh checkout and reported the law going red naming `app`. That is the rung-2 confirmation this skill blocks on. Recorded here so the next reader does not have to reconstruct it: **the second pair of eyes on `9eb45918` is your comment, not the verify ledger.**

**3. `trace/TraceTree.tsx` nested the inline trigger inside the toggle** — **fixed at `f7c8f6c2`, and I certified the fix by mutation rather than reading it.**

Restored `TraceTree.tsx` from `f7c8f6c2^` — the pre-fix nesting, with your new test left in place — and ran the file:

```
FAIL  src/trace/TraceTree.test.tsx > TraceTree > discloses the tokens mark beside
      the toggle, never inside it — one tap, one handler (review of #335)
AssertionError: expected true to be false
 ❯ src/trace/TraceTree.test.tsx:133:38
   133|     expect(toggle.contains(trigger)).toBe(false)
Tests  1 failed | 7 passed (8)
```

Red on exactly the assertion that names the defect, not on a bystander. Restored with `git checkout HEAD --`, unstaged, and `git diff HEAD --quiet` confirmed the tree came back.

The mechanism is the right one, not just a passing test. You chose to move the mark **out** of the toggle rather than suppress the click **inside** inline mode, and that is the choice that does not change the mode for the six sites where it is correct — I checked the count rather than taking it: `grep 'trigger="inline"'` over non-test `packages/web/src` returns **6** sites (`attention` ×2, `ledger`, `collisions`, `drawer/Conversation`, `why/WhySurface`), which is what ADR-0040's dated note now says. The note is accurate and the ADR stayed append-only.

I also asked the question this skill says to ask of a repair — *what does the changed file now participate in that it did not before?* `TraceTree.tsx` moved from the inline trigger to the **default** one, so it joins whatever counts default-trigger buttons, and `font-mono text-inst text-(--ink-body)` moved from the toggle onto a new parent `<div>`. Nothing went red: the full post-merge suite is green, including the three count-and-anchor laws `.swarm/coupling.txt` warns about (`theme/tokens.test.ts`, `theme/legibility.test.ts`, `theme/kind.test.tsx`), each of which fails in both directions and so would have caught a class or literal moving.

Charter §6 parity on the relocated mark still holds through the same helper: `TraceTree.test.tsx:50`'s `discloseText(headline)` finds the trigger by test id rather than by element, so it exercises the new `<button>` exactly as it exercised the old span, and it still refuses to return unless the hover and focus markups are identical.

**4. `main` was red** — closed by someone else. `3cfa15d3` (#349) fixed the `lab/no-live-fleet-law.test.ts` count pin and landed on `main` as `f3107e19`, 43 minutes *after* this PR's CI run. That is why the bar above is red on `build-test-boot` and `windows-suite` and why nothing here is at fault: CI built a merge ref against a red `main`.

## Post-merge gates — re-measured, and the numbers moved in your favour again

Clean install, `origin/main` merged in, node v22.22.2:

| gate | result |
|---|---|
| `npm run lint` | exit 0 |
| `VITEST_MAX_WORKERS=6 npm test` | **423 files / 7900 passed**, 3 expected fail, 2 skipped — **0 failed** |

Your comment on `f7c8f6c2` reports 421 files / 7843. Both are honest; that tree no longer exists, and **423/7900 with zero failures is the current one**. The single failure that stood in round 1 is gone because main fixed it, so this PR now has a genuinely clean suite for the first time.

## Not run, stated rather than left out

The boot smoke, the browser keyboard check, and `scripts/gate.sh` — that last one is the operator's landing step and a lane or reviewer running it is the #408 failure mode. No accessibility checker. CI on `f7c8f6c2` I read with `gh pr checks 335` rather than off the badge; the four red legs are all main's lab count, and re-running them after a merge with current `main` is what will clear them.

Nothing blocking, and I am not asking for a change to either:

- The double tab stop in inline mode — agreed it is #221's, as you said. Noted here only so it stays findable.
- **#334's granularity hole is now the only thing in this PR that a green bar misrepresents.** `lane-page/` holds four adopting files, two covered, and the second law reports the directory proven. It is filed, it is not this PR's regression, and both uncovered files predate #220 on `main`. Fine to land ahead of it; worth not letting it sit long, because the law reads as coverage it does not have — which is the failure mode `cmd_orphans` is the standing example of in `AGENTS.md`.

## Verdict

**Ready to merge, from my side.** All three findings answered, the one that was yours fixed and mutation-certified, gates green on the tree that will actually land. Two things stand between this and `main` and neither is a code change: a re-run of CI on top of current `main`, and the operator's call — including the ledger record, which this PR still does not have and honestly says so in its body.

PR #336, opened 2026-09-07
prd53 w2: one floor and one confound clause live in core; a measured outcome travels per run over a gated route; the launch ceiling refuses by name and override

KelliherL:
**prd-53 wave 2 — two lanes, two commits, one PR.** Closes #323. Closes #324.

After wave 1 an experiment was one forkId and an arm held r runs. Two things still kept the comparison surface honestly empty: the CLI and the console read *different floors* (arms vs completed runs, neither able to see the other), and **no outcome existed on the wire** — measuring was CLI-only, and the web carried an arm-level outcome nothing ever populated. This PR closes both, and makes the launch ceiling an operator's declared act rather than a number with a "raise it" invitation.

## Commit 1 — `#323` one floor and one confound clause live in core, and both surfaces read them (ruling 2)
- `packages/core/src/lab/laws.ts` (new, ADR-0003 pure): `canSummariseArm` (completed runs per arm gate a **summary**), `canRankArms` (arms gate a **cross-arm claim**), `dimensionsOf` / `isCleanlyControlled` / `confoundVoice`, and `COUNTERFACTUAL_CLAUSE` — *what actually happened is one observation, not a distribution*.
- `lab/compare.ts` re-exports `MIN_ARMS_TO_RANK` from core; **`rankable` is gated on distinct arms, not rows** — since #322 each run is a row, so three runs of one arm were "rankable"; rows carry `run`; the table speaks the confound voice, the counterfactual clause below the floor, and a per-arm summary verdict once an arm holds more than one run. A single-run fork prints exactly what it printed before.
- `web/lab/compare/summarise.ts` reads `canSummariseArm`; the number is no longer spelled there.
- **Grep law** (`compare.test.ts`): both consumers import the floor from core, neither restates the number (`MIN_… = 3`, `.length < 3`).

## Commit 2 — `#324` a measured outcome is typed and travels per run over a gated route; a launch above the ceiling refuses by name and override (rulings 3, 6)
- **`fork.measured`** (new, additive, source `lab`): one per run measured — `verified`, `verifiedDetail`, `verifyCommand`, `commits`, typed `source`. Cost and duration stay the fold's own so two numbers about one run cannot disagree. `ForkState` gains `measurements` + `latestOutcomeByLane`; a re-measure appends and re-indexes, never erases.
[...]

Issue #330, opened 2026-09-07
prd53 w5: the lab's documents say what its code does


KelliherL:
## Fence (may touch ONLY)
- `docs/architecture.md`
- `docs/user-guide/the-lab.md`
- `docs/roadmap.md`
- `docs/vision-the-lab.md`
- `docs/design/charter.md`
- `docs/prds/prd-30-the-open-hand.md`
- `docs/prds/done/prd-31-the-bracketed-voice.md`
- `docs/prds/done/prd-32-the-readable-instrument.md`
- `docs/prds/prd-34-the-doorstep.md`
- `docs/prds/done/prd-35-the-operators-hand.md`
- `docs/prds/done/prd-36-the-fleet-surface.md`
- `packages/web/src/lab/the-lab-guide-law.test.ts` (new file)

## Blocked by
#322, #323, #324, #325, #326, #327, #328, #329. Last on purpose: a sweep that runs earlier re-lays ground the other waves are about to dig.
Sibling lanes hold other surfaces: none run beside this one. **`docs/design-notes/lab-launch-ceilings.md` is prd-50's (#237, #238) and is not touched here.** `packages/web/src/lab/branching/perf.test.ts` is wave 4's.
**Model:** sonnet. **Wave:** 5.

## Definition of done
prd-53 rulings 9 and 10.

- `docs/architecture.md` gains the console section it has never had (its walkthrough runs prd13→prd15).
- `docs/user-guide/the-lab.md` covers the 400 (`MAX_ARMS`) and 503 (lock ceiling) refusals an operator actually sees, the estimate's basis, and the measure route.
- `docs/roadmap.md` gains prd-46 through prd-53.
- The **seven** prd-28 citations — `packages/web/src/lab/` *"is prd-28's territory"* in prd-30, 31, 32, 34, 35, 36, and `charter.md`'s — each get a dated note pointing at prd-53. Never a rewrite; a number is an identity.
- `docs/vision-the-lab.md`'s recovered-history header is checked against what shipped and amended if more moved under it.
- **Ruling 10's residuals named with owners or honestly without:** the five `lab/` files on `.windows-known-failures` (owner: prd-25's Windows work); the mid-tool-call cut (unowned, failure stated); `FORK_EXEC_TIMEOUT_MS` and the `install: false` default (unowned).
- **The law:** one assertion per behavioural sentence in `the-lab.md` — prd-43's *the claim is a test*, finally reaching the lab.

[...]

Issue #214, opened 2026-09-02
prd14 w3: the recordings library lists a comparison as its own kind and reopens it


ciaran-slow:
## Fence (may touch ONLY)
- `packages/web/src/recordings/`
- `packages/web/src/lab/compare/`
- `packages/contract/src/contract-coverage-law.test.ts`
- `packages/contract/src/lab-comparisons.contract.test.ts` (new file)
- `packages/contract/src/lab-comparison.contract.test.ts` (new file)
- `packages/contract/src/lab-comparison-save.contract.test.ts` (new file)
- `packages/web/src/replay/mutating-calls-law.test.ts`
- `CHANGELOG.md`

**Fence widened 2026-09-09, before any change**, from the two web directories alone. Every
path added is one this issue's own Definition of done already obliged and no fence named —
the contract law rows and their contract tests, the mutating-modules law row the save
creates, and the one loose sentence in `docs/architecture.md`. Recorded here so
scripts/fence-lint.sh sees the real footprint before dispatch, and recorded again on the
PR before the change, as AGENTS.md requires.

**`docs/architecture.md` was in this widening for about an hour on 2026-09-09 and was taken
back out before dispatch.** scripts/fence-lint.sh 214 330 reported a real OVERLAP: #330
(prd53 w5) is In review and holds that file to give the console the architecture section it
has never had. The nit below — a gated-reads sentence naming the comparison *save* inside an
enumeration of *reads* — is left unfixed here rather than fought over. It belongs to whoever
holds that file next, and it is recorded in the obligations section so it is not lost.

`packages/server/src/comparisons/` is **not** in this fence. It was proposed as a widening
on 2026-09-08; that work is now #376 (prd14 w2) and runs a wave earlier — see Blocked by.

## Blocked by

**#376 (prd14 w2) — the two comparison parsers accept and refuse the same bytes.** This
issue's second Definition-of-done bullet is unsatisfiable until it lands: the server and web
v1 parsers disagree on the run vocabulary, so a comparison saved from the surface either
400s or is stored without its verdict and cannot be reopened. Re-executed on `main` at
`88df0484`, 2026-09-09 — the divergence is live. That fix is entirely server-side and
outside this fence, which is what makes the two a stack rather than a bundle.

**Regroomed 2026-09-09. This issue is w3, not w2** — renumbered when #376 was filed between
it and wave 1. Read `w2` on any prd14 citation written before this date as this issue.

### The three blockers this issue used to name — all cleared

[...]

Issue #213, opened 2026-09-02
prd14 w1: a finished comparison is stored beside the recordings it derives from


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/comparisons/`
- `packages/server/src/api/lab.ts`
- `packages/server/src/api/lab.test.ts`
- `packages/server/src/api/index.ts`
- `packages/server/src/api/route-class-law.test.ts`

**Widened by the plan (2026-09-04).** Every prose site a new route's count moves, plus the two laws that enumerate gated reads by hand. Listed here so fence-lint sees the real footprint before dispatch; recorded again on the PR before the change, as AGENTS.md requires:
- `packages/server/src/api/gated-reads.test.ts`
- `packages/contract/src/contract-coverage-law.test.ts`
- `packages/server/src/server/mutation-guard.ts`
- `SECURITY.md`
- `packages/server/src/api/security.ts`
- `packages/server/src/api/lane-index.test.ts`
- `docs/architecture.md`
- `docs/adr/0008-localhost-only-single-origin-server.md`
- `docs/adr/0036-a-saved-comparison-is-a-sidecar-not-an-event.md` (new)
- `docs/adr/0037-the-server-parses-a-comparison-artifact-with-its-own-copy.md` (new)
- `docs/adr/README.md`
- `CHANGELOG.md`

## Blocked by
**Corrected 2026-09-04 — #23 closed 2026-09-02, so the paragraph below has cleared.** The live blocker is now **#232** (prd43 w6, In progress): its fence holds `packages/server/src/api/route-class-law.test.ts` (`scripts/fence-lint.sh 213 232` reports the OVERLAP — executed 2026-09-04), and it rewrites the count section this issue must edit into one `CLAIMS` table over twelve prose sites. Dispatch after the prd43 w6 bundle merges to `main`; the plan comment below says what to reconcile then. The original text is kept, not deleted, per the convention #214 already follows.

**#23** — prd43 w4 owns `packages/server/src/api/route-class-law.test.ts`, and that file is where this issue's cost lands. Line 99 pins `expect(ROUTE_CLASSES.length).toBe(25)`, an exact count deliberately asserted independently of the array so the walk cannot pass vacuously. Any new route reddens it. #23's whole purpose is to make counts derived from the thing they count, so landing it first is what makes this issue a one-line reconciliation instead of a fight.

[...]

PR #339, opened 2026-09-07
prd53 w3: the workspace scrubs the lab's own axis under a five-position frame; Compare, Trace and Metrics render every state the spec draws

### LINKED ISSUE #325: prd53 w3: the workspace scrubs the lab's own axis, and the frame is one switch over five
## Fence (may touch ONLY)
- `packages/web/src/lab/axis/` (new directory)
- `packages/web/src/lab/frame/` (new directory)
- **Recorded at review 2026-09-08, NOT before the change** (review of PR #339). The paths below were touched by this issue's commit and were missing from this fence. They are reconciliations the commit message names; the record, not the code, was short:
  - `packages/server/src/api/lab.ts`, `packages/server/src/api/lab.test.ts` — **#324's files (wave 2)**: `sessionByteLength` and the three new `LabCheckpointDTO` byte fields the axis places markers by
  - `packages/web/src/lab/types.ts` — **#324's file (wave 2)**: the same three fields on the web type
  - `packages/web/src/lab/api.ts`, `packages/web/src/lab/api.test.ts` — the DTO guard for those fields
  - `packages/web/src/lab/launch/LaunchPanel.tsx`, `packages/web/src/lab/launch/LaunchPanel.test.tsx` — claimed by no issue this wave: `initialCheckpointId` / `onLaunched`, so the seated marker's *fork from here* reaches the panel
  - `docs/prds/prd-53-the-lab.md` — the PRD cites the design-spec artifact

## Blocked by
#322, #323, #324, and the design-spec artifact (drawn between waves 2 and 3; its acceptance criteria are this issue's).
Sibling lanes hold other surfaces: Compare holds `packages/web/src/lab/compare/`; Trace holds `packages/web/src/lab/trace/`; Metrics holds `packages/web/src/lab/metrics/`. **This lane owns `LabPage.tsx` and therefore the wiring of all three into the page** — the siblings export, this lane mounts. `packages/web/src/replay/` and `tide/` are imitated, never imported or edited. `packages/web/src/scene/` is entered only for the frame's scene position, read-only, through its public renderer.
**Model:** opus. **Wave:** 3.

## Definition of done
prd-53 rulings 4 and 8, and S1 of the specification.

- The scrub track: %-of-session over `fork.checkpoint.sessionCutByte` and `eventIndex`; checkpoints as chapter markers — coordinates, not rows. `Scrubber`/`usePlayback` imitated inside `web/src/lab/`.
- The observability frame: one exclusive switch over telemetry · cost · scene · divergence · footprint. Telemetry and cost are one `Sparkline` in two series, solid to the playhead, dashed after, parent ghosted. Scene is one organism through `scene/`'s own renderer, reach truncated at the playhead (charter §8 coexist-by-surface — cite, do not re-argue). Divergence reads 0 % at the fork by construction; a dead arm's line ends. Footprint is overlap, never conflict.
- Every S1 state with its copy: live · empty · loading · error (never conflated with empty) · degraded · replay (the Lab nav entry disabled with its reason) · demo · **partial launch** (failed arms named at the fork marker).
- Keyboard: click seats the playhead; `←`/`→` step markers; `Home`/`End`; `1`–`5` the frame; `Esc` clears.
- **Laws (counts):** every lab surface reading %-of-session calls one function (grep law); the playhead at byte *b* renders at the same x in track, chart and canvas; a failed checkpoints read renders the error copy never the empty copy; the Lab entry is disabled with reason while a recording is loaded.

### LINKED ISSUE #326: prd53 w3: Compare renders a real per-run distribution, and every state it can be in is first-class
## Fence (may touch ONLY)
- `packages/web/src/lab/compare/ComparisonSurface.tsx`
- `packages/web/src/lab/compare/ComparisonSurface.test.tsx`
- `packages/web/src/lab/compare/compare.ts`
- `packages/web/src/lab/compare/compare.test.ts`
- `packages/web/src/lab/compare/types.ts`
- `packages/web/src/lab/compare/index.ts`
- **Recorded at review 2026-09-08, NOT before the change** (review of PR #339). The paths below were touched by this issue's commit and were missing from this fence. They are reconciliations the commit message names; the record, not the code, was short:
  - `packages/web/src/lab/compare/ExperimentComparison.tsx`, `packages/web/src/lab/compare/ExperimentComparison.test.tsx` (new) and `packages/web/src/lab/compare/fromExperiment.ts`, `packages/web/src/lab/compare/fromExperiment.test.ts` (new) — this fence named `compare.ts` / `compare.test.ts`, which do not exist; the lane created these instead, inside its own directory
  - `packages/web/src/lab/compare/attribution.ts`, `packages/web/src/lab/compare/attribution.test.ts` — handed to this lane by #323's commit message (the third copy of the confound logic), never written into the fence

## Blocked by
#322, #323, #324, and the design-spec artifact.
Sibling lanes hold other surfaces: the workspace lane owns `LabPage.tsx` and mounts this surface — this lane exports only. `summarise.ts` is wave 2's and is read, not reopened. `compare/artifact.ts` is **#213's** (prd-14) and is not touched.

## Definition of done
prd-53 ruling 7 and S2 of the specification.

- Arms side by side; each arm's runs as individual points with min–median–max; every spread recomputed when the measure switches; the dimensions line computed, never declared.
- Every S2 state with its copy: below the floor (*"1 of 3 runs completed so far — too few completed to summarise yet"*, runs shown, summary refused as a rendered state) · replicates · confounded · unmeasured · empty · error · **partial launch** (failed arm listed with its failure, excluded from every spread with that stated) · the scoring measure **once, disabled** — *"Scoring — no source yet"* — and nowhere else.
- No sort control exists. Rows are in arm order regardless of value.
- **Laws (counts):** two completed runs → summary absent, refusal present; three → min/median/max present; identical treatments → replicate copy; two-dimension arms → confound copy; rows in arm order (a mutation sorting by cost goes red); an unmeasured run renders `not-run` copy, never a number.

### LINKED ISSUE #327: prd53 w3: Trace diffs an arm against its parent, read and never copied
## Fence (may touch ONLY)
- `packages/web/src/lab/trace/` (new directory)

## Blocked by
#322, #324, and the design-spec artifact.
Sibling lanes hold other surfaces: the workspace lane mounts this surface; `packages/web/src/trace/` (the instrument's own trace furniture — `TraceGantt`, `TraceRow`, `FocusPanel`) is **imported, never edited**; `packages/web/src/drawer/useTranscript.ts` is read for the byte-offset fetch pattern and not edited.

## Definition of done
S3 of the specification.

- An arm's steps against its parent's from the checkpoint forward, each `same` / `diverged` / `added` / `absent`; the parent read at `sessionCutByte` through `GET /api/transcript/:lane?offset=N` and **never copied**.
- Every S3 state: live · at the fork (every row `same` by construction) · dead arm (rows end, nothing dashes) · empty · parent unreadable (its copy; the arm's own rows still render) · replay · partial launch (a failed arm has no trace and says so).
- Keyboard: `↑`/`↓` rows; `Enter` focus panel; `Esc` back.
- **Laws (counts):** row 0 is `same` for every arm; a dead arm's last row has no successor; the parent request carries `offset=`; no lab source file contains transcript text (grep law); a failed parent read renders its copy and the arm's rows remain.
- **Open, not ruled, implementer proposes on this issue before building:** what a *step* is for divergence purposes.

### LINKED ISSUE #328: prd53 w3: Metrics states the basis of every number
## Fence (may touch ONLY)
- `packages/web/src/lab/metrics/` (new directory)

## Blocked by
#322, #323, #324, and the design-spec artifact.
Sibling lanes hold other surfaces: the workspace lane mounts this surface; Compare holds `compare/`; `packages/core/src/selectors/spend.ts` is read, never edited.

## Definition of done
S4 of the specification.

- Every number with its basis line — *"booked $10.50 from each arm lane's own recorded rate"*; distributions on a shared scale across experiments; the verification and provenance table.
- Every S4 state: live · below the floor (an empty track labelled *"refuses to summarise"* on the shared scale — a rendered refusal, not a missing row) · no rate (*"the lane's recent rate cannot be established — no cost has been booked to it in the last hour"*) · unmeasured (`not-run`, no value beside it) · empty · **partial launch** (*"2 of 3 arms dispatched"*, spend is the spend of the two).
- No scoring column exists; none is drawn.
- **Laws (counts):** every rendered figure has an adjacent basis element (DOM law); a below-floor experiment renders a track with the refusal copy; a partial experiment's spend equals the sum of its dispatched arms only; `not-run` rows carry no numeric cell.

### LINKED ISSUE #329: prd53 w4: the lane canvas draws n organisms from real multi-run experiments
## Fence (may touch ONLY)
- `packages/web/src/lab/canvas/` (new directory)
- `packages/web/src/lab/branching/` (retired or reused — decide on this issue before touching it)

## Blocked by
#322, #325. The canvas is placed by the workspace's axis and draws real multi-run experiments — it needs both to exist.
Sibling lanes hold other surfaces: none run beside this one. `packages/web/src/scene/` is read for `ribbon.ts`/`palette.ts` primitives through their public exports and never edited. Violet stays fenced to tissue draws; selection is luminance (law 9a).
**Model:** opus. **Wave:** 4.

## Definition of done
prd-53 ruling 5 — the lane canvas is n organisms.

- An experiment is n parallel realities, drawn as n small organisms (charter §8's coexist-by-surface record — cite, do not re-argue). Growth truncates at the playhead, never extrapolated; dashed-after-playhead is the shared grammar; a dead arm's cord goes necrotic grey with a broken tip, a finished arm keeps a living cord sealed in done's dim green.
- Arms are forkable: the geometry is recursive, grouping on forkId.
- `branching/perf.test.ts`'s stale *"deliberately not mounted"* comment is corrected here if `branching/` is reused, or the file is retired with it.
- **Laws (counts):** n organisms drawn === n dispatch records folded, never a synthesised count; a dead arm's cord has no dashed continuation; selection changes luminance and never hue.
- **Open, not ruled, implementer proposes:** the frame-budget measurement's form (inherits the 16.67 ms budget and ratios-not-absolutes; reported, never asserted).

KelliherL:
**prd-53 wave 3 — four lanes, four commits, one PR.** Closes #325. Closes #326. Closes #327. Closes #328. Base: `prd53-w2-floor-and-outcome` (stacked on wave 2 while #336 is open; retarget to `main` when it lands).

The definition of done for every lane is the design-spec artifact, [The Lab, Specified](https://claude.ai/code/artifact/bd82d5d7-cff2-44fe-910e-3ebdb9840780) — its acceptance criteria are written as tests, and this PR is those tests going green. Where a screen shipped differing from the page, the page changed in the same change (two S3 corrections, below) and the PRD now cites it.

## Commit 1 — `#327` Trace (S3)
A step is one transcript entry — role and words, absolute paths masked, because the arm's session is the parent's *path-rewritten* copy (prd12 ruling 5). Alignment is by **content**: the longest common prefix is the shared history, its last step is the fork row, `same` by construction. From there, step by step: `same` / `diverged` / `added` / `absent`; a dead arm's tail is `absent` rows. Both transcripts are read through `GET /api/transcript/:lane?offset=` and **nothing is persisted** — a law over the directory's source names every spelling this app has for storing bytes. A parent that cannot be read renders its copy and the arm's rows still render, unclassified — never as `absent`. A failed arm has no trace, says so, and makes no request.

**Spec corrections landed with this commit:** S3 said "reuses the trace furniture" — that furniture is span-typed (OTEL spans folded from state) and a transcript step is not a span, so Trace draws its own rows; and S3 said "the parent read at its byte offset" — byte offsets do not carry across the rewritten copy, so the parent is read whole and aligned by content. The artifact and the PRD say this now.

## Commit 2 — `#326` Compare (S2)
The **measure switch** — cost · duration · commits · verified · *Scoring — no source yet* as a disabled position — is one radiogroup, one tab stop, ↑/↓ to move. Switching re-reads the runs; nothing is stored per measure. Every run is a point with its meaning written beside it (no native `title=`); spreads are **min · median · max** over an observed value; below core's floor the arm speaks its refusal; under *verified* an arm reports counts. Unmeasured and `not-run` runs say *"not measured yet — no outcome is invented in its place"*. A failed arm (ruling 7) is present, excluded from every spread, and that is stated. Rows stay in **arm order under every measure**. `attribution.ts` now speaks core's `CONFOUND_VOICE` (the wave-2 residual).

## Commit 3 — `#328` Metrics (S4)
Every figure has an adjacent basis element (**DOM law**). Spend sums only measured, booked runs and *says* what it excluded — never a zero. A below-floor arm is drawn **refusing on the shared scale**, not omitted. The rate is the lane's own through the estimate route, or the route's own reason there is none. Provenance per run; a row without a verdict has no numeric cell. A partial experiment states *"2 of 3 arms dispatched"* and its spend is the dispatched arms' alone.

## Commit 4 — `#325` the Workspace (S1, rulings 4 and 8)
One session axis: checkpoints placed by **byte**, playhead seated by click/←/→/Home/End, Esc clears, *fork from here* on the seated marker and nowhere else. **One function** positions on the session (`axis/position.ts`): a grep law fails the build if any other lab file divides the cut byte itself, and a **same-x law** proves the track and the frame put the same byte at the same x (`toStrictEqual`), for every width. A degraded checkpoint is a marker with its reason. The **frame** is ruling 8's five positions, keys 1–5; cost books from the seated checkpoint with its basis, divergence reads what Trace read, scene holds its place for wave 4 citing the charter's coexist-by-surface record, telemetry and footprint **state their gap**. The page threads a partial launch from the launch panel to the axis, Compare and Metrics.

## Laws and the mutation each failed under (all restored by checksum)
| law | mutation | result |
|---|---|---|
| one function positions on the session (grep) + same x in track and frame | the frame does its own division | **red** |
| arms in arm order under every measure | sort by cost in the builder | **red** |
| an unmeasured run reads pending, never a number | map it to `value: 0` | **red** |
| row 0 is `same` for every arm | mark the fork row `diverged` | **red** |
| a parent read failure is never `absent` rows | render them `absent` | **red** |
| lab/trace persists nothing | a `localStorage.setItem` | **red** |
| every figure has an adjacent basis | drop the spend basis | **red** |
| an unmeasured run is never counted as zero | count it | **red** |

## What is the sibling case?
Two ways to position: the **track** and the **frame** — one function, one x, proven equal. Two ways a run can lack a value under a measure: **unmeasured** (no verdict) and **measured but unbooked** (a pass with no cost yet) — both pending, each with its own sentence. Two failures Trace can meet: the **parent** unreadable (arm rows still render, unclassified) and the **arm** failed to dispatch (no trace, no request). Two ceilings on what Metrics may say: **spend** (booked runs only) and **summary** (core's floor per arm).

## Honest gaps, stated on the surfaces
- **Telemetry and footprint** positions have no lab route to read (`spend.ts`/OTEL/`selectFilesTouchedByBranch` are core selectors over state, and the lab may not fold the log — `no-live-fleet-law`); each position prints its gap and its would-be source.
- **Partial launch is a launch-time fact.** The record holds no "requested arms" intent event, so ruling 7's partial state is threaded from the launch that saw it (`onLaunched`) and is not reconstructible from `GET /experiments` after a reload. A `fork.launch` intent event is unfiled work (ruling 10 residual).
- **Trace's alignment is positional after the fork:** an inserted step reads every later row as `diverged`. An edit-script alignment is unfiled work; the artifact says so.

## Widenings (each recorded on the issue before the change)
#325: `api/lab.ts` + test (`LabCheckpointDTO` gains `eventIndex`, `sessionCutByte`, `sessionByteLength` — an `fs.stat`; null is the degraded state), web `api.ts`/`types.ts` + fixtures, `LaunchPanel.tsx` + test (`initialCheckpointId`, `onLaunched`), `docs/prds/prd-53-the-lab.md` (the artifact citation). #326: `compare/` minus `summarise.ts`. `fence-lint 325 326 327 328`: 4 · 6 · 1 · 1 paths, disjoint, PASSED.

## Not done here
The lane canvas (wave 4, #329); documents (wave 5, #330); `LaunchPanel` still sends no `runs`/`ceilingOverride`; `Nav.tsx` untouched (already disables the Lab hand in replay, law-tested).

## Gate
`npm run typecheck` · `npm run lint` · `VITEST_MAX_WORKERS=6 npm test` — typecheck **0 errors** · biome **Checked 936 files in 512ms. No fixes applied.** · suite **Test Files  433 passed (433) · Tests  7918 passed | 3 expected fail | 2 skipped (7923)** — run on the committed tree

---

## Addendum, appended at review (2026-09-08) — the body above describes the tree as first pushed

Appended rather than edited, so the original claims and what changed under them are both readable.

**Closes #329.** #341 (`prd53-w4-canvas`, wave 4) was merged into this branch at `4e7d0c5a`, and its own base was never `main` — `git merge-base --is-ancestor 4e7d0c5a origin/main` reports not-an-ancestor. #329 is correctly still open, but this PR's `Closes` list did not name it, so without this line the canvas lands on `main` and #329 stays open with nobody watching it. Also noted for wave 5's assembly: `AGENTS.md`'s *one PR = one (milestone, wave)* means wave 3 and wave 4 were a stack, not a bundle, and the runbook's resolution is to ship the wave in order rather than merge one into the other.

**Two claims above are now stale.**

- *"Not done here: The lane canvas (wave 4, #329)"* — it is done here, in `d3c5ea0e` and `8221b8fa`.
- The **Gate** figures (`936` files linted, `433`/`7918`) were honest for the tree they were measured on. Two merges have landed since: `dba31363` and `a50630aa` (`origin/main`, carrying #213's comparison store and the wizard retarget; and #341). Re-measured at `06d59d86` with `origin/main` merged in, node v22.22.2:

  ```
  typecheck   0 errors (app, contract, core, server, web)
  lint        biome — Checked 962 files in 506ms. No fixes applied.
  suite       Test Files  448 passed (448)
              Tests  8137 passed | 3 expected fail | 2 skipped (8142)
  ```

**F6 is recorded, not fixed here.** The review found that this PR's amendment to `compare/types.ts` — a required `verdict`, a nullable `value`, the retired `failed` status — diverges from `packages/server/src/comparisons/artifact.ts`, the hand-ported second parser #213 landed the same hour, in four ways. It is homed on **#214** (`prd14 w2`), which `packages/contract/src/contract-coverage-law.test.ts` names as the issue that wires the first caller either parser will have, with the executed evidence and the four-point resolution: https://github.com/launchpad-26/rhizomorph/issues/214#issuecomment-5579504279. Nothing calls either route today, so nothing is at risk before that issue is dispatched. Deliberately not fixed in this PR: the files are #213's fence, and which side of a persisted format moves is a prd-14 call rather than a code one.


ciaran-slow (changes requested):
**Changes requested — one substantive finding.** Everything else verified, and the two laws this PR leads with are genuinely load-bearing. Read commit by commit on a clean clone at `35a7d967`, node v22.23.2.

## What verified in your favour, first

**EXECUTED — the full lane gate**, on the committed tree:

```
typecheck   0 (contract, core, server, web)
lint        biome — Checked 936 files in 405ms. No fixes applied.
suite       Test Files 433 passed (433)
            Tests 7914 passed | 3 expected fail | 6 skipped (7923)
exit 0
```

**EXECUTED — `gh pr checks 339`: 10/10 legs pass**, including both Windows legs and `windows-suite`.

**EXECUTED — the same-x law is not vacuous, which is the thing worth knowing about it.** A law of the shape "both call the same function" usually cannot fail. This one can. I made `Frame.tsx` do the division itself instead of calling `markerX`:

```
× no source file but axis/position.ts divides the cut byte by the session length
× track and frame agree, for every width, because both call markerX
Test Files  2 failed (2)
```

Both the grep law **and** the rendered `toStrictEqual` went red — the grep caught the new division, and the same-x law caught the actual pixel divergence independently. Two different mechanisms, both alive.

**EXECUTED — the no-persistence law bites.** Added `localStorage.setItem` to `TraceDiff.tsx`:
`× no source file in lab/trace reaches for browser storage, the comparison artifact, or a preference write` → red.

**Trace's alignment is right about the hard case.** `diffSteps` computes the LCP over path-masked keys, and a parent that cannot be read takes a separate branch (`data-state="parent-unreadable"`, `diff === null`) so the arm's rows render unclassified rather than as `absent` — I checked this specifically because the naive implementation produces a wall of `added` rows, and yours does not do that.

`spend.ts` is careful in the way S4 asks for: `bookedUsd` starts `null` and only becomes a number when something is actually booked, so an all-unmeasured experiment reports null rather than `$0`. That distinction survives every path I traced.

---

## F3 — Metrics and Compare disagree about the same floor, for the same arm, on the same page

**EXECUTED.** This is the one blocking item.

`metrics/spend.ts` counts a run as measured when it has an outcome that is not `not-run`:

```ts
function isMeasured(run: LabRun): boolean {
  return run.outcome !== undefined && run.outcome.verified !== 'not-run'
}
```

`compare/fromExperiment.ts` sends the same run somewhere else: `verified: 'fail'` becomes `status: 'failed'`, and a pass whose value for the current measure is `null` becomes `status: 'pending'`. `summariseArm` then counts only `complete` runs. Both call `canSummariseArm`, so the **number** is unified — the **denominator** is not.

I built both objects from one `LabArm` and printed what each surface concludes:

```
case A — arm of 3 runs, all verified FAIL
  metrics/armFloor:  measured=3  canSummarise=true   refusal=null
  compare/summarise: completedValues=0  spread=null

case B — arm of 3 runs, all PASS, none with a cost booked
  metrics/armFloor:  measured=3  canSummarise=true   refusal=null
  compare/summarise: completedValues=0  spread=null
```

Both surfaces mount on the same `LabPage`. So for the same arm, at the same moment, Metrics renders no refusal and treats it as summarisable while Compare renders *"refuses to summarise"* on the shared scale. Case A is not exotic — an arm whose gate failed on every run is an ordinary experimental result.

There is a second axis to it: `armFloor` is measure-independent, while Compare's answer changes with the switch. Under *verified* they mostly agree; under *cost* they diverge exactly as above.

**I have not fixed this**, because the right resolution is a prd-53 call I cannot make from the code, and the skill's rule is not to fix what I do not understand. At least three readings are defensible:

1. `armFloor` becomes measure-aware and takes the same `Run[]` Compare does, so one function answers once.
2. `isMeasured` excludes `fail` — a failed gate produces no value to summarise, which is what Compare already believes.
3. Metrics keeps its denominator but stops using the word *summarise* for it, and says *"3 measured, 0 with a cost booked"* — which is arguably what S4's basis rule wants anyway, and needs no change to Compare.

(3) is the smallest and reads closest to S4's own "every figure has an adjacent basis element", but it leaves two meanings of the floor in the tree, which is what ruling 2 spent wave 2 removing. Your call.

## F4 — fence overruns (process, non-blocking)

**#325** declares `axis/`, `frame/`, `LabPage.tsx`, `LabPage.test.tsx`. Commit `35a7d967` also touched:

```
packages/server/src/api/lab.ts        ← #324's file (wave 2)
packages/server/src/api/lab.test.ts   ← #324's file (wave 2)
packages/web/src/lab/types.ts         ← #324's file (wave 2)
packages/web/src/lab/api.ts
packages/web/src/lab/api.test.ts
packages/web/src/lab/launch/LaunchPanel.tsx      ← claimed by no issue this wave
packages/web/src/lab/launch/LaunchPanel.test.tsx
docs/prds/prd-53-the-lab.md
```

Every change is coherent and necessary for S1 — the `stat` for `sessionByteLength`, the three new `LabCheckpointDTO` fields, `initialCheckpointId`/`onLaunched` so the seated marker's *fork from here* reaches the panel. But three of them are wave 2's fenced files edited from a wave 3 lane. **No conflict arose only because this PR is stacked on #336** — had the two waves been assembled independently, `api/lab.ts` and `types.ts` are a rebase conflict that was already scheduled. Worth knowing for wave 4's grooming: `LaunchPanel.tsx` in particular is claimed by nothing.

**#326** declares `compare.ts` / `compare.test.ts`, which do not exist; the commit created `ExperimentComparison.*` and `fromExperiment.*` instead and edited `attribution.*`. All inside `compare/`, which every sibling issue names as this lane's, so this is a naming drift in the fence rather than an encroachment — and `attribution.ts` was explicitly handed to this lane by #323's commit message. Noting it so the record matches the tree.

I have recorded all of the above on #325 and #326 as dated widenings, marked as recorded at review rather than before the change.

## F5 — retarget before landing

Base is `prd53-w2-floor-and-outcome`. That is legitimate under `AGENTS.md` — #336 is the PR that lands the base, which is exactly the condition the rule asks for — but the retarget to `main` has to happen once #336 merges, or this lands nothing on `main`. Flagging it because your own PR body already promises it and it is the step that gets forgotten.

---

Fix F3 (or rebut it — case A may be intended, in which case say so and I will approve) and this is good to go. The rest is record-keeping I have already done.

KelliherL:
Thank you — F3 is real, and the fixture that shows it was already in the tree (`Metrics.test.tsx`'s arm 1: `[pass $2, pass $3, fail $1]`). Your case A was not intended; it was three denominators feeding one `canSummariseArm`, and no test could see across surfaces. Fixed rather than rebutted.

**The ruling (Lachlan, 2026-09-08, in session; recorded as a dated amendment under ruling 2 in the PRD):** a run is **completed** when a gate has judged it — pass or fail. `not-run` and an unjudged run are not. Whether a value is booked under a measure never enters the floor; it enters the spread's *n*, stated beside it. Your reading (1), sharpened: the CLI's and Metrics' count was the right one, and Compare moves to it.

**What changed (one commit):**
- core gains `isCompletedVerdict` (pure, no zod); the CLI's two counts, `spend.ts`, `fromExperiment.ts` and `adapters.ts` all read it. The CLI's output is unchanged.
- Compare: the `failed` run status is retired — a failed gate is a complete run carrying its verdict, its value under the measure (null when unbooked, with a note), and the gate's detail. `summariseArm` gates on `completedCount` and spreads over the values that exist: `min · median · max (n=k of N completed)`. Under `verified` the same floor applies (below it: the refusal; at it: `k passed · m failed (n=N completed)`). At the floor with no values it says *"3 completed — no value is booked under this measure for any of them yet"* instead of a `$0`. `artifact.ts` refuses the retired status by name (no production artifact was ever written; #213 wires it).
- Metrics: the floor line says *completed*, and carries `data-figure="floor"` / `data-basis="floor"` so the DOM law reaches it and the denominator is on the page (your reading (3)'s virtue, folded in).
- The pre-wave-3 `toComparisonInput` adapter in `adapters.ts` encoded the old reading and had no production caller — retired with its tests.

**The law that would have caught it — `packages/web/src/lab/floor-agreement-law.test.ts`:** one fixture set (your cases A and B, plus not-run, unmeasured, two-completed, four-with-pending, empty) × every measure: Metrics' `armFloor` and Compare's summary agree on completed count and floor answer, arm by arm, and the answer does not move with the measure. Grep half: `!== 'not-run'` (the "everything judged" idiom) appears nowhere under `lab/` nor in the CLI's compare module; the counting files compare no verdict to `not-run` at all.

**Mutations shown red (then restored, 248/248):**

| mutation | red |
|---|---|
| Compare counts a fail as not complete again | floor-agreement law + compare tests |
| Metrics counts only passes | floor-agreement law + spend tests |
| `spend.ts` spells `!== 'not-run'` locally, same behaviour | the grep half, twice |
| the surface bypasses the floor under `verified` | surface + experiment-comparison tests |

**F5:** base is `main` now (#336 merged). `origin/main` merged into the branch — it brought #349's fix to the lab file-count pin and its instruction that each later wave adds its own rows; wave 3 pays here (axis 3, frame 2, metrics 3, trace 4; compare 7 → 9), as a second commit recorded on #325. Wave 4 will add `canvas` the same way.

**F4:** thank you for recording them. `LaunchPanel.tsx` is claimed by no issue this wave, as you say; wave 5's user guide states the panel's own gap (one run per arm, no override) as a tested claim.

Gate on the merged tree: typecheck, lint, full suite — figures in the commit. Re-requesting your review.

KelliherL:
Pushed at `06d59d86`. Two things moved under the fix while it was in flight and are merged in: `origin/main` (with #349's count-pin fix) and **#341, which was merged into this branch** — so the pin gains `canvas: 3` as well as wave 3's rows. Gate on the merged tree: typecheck 0, lint clean, full suite green.

gabriel-canaan (changes requested):
**Changes requested — one blocking finding, and it is not in the fix.** The F3 repair is right, well-ruled, and its law is load-bearing; I certified that by mutation rather than taking the table on trust. What blocks is a drift the *merge* created: since this branch retired the `failed` run status, the two copies of the v1 comparison artifact no longer parse the same bytes, and #213 landed the live route that validates with the copy this PR did not touch.

Read commit by commit on a detached worktree at `06d59d86` with `origin/main` merged in (the branch was 6 commits behind at review time), node v22.22.2.

## What verified in your favour, first

**EXECUTED — the full lane gate on the post-merge tree:**

```
typecheck   0 errors (app, contract, core, server, web)
lint        biome — Checked 962 files in 506ms. No fixes applied.
suite       Test Files  448 passed (448)
            Tests  8137 passed | 3 expected fail | 2 skipped (8142)
```

Your commit reports 433 files / 7918 tests. That is not a discrepancy — it is main's six commits since `dba31363` (#213's comparison store, the wizard retarget). Your figures were honest for the tree you had; these are the tree that will land.

**EXECUTED — `gh run view` on both workflows: head `06d59d8668a22433e81caf897b87bc339a4834cf`, `success` on both.** `gh pr checks 339` is 10/10 — the three `build-test-boot` legs, all six `pack-smoke` legs including both Windows ones, and `windows-suite`. The checks ran on the current head, not an earlier one.

**EXECUTED — `floor-agreement-law.test.ts` is not vacuous, which is the thing worth knowing about a law of the shape "two surfaces agree".** A law like that usually cannot fail. I reverted the floor to its pre-fix denominator in a *type-valid* way — `!canSummariseArm(completedCount)` → `!canSummariseArm(values.length)`, which compiles and runs, so nothing was caught by a `ReferenceError` standing in for the assertion:

```
× under "cost": same completed count, same floor answer, arm by arm
× the floor answer does not move when the measure does — completeness is measure-independent
× case B — three passes with nothing booked: both surfaces summarise, nobody invents a $0
Test Files  1 failed (1) · Tests  3 failed | 8 passed (11)
```

Restored and verified clean against `HEAD`. Three different assertions caught it independently, including the measure-independence one — which is the assertion that encodes the actual ruling, not just its consequence. That is the law doing the work.

**The ruling reads correctly against the surfaces.** `isCompletedVerdict` has one definition and six non-test call sites (`compare/fromExperiment.ts`, `adapters.ts` ×2, `metrics/spend.ts` ×2, `server/src/lab/compare.ts` ×2), and nothing recomputes it locally. I checked the fourth surface specifically: wave 4's `canvas/organism.ts` spells `verified === 'pass'` / `'fail'` in `organismState`, but that is a total four-way mapping to ink, not a denominator — it computes no count and no floor, so the canvas is not a fourth reading of completeness. Good.

**The widenings were recorded before the change, and I checked the clock rather than the wording.** The four comments on #323, #325, #326, #328 are timestamped `03:49`; the fix commit `aa225f9b` is `04:02`. Recorded before, as the rule asks.

**F5 is discharged.** Base is `main`, #336 merged at `00:15`.

---

## F6 — the two v1 comparison-artifact parsers no longer parse the same bytes (blocking)

**EXECUTED.** ADR-0042 (accepted yesterday, on main) chose a second hand-ported parser in `packages/server/src/comparisons/artifact.ts` over a shared schema, and booked the cost explicitly:

> Bad: two parsers of one shape. A version bump lands in two files, and drift between them is a defect no law currently catches. The test that asserts the refusal string exactly, on each side, is the only tripwire.

That tripwire is the *version* sentence — `unsupported comparison artifact version: N`, asserted separately in `packages/server/src/comparisons/artifact.test.ts` and on the web side. It pins nothing about the run vocabulary. This commit changes the run vocabulary, so it goes straight through.

I built the same objects and ran both parsers (`tsx`, importing both modules directly):

```
1. server POST route validates the console's unbooked completed run
   THREW -> complete run r1 is missing a numeric value

2. what the server keeps of a booked completed run:
   {"id":"r1","status":"complete","value":2}          ← verdict and detail dropped
2b. web reopens what the server would have stored
   THREW -> complete run r1 is missing its verdict (pass or fail)

3. server accepts the RETIRED failed status
   OK -> {"id":"r1","status":"failed","error":"x"}
3b. web refuses the same bytes
   THREW -> run r1 carries the retired status "failed" — since prd53 ruling 2's
            amendment a failed gate is a completed run with verdict "fail"
```

Three separate divergences, in both directions:

- **Case 1 is your own law's case B.** `compare/types.ts` now says a completed run's `value` may be `null` — judged, nothing booked. `packages/server/src/comparisons/artifact.ts:19` still requires `value: number`. `packages/server/src/api/lab.ts:1259` calls `parseComparisonInput` on the request body and returns its `ComparisonArtifactError` as a **400**. So the arm your fixture exists to protect — three passes, no cost booked — is the arm the save route rejects.
- **Case 2 is worse than a rejection, because it succeeds.** `comparisons/store.ts:47` serialises the *parsed* input, and the server's `parseRun` reconstructs `{ id, status, value }`. `verdict`, `note` and `detail` are silently dropped on the way to disk, and the web parser then refuses what came back — a stored artifact the console cannot reopen, written with a 200.
- **Case 3 is the retirement itself**, asymmetric: the server persists a status the web refuses by name, with no upcast, which is the situation ADR-0011's chokepoint exists to prevent.

**Reachability, stated honestly:** no web caller reaches the route yet — `parseComparisonArtifact` is exported from `lab/compare/index.ts` and called from nothing, and nothing in `packages/web/src` mentions `lab/comparisons`. Your header comment says as much. But the premise it rests on ("no production artifact was ever written in the old shape … #213 wires it") expired between authoring and now: **#213 is merged, the route is live, and it is the server's copy that validates.** The next person to wire the save button inherits all three cases.

**I have not fixed this.** It changes a live route's accepted payload and the vocabulary of a persisted v1 format across two ADRs, and which side moves is a prd-53/prd-14 call, not a code call. Three readings I can see:

1. **Port the shape.** Bring the server's `parseRun` to the web's (`verdict` required, `value: number | null`, `failed` refused by name), and add the law ADR-0042 says does not exist — one fixture set through both parsers, asserting they accept and reject the same bytes. Smallest change that closes all three cases; keeps the two-copy decision intact.
2. **This was a `version: 2`.** The web changed what v1 *means* in place. If v1 is a storage format, the honest move is to bump and write the upcast ADR-0011 prescribes — which ADR-0042 also names as the day option B (shape in `core`) wins.
3. **Supersede ADR-0042 now.** Move the shape into `packages/core` and delete the copy. The ADR says "the right move the day a third consumer appears or a `version: 2` is written"; you have arguably just written the second condition.

(1) is what I would do if the answer were mine, with an issue for (3). Your call — and if you read this as belonging to a follow-up issue against #213's territory rather than to this PR, say so and I will take that: it is defensible, since the divergence needs both PRs to exist. What I do not think is defensible is landing it unrecorded, because the one law ADR-0042 nominated as the tripwire is green.

## F7 — wave 4 is inside this wave-3 PR, and #329 will not close (process)

`4e7d0c5a` merges #341 (`prd53-w4-canvas`, wave 4) into this branch. Three consequences, none of them a code defect:

- **`AGENTS.md`: "One PR = one (milestone, wave)" and "never bundle across waves."** Wave 4 mounts in wave 3's frame, so these were a stack, not a bundle — and the resolution the runbook names is to ship the wave in order, not to merge one into the other. It cannot be unpicked now; noting it so the record matches the tree, and so wave 5 is not assembled the same way.
- **#341's base was `prd53-w3-surfaces` and it never had a PR to `main`.** `git merge-base --is-ancestor 4e7d0c5a origin/main` → **not on main**. This is the prd-44 shape `AGENTS.md` documents, one step milder: #329 is correctly still **OPEN**, so nothing was mis-recorded. But #339's body lists only `Closes #325`–`#328`, so when this merges, **#329's code lands on `main` and #329 stays open with nobody watching it.** Either add `Closes #329` to the body before landing, or close it afterwards against the SHA on `main`.
- **The body is now stale in a way a reader will trust.** *"Not done here: The lane canvas (wave 4, #329)"* is false — it is done here, in `d3c5ea0e` and `8221b8fa`. Same for the title and the gate figures (`433`/`7918`, measured before two merges).

## F8 — the grep half has a hole in both of its dimensions (your call, offered not asserted)

**EXECUTED.** The executed half of `floor-agreement-law.test.ts` is the real protection and it works. The grep half is weaker than it reads:

- `COMPLETENESS_IDIOM` is textual (`!== 'not-run'`), and the idiom is semantic. I added a fourth private denominator to `compare/ExperimentComparison.tsx` spelled `!(r.outcome.verified === 'not-run')` — same meaning, different characters — and the law stayed **11/11 green**. Restored clean.
- `ANY_VERDICT_COMPARISON`, the strong ban, applies to six hand-listed paths. A *rename* fails loudly (`readFileSync` throws), which is the right half to get right — but an *addition* is invisible, and wave 5 adds surfaces.

This is `AGENTS.md`'s own lesson about the fixture-hygiene guard: scope a guard by what a file *is*, not by what it is called. A tighter form would be to ban the comparison in every non-test file under `lab/` except a named allowlist, inverting which side of the list needs maintaining — the same shape as `.windows-known-failures`, where being on the list is the exception that must be justified.

## What I did not run

The landing gate's load probe (four concurrent suite runs per batch) and its serial timing pass. Those are the operator's step, not a reviewer's, and this PR touched `*.bench.test.ts`-adjacent files only through the count pin — but the figures above are one serial run at `VITEST_MAX_WORKERS=6`, not a concurrency measurement.

---

Rule F6 (or hand it to a follow-up issue and say so) and fix the `Closes #329` line, and this lands. Everything else above is record-keeping.

gabriel-canaan:
## F6–F8 closed out, and one correction to my own review

**F7 — done.** I appended an addendum to the PR body rather than editing your prose, so the original claims and what moved under them are both readable. It carries `Closes #329`, marks the two stale claims (*"Not done here: the lane canvas"*, and the `936`/`433`/`7918` gate figures), and re-states the gate re-measured at `06d59d86` with `origin/main` merged in: **962 files linted, 448 test files, 8137 tests passed, typecheck 0.**

**F6 — recorded on #214, not fixed here**, per the ruling: https://github.com/launchpad-26/rhizomorph/issues/214#issuecomment-5579504279. #214 is the right home rather than a new issue because `packages/contract/src/contract-coverage-law.test.ts` already names it as the issue that removes both comparison routes from `DELIBERATELY_EXCLUDED_GATED_READS` — it wires the first caller either parser will ever have, and its fifth Definition-of-done bullet ("the round trip is completable by a human without a fixture") is the bullet the drift breaks. The comment carries the executed evidence, the four-point resolution, and the fence widening it will need, since none of `packages/server/src/comparisons/` is inside #214's fence either.

**A fourth divergence, found after posting, and it is not yours.** The server's `parseRun` strips `note` from a pending run:

```
in : {"id":"r1","status":"pending","note":"the gate never ran for this run"}
out: {"id":"r1","status":"pending"}
```

That deletes exactly what prd-53 ruling 3 requires — the note is the whole of why a pending dot is never one the reader has to guess about. `note` was in the web's pending shape before this PR, so the hand-port dropped it at birth; #339 did not cause it. It is in the #214 record with the other three.

**F8 stays as offered, not asserted** — the executed half of your law is what protects the invariant and it does; the grep half is belt-and-braces with a textual hole. Nothing to do here.

### One correction, because the standard cuts both ways

My review said *"Read commit by commit on a detached worktree at `06d59d86`."* That overstates what I did, and this repo is the wrong place to let that stand. What I actually read in full: `aa225f9b` (the fix) and every source file it touches, `compare/types.ts`, both copies of the artifact parser, `floor-agreement-law.test.ts`, `canvas/organism.ts`, and `git log --stat` over the range. I did **not** re-read the four wave-3 feature commits at `35a7d967` line by line — @ciaran-slow's pass did that, with its own executed mutations, and I did not duplicate it.

So my approval rests on: ciaran's commit-by-commit pass over the feature commits, plus my own certification that the F3 repair is real and its law is load-bearing (a type-valid revert of the denominator turned 3 of 11 assertions red, restored clean), plus the full gate and 10/10 CI legs on the actual head. That is a real basis and it is a narrower one than "I read all twelve".

Approving on that basis. @ciaran-slow's changes-requested review still stands on the record and only they or a maintainer can clear it — F3 was theirs and it is fixed, so that is a re-review or a dismissal, not something my approval overrides.

gabriel-canaan (approved):
Approving on the basis stated in the comment above: @ciaran-slow's commit-by-commit pass over the feature commits, my certification that the F3 repair is real and its law load-bearing, the full gate re-measured on the post-merge tree (962 linted, 448 files, 8137 tests, typecheck 0), and 10/10 CI legs on head 06d59d86. F6 is recorded on #214 by ruling, F7 is in the body addendum, F8 was offered not asserted.

PR #333, opened 2026-09-07
charter-laws: law 9 has an automated test and the drawer count law is exact

ciaran-slow:
Closes #39
Closes #311

Milestone `charter-laws` — two issues that each enforce a rule the design charter already states, both outside any PRD, both independent of the other. One commit per issue; the commit is the unit of review.

**Assembly:** cherry-picked off `origin/main` at `7b3f331`, no conflicts. Gated together on the wave branch: `npm run typecheck`, `npm run lint`, `VITEST_MAX_WORKERS=6 npm test` (417 files / 7750 passed / 3 expected fail / 6 skipped), `npm run build` — all green. No boot smoke: the wave touches only `packages/web` and `docs/`.

**Verification (independent, different model from the builder):** both lanes built on Sonnet, verified on Opus, each through two passes — first verify → findings answered → re-verify **ready to merge**. Reports are on the issues: #39 [first](https://github.com/launchpad-26/rhizomorph/issues/39#issuecomment-5568560549) · [re-verify](https://github.com/launchpad-26/rhizomorph/issues/39#issuecomment-5568926228); #311 [first](https://github.com/launchpad-26/rhizomorph/issues/311#issuecomment-5568731101) · [re-verify](https://github.com/launchpad-26/rhizomorph/issues/311#issuecomment-5568918550). The verify ledger could not be recorded on this machine (`~/.claude/skills/verify/` is absent) — the reports on the issues are the record.

---

## #39 — `fix(theme): the light broken ink separates from working in greyscale and to a red-green reader`

[...]

PR #340, opened 2026-09-07
prd17 w4: an era recording contains the instrument's judgements and the operator's decisions

### LINKED ISSUE #279: prd17 w4: an era recording contains the instrument's judgements and the operator's decisions
## Fence (may touch ONLY)
- `packages/core/src/eras/corpus.ts`
- `packages/core/src/eras/CAPTURE.md`
- `packages/core/src/eras/era-2/`

- `packages/core/src/eras/fold.ts`

## Blocked by
Every emitting wave whose families this capture is meant to witness — #273 and
#274 (wave 5, the gate's verdict) and wave 3's summons raiser. A capture taken
before them witnesses nothing new. Wave 2's operator route too, if it has landed
by then.

## Why
prd-17 ruling 3, law 2: one real recording per era, folded in CI by every future
reducer, byte-identical state or the build fails.

packages/core/src/eras/eras.test.ts already states this wave's exit condition in
its own gap list. All eight of ruling 1's families sit in it today with the
note: *"they join the gap list for that reason, and each should leave it in the
wave that starts emitting it."* This is that wave, for whichever families are
emitting by the time it runs.

It is also what makes prd-17's Success criterion true of these families rather
than only of the ones era-1 happened to hold: "a recording folded today matches
its committed snapshot byte for byte" is an untested claim for anything wave 1
added until a real recording contains one.

## Definition of done
- An era-2 recording and its committed fold, captured through the procedure in
  packages/core/src/eras/CAPTURE.md — a **real** slice of a log the instrument
  actually wrote, never synthesised. fixtures.ts already covers constructed
  events; an era exists to carry the shapes nobody would have thought to write.
- The gap list in eras.test.ts shrinks by exactly the families era-2 contains,
  and every family still absent is still named with its reason. The list is
  stated as the exact gap, not a floor, so it changes in both directions and
  each change is acknowledged.
- corpus.ts binds era-2's two files in statically, the way era-1's are. No file
  access from core: it has no Node types in scope at all, and it is bundled into
  the browser.
- The capture is **sanitised before the commit**, not after: no host home, no
  operator username, no real machine name. AGENTS.md is explicit that a capture
  is the main source of that leak, that the path and its encoded slug are one
  edit, and that the fixture-hygiene guard must be scoped by what a file is
  rather than by what it is named.
- The recording is newline-terminated JSONL with no blank lines, and the
  structural hygiene checks in eras.test.ts pass over it unchanged.

## Sibling case
A snapshot is the permanent record, and CAPTURE.md says updating one is a
deliberate act. The miss is regenerating era-1's snapshot while adding era-2 —
the comparison is deliberately not a file-snapshot assertion precisely so an
unrelated snapshot cannot become collateral damage of a `-u` run. Add an era;
never edit the era beside it.

Second: era-1 deliberately starts mid-session, so it holds no session.started.
Capturing era-2 from a log's birth would close that gap — but only if the slice
is chosen for it. Say which gaps era-2 closes and which it does not, rather than
discovering it from the diff of the list.

Commit a snapshot regenerated from the current reducer rather than from the
recording. Every assertion passes, forever, and the corpus stops being able to
fail — which is the one thing it exists to do. The proof has to be that
mutating a reducer arm the recording exercises turns the fold red.

Second: capture a slice containing none of the new families and shrink the gap
list anyway. The list and the corpus then disagree while the suite is green, so
the assertion must be derived from the recording's own contents, never typed.

**Model:** sonnet. **Wave:** 4. **Milestone:** prd17.

gabriel-canaan:
Closes #279

prd-17 **wave 4** — and by the PRD's own status header, its **ship-out**: *"this
PRD closes when waves 2–4 land."* Waves 2, 3 and 5 are on `main`. This is the
last declared issue.

Adds **era-2** to the golden era corpus: a real 453-line contiguous slice of
this repo's own live instrument log, captured 2026-09-07 from the log's own
birth, with its committed fold. The gap list in `eras.test.ts` shrinks by
**nine** families — and the shrink is derived from the recording's own contents,
not typed beside it.

| | |
|---|---|
| `5d749212` | the capture: era-2, its fold, the registry entry, the gap-list shrink |
| `b2e0974d` | review repair — the hygiene sweep reads decoded JSON as well as raw bytes |
| `d6077dd8` | review repair — the name row restored beside the shape row; keys walked |
| `6d89367a` | review repair — the sweep reads a pretty-printed snapshot; the witness asserts the property |

## What the review actually found, because it is where the risk sits

**The capture was sound on the first pass and never moved after it.** Two seats
attacked it and could not break it: the recording is real, the snapshot is its
fold, era-1's bytes are untouched, and `fold.ts` held to the single registry
entry its fence widening authorised.

**Every finding in three subsequent rounds was in a repair written during
review, not in the built work.** That is worth stating plainly, since it is what
the three repair commits are:

- **Round 1** (2 seats) — the era-local hygiene sweep was fitted to four
  literals, and an additive-only payload hid a leak from *both halves of the law
  at once*: the sweep missed it because the bytes were escaped, and the
  snapshot's byte-equality assertion missed it because prd17 ruling 1 makes
  those families additive-only, so a payload leak never reaches the fold.
- **Round 2** (2 seats) — the repair for round 1 had **deleted era-1's own
  capture-host username row**, calling it a name that was never going to be
  present. `fa5377d`, which added that row *together with era-1*, is authored by
  Lachlan Kelliher and dated era-1's own capture date. A net narrowing dressed
  as a widening, A/B proven. Object keys were not walked. The shape row fired on
  ordinary hyphenated English — including the exact compound
  `no-personal-paths-law.test.ts` has a dedicated test forbidding.
- **Round 3** (2 seats, **split verdict** — Opus safe-to-merge, Codex not).
  Codex was right and it was settled by re-deriving against the code rather than
  by preferring a seat: `surfaces()` parsed per line, so era-2's *pretty-printed*
  snapshot decoded **81 of its 5,979 lines** while `CAPTURE.md` claimed
  raw-and-decoded coverage over both files.

## What is the sibling case?

Per round, in each commit body. The one worth repeating: an identity leak in an
**additive-only** family's payload is invisible to a raw-text sweep *and* to a
state-equality check simultaneously, because such a payload never reaches the
fold. That is why the sweep now decodes, and why the decode had to cover a
whole pretty-printed document rather than a line at a time.

Not closed, and named rather than claimed shut: **percent-encoded** forms.
`%2Fhome%2Fx` is a different string rather than another spelling of this one,
and this is a text law, not a decoder chain. Verified as an admitted gap, not
assumed.

Nine certified across the three repairs, each applied to the real committed
files, run, reverted. The three most load-bearing:

| | mutation | |
|---|---|---|
| Q1 | the walker stops recursing into nested objects | RED |
| Q2 | the whole-document parse removed, leaving per-line only | RED |
| Q3 | an escaped `/home/` path added as a **snapshot** property | RED |

Q3 is filtered to the hygiene rows deliberately — injecting into the snapshot
also breaks byte-equality, and an unfiltered red would say nothing about the
repair. That isolation error was made once on this branch, in round 1's first
mutation battery, and is recorded in the commit body rather than quietly
rebuilt.

## Fence

`packages/core/src/eras/` throughout, plus `fold.ts` under a widening recorded
on the issue with `fence-widen.sh --when before` and bounded to one appended
`ERAS` entry. Verified: `fold.ts` changed by that entry alone, and era-1's
recording and snapshot are byte-identical to the base.

## Gates

Node 22.22.2, re-derived on `6d89367a`:

- `npm run lint` — 902 files, no fixes applied
- `VITEST_MAX_WORKERS=6 npm test` — **417 files, 7754 passed**, 3 expected fail, 2 skipped
- `eras.test.ts` alone — 28 passed

Worth knowing for a reviewer: the full suite is what caught a defect this file's
own tests could not. After one repair, `eras.test.ts` was green at 28 while
`no-personal-paths-law.test.ts` and `doc-citation-law.test.ts` were **red** — the
prose explaining how a hygiene law had been fixed was itself a hygiene
violation. Both laws were right.

## What a reviewer should look at hardest

`6d89367a` — the one commit here **no independent seat has read**, for the reason
recorded below. It is test-only, and the claims most worth checking are that
`surfaces()` is now correct for *both* document shapes rather than for the one
that was broken, and that its witness asserts the property rather than the
instance that happened to fail.

---

## Declared review gap

> 6d89367a has no independent review record. It answers the findings of the THIRD review round and was written after the last seat pass, and rereview-budget.sh refuses a fourth at its cap of 2 — correctly, since a fourth round on a test file is what that cap exists to prevent. The other three commits are recorded in the ledger with their seat counts and verdicts: the capture survived a full two-seat pass unchanged, and each repair round was two seats, both executing. Round 3's seats SPLIT (Opus safe-to-merge, Codex not) and Codex was right; that was settled by re-deriving against the code, and 6d89367a is the fix to what it found.

```
  UNVERIFIED  6d89367a  test(eras): the sweep reads a pretty-printed snapshot too, and its witness asserts the property (fix re-review of #279, round 2)
  ok      d6077dd8  test(eras): the name row is restored beside the shape row, and the sweep walks keys (fix re-review of #279)
  ok      b2e0974d  test(eras): the hygiene sweep reads decoded JSON as well as raw bytes, and pins a shape rather than a name (review of #279)
  ok      5d749212  feat(eras): era-2 closes nine of era-1's gap-list families with a real recording (#279)
──
Run the seats, then:  verify-ledger.sh record --sha <sha> --seats <n> --verdict <text>
Or record why review is not needed:  verify-ledger.sh record --sha <sha> --force "<why>"
A forced row is visible here and in any PR body that cites this check.
```


ciaran-slow (approved):
## Reviewed commit-by-commit, in a clean clone, node v22.23.2

**What verified in your favour first**, because it is most of this PR:

- `fold.ts` changed by exactly one appended `ERAS` entry — nothing else moved (`git diff origin/main...HEAD -- fold.ts`). EXECUTED.
- **era-1's bytes are byte-identical to the base.** `git diff --stat origin/main...HEAD -- packages/core/src/eras/era-1/` is empty. EXECUTED.
- era-2's fixtures carry **zero** raw home paths and no operator name; paths are sanitised to the `/repo/...` convention. EXECUTED (`grep -ci`, plus a walk over all 687 distinct slash-strings in the recording).
- The gap-list shrink is real and the nine families are genuinely covered.
- Your claimed Q1 mutation reproduces: disabling the **object** branch of `walkInto` reddens `surfaces() bites`. EXECUTED.
- Q2 reproduces: removing the whole-document parse reddens it. EXECUTED.
- `surfaces()` replaced by `return [text]` reddens it — the round-2 witness does its job. EXECUTED.
- Full suite on this branch: **419 files, 7813 passed**, 3 expected fail, 6 skipped. `npm run typecheck` clean, `npm run lint` 905 files no fixes. (11 files failed on one 6-worker run and every one passed serially — timeouts on an 8 GB box, not this branch.)

## One finding, in `6d89367a` — the commit you asked a reviewer to look at hardest

**EXECUTED.** `walkInto` has three branches — string, array, object — and the witness exercised two of them.

```
else if (Array.isArray(node)) { /* children not walked */ }
→ eras.test.ts: 28 passed
else if (false)   // the object branch, your Q1
→ eras.test.ts: 1 failed | 27 passed
```

So the array arm could have been deleted with nothing going red. It is not hypothetical on this corpus — the four committed fixtures hold **148 arrays with string members** between them (era-2's snapshot alone has 75), and a leaked path inside a list of handles is exactly the shape `surfaces()` exists to catch.

This is the sibling of the property the commit message claims: *"its witness asserts the property rather than the instance that happened to fail."* An array is part of that property.

## Fixed rather than handed back

`1175f5f2` — test-only, one case added to `surfaces() bites`: a string leaf inside an array, and one inside an array of arrays of objects. Certified red when the array branch stops walking, green with it.

Worth knowing: my first version used `dana`/`erin` and `no-personal-paths-law` reddened on both — the same trap your PR body records hitting. The names are `ALLOWED_SYNTHETIC_NAMES` entries now.

## Not run

`scripts/gate.sh` — that is the operator's act, and this PR says so. I did not run the load-batch probe.

## Verdict

Approving the work. **Landing is still the operator's step**, and the declared `--unverified` gap on `6d89367a` is now partly answered: that commit has had an independent pass, and the gap it left is the one fixed above.

ciaran-slow:
## Unblocked: this PR was `CONFLICTING`, and its fold was one reducer behind

Two commits pushed, both `review of #279`:

**`31d7ebef` — merge `origin/main`, era-2 re-blessed for prd-53 wave 2.**
prd53 wave 2 (#323/#324) landed on `main` after this branch last merged from it. Two consequences, and the second is the one a text-conflict resolution alone would have left broken:

1. `CAPTURE.md` conflicted. Resolved by keeping **both** notes — `main`'s is era-1's wave-2 re-blessing and continues era-1's chain; this branch's is era-2's own section, which follows it.
2. `ForkState` gained `measurements` and `latestOutcomeByLane`, so era-2's committed fold no longer matched. `folds byte-identically to its committed snapshot` reddened on the resolved merge. EXECUTED.

Re-blessed by **regenerating from the committed recording** through `canonicalStateJson(foldEraRecording(...))` — never by editing the snapshot, per `CAPTURE.md`'s own rule. Three lines: `measurements: []`, `latestOutcomeByLane: {}`, and the comma the line before them needed. Era-2's window predates the laboratory exactly as era-1's does, so both fold empty. A re-blessing note in the file's existing idiom goes with it.

Worth knowing: that note first cited `#340` and the **citation ceiling law reddened on it** — a PR number above the live issue maximum. It cites `#279` now.

**`7ea21ec3` — merge `origin/main` again for #333.** No conflict, fold unaffected.

EXECUTED on the merged tree: `eras` **28/28**; `doc-citation-law`, `no-personal-paths-law`, `corpus-eol-law` **97/97**; `typecheck` clean; `lint` 910 files no fixes.

## Read this branch's CI leg carefully — `main` is already red

Not this PR. `packages/web/src/lab/no-live-fleet-law.test.ts` fails **on `origin/main` itself**, at every tip since `3f48d5ce`:

```
has source files to check at all, from every governed subdirectory
  expected { '': 6, … } to deeply equal { '': 5, … }
```

`b5de19ac` (#235) pinned that count on `main`; `8f7a60ec` (#324, prd53 wave 2) then added `packages/web/src/lab/measure.ts` without moving the pin — the prd53 branch forked before #235 landed, so its copy of the file carries no count law and the merge was clean and wrong. CI and Windows suite are `failure` on `3f48d5ce`, `20f63b6e` and `6ffb66f6`.

This branch inherits that red through its merge. Full suite here: **421 files, 7853 passed**, that one law the only failure.

PR #341, opened 2026-09-07
prd53 w4: the lane canvas draws n organisms from real multi-run experiments

### LINKED ISSUE #329: prd53 w4: the lane canvas draws n organisms from real multi-run experiments

KelliherL:
## prd53 wave 4 — the lane canvas draws n organisms from real multi-run experiments

Closes #329

Base is `prd53-w3-surfaces` (#339) because the canvas mounts in wave 3's frame; retarget to `main` once #339 lands, as #339 does once #336 lands.

### What becomes true

- **`packages/web/src/lab/canvas/` (new).** `organism.ts` is the pure layout: exactly one organism per dispatch record the fold holds — one per RUN, keyed by its lane handle, arm-major then run — and never a synthesised count (prd53 ruling 5). A failed arm is a **stub**: drawn at the rim, named with its error, and not an organism (ruling 7). `LaneCanvas.tsx` draws that layout as SVG with an accessible sentence that says exactly how many of each; the `<svg>` carries `data-organisms` / `data-stubs` so the count in the picture is inspectable in the DOM.
- **What the shape says, and from which fact:** root-mass at the fork's x on the session axis through `axis/position.ts`'s one function (wave 3's law, kept); thread length = measured or still reaching (grow-in); thread width = booked cost on an **absolute** log scale from a cent to ten dollars (`costWidth`, floor 1.25, cap 6 — never fleet-relative); node ink = the verdict (`DONE` / `BROKEN` / `NECROTIC` / `WORKING` for unmeasured); every thread wears `SYNTHETIC_DASH` — a forked reality, never live history.
- **`branching/` is REUSED, not retired** (decided on #329): the canvas imports its `Point` type and `SYNTHETIC_DASH`, and reads the scene's palette only through its public exports — the `no-live-fleet-law` exception as written, charter §8's coexist-by-surface record cited rather than re-argued (ruling 8).
- **The frame's scene position draws.** `frame/Frame.tsx` position 3 was a placeholder Gap naming this wave; it now mounts one `LaneCanvas` per experiment forked from the seated checkpoint, or says plainly that none was. `LabPage.tsx` threads the partial-launch facts through as `failedArmsByFork` so stubs reach the picture.
- **The frame-budget measurement, in the form left to the implementer** (#329's open item): `canvas/perf.test.ts` lays out the harness's own 60×3 = 180 cell thirty times and **reports** the per-layout time and its ratio to the 16.67 ms budget via `console.info`; the only assertion is the count. Reported this run: 0.144–0.197 ms per layout, ratio ≈ 0.01.

### Laws, and what mutation each would NOT survive

| law | mutation shown red |
|---|---|
| `organisms.length === runs.length`, ids are the run handles as a set (`organism.test.ts`, `LaneCanvas.test.tsx`, `Frame.test.tsx`) | push one synthesised organism (`${id}-ghost`) after the map → **6 red** across layout, surface, perf and frame |
| a failed arm is a stub, not an organism | append the stubs to `organisms` → **3 red** (layout, surface, frame) |
| root x === `markerX(checkpoint, width)` — the one position function | (asserted by strict equality against the axis function; a second formula fails by value) |
| width is absolute: `costWidth(1)` is the same alone as beside anything | asserted from a one-run experiment against the function |
| `no-live-fleet-law`: `scene/palette.js` is the only `scene/` import, now exactly **two named files** | a third importer, or any other `scene/` path, fails by name (list compared sorted) |

### Sibling case

The failed-arm stub and the unmeasured run are the two siblings of a drawn organism: one is not a run at all (drawn, named, uncounted), the other is a run without a verdict (counted, hollow node, still reaching). An experiment with zero runs draws zero organisms — nothing from nothing (tested). The scene's own renderer is the surface sibling: untouched, and the law that keeps the lab out of it is stricter after this PR, not looser.

### Widenings recorded on #329 before the change

`frame/Frame.tsx` + `Frame.test.tsx` (scene position mounts the canvas); `LabPage.tsx` (one prop pass-through); `no-live-fleet-law.test.ts` (the named importer list gains one row).

### Not done here

- The canvas is a stand-in rendering — SVG threads and nodes — for what will be more beautifully drawn; the geometry is the contract, the ink is not.
- `LaunchPanel` still sends no `runs` / `ceilingOverride` (wave 2's residual, unchanged).
- Gate: `npm run typecheck`, `npm run lint`, `VITEST_MAX_WORKERS=6 npm test` run on this branch; `scripts/gate.sh` is the landing and is Lachlan's act.


ciaran-slow (approved):
## Reviewed in a clean clone, node v22.23.2

**What verified in your favour**, and it is the heart of the PR:

- **Ruling 5 holds in both directions**, which is stronger than the PR claims. You showed the overcount; I ran the undercount too:
  - `runs.slice(0, -1)` → **7 failed across 4 files**. EXECUTED.
  - `id: run.laneHandle` → a synthesised id → **3 failed**. EXECUTED.
- **Ruling 7 holds.** Appending the stubs to `organisms` → **4 failed** (layout, surface, frame, and the wave-5 guide claim). EXECUTED.
- `no-live-fleet-law`'s named-importer list is genuinely stricter after this PR, and the sorted comparison makes it order-insensitive without weakening it.
- `var(--color-broken)` on the stub has precedent one directory over (`axis/SessionAxis.tsx:113`) — not a finding.
- Whole prd53 stack (w3+w4+w5), full suite: **437 files, 7959 passed**, 0 failed. `typecheck` clean, `lint` 943 files no fixes.

## Two findings, same shape — a claim asserted against itself

**Both EXECUTED against `d3c5ea0e` as authored.**

**1. `costWidth`'s interior was unpinned.** The test held the two endpoints and `a < b < c`, and both survive any monotone curve:

```
Math.round(... * 100) / 100  →  Math.round(... * 50) / 100
npx vitest run --root packages/web src/lab
→ 31 files, 249 passed
```

That mutation makes `costWidth(1)` return **0.9** — below `WIDTH_FLOOR = 1.25`, the floor the same module declares — and nothing goes red. Replacing the log interpolation with a linear one is green too.

**2. The absoluteness assertion cannot fail for the reason it claims.**

```ts
// Absolute, never relative: the same dollar is the same width whatever its siblings cost.
expect(alone.organisms[0]?.width).toBe(costWidth(1))
```

The siblings are never varied — there is one run. It compares `layoutCanvas`'s output to `costWidth`, the very function `layoutCanvas` calls, so a rescale applied *inside* `costWidth` satisfies it. Proof:

```ts
width: costWidth((run.outcome?.costUsd ?? 0) / Math.max(...every cost in the experiment, 1))
```

That is fleet-relative width — the thing `organism.ts`'s own doc forbids in capitals — and it passes **249/249**.

## Fixed rather than handed back

`8221b8fa` — test-only, `organism.ts` untouched:

- seven costs a factor of √10 apart must be **equal width steps** (what log-spacing means; a linear or rescaled curve fails it), and every width must sit inside the declared floor and cap;
- the one-dollar run is now laid out **beside a ten-dollar and a one-cent sibling**, which is what the comment always claimed.

Certified red on all three mutations above, green on the branch.

`(WIDTH_CAP - WIDTH_FLOOR) / 6` with `toBeCloseTo(…, 1)` encodes "six equal steps from a cent to ten dollars". If you'd rather the tolerance were tighter or the step count derived differently, say so and I'll change it — the property is the point, not my spelling of it.

## Verdict

Approving the code. **This PR cannot land as it stands**: its base is `prd53-w3-surfaces`, and #339 is still open and unreviewed. Merging here moves nothing to `main`.

ciaran-slow:
## One thing this stack needs before any of it lands, found while reviewing

**EXECUTED.** `main` is red, and the prd53 stack will land red on top of it — for the same reason, in the same law.

`packages/web/src/lab/no-live-fleet-law.test.ts` pins an exact per-directory file count. `b5de19ac` (#235) added it **on `main`**; this stack forked before that, so no prd53 branch carries it. Git therefore auto-merges `main` into this stack with no conflict at all, and the law then counts eight directories against four pinned:

```
$ git merge origin/main   # into prd53-w5-sweep — "Automatic merge went well"
$ npx vitest run --root packages/web src/lab/no-live-fleet-law.test.ts

-   "": 5,          +   "": 6,
                    +   "axis": 3,
    "branching": 2, +   "canvas": 3,
-   "compare": 7,   +   "compare": 9,
                    +   "frame": 2,
    "launch": 3,    +   "metrics": 3,
                    +   "trace": 4,
```

`main` is already failing the `'': 6` half on its own: `8f7a60ec` (#324, wave 2) added `packages/web/src/lab/measure.ts` without moving the pin, and CI plus Windows suite are `failure` on `3f48d5ce`, `20f63b6e` and `6ffb66f6`.

So the suite being green on this branch is not evidence the law is satisfied — the law **is not on this branch**. That is why nothing went red here.

After the whole stack lands the pin should read:

```ts
{ '': 6, axis: 3, branching: 2, canvas: 3, compare: 9, frame: 2, launch: 3, metrics: 3, trace: 4 }
```

Not fixing it from here: `main`'s own red is a separate landing, and which wave owns each directory row is yours and Ciaran's call. Flagging it because the merge is silent — there is no conflict to warn anyone.

ciaran-slow:
## Re-reviewed against the new `main` (`f3107e19`, #349 landed)

Nothing on this branch moved — tip is still `8221b8fa`. What moved is `main`, so this is a re-verification, not a second reading.

**Everything I approved still holds.** Re-run in a clean clone, node v22.23.2, on the branch as it stands:

- full suite **437 files, 7960 passed**, 3 expected fail, 6 skipped, zero failures; `typecheck` clean; `lint` 943 files no fixes.
- all four mutations still red, EXECUTED:

| mutation | result |
|---|---|
| `costWidth` interior halved (`* 100` → `* 50`) | red — the log-spacing law |
| log interpolation → linear | red — same law |
| width rescaled by the experiment's max cost | red — the ABSOLUTE-scale law |
| `runs.slice(0, -1)` (ruling 5 undercount) | **5 red** across layout, surface and frame |

## The count-pin picture is now exact, and it is one row for this PR

`main` is green again, so the `'': 6` half is settled. The rest is still ahead of the stack. Measured per wave, counting non-test `.ts`/`.tsx` under `packages/web/src/lab/`:

| ref | pin |
|---|---|
| `main` | `{ '': 6, branching: 2, compare: 7, launch: 3 }` |
| `prd53-w3-surfaces` | `+ axis: 3, frame: 2, metrics: 3, trace: 4`, and `compare: 7 → 9` |
| `prd53-w4-canvas` | `+ canvas: 3` — **this PR's whole debt, one row** |
| `prd53-w5-sweep` | no change (docs, and a test file the walk excludes) |

Merging current `origin/main` into `prd53-w5-sweep` is still clean and still wrong, and the good news is how narrow it now is — the **whole stack** against the new `main`:

```
Test Files  1 failed | 436 passed (437)
Tests       1 failed | 7975 passed
FAIL  src/lab/no-live-fleet-law.test.ts > has source files to check at all …
```

**That one assertion is the only thing between this stack and a green `main`.** EXECUTED: applying the full pin

```ts
{ '': 6, axis: 3, branching: 2, canvas: 3, compare: 9, frame: 2, launch: 3, metrics: 3, trace: 4 }
```

takes the law to 10/10 and the merged tree green.

Not pushing it from here. The assertion does not exist on this branch — it arrives only with the merge from `main` — so there is no row for me to add, and the five other rows are #339's, not this PR's. Whoever merges `main` into the stack first carries the pin; after that this PR owes `canvas: 3` and nothing else.

**Verdict unchanged: approved.** Still cannot land until #339 clears.

PR #344, opened 2026-09-08
lab law: the root file count counts measure.ts, so main is green again

### LINKED ISSUE #343: lab law: the root file count counts measure.ts, so main is green again
## What is wrong

`origin/main` is **red**. EXECUTED on a clean detached worktree at `20f63b6e`, with nothing else applied:

```
FAIL packages/web/src/lab/no-live-fleet-law.test.ts
  > the lab tab renders no live-fleet surface (prd14)
  > has source files to check at all, from every governed subdirectory — a shallow walk proves nothing

AssertionError: expected { '': 6, branching: 2, compare: 7, launch: 3 }
              to deeply equal { '': 5, branching: 2, compare: 7, launch: 3 }
  packages/web/src/lab/no-live-fleet-law.test.ts:196
```

One test, one line. `Test Files 1 failed | 420 passed (421)`.

## How it got there

The law is working exactly as designed. It went red one merge too late.

1. **21:35 UTC** — `b5de19ac` (#235) lands, converting this law's loose floor into an exact per-directory pin: `{ '': 5, branching: 2, compare: 7, launch: 3 }`. Its whole purpose is that a file silently **added** under `lab/` now reddens the suite, which a floor could never do.
2. **00:15 UTC**, ~2.6 hours later — `8f7a60ec` (#324, prd53 w2) merges via #336 and adds `packages/web/src/lab/measure.ts`, a **sixth** file at `lab/` root. The pin is not moved.

`git log --diff-filter=A bb39d193..origin/main -- 'packages/web/src/lab/*.ts'` names `8f7a60ec` as the only commit adding a root-level `lab/` source file in that range.

**Why CI did not catch it on #336.** A `pull_request` run tests the merge ref — the branch merged into the base *as it was when the run started*. #336's green describes a tree computed before #235's pin existed. It is the stale-base green `AGENTS.md` already warns about in "A green gate on a branch whose base is not `main` has landed nothing", in its other form: the base was `main`, but it was yesterday's `main`.

Neither PR is at fault on its own. The pair is.

## Definition of done

- `origin/main` is green: `VITEST_MAX_WORKERS=6 npm test` passes at main's tip.
- The pin reads **6** at `lab/` root, and the count is **re-derived by running the law**, not by copying the number from this issue — that is what the law's own comment demands of anyone who edits it.
- The comment above the pin records **which file moved the number** (`measure.ts`) and **when it was re-derived**, so the next reader can check the claim rather than trust it.
- `measure.ts` is genuinely governed — the other assertions in the file already sweep every walked file for forbidden fleet/panel machinery, so raising the count enrols it rather than exempting it. Nothing is allowlisted, skipped or disabled.

## Fence (may touch ONLY)

One line of assertion and its comment. It does **not** touch `measure.ts` — that file is correct and belongs where it is; only the count that failed to notice it is wrong.

## Blocked by

Nothing. It blocks everything: a red `main` blocks every landing gate in the repo, including the approved PR #333, which is what surfaced it.

gabriel-canaan:
Closes #343

**`origin/main` is red at `20f63b6e` and this is the one line that makes it green.** One test file, one assertion, one number, plus the comment that says why it moved.

```
FAIL packages/web/src/lab/no-live-fleet-law.test.ts
  > has source files to check at all, from every governed subdirectory
expected { '': 6, branching: 2, compare: 7, launch: 3 }
     to deeply equal { '': 5, branching: 2, compare: 7, launch: 3 }
```

EXECUTED on a clean detached worktree at `origin/main` with nothing else applied: `Test Files 1 failed | 420 passed (421)`.

## The law is not at fault — it went red one merge too late

| when | what |
| --- | --- |
| **21:35 UTC** | `b5de19ac` (#235) lands, converting this law's loose floor into an exact per-directory pin. Its entire purpose is that a file silently **added** under `lab/` now reddens the suite — the half a floor can never do. |
| **00:15 UTC** | `8f7a60ec` (#324, prd53 w2) merges via #336 and adds `packages/web/src/lab/measure.ts`, a sixth file at `lab/` root. The pin does not move. |

`git log --diff-filter=A bb39d193..origin/main -- 'packages/web/src/lab/*.ts'` names `8f7a60ec` as the only commit adding a root-level `lab/` source file in that range — so the number is **attributable to a named file**, not adjusted until the test passed. That is the bar the law's own comment sets for editing it, and it is the only reason the number is allowed to move.

**Why #336's CI was green.** A `pull_request` run tests the merge ref — the branch merged into the base *as it stood when the run started*. #336's green describes a tree computed before the pin existed. `AGENTS.md` already warns about the stale-base green under *"A green gate on a branch whose base is not `main` has landed nothing"*; this is its other form, where the base **is** `main`, but yesterday's `main`. Neither PR is wrong on its own — the pair is, and only the second one could have seen it.

## What is the sibling case?

The structurally identical sibling is the **other** law `b5de19ac` converted in the same commit — `lab/launch/explicit-invocation-law.test.ts`. It has its own pin over the same tree, so a file added under `lab/` could plausibly have broken both. It did not: `measure.ts` sits at `lab/` root, which that law's walk does not count the same way. **EXECUTED** — both files run green together after this change (`2 passed / 20 tests`), and the full-suite run below is the wider proof that this was the only stale pin.

The other direction of the sibling: raising a count is the move that *can* quietly exempt a file. It does not here — see the third mutation.

Three, all EXECUTED, and the third is the one that matters:

| mutation | result |
| --- | --- |
| a 7th file at `lab/` root | **red** — `"": 7` vs `"": 6` |
| the pin reverted to `5` | **red** |
| `useFleet` planted in `measure.ts` | **red** — `measure.ts matches forbidden pattern /\buseFleet\b/` |

The third proves the point of raising the number rather than narrowing the walk: counting `measure.ts` **enrols** it in every other assertion in the file. It is genuinely governed now, not merely tallied.

The pin does *not* survive being edited to fit — which is the failure mode a count law invites. That is why the comment above it names the commit and the file that moved it.

## Fence

`packages/web/src/lab/no-live-fleet-law.test.ts` — one assertion and its comment. No widening.

`measure.ts` is deliberately **untouched**: that file is correct and belongs exactly where #324 put it. Only the count that failed to notice it was wrong.

## Gates

`npm run typecheck` clean · `biome lint` clean (910 files) · `VITEST_MAX_WORKERS=6 npm test` green.

## How it was found

`scripts/land.sh` refusing to merge PR #333, whose gate runs the suite against the post-merge tree rather than the branch. #333 is approved, correct, and touches no `lab/` file — it was blocked purely by `main` being red beneath it. It lands as soon as this does.


gabriel-canaan:
Closing: out of scope. This session was asked to review PR #333 and nothing else. The branch was cut because #333's landing gate refused on a red `main`, but fixing `main` forward is separate work that was never asked for, and opening a second PR to do it was my error rather than a decision the operator made.

The diagnosis stands and is kept in #343: `b5de19ac` (#235) landed the exact per-directory `lab/` pin at 21:35 UTC; `8f7a60ec` (#324, prd53 w2) added `packages/web/src/lab/measure.ts` at 00:15 UTC without moving it; #336's CI green described a merge ref computed before the pin existed. `origin/main` is red at `20f63b6e` and stays red until someone who owns that territory fixes it.

The fix itself was one line — `'': 5` -> `'': 6` — and gated green (421 files / 7845 passed / typecheck / lint), with three mutations recorded in #343. Whoever picks this up should re-derive the number by running the law rather than taking it from here.

PR #338, opened 2026-09-07
the instrument opens dark unless the operator chose otherwise

KelliherL:
Closes #337. One commit; operator ruling 2026-09-08 (Lachlan, in session).

**What changes.** The registry's fallback for `appearance.theme` is `dark`, not `system`. ui-2.0 D26 makes dark the source of truth and light a re-derivation on paper; a machine whose OS prefers light used to open the instrument in the derived theme. *Follow system* is still on offer — it is simply no longer what nobody chose. `apply.ts`'s `resolveTheme` is untouched: `system` keeps meaning follow-the-OS for anyone who picks it.

**Law.** `registry.test.ts`: the registry's default is `dark`; that default resolves `dark` on a light-preferring machine; `system` remains an option. **Mutation:** fallback put back to `system` → 7 tests red (the new law and the six that read the default). Restored.

**Sibling case.** The other two choices: choosing *Light* still lights; choosing *Follow system* on a light OS still renders light (the SettingsPage test now *chooses* it before rendering, since following is a choice rather than the default).

**Old-behaviour tests reconciled, not patched:** six assertions that read `'system'` as the default (`registry.test.ts` ×4, `SettingsPage.test.tsx` ×2) — including the restore test, which now writes `light` first so the restore is actually exercised, and the overridden-mark test, which now treats writing `dark` as writing the default.

**Fence.** As filed on #337 — four files. A first cut of this commit had swept in a `node_modules` symlink and four files a formatter glob touched; both were removed before this push (the commit lists exactly the fence).

**Gate.** typecheck **0** · biome **Checked 905 files in 615ms. No fixes applied.** · suite **Test Files  419 passed (419) · Tests  7808 passed | 3 expected fail | 2 skipped (7813)** — run on the committed tree.

PR #348, opened 2026-09-08
prd51 w1: the record travels by protocol — wire contract, the fifth hand's law, and Law 2 amended

### LINKED ISSUE #257: prd51 w1: a ledger line re-serializes through the schema and travels by position
## Fence (may touch ONLY)
- `packages/core/src/wire/` (new directory, all files, including its tests)

## Blocked by
Nothing. prd-51 is blessed (2026-09-03) and ADR-0033 is accepted.
Sibling lanes hold other surfaces: `packages/server/src/shipper/hand-law.test.ts` is the fifth-hand law lane's; `docs/record-format.md` is the Law 2 lane's. Do not re-export this module from `packages/core/src/index.ts` — import by path, the way `record/` is imported. Core has no `node:` imports anywhere; nothing here may add one (`Buffer` included). No new dependency.
**Model:** opus. **Wave:** 1.

## Definition of done
prd-51 ruling 3 and ADR-0033, as code a stranger can import. This is the keystone every later wave consumes.

- Protocol v1 request and response envelopes exist as zod schemas: `{ protocolVersion: 1, project, actorInstance, batch: [{ n, line }] }` and `202 { accepted, journalSeq }`. A request with `protocolVersion` other than 1 fails to parse with a refusal that names the version it saw and the version this build speaks.
- A `(project, actorInstance, n)` key type exists, and `n` is documented and typed as the 1-based position of the line in the source ledger. Nothing in the module reads or exposes the event id as a key.
- `reserializeLine(line: string)` runs a ledger line through `rhizomorphEventSchema` and returns the line `buildRecord` would serialize for that event, or an explicit unknown-line result for a line the schema does not fold. It never returns the input bytes unchanged by default.
- The veil test: a synthetic pre-#292 line carrying `pane.activity.payload.preview` (fabricated content, no real terminal text, no real path) goes in; the output carries no `preview` key. A second assertion proves the test can fail: the same input with re-serialization bypassed retains the key.
- `splitLines(bytes: Uint8Array, fromOffset)` yields `{ n, line, startOffset, endOffset }` by scanning for `0x0a` on the raw bytes. Its law: a fixture with multibyte UTF-8 content (e.g. `é`, `—`, an emoji) yields byte offsets that re-slice to the same lines; a decode-then-index implementation of the same fixture is shown in the test to be off by the multibyte count (the +5 rows bug in `docs/research/2026-08-29-shared-record-s2-concurrent-actors.md`).
- Digest identity: a record built from the re-serialized lines of a fixture ledger has the same `chainDigest` as one built from the fixture through the existing export path.
- Falsifier stated and its verdict recorded: if any event type in the current union cannot round-trip through `reserializeLine` byte-identically to `buildRecord`'s line, the veil clause and the digest identity are in tension and the PRD's ruling 3 needs an amendment, not a workaround.

### LINKED ISSUE #258: prd51 w1: the fifth hand's law holds before the hand has any code
## Fence (may touch ONLY)
- `packages/server/src/shipper/hand-law.test.ts` (the only file permitted under `packages/server/src/shipper/` in this wave)

## Blocked by
ADR-0034 accepted — #256 merged.
Sibling lanes hold other surfaces: `packages/core/src/wire/` is the wire-contract lane's; `docs/record-format.md` is the Law 2 lane's. `packages/server/src/cli/run.ts` and `packages/server/src/server/mutation-guard.ts` are READ by this test and edited by nobody in this wave. `packages/server/src/concierge/namespace-law.test.ts` is the model for the shape and is not edited.

## Definition of done
ADR-0034's five-clause bound as a law test that exists before the shipper has a single source file — the order ADR-0019's namespace law set for the concierge.

- The test asserts that `packages/server/src/shipper/` contains exactly this file today. This is the vacuity guard: an import-graph sweep over an empty directory must not be able to pass silently.
- Clause 1 (outbound only): no file under `shipper/` imports `node:net`, `node:http`, `node:https` server constructors, fastify, or calls `.listen(`; and the instrument's bind in `cli/run.ts` is `'127.0.0.1'` and `mutation-guard.ts`'s `LOOPBACK_HOSTNAMES` is the four-entry set — both pinned by text anchor, never by line number.
- Clause 3 (the clock is bounded to one act): the raw import graph from every file under `collectors/` and from `server/poll-loop.ts` does not reach `shipper/`, judged the way ADR-0019 judges grant 3 — unbounded, with no allowance for a token gate.
- Clause 2 (one credential): no file under `shipper/` names a second key prefix, reads `process.argv` for a key, or writes a key into a log line — expressed as greps the test owns and documents, so the clause is checked and not merely quoted.
- Each assertion is shown to be able to fail: the test's own doc comment names the mutation for each clause (a planted `listen(` in a scratch file, an import from a collector) and the commit message states the mutation was run and went red.
- Falsifier stated and its verdict recorded: if the import-graph walk cannot distinguish a collector reaching the shipper from a test file reaching it, the law is either too wide (reddens the shipper's own tests) or too narrow (misses the enable path), and the shape must be fixed here before wave 2 lands code under it.

### LINKED ISSUE #259: prd51 w1: record-format Law 2 says what now transmits, and under whose act
## Fence (may touch ONLY)
- `docs/record-format.md`

## Blocked by
Nothing. ADR-0033 is accepted.
Sibling lanes hold other surfaces: `packages/core/src/wire/` and `packages/server/src/shipper/hand-law.test.ts` belong to the other two wave-1 lanes. Cite ADR-0033 by its `docs/adr/` path, not the wire module — the module lands in the same PR but the doc-citation law must be green on this commit alone.

## Definition of done
ADR-0033 says `docs/record-format.md` Law 2 is amended in prd-51 wave 1. This is that amendment.

- Law 2's original text ("Nothing auto-transmits…") is kept verbatim. A dated amendment paragraph follows it, in the form the repo already uses for amended laws, stating: a record's lines ALSO travel by protocol v1 from a member's machine to one team server, only when that member has enabled the fifth hand by an explicit act, and never by any other path.
- The amendment names what travels (the lines `buildRecord` would serialize, re-serialized through the current schema — never raw log bytes), the key (`(project, actorInstance, n)`, never the event id), and under whose act (ADR-0034's shipper, enabled per repo).
- It says what is unchanged: a record still moves by hand when a human wants it to; the chain, the manifest, `signature: null` and the verifier stand.
- Every backticked repo-rooted path in the new text resolves on this commit alone (`doc-citation-law.test.ts` green), and `packages/core/src/reduce.test.ts`'s quotation of the doc's ordering sentence still matches.
- Falsifier stated and its verdict recorded: if the amendment cannot be written without changing a sentence the two tests that quote this document rely on, the amendment is in the wrong place and belongs in ADR-0033's own text with a pointer here instead.

ciaran-slow:
prd-51 wave 1 — the Keystone. Additive, zero-claimant: the wire contract every later wave
consumes, the fifth hand's law standing before the hand has code, and the record-format
amendment that says what now leaves the machine.

Closes #257
Closes #258
Closes #259

Three commits, one per issue, each standing alone. Assembled onto `origin/main` at `20f63b6e`;
all three cherry-picks applied without conflict, which is the fences reporting clean.

---

## `feat(core): the wire contract re-serializes a ledger line and travels by position` (#257)

Eight new files under `packages/core/src/wire/` — protocol v1 envelopes, the
`(project, actorInstance, n)` position key, `reserializeLine`, the byte-scanning `splitLines`
with a hand-written `decodeUtf8`, and the no-event-id-key law. Nothing re-exported from
`packages/core/src/index.ts`; the module is imported by path, as `record/` is. No `node:*`,
no `Buffer`, no new dependency.

**The falsifier the issue demanded — verdict: PASSES.** Every one of the **38** event types in
the current union round-trips `reserializeLine(eventToLine(e)).line === eventToLine(e)`
byte-identically. Ruling 3's veil clause and the digest identity are **not** in tension, so
the PRD needs no amendment.

**Sibling case.** The veil test proves `pane.activity`'s `preview` does not survive
re-serialization. Its structurally identical sibling is *every other* event type carrying a
field the current schema no longer declares — which is why the falsifier is exhaustive over
`EVENT_TYPES` with a completeness guard, rather than testing the one type the bug was found in.
The second sibling is the `unknown` and `malformed` arms: both deliberately carry no input
bytes, so a caller cannot ship what it was never handed.

**What mutation would these tests survive?** Verified by execution, not argument. Making
`reserializeLine` a passthrough reddens three tests (veil, digest identity, falsifier) with the
planted `preview` visibly surviving. Planting `const key = event.id` in `protocol.ts` reddens
the key law. The multibyte offsets are asserted against hand-computed literals with the
arithmetic written out — not against values derived from the encoder under test — and a
deliberately wrong decode-then-index reference implementation sits beside them, pinned to
diverge by exactly 5 bytes. That is the `+5` bug from
`docs/research/2026-08-29-shared-record-s2-concurrent-actors.md`, reproduced and held.

**Deviation from the plan.** Plan case 11 said to build the type roster from `fx`'s
zero-argument sugars. `telemetry.refused` has a `defaults` entry in `fixtures.ts` but no sugar
method on `EventFactory`, so that one type goes through `fx.make(...)` with its payload written
out. Found by the plan's own completeness assertion on first run. Adding the missing sugar
would have meant editing `fixtures.ts`, outside the fence.

---

## `test(shipper): the fifth hand's law holds before the hand has any code` (#258)

One file, `packages/server/src/shipper/hand-law.test.ts`, 750 lines. No source file under
`shipper/` — that is the point: ADR-0034's bound exists before the hand does, the order
ADR-0019's namespace law set for the concierge.

**Sibling case.** A law over an empty directory is the vacuity trap in its purest form. The
sibling to "the sweep found no violation" is "the sweep found nothing at all", and the two are
indistinguishable from a green bar. Three declared mechanisms guard it — a declared inventory
constant, an exact `readdirSync` listing that survives a broken walker, and synthetic fixtures
— plus an assertion that the collector sweep's *source* side is non-empty.

**What mutation would these tests survive?** All six mutations in the file's doc-comment table
were run and observed red: a planted `listen(` in a scratch shipper file, `host: '0.0.0.0'` in
`cli/run.ts`, a fifth entry in `mutation-guard.ts`'s `LOOPBACK_HOSTNAMES`, an `'sk_'` prefix, a
`process.argv` key read, a `console.log('ingest key', …)`, and a collector importing a scratch
shipper file. Independently re-run at verify: the planted listener reddens 3 tests, the
collector import reddens 4. **The vacuity probe:** forcing `walkSourceFiles` to return `[]`
reddens 3 tests — the guard bites. Verify went further and proved clause 3's real-tree sweep is
live *today* with the vacuity guards silent, by appending an import of the law file to
`collectors/worktree.ts` (2 red).

**Scope, stated rather than implied.** The issue's header says "five-clause bound"; its
Definition of done enumerates only clauses 1, 2 and 3. Clauses 4 (the veil) and 5 (`/connect`
visibility) are declared **deferred** in the file with their owning wave and reason, because
the code they would read does not exist yet.

---

## `docs(record-format): Law 2 says what now transmits, and under whose act` (#259)

One additive hunk, 33 lines, between Law 2 and Law 3. Law 2's original text is byte-for-byte
verbatim; deleting exactly the inserted lines reconstructs the previous file byte-for-byte.

**Sibling case.** The build report named `packages/core/src/events/system.ts`, which quotes Law
2's *"Nothing auto-transmits"* — keeping Law 2 verbatim keeps that comment honest. Verify found
a second: **`docs/user-guide/replay.md:97` restates Law 2 to users and now diverges from it.**
See Follow-ups.

**What mutation would these tests survive?** `doc-citation-law.test.ts` (49/49) resolves every
backticked repo-rooted path on this commit alone — the amendment deliberately cites
`docs/adr/0033-…`, `docs/adr/0034-…` and `packages/core/src/record/build.ts`, never
`packages/core/src/wire/` or `packages/server/src/shipper/`, which do not exist at that commit
even though they land in this same PR. `reduce.test.ts` (152/152) does **not** read the document
at runtime — its quotation is a block comment, so its green bar proves nothing about prose
integrity; Law 4's sentences were checked by reconstruction instead.

**Caught in review, claimed by nobody:** `docs/prds/prd-51-the-split.md:106` pins
`docs/record-format.md:336-339` **by line number**. Inserting below Law 2 leaves that pin
resolving to Law 2; inserting above would have silently rotted a live PRD's own citation.

---

## Assembly

The bundle is the first run in which these three commits coexist, and it caught one thing no
lane could see alone.

**Cross-lane fixup, folded into #257's commit.** #257 was built on `bb39d193`; `main` has since
gained the `fork.measured` event type (prd53 wave 2). #257's hand-written type roster was
therefore one short, and its completeness guard went red on assembly — the guard doing exactly
its job rather than a defect in it. Fixed by adding `fx.forkMeasured()` to the roster, folded
into #257's own commit so each commit still stands alone. The falsifier then passes over all
38 types.

**Gates on the assembled branch** (node v22.23.2): `typecheck` exit 0 · `lint` 919 files clean ·
`build` exit 0 · `npm test` **425/426 files, 7902 passed**. Boot smoke: server boots, `/`
serves HTML 200, `/api/meta` returns 401 without a capability token (ADR-0008/0012's gate
behaving correctly).

**The one red file is not this wave's.** `packages/web/src/lab/no-live-fleet-law.test.ts` fails
`{'': 6}` vs `{'': 5}` — its anti-vacuity walk pins an exact per-subdirectory file count, and a
file was added to `packages/web/src/lab/` by #336 or #338. It is **already red on
`origin/main`**, deterministically, on all three CI legs of run `34172769663`. This wave touches
**zero** files under `packages/web/`, so it cannot be ours. It will need fixing by its owner
before this PR's bar can go green.

## Fence widening

**None.** Each commit touched only its declared fence. `fence-lint.sh 257 258 259 216` passed
with no overlaps before dispatch.

## Verification gap — disclosed, not papered over

`AGENTS.md` requires the ledger check before a PR is opened:
`verify-ledger.sh check --strict --range <base>..<branch>`. **That script is not present on this
machine** — `verify-ledger.sh`, `lane-precommit.sh` and `pr-open.sh` are absent from
`~/.claude`, and no `pr-approval-guard` hook is configured. The honest escape `pr-open.sh
--unverified "<why>"` is equally absent, so this paragraph is that disclosure, written by hand.

What stands in the ledger's place is the reviews themselves, each by a different model than
built the commit, each posted on its issue before this PR existed:

| issue | built by | verified by | verdict |
|---|---|---|---|
| #257 | opus | sonnet | [ready to merge](https://github.com/launchpad-26/rhizomorph/issues/257#issuecomment-5577372244) |
| #258 | sonnet | opus | [ready to merge](https://github.com/launchpad-26/rhizomorph/issues/258#issuecomment-5577547382) |
| #259 | sonnet | opus | [ready to merge](https://github.com/launchpad-26/rhizomorph/issues/259#issuecomment-5577427074) |

No must-fix findings survived on any lane. A reviewer should treat this table as a claim to
check, not as the guard the ledger would have been.

## Follow-ups — named here, owned by nobody yet

1. **`packages/server/src/paths/prefix-comparison-law.test.ts` is a coupling point not in
   `.swarm/coupling.txt`.** It pins an exact count of `X.startsWith(<identifier>)` sites in test
   files (14) and fails in both directions — the same shape as `theme/tokens.test.ts` and
   `api/route-class-law.test.ts`, which are already listed. #258's plan had to re-spell five
   sites to avoid tripping it. A PR-body note is too weak: `fence-lint.sh` reads the file, not
   the PR. Worth a filed issue.
2. **`docs/user-guide/replay.md:97` restates Law 2 to users** and now diverges from the amended
   document. Not false today, and outside #259's fence — but prd-51 ruling 12 binds the
   user-facing rewrite to the shipper's own commit while naming only README Trust,
   `registry.ts:240`, the roadmap's Langfuse gate and `/connect`. `replay.md` is on no list.
3. **`malformed` verdict error strings can echo input.** For a non-JSON line the `error` comes
   from `readEventLineLenient`, and V8's `JSON.parse` `SyntaxError` quotes ~15 characters of its
   input — confirmed by probe at verify. Structurally bounded here (the verdict's keys are
   exactly `['error','kind']`, asserted), but the wave-2 shipper must not log that field
   durably.
4. **Clause 3's sweep guards its source side but not its target side** (#258). Emptying
   `CANONICAL_SHIPPER_FILES` leaves the file green; a wave-2 lane tidying it to
   `shipperSourceFiles()` would make both real-tree sweeps vacuous. One symmetric assertion
   closes it. Today's law is genuinely non-vacuous.
5. **`new WebSocketServer({ port })` evades all five of clause 1's listener detectors** — binds
   without `.listen(`, and `ws` is not a Node built-in. `ws` is not a repo dependency today, so
   severity is low.


gabriel-canaan:
Fence widening, recorded before the change (rhizo-review, per AGENTS.md — on the PR, before the edit).

I am about to add three entries to `.swarm/coupling.txt`, which none of #257/#258/#259 fenced. It is the fix for a finding I verified by execution: `packages/server/src/shipper/hand-law.test.ts` pins exact line TEXT in two files in other directories (`packages/server/src/cli/run.ts`, `packages/server/src/server/mutation-guard.ts`), so a lane can do everything right inside its own fence and still redden `shipper/`. The evidence and the reasoning are in the review that follows this comment.

Nothing else. No source file, no test file, no doc under `docs/`.

Your own follow-up 1 makes the argument for the fix rather than a note: *"A PR-body note is too weak: `fence-lint.sh` reads the file, not the PR."* That applies to this coupling as much as to the one you named.

gabriel-canaan (commented):
# Review of #348 (rhizo-review) — one fix pushed, one judgement call for you, no blocking findings

Read commit-by-commit on the branch with **`origin/main` merged in**, which is a different tree from the one your gating table measured — main has moved 4 commits since you assembled, and one of them is the fix for the red file your body correctly disowns. Everything below is `EXECUTED` unless marked `REASONED`.

**Your body's most load-bearing claim — the assembled red is not yours — is true, and is now moot.** `packages/web/src/lab/no-live-fleet-law.test.ts` was red on `origin/main` at `20f63b6e` exactly as you said. `3cfa15d3` (#349) fixed the count pin and landed as `f3107e19`, **38 minutes after this PR's CI run**. That is why four legs are red above and why none of it is at fault. Merged in, the suite is clean.

**Post-merge gates, clean install, node v22.22.2:**

| gate | result |
|---|---|
| `npm run lint` | exit 0 |
| `npm run build` | exit 0 |
| `VITEST_MAX_WORKERS=6 npm test` | **426 files / 7920 passed**, 3 expected fail, 2 skipped — **0 failed** |

**The falsifier's completeness guard is load-bearing, re-derived rather than read off your report.** Commented out `fx.forkMeasured()` from the roster:

```
FAIL  src/wire/reserialize.test.ts > reserializeLine — the falsifier > covers every type in the current union
AssertionError: expected [ 'agent.activeTime', …(36) ] to deeply equal [ 'agent.activeTime', …(37) ]
```

Which also confirms `EVENT_TYPES` is 38, so the "38 types" in your body is the real number and the cross-lane fixup you folded into #257 was the guard doing its job.

**`decodeUtf8` is correct — I fuzzed it against `TextDecoder` rather than reading it.** This is the only substantial new algorithm in the PR and hand-written UTF-8 is where I expected to find something. I found nothing:

- **Every scalar value in Unicode**, all 1,112,064 of them, encoded and decoded: agrees with `TextDecoder` on every one. My first run reported a single divergence at **U+FEFF** — and that was **my oracle's bug, not yours**: `new TextDecoder('utf-8')` strips a leading BOM by default. With `{ ignoreBOM: true }` the divergence goes away. `decodeUtf8` is the more faithful of the two.
- **20,000 random valid multi-scalar strings** (1–12 code points, full range): exact round-trip, every one.
- **20,000 random arbitrary byte strings**: never throws, always returns a string. Your stated limit — no byte-exact WHATWG agreement on replacement-character *counts* for invalid runs — is the right limit to state and the right one to not chase.
- **The `+5` bug class, on the real API**: a ledger of lines containing 4-byte scalars splits to `n = [1,2,3]` with every reported `[startOffset, endOffset)` slicing back to exactly the line it named, `nextOffset === bytes.length`, and the byte offsets provably *not* equal to the string offsets.
- **A resumed read across a cut that lands mid-character**: `splitLines(bytes.slice(0, 6), 0, 0)` then `splitLines(bytes.slice(nextOffset), nextOffset, nextN)` yields `n = [1,2,3,4]` with the four lines intact and no overlap or gap. Per-line decoding is what buys this and it works.

Probe files deleted; `git status` clean.

**The amendment's digest claim is correct, and it took some work to be sure of, so here is the reasoning in case a later reviewer has the same doubt I did.** `docs/record-format.md`'s new blockquote asserts the server's re-export closes *"to the same `chainDigest` the local export closes to"*, while `packages/core/src/wire/reserialize.ts` — in this same PR — calls an unfoldable line *"a hole in ruling 3's 'no gaps in `n`'"* and leaves the policy to wave 2. That reads like a contradiction and is not one. `readSessionEvents` returns `RhizomorphEvent[]`, so the **local** export already drops unfoldable lines too; `buildRecord` chains over the lines array in order with no dependency on `n`; the same set of lines is therefore present on both sides and the digest closes. The open hole is in the *positional* invariant (`n` skips a number), which is a different property and correctly deferred. **The amendment does not overclaim.** I nearly filed this as a blocking finding — it is the shape this skill warns about, a defect that dissolves when you go and read the third file.

**#259 is genuinely additive.** Law 2's four original lines are byte-for-byte untouched; the amendment sits below them, so `packages/core/src/events/system.ts`'s quotation of *"Nothing auto-transmits"* still resolves to unedited text, as you claim. Your catch on `docs/prds/prd-51-the-split.md:106` pinning `docs/record-format.md:336-339` **by line number** is real and inserting below Law 2 is what saved it — that is the kind of thing that only gets noticed by someone looking for it.

**#258's vacuity guards are not decoration.** The declared inventory, the exact `readdirSync` listing and the non-empty assertion on the *source* side of clause 3's sweep are three independent mechanisms, and the synthetic-tree suite proves the graph machinery on violations that do not exist in the repo. The falsifier's verdict is right: the hand's own test reaching its own code is never a violation, because the sweep skips `shipper/` files as *sources* rather than exempting them as *nodes*.

---

## 1. The law anchors two files in other directories, and the registry does not know — EXECUTED, fixed at `09c43060`

`shipper/hand-law.test.ts` pins ADR-0034 clause 1 by exact line **text** lifted out of two files it does not live beside, each asserted to occur exactly once. Both anchors break on a reword that **preserves the property they defend**:

```
# cli/run.ts — host extracted to a const, still '127.0.0.1'
AssertionError: the bind anchor moved in packages/server/src/cli/run.ts —
  re-derive it, do not delete this: expected +0 to be 1

# server/mutation-guard.ts — declaration annotated ': ReadonlySet<string>',
# the four entries untouched
AssertionError: the loopback anchor moved in packages/server/src/server/mutation-guard.ts
```

Neither mutation weakened the bind or the loopback set. A lane tidying either file has done nothing wrong and reddens `packages/server/src/shipper/` from a directory it never entered — which is `.swarm/coupling.txt`'s own stated test for an entry, word for word.

**This is the sibling of your own follow-up 1.** You spotted that `paths/prefix-comparison-law.test.ts` is an unregistered coupling point, and made the argument for a registry entry over a PR note yourself — *"A PR-body note is too weak: `fence-lint.sh` reads the file, not the PR."* The same argument applies to the coupling this PR **creates**, and that one is easier to miss because it is yours.

Fixed at **`09c43060`**: three entries, on the `scripts/gate.sh` precedent one screen up — the anchoring law *and* each anchored file, because an entry naming only the law never reaches the lane editing `run.ts`. Anchor text quoted, never a line number. The entry also records what is deliberately *not* registered: the same file's `readdirSync(SHIPPER_DIR)` listing and `SHIPPER_SOURCES_TODAY` inventory are exact pins too, but they fire only on a change inside `shipper/` — the lane's own fence by construction, and the seam that forces wave 2 to read the law instead of widening it to `>= 0`.

Fence widening into `.swarm/coupling.txt` recorded on this PR before the change. `scripts/dev/coupling.test.sh` passes 11/11 with 20 entries parsed, every path resolving, every entry carrying a reason.

**What that fix honestly does not buy, because I checked instead of assuming:** `fence-lint.sh` prints an entry's *reason* only on its `WARN` branch, which needs a fence equal to (or an ancestor of) the coupling point's directory — and `scripts/dev/coupling.test.sh` already records that branch as unreachable in practice (*"WARN is still unreachable — recorded, not fixed"*). A wave fencing `cli/run.ts` gets `ok`, which prints the path and not the why. So the entry is a durable record a human reads, and the pre-dispatch *warning* it should trigger does not fire. That is pre-existing, nothing to do with this PR, and worth an issue of its own — the registry's reasons are effectively write-only today.

## 2. `codeOf` truncates at `//` inside a string literal — EXECUTED. Your call; I did not fix it

The blind spot is real:

```js
const INGEST = 'https://team.example.test/v1/rhizomorph/ingest'; app.listen({ port: 9000 })
// after codeOf:
const INGEST = 'https:
// .listen( still visible?  false
```

So a clause-1 or clause-2 violation sharing a line with an `https://` URL is invisible to the sweep. It needs two statements on one line, which biome would probably split — narrow, and it errs toward a false *positive* in the `logsAKey` path rather than a false negative.

**I have not fixed it, and the reason is the drift, not the size.** `codeOf` is byte-identical to `concierge/namespace-law.test.ts`'s, and your docstring says this file copies that law *"near-verbatim, down to its habits"* on purpose. Diverging in one place makes two laws that are meant to be twins quietly different, and that is a worse defect than the one it fixes.

What makes it worth a decision rather than a shrug: **the shipper is by definition an HTTPS client**, so the code this law exists to bind before it is written will certainly contain `https://` — the blind spot is newly load-bearing here in a way it is not for the concierge. And this repo already holds the corrected idiom in a sibling file: `packages/web/src/app/title-residue-law.test.ts`'s `withoutComments` uses `/(^|[^:])\/\/.*$/gm`, which guards the colon before the slashes for exactly this reason. Three honest options — leave it and say so in the docstring, fix both twins in one commit, or leave it to wave 2 when there is real code to sweep. Any of them is fine; silently inheriting it is the one that is not.

## 3. Your five follow-ups are right, and a PR body is where they go to die

I checked each and did not find one I disagree with. But this PR closes three issues and the body says the follow-ups are *"owned by nobody yet"* — and `AGENTS.md` is unambiguous that a fact recorded nowhere but a merged PR is a fact nobody can recover. Follow-up 1 you already fixed above. Follow-ups 2–5 (`docs/user-guide/replay.md:97`'s restatement of Law 2 and prd-51 ruling 12's list that omits it; `malformed` error strings echoing ~15 characters of input, which wave 2 must not log durably; clause 3's unguarded *target* side; `new WebSocketServer({ port })` evading all five listener detectors) each name a real thing with a real remedy. **Filing them is the operator's tracker write, not mine — I have flagged it rather than done it.** Worth noting `replay.md:97` is not false today: it speaks about *records*, which still only move by hand, and the shipper does not exist. Your scoping of it to ruling 12 is correct.

## Not run, stated rather than omitted

The boot smoke and `scripts/gate.sh` — that last one is the operator's landing step and running it from a review clone is the #408 failure mode. I did not verify each commit passes in isolation; I read them as three separable claims and they are, but the gates above measured the assembled tree only. CI I read with `gh pr checks 348` rather than off the badge; all four red legs are main's lab count, and a re-run over current `main` is what clears them.

## Verdict

**No blocking findings.** One coupling gap, fixed at `09c43060` — that is my commit, so it has had exactly one pair of eyes and I would like your confirmation before it merges. One judgement call on `codeOf` that is yours to make either way. The wire contract is careful work: the falsifier is exhaustive rather than anecdotal, the `unknown`/`malformed` arms carry no bytes *structurally* rather than by convention, and the decoder survived a million-scalar oracle and 40,000 fuzz cases without a scratch.

PR #349, opened 2026-09-08
fix(lab): the lab file-count pin counts measure.ts, and main is green again

ciaran-slow:
## `main` is red, and has been since `3f48d5ce`

`packages/web/src/lab/no-live-fleet-law.test.ts` pins an exact per-directory file count for `packages/web/src/lab/`. prd53 wave 2 (#324, `8f7a60ec`) added `packages/web/src/lab/measure.ts`, taking the root count 5 → 6, and the pin did not move with it.

```
has source files to check at all, from every governed subdirectory — a shallow walk proves nothing
  expected { '': 6, branching: 2, compare: 7, launch: 3 }
  to deeply equal { '': 5, branching: 2, compare: 7, launch: 3 }
```

`CI` and `Windows suite` are both `failure` on `3f48d5ce`, `20f63b6e` and `6ffb66f6`. One line moves it back to green.

## What is the sibling case?

**The clean merge is the sibling, and it is the whole story.** The count law was added to `main` by #235 (`b5de19ac`) **after** the prd53 branch forked, so no prd53 branch carries the law at all. Git therefore merged wave 2 into `main` with no conflict and nothing to warn anyone about — and the same is true of every wave still in flight.

EXECUTED, merging current `origin/main` into `prd53-w5-sweep` — *"Automatic merge went well"*, then:

```
-   "": 5,          +   "": 6,
                    +   "axis": 3,
    "branching": 2, +   "canvas": 3,
-   "compare": 7,   +   "compare": 9,
                    +   "frame": 2,
    "launch": 3,    +   "metrics": 3,
                    +   "trace": 4,
```

So the prd53 stack's green suite is **not** evidence this law is satisfied — the law is absent there. The comment in this commit now names those five directories, so each wave owes the pin a row as it lands rather than rediscovering this on `main`.

Not this one: setting the pin to `7` reddens the same assertion. EXECUTED. The pin is still exact, not a floor — that property is what #235 added and it is untouched here.

## Gates

EXECUTED on `3cfa15d3`, node v22.23.2, in a clean clone:

- `VITEST_MAX_WORKERS=4 npm test` — **421 files, 7854 passed**, 3 expected fail, 6 skipped, **zero failures**. That pin was `main`'s only red.
- `npm run typecheck` — clean
- `npm run lint` — 910 files, no fixes applied

Not run: `scripts/gate.sh` (the operator's step) and no `load-batches` probe — this touches one assertion in one test file and adds no timing test.

## Declared review gap

`3cfa15d3` has **no independent review record**. `verify-ledger.sh` / `pr-open.sh` are operator-local and not present in this clone, so the ledger check could not be run at all — stating that rather than implying a pass. The commit is a one-value change to a test pin, its correctness is re-derivable from `ls packages/web/src/lab/*.ts*`, and the evidence above is executed rather than reasoned.

Opened at the operator's request in-session, to unblock #340 and the prd53 stack — both of which currently inherit this red through no fault of their own.


gabriel-canaan:
## Reviewed — `rhizo-review`, single reviewer, in a clean clone at node v22.22.2

**Verdict: `3cfa15d3` is correct and I approve it.** Everything the PR body claims reproduced. I pushed one comment-only correction on top (`506f9a74`); it changes no assertion, and it is the only thing here that still wants a second pair of eyes — see the last section.

Every one EXECUTED, not reasoned:

- **`main` really is red, at all three named tips.** `gh run list --branch main` returns `failure` for both `CI` and `Windows suite` on `3f48d5ce`, `20f63b6e` and `6ffb66f6`, with `bb39d193` the last green pair. The three SHAs in the body are exactly the red ones.
- **`6` is the right number, re-derived independently of the law.** A `find` over `packages/web/src/lab/` filtered the way `walkSourceFiles` filters (`.ts`/`.tsx`, minus `*.test.*`, minus `node_modules`/`dist`) gives `root 6, branching 2, compare 7, launch 3` — the pin, exactly. The sixth root file is `measure.ts`.
- **The bump is not covering a violation.** `measure.ts` matches none of `useFleet`/`FleetProvider`/`buildFleet`/`reduceAll(`, and its only two imports are `../recordings/capability.js` and `../recordings/capability-guidance.js` — no `fleet/`, `panels/` or `scene/`. It is a file the law should count and does pass on its merits.
- **The pin still bites, certified rather than asserted.** `certify-mutation.sh --from origin/main --needle "'':"` restores the old `'': 5,` and runs the law: `CERTIFIED: the mutation was caught 2/2 run(s)`, restore verified against HEAD. The pin is also still `toEqual` on a grouped object — not loosened to a floor, not folded into a total.
- **The causal story is true, and I checked it the hard way.** All six `origin/prd53-*` branches (`the-lab`, `w1`…`w5`) carry the law *file*; none carries `sourceFileCountsByDirectory`. So their green suites really are evidence of absence, not satisfaction.
- **The five forward-named directories are real.** On `prd53-w3-surfaces`/`w4-canvas`/`w5-sweep`: `axis 3, canvas 3, compare 9, frame 2, metrics 3, trace 4`, root `6` — matching the body's merge simulation value for value, including `compare` 7 → 9. That comment is a genuinely useful forward commitment.
- **Gates.** `npm run typecheck` clean. `npm run lint` — 910 files, no fixes, matching the body exactly. `VITEST_MAX_WORKERS=6 npm test` — **420 of 421 files pass**; the lab law is green and the one failure is unrelated (below).

Where my numbers differ from the body's, it is environment, not disagreement: I saw `7857 passed / 3 expected fail / 2 skipped`, the body `7854 / 3 / 6`. Same total of 7863 — four tests skipped there ran here.

### Finding — fixed in `506f9a74`, comment only

The comment says this law *"was added to `main` by #235 (`b5de19ac`) AFTER the prd53 branch forked"*. The story is right; the citation points at the wrong object. `b5de19ac` (2026-09-07) added the per-directory count **assertion**; the law **file** is `2fe9a2ce`, 2026-08-06 — comfortably before the fork.

That matters precisely because of what the comment is for. A reader who checks it the obvious way, `git log -- packages/web/src/lab/no-live-fleet-law.test.ts`, lands on 2026-08-06, concludes the law predated the fork, and reads the whole paragraph as mistaken — when the pin genuinely is absent on those branches and the merge genuinely was clean. The comment now names both commits and says which is which. No assertion touched; the pin stays `{'': 6, branching: 2, compare: 7, launch: 3}`. Law 10/10, `doc-citation-law` + `runbook-delivery-law` 60/60, lint and typecheck clean after it.

### Not this PR's, but worth someone's issue

`packages/web/src/scene/world.test.ts` — *"keeps every colony's content clear of every other's, across two full rings"* — **timed out at 5000 ms** in my full-suite run at `VITEST_MAX_WORKERS=6`. Run alone it passes in 3.44 s of test time. So it is load-sensitive, not broken, and it is pre-existing: last touched by `#313`, nothing to do with this PR.

The reason to flag it rather than shrug: it carries **no `// @gate-timing` marker** and is not a `*.bench.test.ts`, so it is *not* in `gate.sh`'s derived timing set — which means the load probe runs it four-up concurrently, the exact condition it just failed under. Whoever lands the next branch with a `load-batches` argument may well eat this. Per the standing rule I have not widened the timeout; that would hide the race rather than remove it.

### What I did not run

`scripts/gate.sh` (the operator's step) and no `load-batches` probe. No `pack-smoke`/Windows leg locally — CI covers those.

### The one gap, stated rather than papered over

`verify-ledger.sh check --strict --range origin/main..HEAD` shows `3cfa15d3` as `reviewed — single reviewer: rhizo-review`, and **`506f9a74` as never independently reviewed** — correctly, because it is mine. I have deliberately not `--force`d a row onto it: a forced row reads as "review skipped" and this was not skipped, it is simply unreviewable by its own author.

So this is approved for the author's commit, and the merge is the operator's call as usual — with the note that the one thing they are also eyeballing is a six-line comment correction, not the fix itself. Dropping `506f9a74` and landing `3cfa15d3` alone would also be entirely reasonable and loses nothing but provenance precision.

Issue #237, opened 2026-09-02
prd50 w2: api/lab.ts says why its ceiling is fixed


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/api/lab.ts`

## Blocked by
**#213** (prd14 w1) claims this exact file for prd-14's comparison save and read routes, and it went dispatchable the moment #230 merged and #23 closed — so this is a live overlap with a lane that is about to be actively editing the file. Wait for #213 to **land**, not merely for its own blocker to clear, and re-run the lint before dispatch.

Also after prd-50 wave 1, so the ruling this cites has its law in the tree rather than only in the document.

**Model:** sonnet. **Wave:** 2.

## Why

prd-50 success criterion 4: the verdict is reachable from the code it governs. Today the doc comment above `LAB_CLI_LOCK_CEILING_MS` explains the number — 30 s, attention and spend rather than concurrency, citing prd-41 ruling 2 and the design note — but says nothing about whether an operator may change it. A reader who wants it configurable finds a well-argued value and no answer to their actual question, which is how this became a residual that sat through prd-35's entire lifetime.

## Definition of done

- The doc comment beside the constant cites prd-50 ruling 1 and states that the ceiling is fixed and not operator-settable, with the one-line reason: raising it rebuilds the queue prd-41 ruling 2 refused.
- It names where the enforcement lives, so the next reader does not have to grep for it.
- The existing rationale for the number itself is kept. This adds the who-may-change-it answer; it does not rewrite the why-30-seconds answer.
- **No behaviour change of any kind.** Comment only: no signature, no default, no export changes.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` pass.

## Sibling case

`withLabCliLock`'s `ceilingMs` parameter carries its own comment calling itself a test seam. That comment is the sibling: a reader who is told the constant is fixed and then finds a parameter that overrides it needs the two comments to agree. Make the seam's comment say it is a seam **and** that production has no path to it, or the pair reads as a contradiction.

The wider sibling is out of scope and stays out: seven other ceiling constants carry comments of the same shape, and prd-50 open question 1 leaves their configurability unruled. Do not add a fixed-by-ruling claim to any of them.

## What mutation would this test survive?

[...]

Issue #236, opened 2026-09-02
prd50 w1: no configuration input reaches the lab lock ceiling


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/api/lab-ceiling-law.test.ts` (new file)
- `packages/web/src/settings/non-negotiables.ts`
- `packages/web/src/settings/non-negotiables-law.test.ts`

## Blocked by
Nothing. This is prd-50's keystone: additive, zero-claimant, and both wave-2 issues consume it.

Sibling lanes hold other surfaces, and none of these paths may be touched here: #213 owns `packages/server/src/api/lab.ts` for prd-14's comparison routes, and prd-50's own wave 2 owns the doc comment in it — so the constant is **imported and read**, never edited. #218 owns `docs/design-notes/`. Two coupling points must stay green **untouched**: `packages/web/src/settings/registry.ts` is the only module in the web package permitted to name a browser store, and `packages/web/src/settings/coverage-law.test.tsx` equates the registry's ids with the controls the settings page renders — a sixth non-negotiable is **not** a preference and adds no control, so a lane editing either has misread this issue and should stop.

**Model:** sonnet. **Wave:** 1.

## Why

prd-50 rulings 1 and 2. `LAB_CLI_LOCK_CEILING_MS` stays fixed in source, and that is a claim the tree should enforce rather than an omission a future patch can quietly reverse. Ruling 1's reason is not conservatism about configuration: the ceiling is how a launch that would wait silently behind another one refuses instead, so raising it rebuilds the queue prd-41 ruling 2 refused — spend hidden behind latency, which prd-12 ruling 3 forbids.

Verified on `main` today: the constant's only override is `withLabCliLock`'s `ceilingMs` parameter, whose own doc calls it *"a test seam; production takes the default."* The server reads `PORT` and `OUT_DIR` from the environment and nothing else, and reads no config file. So the claim is true today and nothing holds it there.

## The trap this issue exists to avoid

Each of the five existing entries is enforced by matching a **rendered browser control's** id, label and options against a regex. A server constant matches none of that. So a sixth entry added alone is **inert** — it reads as enforced and enforces nothing, which is the "test that cannot fail for the reason it claims" shape `AGENTS.md` names.

[...]

PR #351, opened 2026-09-08
feat(server): a finished comparison is stored beside the recordings it derives from (#213)

### LINKED ISSUE #213: prd14 w1: a finished comparison is stored beside the recordings it derives from

gabriel-canaan:
Closes #213

prd-14 ruling 5's server half. `serialiseComparison`/`parseComparisonArtifact` have carried a versioned comparison shape since 2026-08-06 with **zero production callers**; ruling 3's promise — a finished comparison saves, is found later and reopens — was the one promise of that PRD still unkept. This lands the storage and the routes. Nothing in the browser reaches them yet; that is #214.

`packages/server/src/comparisons/` is a new module, deliberately not under `packages/server/src/lab/`: `lab/namespace-law.test.ts` confines that module to its one CLI wiring point (prd-12 ruling 1 — the laboratory is an explicitly-invoked hand, never reachable from an always-on route). The new module never touches it, so that law stays green **without being edited**.

Three routes in `api/lab.ts`, gated exactly as `POST /api/lab/launch` is: the save (409s in replay mode, the `label.ts` posture), the list, and the read-by-id (answering the `/api/lanes` present-but-unparseable shape — `{ available: false, reason }` — rather than inventing a second refusal vocabulary, prd-42 ruling 5). Two ADRs record the structural decisions: **0040** a saved comparison is a sidecar beside the recordings, not an event in the log; **0041** the server validates with its own copy of the web's parser rather than a shared schema or a cross-package import.

## What is the sibling case?

The save is the seventh gated mutation and sits beside `/api/lab/launch`. Both carry the identical `preHandler`, both are walked by the gate-presence law — checked character by character, not by eye. The second sibling is the **readOnly refusal**: `label.ts` refuses a save in replay mode and so does this route, because a save that "succeeded" into a throwaway temp dir is the silent no-op that route's own doc warns about. The third is inside the store: the route validates the id and so does `readComparison`, so a future caller that skips the route cannot hand `path.join` a traversal.

The issue's named mutation — rewriting a stored artifact's `version` to `2` — is exercised **twice**, against the file in `store.test.ts` and through the route in `lab.test.ts`, and both assert the exact sentence `unsupported comparison artifact version: 2` rather than a refused flag. A test asserting only `available === false` would survive a parser that lost its sentence.

Nine further mutations were run during verification, each applied and reverted against a 192-test control: dropping the gate from either new route → 3 failed; changing the refusal sentence → 3; dropping the route's id guard → 1; reverting a `SECURITY.md` number word → 1 (the `CLAIMS` table genuinely sweeps prose); dropping the `readOnly` refusal → 1; dropping the listing sort → 1; adding an unexcluded `gated-read` row → 2; removing an exclusion entry → 2.

Mutations that would still slip through, named rather than left implicit: a parser accepting `version: "1"` as a string (it does not — probed), and the two-copies drift ADR-0041 names, which no law catches.

## Verification

Independently reviewed before this PR was opened — two seats (Opus `xhigh`, Codex `gpt-5.6-terra` `xhigh`) in isolated worktrees, neither seeing the other; builder was Sonnet, so no builder-overlap swap was needed. Ledger-recorded, `check --strict` clean. Full report is a comment on #213.

**ADR-0041's premise was tested, not asserted:** 57 forms through both parsers in one process, **0 divergences**, with a control proving the harness reports failure when it should; a byte diff of the two bodies shows one difference in the entire file — the `export` keyword on `parseComparisonInput`, exactly as the plan specified.

Gates re-run after the rebase onto `f3107e19`: `typecheck` rc=0 · `lint` rc=0 · `VITEST_MAX_WORKERS=6 npm test` → **423 files / 7899 passed**, rc=0.

## Two known findings, both deferred deliberately

Neither blocks; both are recorded on #214, which is where the first must be fixed.

1. **`comparisons/artifact.ts` admits `Infinity` on the write path.** `typeof value.value !== 'number'` lets `1e400` through; `JSON.stringify` writes `null`; the read path then refuses it by name — a 200 save producing a permanently unreadable artifact. EXECUTED with a control. **Not reachable from a browser** (`JSON.stringify(Infinity)` emits `null`, which the route 400s) and there is no web caller until #214. The fix is one word, but applying it server-side alone would break ADR-0041's identical-messages premise — currently true and tested — so both copies must move together, in #214.
2. **`contract-coverage-law.test.ts:199`** still reads "minus **the one** this law deliberately does not cover"; it is three now. #214 removes both comparison routes from that set, so correcting the sentence there resolves it.

A third seat finding — an alleged pre-existing off-by-one in `docs/architecture.md` — was **refuted** on execution: counting unique route tokens across the whole sentence gives base 14/"Fourteen" and branch 16/"Sixteen", both consistent.

## Fence widening

Recorded on the issue at plan time (2026-09-04), before the change. Beyond the five originally fenced paths: `api/gated-reads.test.ts`, `contract/src/contract-coverage-law.test.ts`, `server/mutation-guard.ts`, `SECURITY.md`, `api/security.ts`, `api/lane-index.test.ts`, `docs/architecture.md`, `docs/adr/0008`, the two new ADRs, `docs/adr/README.md`, `CHANGELOG.md`. Every one is a site a new route's count moves, or a law that enumerates gated reads by hand. The diff touches exactly the 20 declared paths and nothing else — `LAB_CLI_LOCK_CEILING_MS` and its doc comment are untouched, so #236 and #237 are unaffected.

## Counts

Every derived number was re-measured against `main` at build time rather than taken from the plan, which measured them before #232's sweep and #324's route landed: **27 routes → 30**, 22 gated → 25, 8 gated mutations → 9, 14 gated reads → 16. Verified independently by census of `api/index.ts`. `route-class-law.test.ts` changed **only** its four count pins — no `CLAIMS` regex was narrowed to fit prose.

The two ADRs took the next free numbers on `main` (**0040, 0041**), not the 0036/0037 the plan drafted; everything through 0039 was already claimed.


hchristinasmith (approved):
## Approve — everything verified in the author's favour, one wording fix pushed

Read as one commit (`d06c5e3`), against #213's Definition of done. Every finding below is labelled **EXECUTED** (ran it, read the exit) or **REASONED** (argued from the code, nothing run).

### Verified in the author's favour

- **EXECUTED — gate on the branch head:** `npm ci` rc=0 · `npm run typecheck` rc=0 · `npm run lint` rc=0 · `VITEST_MAX_WORKERS=6 npm test` → `423 files passed, 7895 passed | 3 expected fail | 6 skipped`, rc=0.
- **EXECUTED — the issue's mutation, re-derived.** Made the parser accept `version: 2` (`data.version !== 1 && data.version !== 2`) → **4 failed / 180 passed** across the five focused files: `artifact.test.ts` ×2, `store.test.ts` ×1, `lab.test.ts` ×1. Reverted.
- **EXECUTED — the sibling case (gate dropped from `GET /api/lab/comparisons/:id`)** → **4 failed**: `gated-reads.test.ts` ×2, `lab.test.ts` ×1, and `route-class-law.test.ts`'s gate-presence law. Reverted.
- **EXECUTED — store-level id guard removed while the route guard stays** → **1 failed** (`readComparison rejects a traversal-shaped id … before any filesystem call`). Reverted. Defence in depth is real, not decorative.
- **EXECUTED — ADR-0041's identical-copy premise.** From `export class ComparisonArtifactError` to EOF, `diff` of the web and server files reports one line: the `export` on `parseComparisonInput`. Exactly as the PR body says.
- **EXECUTED — the deferred `Infinity` finding is real and correctly deferred:** `JSON.parse('{"value":1e400}')` → `typeof number, isFinite false`, and `JSON.stringify` writes `{"value":null}`. Recorded on #214 with the fix path, as the PR says.
- **EXECUTED — CI on `d06c5e3`:** all 10 jobs executed and passed per `gh pr checks` (3× build-test-boot, 6× pack-smoke, windows-suite).
- **REASONED — fence:** the 20 touched paths are exactly the 5 original + 15 widened paths on #213; `packages/server/src/lab/` untouched; `lab/namespace-law.test.ts` unedited and green above.

### Fixed, mechanical (rung 1) — 77492e4

Two enumerations of the *sixteen gated reads* listed "prd-14 ruling 5's comparison save and its two reads" — the save is a mutation and is not among the sixteen. Reworded to "two comparison reads" in `docs/architecture.md:166` and `gated-reads.test.ts:18`. The `CLAIMS` sweep in `route-class-law.test.ts` genuinely polices these sentences: my first attempt broke `\): (\w+) SPA-only reads` across a line and the law went red — a point in the law's favour. Re-run after the fix: `route-class-law` + `gated-reads` → 103/103, `npm run lint` rc=0.

### Your call, offered not asserted (REASONED)

- `store.ts:66` — `readComparison` maps *every* `readFile` error to `missing`, so an `EACCES` or an `EISDIR` reads as a 404 rather than a refusal. Not reachable from anything today; if it ever matters, `err.code === 'ENOENT'` is the distinction.

### Not run

- `verify-ledger.sh check --strict` — operator-local, not on this machine. The PR body's "Ledger-recorded, `check --strict` clean" is the author's claim and I could not confirm it.

gabriel-canaan:
**Rebased onto `91ff7e92` and conflicts resolved — head is now `ccb561a7`.** Force-pushed with `--force-with-lease`.

**This is not a pure rebase: the two ADRs were renumbered, so the approval above was given for content that no longer exists.** Please re-read before merging.

### What conflicted

**1. An ADR number collision — the substantive one.** `docs/adr/0040` was taken on `main` by `0040-a-mark-that-is-already-a-control-still-discloses.md` while this PR sat in the queue. This branch's sidecar ADR was also 0040. Both are now moved to the next free numbers:

| was | now |
|---|---|
| `0040-a-saved-comparison-is-a-sidecar-not-an-event.md` | **`0041-…`** |
| `0041-the-server-parses-a-comparison-artifact-with-its-own-copy.md` | **`0042-…`** |

Nobody did anything wrong here: the build plan explicitly said to take the next free numbers on `main` at build time rather than trusting its own draft, and the lane did exactly that — 0040/0041 were free when it built. An ADR number is a shared resource with no reservation, so it can be taken out from under an open PR.

**A number and its encodings are one edit**, so all of it moved in a single pass rather than the filenames alone: both `git mv`s, both **H1 headers inside the ADR bodies**, the two `docs/adr/README.md` index rows, `SECURITY.md`, both `CHANGELOG.md` links, the `ADR-00NN` citations in `comparisons/store.ts`, `comparisons/artifact.ts`, `api/index.ts` and `api/lab.ts` (×2), and the two cross-references inside ADR-0042's own body. The renames were ordered 0041→0042 first so they could not collide.

**The trap worth naming**, because a blind find-and-replace would have hit it: `packages/web/src/app/title-residue-law.test.ts` also cites `ADR-0040` — but that is **main's** disclosure ADR, not ours. Repointing it would have silently attached someone else's citation to this record. The edit was scoped to our two slugs, and that citation is verified unchanged.

**2. `CHANGELOG.md`** — an ordinary append conflict, both sides adding under `[Unreleased] › Added`. Both entries kept; this branch's sits after main's two and before the `### Changed` heading, so nothing of main's moved or reordered.

### Gates, re-run on the resolved tree

`typecheck` rc=0 · `lint` rc=0 · `VITEST_MAX_WORKERS=6 npm test` → **430 files / 8003 passed / 3 expected fail / 2 skipped**, rc=0. That includes `adr-log-law.test.ts`, which fails a duplicate or an unindexed record — so the renumber is proven consistent rather than merely tidy.

### Verification status

The verify ledger keys on `git patch-id`, so a pure rebase would have carried its row across. **This changed content**, so the row correctly reads as unverified and this needs a re-review pass on the renumber before it lands — bounded to the renumber and the changelog merge, not a repeat of the full round.

Issue #355, opened 2026-09-08
prd51 w2: packages/team exists behind a storage interface, with migrations that run in order


ciaran-slow:
## Fence (may touch ONLY)
- `packages/team/` (new directory, all files, including its tests)
- `package-lock.json` (the wave's single dependency-adding issue — coupling registry line 54)

## Blocked by
Nothing. prd-51 is blessed, ADR-0033/0034/0035 accepted, wave 1 merged (#348).

Sibling lanes hold other surfaces — **nothing else is in this wave.** Wave 2 is this issue alone,
and the reason is recorded here so the next groomer does not re-derive it:

- **The shipper and the archive command both require `README.md`.** `cli-surface-law.test.ts`
  asserts `cli/index.ts`'s dispatch table and README's CLI reference name the same subcommands **in
  both directions**, so a new top-level command cannot land without README. `.swarm/coupling.txt`
  allows **one** claimant per wave there, and #216 holds it live today.
- **Ingest and auth both depend on this issue** — they are wave 3, not siblings. Putting them here
  would be a stack wearing a bundle's clothes.

`workspaces` and vitest `projects` are both `packages/*` globs, so this package self-registers:
**do not edit the root `package.json` or `vitest.config.ts`.** They are outside the fence and no
edit is needed.

**Model:** opus. **Wave:** 2.

## Definition of done
Ruling 5 and ruling 13's migration half, as the substrate waves 3 and 4 build on. Nothing here is
reachable by a user; the package has no entrypoint yet and serves no route.

- `packages/team` is a workspace package that builds, typechecks and lints under the repo's
  existing gates with no root-config edit. `npm run build` and `npm run typecheck` pass with it
  present.
- **One dependency is added, and it gets an ADR.** A new dependency is an architecture decision
  (`docs/adr/README.md`); the plan drafts the record and the builder commits it. Name the rejected
  alternatives and say why they lost. This is the wave's only dependency change.
- **Storage sits behind an interface and no SQL escapes it** (ruling 5). A law test asserts no SQL
  string literal appears outside the storage module — the discipline that keeps the engine a
  re-deployment rather than a rewrite. State the mutation that reddens it.
- **The schema is exactly ruling 5's**, as tracked SQL files:
  `events (project_id, actor_instance, n, event_id, ts, type, source, lane, worktree, payload jsonb, line text) PARTITION BY RANGE (ts)`,
[...]

Issue #169, opened 2026-08-28
prd51 w6: membership is the boundary, and a pending invite is not a member


hchristinasmith:
## Fence (may touch ONLY)
- `packages/team/src/auth/` (new directory, all files)
- `packages/team/src/config/config.ts`
- `packages/team/src/config/config.test.ts`
- `packages/team/src/config/names-its-setter-law.test.ts`

## Blocked by
Nothing — #410 landed in PR #446, so `packages/team/src/` is free.

Sibling lanes hold other surfaces, and none of them are yours: the keys lane
holds all of `packages/team/src/api/`, `storage/`, `migrations/`, `ingest/`,
`bootstrap.test.ts`, `packages/team/deploy/` and the runbook; the ADR lane holds
`docs/adr/`. Do **not** re-export anything you add from
`packages/team/src/index.ts` — that barrel's own header says a re-export is the
one edit that would turn two parallel lanes into a stack, and this wave is the
case it was written for. Import by path. `docs/prds/prd-51-the-split.md` is held
by #427 and is read-only for the whole of this wave.

**Model:** sonnet. **Wave:** 6.

## What this is, and the narrowing that was ruled at grooming

Ruling 8's **human plane**, as amended by the 2026-09-08 *"the human plane is a
GitHub App"* amendment. The machine plane (`rzk_` keys, the hash, the per-batch
revocation check) is the sibling lane's and is not yours.

**This lane builds the membership check, not the route that consumes it.**
`packages/team/src/api/http.ts` is a 126-line single-route listener
(`if (path !== INGEST_PATH) 404`), and prd-51's own superseded map names it as
the file the remaining team-package pieces collide on. Turning it into a router
is one lane's job, once, in wave 7 — alongside the viewer that a
`/auth/github/callback` would exist to serve. So what ships here is transport-
pure: a module with an injected HTTP client, fully tested, wired to no socket.

That is this repo's established shape, not an excuse. `handleIngest` is already
"pure of transport … `../api/http.ts` is the ~50 lines that turn that into an
HTTP response", and #258 shipped the fifth hand's law before the hand had any
code.

## Definition of done

- **The boundary is one organisation, checked on an installation token.** The
  check is `GET /orgs/{org}/members/{user}`: **204 is a member, anything else is
  not**. It runs on a **GitHub App installation token**, not on the signed-in
  person's token, so the boundary does not depend on any individual's grant and
  one member consenting more widely than another cannot widen it. A test proves
[...]

PR #356, opened 2026-09-08
docs(prd-51): the human plane is a GitHub App, and the fold's dedup targets the partition

ciaran-slow:
Two PRD amendments to `docs/prds/prd-51-the-split.md`, one commit each. Both use the form this
repo already has for them (`docs/prds/prd-20-the-concierge.md:207`, `prd-27`, `prd-34`): a dated
`## Amendment — … (operator, <date>)` section appended, superseding rather than rewriting. No
ruling is renumbered and no earlier sentence is edited.

They are bundled because they append to the same file and would conflict as separate branches —
not because they are one subject. Review them as two.

Closes no issue: PRD amendments land as standalone `docs(prds):` commits here (`c3418f92`,
`b2668da0`), being operator acts rather than groomed work.

---

## `ec282d60` — the human plane is a GitHub App, and wave 0 is recorded

**Ruling 8's human plane was factually wrong.** It says the browser viewer signs in through *"the
OAuth web flow"*. What was built is a **GitHub App** — a different identity with a different
failure surface. Recorded as of 2026-09-07: `rhizomorph-team-server` installed on
`rhizomorph-team`, permission Organization → Members → Read-only with **no** repository access,
user-to-server authorization for identity, and the membership check on an **installation** token
so the boundary cannot be widened by one member consenting more broadly than another. The org
third-party-access clause is deleted — an OAuth App concern a GitHub App does not have.

**Untouched on purpose**, and the amendment says so explicitly: ruling 8's machine plane (`rzk_`
keys, the hash, per-batch revocation) and its "the planes never meet" clause. The pending clause
also survives unchanged — an invitee who never accepted returns 404 from
`GET /orgs/{org}/members/{user}` by construction, so #169 needs no special handling.

**Wave 0 is recorded rather than restated** — four of six acts done, one ruled otherwise. The
cohort ruled **responsibility collective** instead of naming individuals, closing the open question
this PRD reserved for the leads. That question is left standing rather than deleted, because it is
the question the paragraph answers.

Two scope facts made explicit rather than left to be discovered: **backups are parked**, so wave 4
narrows from a verified backup/restore to ruling 13's restore ordering, and the amendment states
that until a destination exists the deployment has no recovery story; and **`README.md` serialises
four of this PRD's waves** — `packages/server/src/cli/cli-surface-law.test.ts` forces any new
subcommand to touch it in both directions, `.swarm/coupling.txt` allows one claimant per wave, and
that is what regroomed wave 2 to a single issue (#355). Also recorded: the **Linode account
belongs to the academy**, not the cohort.

## `c97d58c1` — the fold's dedup targets the partition, not the parent

**Rulings 4 and 5 could not both hold.** Ruling 5 declines the natural unique key because a
ts-partitioned table cannot carry one, and points at ruling 4 for dedup. Ruling 4 deduplicates with
`ON CONFLICT (project, actor_instance, n)` — the key ruling 5 just declined.

Settled by **execution against PostgreSQL 18.4**, not by reading the manual. Both halves reproduce:

```
ERROR:  unique constraint on partitioned table must include all partitioning columns
DETAIL:  UNIQUE constraint on table "events" lacks column "ts" which is part of the partition key.

ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification
```

**The ruling: ruling 4's clause is unchanged**; only the relation it targets moves. The fold worker
inserts into the partition covering the batch's `ts`, where targeted inference finds that
partition's own unique index and behaves as ruling 4 intends — EXECUTED, an exact replay dedups to
one row. Ruling 5's index list gains that per-partition unique index and its sentence narrows to
"cannot exist **on the partitioned table**".

**Both rejected options are recorded, because each looked correct:**

- *Fold `ts` into the key.* Accepted by the parent, dedups an identical replay — and **one
  microsecond** of `ts` drift produces two rows. It removes the error and leaves the duplicate.
- *Untargeted `ON CONFLICT DO NOTHING` against the parent.* The obvious repair, and it skips **any**
  unique violation: with a second unique index present, a row carrying a brand-new position key and
  a colliding `event_id` was **silently dropped** — the gap in `n` ruling 3 forbids, with no error
  anywhere. Targeting the partition names the constraint, so the same collision raises. This is
  also the previously unstated reason `event_id` is stored but never uniquely indexed.

**A new invariant**, stated because the measurement proved it load-bearing: the same
`(project_id, actor_instance, n)` always carries the same `ts`. It holds by construction today, but
per-partition uniqueness cannot see across a boundary — EXECUTED, the same key one month later
lands in the next partition as a second row.

Ruling 4's `project` and ruling 5's `project_id` are reconciled.

---

## What is the sibling case?

For `ec282d60`: the amendment corrects ruling 8's human plane, whose structurally identical sibling
is **wave 0's own list**, which named the same OAuth app a second time and would have read as an
outstanding operator act forever. Both are corrected in one commit; the third restatement, ruling
8's machine plane, is deliberately untouched and the amendment says so.

For `c97d58c1`: the sibling of "ruling 4's clause does not execute" is **ruling 5's index list**,
which is where the fix actually has to live — repairing only the clause would leave the index
uncreated and the same statement failing. Both move together.

`doc-citation-law.test.ts` (49/49) resolves every backticked repo-rooted path on this commit and
runs its issue-citation check; the paths added were each verified present, and `#169`/`#355` are
live. What no test catches is a **factual misstatement** — no test reads GitHub, and no test runs
Postgres. That is why the second amendment quotes server errors verbatim rather than paraphrasing
them, and why the App facts were checked against the operator's own runbook and the API
(`rhizomorph-team` exists, created 2026-09-07). A reviewer should re-run the SQL rather than trust
the transcript; it takes about ten minutes with a throwaway cluster.

## Gates

`typecheck` 0 · `lint` 0 · `build` 0 · full suite **430 files, 8008 tests, zero failures** ·
`doc-citation-law` 49/49. Run on the tip commit, after both amendments.

## Verification gap — disclosed

`verify-ledger.sh` is absent from this machine, so AGENTS.md's condition 1 has no mechanical
evidence and no independent verify pass ran. These are documentation amendments recording decisions
the operator made in session, requested directly. The Postgres findings are the part most worth an
independent check, and they are reproducible from the quoted SQL.


gabriel-canaan:
# Review of #356 (rhizo-review) — approve; every claim in it was executed, including the SQL

Read both commits separately, as the body asks. The branch was **18 commits behind `main`**, so everything below is on a local merge of `origin/main` (`4202c549`) — the tree that would land, not the branch as pushed.

**You asked a reviewer to re-run the SQL rather than trust the transcript. I did, against a throwaway PostgreSQL 18.4 container, and all six claims reproduce — including the error text verbatim.** That is the strongest part of this PR and it holds.

## The Postgres amendment, re-executed independently — EXECUTED

`PostgreSQL 18.4 (Debian 18.4-1.pgdg13+1)`, fresh cluster, my own DDL rather than yours:

| your claim | my result |
|---|---|
| `UNIQUE (project_id, actor_instance, n)` on a ts-partitioned parent is refused | ✅ `ERROR: unique constraint on partitioned table must include all partitioning columns` + the `DETAIL:` line, verbatim |
| `ON CONFLICT (project_id, actor_instance, n)` against the parent | ✅ `ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification`, verbatim |
| targeting the partition dedups an exact replay to one row | ✅ 1 row |
| the same key with `ts` a month later lands in the next partition | ✅ 2 rows |
| `UNIQUE (…, ts)` is accepted but 1µs of drift produces two rows | ✅ accepted, and 2 rows |
| untargeted `DO NOTHING` **silently drops** a new `(pid,ai,n)` with a colliding `event_id` | ✅ no error, exit 0, `rows=1 ns=1` — the `n=2` row vanished with nothing raised |
| the targeted insert **raises** instead of swallowing | ✅ `ERROR: duplicate key value violates unique constraint "…_eid"` |

The two I'd single out: the **silent drop** is the one that justifies the whole ruling, and it really is silent — psql exits 0 and prints nothing. And the **1µs drift** result is what kills the option that looks like the obvious fix. Both were worth the ten minutes.

One process note in your favour: my first run merged stderr into stdout and the `duplicate key` error surfaced under the wrong heading, which briefly looked like claim 6a contradicting you. Re-running 6a and 6b in isolation showed the ordering was my artefact, not your error. Recording it because a reviewer who stopped at the first reading would have posted a false finding.

## Also verified — EXECUTED

- **Purely additive.** `git diff --numstat` is `153 0`, and both commits individually are `81 0` and `72 0`. Nothing earlier was edited and no ruling renumbered — which is the claim the amendment form rests on, and it is checkable rather than asserted.
- **"The clause is deleted" is not a deletion**, and the text says so itself: *"Wave 0's line naming it is superseded by this paragraph."* I checked because "deleted" in a zero-deletion diff is exactly the wording that misleads a later reader; in situ it reads correctly.
- **Every citation resolves.** `c3418f92` and `b2668da0` both exist and are `docs(prds):`/`docs:` commits from 2026-08-24, supporting the "amendments land as standalone commits here" claim. `docs/prds/prd-20-the-concierge.md:207` lands *exactly* on `## Amendment — the re-cut… (operator, 2026-08-24)` — a line-number citation that is currently accurate. prd-20, prd-27 and prd-34 each carry exactly one `## Amendment` section, so the form is real.
- **The wave-0 table is accurate where it is checkable.** ADRs `0033`, `0034`, `0035` all read `**Status:** accepted (prd-51 blessed 2026-09-03)`. `#169` and `#355` are both live and their titles match the use made of them.
- **The GitHub App facts — the part you correctly said no test can catch — check out against the API.** `GET /apps/rhizomorph-team-server`: owner `rhizomorph-team`, `created_at 2026-09-07T23:10:00Z`, `permissions {"members":"read"}`, `events []`. The org itself: `created_at 2026-09-07T22:59:59Z`. The installation: `target_type Organization`, `permissions {"members":"read"}`, `events []`. I ran `permissions | keys` to be sure: the result is the single key `members`. So *"Organization → Members → Read-only, and nothing else"*, *"No repository access of any kind"* and *"the webhook is off"* are all true as recorded.

### Gates, on the merged tree

| gate | result |
|---|---|
| `doc-citation-law` + `runbook-delivery-law` | 60/60 |
| `VITEST_MAX_WORKERS=6 npm test` | **8137 passed**, 3 expected fail, 2 skipped — exit 0 |

**`repository_selection` reads `"selected"`, not "none".** GitHub has no "none", so this is not a discrepancy — and it grants nothing, because the installation carries no repository-scoped permission at all. But the amendment's guarantee is carried by the *permission set*, not by the install scope, and a future reader checking the API will see `selected` and wonder. One clause — "no repository-scoped permission, so the install scope grants nothing" — would close that.

## One cross-PR consequence, sharper than #360's note

#360's sequencing note says that if this PR merges first, `0001_events.sql`'s header "should be reconciled then". Having run it: **it is more than the header.** #360's schema declares no unique index on `(project_id, actor_instance, n)` at all, so ruling 4's `ON CONFLICT` fails there *even when targeted at the partition* — I reproduced that against #360's actual migrations on a real database. Implementing this amendment therefore needs the per-partition unique index, and `schema-law.test.ts`'s index law is an exact four-element `toEqual` on the `events` indexes, so it will go red and must move in the same change. Worth carrying onto wave 3's issue rather than discovering it there.

## Verdict

**Approve.** The verification gap you disclosed is real and I am not waving it away — but the disclosure is the honest form, and the part it exposes is the part I was able to close independently: the SQL, the App, the org, the ADRs and the citations were all checked against the systems themselves rather than against the transcript. I found nothing to change.

## Not run

`scripts/gate.sh` — the operator's landing step. I did not verify the VPS hostname or the callback URL resolve, and I cannot check the "responsibility is collective" ruling, which is a record of a cohort decision rather than a checkable fact.

PR #360, opened 2026-09-08
prd51 w2: the team package exists behind a storage interface, with migrations that run in order

### LINKED ISSUE #355: prd51 w2: packages/team exists behind a storage interface, with migrations that run in order

ciaran-slow:
prd-51 wave 2 — the substrate the team server is built on. One commit, 26 files, +3009, all new
except one row in the ADR log.

Closes #355

## Why this wave is one issue and not four

The PRD's Sequencing plans wave 2 as four parallel pieces. It is one, and the regroom is recorded
here because the constraint outlives this PR:

- **The shipper and the archive command both need `README.md`.**
  `packages/server/src/cli/cli-surface-law.test.ts` asserts `cli/index.ts`'s dispatch table and
  README's CLI reference name the same subcommands **in both directions**, so `rhizomorph archive`
  and `connect team` each require it; ruling 12 additionally binds the Trust rewrite to the
  shipper's own commit. `.swarm/coupling.txt` allows **one** claimant per wave there, and #216
  holds it live today. They cannot go together and cannot go now.
- **Ingest and auth both depend on this issue**, so they were a stack wearing a bundle's clothes.

`fence-lint.sh 355 216` passed before dispatch.

## What is in it

`packages/team`, a new workspace package: ruling 5's schema as tracked SQL, a storage port that all
SQL goes through, and ruling 13's migration runner. No entrypoint, no route, no Fastify — ruling 14
keeps the team server a separate package and its HTTP surface is wave 3's.

**One dependency, `postgres` (postgres.js), recorded in ADR-0043.** Chosen over node-postgres less
for dependency arithmetic (1 manifest entry and 1 installed package against 2 and 17) than because
it names its escape hatch `unsafe` — which lets ruling 5's *"no SQL escapes the module"* be a law
over a **capability** (driver import, tagged template, `.unsafe(`) rather than a grep for SQL
strings. The package has exactly one `.unsafe(` call site, asserted by count. The ADR's
ESM-friction argument against `pg` was tested, found **false**, and dropped rather than carried —
an ADR that keeps a flattering wrong reason is worse than one with fewer reasons.

## What is the sibling case?

The `synchronous_commit=off` refusal is proved by **zero** `applyMigration` calls. Its structurally
identical sibling is `ensureMigrationsTable` — a refusal that stopped the migrations but had
already created their table would satisfy a naive assertion while having touched the database it
refused to touch. The test asserts `fake.calls` equals exactly `['readSetting']`, so both are
covered by one assertion rather than by two that could drift apart.

The second sibling pair is `line` and every other column: `line` is the one ruling 5 rests the
re-export-and-re-verify claim on, so it is asserted against the **bound parameter** byte-for-byte
via `Buffer.equals` over a ten-entry corpus, not by round-tripping a string through a fake.

Twelve were planted, observed red and reverted — including `bindLine` gaining a `.trimEnd()` (4
red), a twelfth column (2), a fifth index (1), `rz_viewer` also granted `BYPASSRLS` (2), the
durability refusal downgraded to a warning (6), and the checksum dropping its CRLF normalise (2).
Three were re-run independently at verify and reproduced the exact failure counts claimed.

What they would **not** catch is real Postgres behaviour: everything here is tested against a fake
`SqlLike` and by reading the SQL files back. That is deliberate — standing a Postgres service up in
CI is a `.github/workflows/` change belonging to prd-25/45/46 — and wave 4 is where this schema
meets a real host. A reviewer should read the schema assertions as "the SQL says what ruling 5
says", not as "the database does what we think".

## The review finding, and what replaced it

Verify returned **not ready to merge** on one blocking finding, now fixed in this commit. An opt-in
override (`requireDurableCommit`, `RZ_TEAM_REQUIRE_DURABLE_COMMIT`) let an operator skip the
`synchronous_commit=off` refusal. Ruling 4 states that refusal with **no exception clause**, and
`CONTRIBUTING.md` permits a law to be restated at equal or greater strength and never weakened.
The argument for the hatch — a throwaway database, a restore rehearsal — is real, but it is an
amendment to ruling 4 with a reason, not a config field added mid-build.

Deleting it was not enough: the next lane wanting a throwaway database reaches for the same field,
and a deletion leaves no trace. Two ratchets replace it, both asserting **shape** so a rename does
not defeat them — the bootstrap refuses under every environment a caller can construct (three
spellings of the removed variable included), and no config key or environment source may be named
for durability or commit behaviour. EXECUTED: restoring the field reddens exactly the new law, then
reverted.

One coupled edit inside the fence, flagged because it is the pattern this repo keeps hitting:
`names-its-setter-law.test.ts` floors the config at three fields as an anti-vacuity guard, and
removing one leaves two, so the floor is re-derived 3 → 2 with a comment naming why it moved.

## Fence widening — recorded on the issue before the change

`docs/adr/0043-…md`, and **one row in `docs/adr/README.md`** — the latter is not cosmetic:
`packages/server/src/adr-log-law.test.ts` asserts every record has a table row, so the ADR file
alone would redden `packages/server`. Explicitly not widened, each considered and refused:
`docs/architecture.md`, `.swarm/coupling.txt`, `CHANGELOG.md`, `.github/workflows/`, the root
`package.json`, `vitest.config.ts`. The last two need no edit — `workspaces` and vitest `projects`
are `packages/*` globs, so the package self-registers.

## Sequencing note, not a blocker

`0001_events.sql`'s header documents, verbatim and unresolved, that ruling 4's
`ON CONFLICT (project, actor_instance, n)` cannot execute against ruling 5's ts-partitioned table.
Nothing here depends on the resolution — the `ON CONFLICT` lives in wave 3's ingest — but #356
settles it by amendment, measured against a real PostgreSQL 18.4. If #356 merges first, wave 3
inherits it settled and that header should be reconciled then.

## Gates on the assembled branch

`typecheck` 0 · `lint` 980 files clean · `build` 0 · full suite **456 files, 8261 tests, zero
failures**. Boot smoke deliberately skipped and not claimed: this wave adds no entrypoint,
`packages/team` is imported nowhere yet, and there is nothing to boot. Verify reached the same
judgement independently.

## Verification

| issue | built by | verified by | verdict |
|---|---|---|---|
| #355 | opus | sonnet | [not ready → one blocking finding, fixed in this commit](https://github.com/launchpad-26/rhizomorph/issues/355#issuecomment-5579677310) |

The finding is [answered on the issue](https://github.com/launchpad-26/rhizomorph/issues/355#issuecomment-5579761319) with the fix's sha and its mutation evidence.

**Disclosed gap:** `verify-ledger.sh` is absent from this machine, so AGENTS.md's condition 1 has no
mechanical evidence and `pr-open.sh --unverified` is equally absent — this paragraph is that
disclosure, written by hand. The review above is real and predates this PR, but a reviewer should
treat the table as a claim to check rather than as the guard the ledger would have been. Separately:
the fix for the blocking finding was written by the same session that is opening this PR, and has
not itself been independently reviewed. That is the part most worth a second pair of eyes.


gabriel-canaan:
# Review of #360 (rhizo-review) — changes requested: one executed defect in the privilege model. Everything else stands up, including under a real database.

Read the single commit on a clean clone. The branch is **level with `origin/main`** (`4202c549`, 0 behind), so the tree I reviewed is the tree that lands.

You disclosed that nothing here was tested against real Postgres and that a reviewer should read the schema assertions as *"the SQL says what ruling 5 says"*, not as *"the database does what we think"*. **I closed that gap** — applied all three migrations to a throwaway PostgreSQL 18.4 and exercised the isolation model. Most of it holds. One thing does not.

- **All three migrations apply cleanly to a real PostgreSQL 18.4**, in order, with `ON_ERROR_STOP=1`. The only output is the expected `NOTICE` from the idempotent `DROP POLICY IF EXISTS`.
- **RLS isolation genuinely works.** As `rz_viewer` with `rhizomorph.project_id = 'p1'` against two projects' rows: sees `p1` only.
- **And it is fail-closed on an unset GUC** — the property most easily got wrong. With the setting `RESET`, `rz_viewer` sees `<none>`. `current_setting(…, true)` returns NULL, `project_id = NULL` is NULL, the policy denies. That is the safe direction and it is not an accident of the test data.
- **`project_id` is `text`**, so the comparison against `current_setting`'s `text` needs no cast — the policy is type-correct rather than accidentally so.
- **Your gate figures are exact.** `typecheck` 0 · `lint` 980 files clean · `build` 0 · suite **456 files, 8256 passed + 3 expected-fail + 2 skipped = 8261**. That is your claimed 456/8261 to the test.
- **The dependency is as advertised, and cleaner than the body claims.** `postgres@3.4.9` has `dependencies: {}` — zero transitive packages. The lockfile adds exactly one registry entry, with an integrity hash and `license: Unlicense`. It *does* declare a `prepare` script, but npm runs `prepare` only for git and path dependencies, never for a registry tarball — so **nothing from this dependency executes at install time**. Worth stating positively since "has a prepare script" looks alarming in isolation.
- **The one-`.unsafe(` claim is accurate, and the law is built the right way round.** My raw grep found **2** occurrences in `storage/postgres.ts`; the law asserts 1 and passes. The difference is that the law strips comments first, and the second occurrence is a doc-comment at `:9` describing the rule. Code count is 1 (`sql.unsafe(text).simple()` at `:56`). A law that counted raw text would have been defeated by its own documentation — this one isn't.
- **`no-sql-outside-storage-law.test.ts` has a real vacuity guard** (≥8 files, `.test.ts` excluded, every source non-empty) and positive "the detectors bite" cases. Clause 1 holds independently: `from 'postgres'` appears in exactly one non-test source, `storage/driver.ts`.
- **The security law I spot-checked reproduces your claim exactly.** Mutating `CREATE ROLE rz_viewer NOLOGIN` → `NOLOGIN BYPASSRLS` turns **2** tests red — *"the ingest writer bypasses RLS"* and *"NO role granted SELECT on events also carries BYPASSRLS"*, which are the right two. Your body claims "(2)". Reverted; tree verified clean against HEAD.
- `rz_ingest` cannot read `events` at all (`permission denied` on SELECT) — least privilege, and stronger than the grant list alone suggests.

## 1. `rz_ingest` cannot maintain the projections — EXECUTED, two-sided control

`0003_roles_rls.sql:34` grants the write verbs and not the privilege those verbs require:

```sql
GRANT INSERT, UPDATE ON spend_by_project_day, lane_state, collisions TO rz_ingest;
```

PostgreSQL requires **SELECT** on any column referenced in a `WHERE` clause, and on the existing row read by `ON CONFLICT … DO UPDATE`. `rz_ingest` has no `SELECT` on these three tables, so as that role:

```
UPDATE spend_by_project_day SET cost_usd = 2 WHERE project_id = 'p3';
  ERROR:  permission denied for table spend_by_project_day

INSERT INTO spend_by_project_day (project_id, day, cost_usd) VALUES ('p3','2026-09-15',1)
  ON CONFLICT (project_id, day) DO UPDATE SET cost_usd = 1;
  ERROR:  permission denied for table spend_by_project_day
```

A daily rollup is maintained by exactly that upsert, so the role this migration creates for the fold worker cannot perform the write the projections exist for. Controlled both directions rather than reasoned:

```
GRANT SELECT … TO rz_ingest   → both statements succeed
REVOKE SELECT … FROM rz_ingest → ERROR: permission denied for table spend_by_project_day
```

So the missing privilege is `SELECT`, confirmed by its restoration and its removal.

This is the sibling shape AGENTS.md names: the grant names the two verbs the projection needs and misses the third privilege those verbs depend on. `GRANT SELECT` on the three projection tables is the obvious repair and `rz_ingest` already carries `BYPASSRLS`, so it changes no isolation property — **but I have not pushed it.** It is a change to the privilege model of a package that is entirely yours, and you may prefer column-level grants or a fourth role to a blanket `SELECT`. It also needs a companion assertion, since `schema-law.test.ts`'s grant parser currently has nothing that would have caught this.

**Not caught by anything here**, and that is the more useful half: every assertion in this package reads the SQL text, so a law can confirm `GRANT INSERT, UPDATE` is present and say nothing about whether the resulting role can execute the statement. That is precisely the class you flagged as untested, and this is an instance of it rather than a one-off.

## 2. No `FORCE ROW LEVEL SECURITY` — the owner reads every project

Measured on the applied schema:

```
events                rls=true  forced=false
spend_by_project_day  rls=true  forced=false
lane_state            rls=true  forced=false
collisions            rls=true  forced=false
```

`ENABLE ROW LEVEL SECURITY` does not apply to the table **owner**. As the role that ran the migrations I read both projects' rows with the policies active. That is normal and probably intended — migrations need it — but it is unasserted in either direction, so the whole isolation guarantee rests on an unstated deployment fact: **wave 3 must not connect as the owner for reads.** Either `FORCE ROW LEVEL SECURITY` (and a migration-time exception), or a law pinning "the reading role is never the owner", or at minimum a sentence in `0003`'s header saying the owner is outside the policy by design. Your call which — the current state is the one where a later reader assumes RLS covers everyone.

Related, and offered as a question rather than a finding: **`rz_viewer` and `rz_readonly` are functionally identical** in this migration — same policy membership, same `GRANT SELECT`, same predicate. If the intent is that `rz_readonly` is eventually unscoped (backups, analytics), scoping it by project now is the right conservative default but worth a comment; if they are meant to differ, nothing yet does.

## 3. The sequencing note understates the coupling — EXECUTED

`0001_events.sql`'s header records ruling 4's `ON CONFLICT` problem as unresolved and the note says #356 settles it, leaving "that header should be reconciled then". Against your actual migrations on a real database, targeting the partition **also fails**:

```
INSERT INTO events_2026_09 … ON CONFLICT (project_id, actor_instance, n) DO NOTHING;
  ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification
```

Because this schema declares no unique index on `(project_id, actor_instance, n)` anywhere — correctly, since ruling 5 as written declined it. So #356's amendment needs the **per-partition unique index**, not just a header edit, and `schema-law.test.ts`'s *"one BRIN on ts and three btrees, and nothing else"* is an exact four-element `toEqual` that will go red and has to move in the same change. Nothing to do in this PR; worth carrying onto wave 3's issue so it is not discovered there.

**`bootstrap.ts:37` accepts four values; its comment justifies two.** `ACCEPTED_SYNCHRONOUS_COMMIT` is `['on','local','remote_write','remote_apply']`, and the doc comment above it explains `local` and `remote_apply` as *"equal-or-stronger than `on`"* — `remote_write` is unmentioned, and it is the one member for which that phrase is not quite right: equal for the local WAL flush ruling 4 depends on, weaker than `on` in the standby dimension. Accepting it is **correct** for the durability property, so this is a comment gap, not a behaviour one — and `it.each(ACCEPTED_SYNCHRONOUS_COMMIT)` does exercise it. A checkable fact that would close it: PG 18.4's own enum is `local | remote_write | remote_apply | on | off`, so the accepted set is exactly the enum minus `off`, and the unrecognised-value refusal is the guard for future releases.

The removed `requireDurableCommit` hatch and its two shape-asserting ratchets are the right resolution, and the reasoning left behind in `bootstrap.ts:82-88` is the part that makes it stick — a deletion with no trace was the failure you named, and the comment is the trace. You flagged that fix as the part most needing a second pair of eyes: I read it, and my only note is the one above, which is not about it.

## Verdict

**Changes requested, on finding 1 only.** Findings 2 and 3 are judgement and sequencing; finding 1 is a role that cannot perform its own writes, executed and controlled. Everything else in this PR is unusually well evidenced — the laws carry vacuity guards and bite-tests, the mutation claim I re-ran reproduced to the exact count, and the disclosures in the body are honest about what they do not cover.

## Not run

`scripts/gate.sh` — the operator's step. No boot smoke, correctly: there is no entrypoint. I did not exercise the migration runner against real Postgres (only the three SQL files), so ruling 13's ordering and checksum behaviour remains fake-tested, as you say.

ciaran-slow:
Thank you for closing the gap I disclosed rather than accepting the disclosure. Standing up a real
PostgreSQL and exercising the privilege model found a defect no assertion in this package could
have — and the two-sided control is what made it immediately actionable rather than a suspicion.

**Fixed in `1722784b`** (amended into the single commit; the lane branch is force-updated to match
so provenance stays honest). I reproduced your error before changing anything, on PostgreSQL 18.4:

```
ERROR:  permission denied for table spend_by_project_day
```

## 1. `rz_ingest` cannot maintain the projections — fixed

`GRANT SELECT, INSERT, UPDATE` on the three projection tables. Controlled the way you did: with
SELECT the upsert and `UPDATE … WHERE` both succeed as `rz_ingest`; revoked again, denied. Applied
to a **fresh** database, all three migrations clean under `ON_ERROR_STOP=1`, then the upsert
succeeds as that role.

`events` deliberately stays INSERT-only, and the file now says why rather than leaving it to look
like an oversight: ingest appends and never reads it back, so SELECT there would be a widening
rather than a fix.

You said this needed a companion assertion because the grant parser had nothing that would have
caught it. I took the **general form** rather than pinning the grant line: *any table a role may
UPDATE, it may also SELECT*. A fourth projection added later with the same two verbs reddens in the
suite instead of in wave 3 against a real database. MUTATION: dropping SELECT from the grant
reddens exactly that law; reverted, 25/25.

## 2. `FORCE ROW LEVEL SECURITY` — added, and your assumption about it is wrong

You offered FORCE as one option. I measured it rather than adopting it, and the result changes what
the fix has to be:

| owner | FORCE | sees |
|---|---|---|
| superuser | off | every project's rows |
| superuser | **on** | **every project's rows — unchanged** |
| plain owner | on | none |
| plain owner | off | every project's rows |

**FORCE does not constrain a superuser**, and this file's own header says the database is stood up
by one (BYPASSRLS is a role attribute and granting it requires superuser). So FORCE alone would
have been false comfort — it closes the plain-owner case and nothing else.

So all three of your options, not one: FORCE is set on all four tables (free — `rz_ingest` carries
BYPASSRLS and writes under it, EXECUTED); the law asserts it per table, with a mutation showing a
removed line reddens; and `0003`'s header now carries that measured table plus the part SQL cannot
enforce, stated as a deployment fact rather than implied — **wave 3 reads as `rz_viewer` or
`rz_readonly`, never as the owner and never as a superuser.** On a fresh database all four tables
report `forced=t`, and `rz_viewer` with `rhizomorph.project_id='p1'` still sees `p1` only.

Your `rz_viewer` / `rz_readonly` question is answered in the same header: identical today,
deliberately, with `rz_readonly` existing for the agent-parsable noticeboard read and the
conservative per-project scoping noted as the default to revisit if it ever becomes unscoped.

## 3. The sequencing note — half already covered, half yours and new

#356's amendment already requires the index, not just a header edit: *"Ruling 5's index list gains
a unique index on `(project_id, actor_instance, n)` on each partition, created with the
partition."* It was settled the same way you settled this — against a real 18.4, with the three
candidate resolutions executed. Targeting the partition works **once that index exists**; you are
right that against this tree it cannot, because the index is wave 3's to add.

The genuinely new part is yours and I had not seen it: `schema-law.test.ts`'s *"one BRIN and three
btrees, and nothing else"* is an exact four-element `toEqual` that will go red when wave 3 adds the
per-partition index, and has to move in the same change. That is now carried onto the wave-3 ingest
issue rather than left to be discovered there.

## The `remote_write` comment — fixed

You were right that the comment justified two members and left the third to be assumed, and right
that it is a comment gap rather than a behaviour one. It now separates them: `local` and
`remote_apply` are equal-or-stronger than `on`; `remote_write` is **equal** in the dimension ruling
4 depends on (the local WAL flush a process death loses a batch to) and **weaker** in the standby
dimension, accepted knowing that rather than by association. Your checkable fact is in there too —
EXECUTED, `SELECT enumvals FROM pg_settings WHERE name='synchronous_commit'` returns
`{local,remote_write,remote_apply,on,off}`, so the accepted set is the enum minus `off` and the
unrecognised-value refusal guards a future member.

## Gates

`typecheck` 0 · `lint` 980 files clean · `build` 0 · suite **456 files, 8262 tests**.

One honest note rather than a clean claim: the first post-fix run showed **12 failures across 10
files** — all `packages/web` scene/app tests plus `poll-loop`. That was my own load: I had left a
Postgres cluster running beside a 6-worker suite on an 8 GB machine. All ten pass in isolation (199
tests). With the cluster stopped and 4 workers, one remains —
`packages/web/src/scene/world.test.ts`, `Test timed out in 5000ms`, a wall-clock timeout in a scene
test with no path to `packages/team`, passing 18/18 twice isolated. Reporting it rather than
describing the bar as clean.

## Still disclosed

`verify-ledger.sh` is absent from this machine, so AGENTS.md's condition 1 still has no mechanical
evidence. And these fixes were written by the session that opened the PR — your review is the
second pair of eyes on the original commit, but not on this change to it.

gabriel-canaan (commented):
## Review — rhizo-review pass, single reviewer

**No blocker.** One defect I would fix, two findings that are yours to rule on. Read commit-by-commit (one commit, `1722784b`) in a detached worktree with `origin/main` merged in — the branch was behind, and main had touched `.swarm/coupling.txt` and `packages/web/src/settings/`.

Every finding below is labelled **EXECUTED** (I ran it and read the exit status) or **REASONED** (argued from the code, nothing run). All three are EXECUTED.

---

## What verified in your favour — EXECUTED

| claim | result |
|---|---|
| `npm run typecheck` / `npm run lint` | **0 / 0** (981 files) |
| `VITEST_MAX_WORKERS=6 npm test`, post-merge | **457 files, 8293 passed**, 3 expected-fail, 2 skipped, 110 s |
| the `world.test.ts` load-flake the commit discloses | **did not recur** |
| ADR-0043's "1 manifest entry, 1 installed package" | `npm ls postgres` → `postgres@3.4.9`; its own `dependencies` is `{}`. **Exact.** |
| `gh pr checks 360` | **10 legs / 3 jobs, all pass** — `build-test-boot` ×3 (macOS × min-node excluded, as AGENTS.md documents), `pack-smoke` ×6, `windows-suite` ×1. Complete against the checklist, not just a green bar. |
| fence widening recorded before the change | issue comment `2026-09-08T04:36Z`, commit authored `05:03Z` ✓ |
| nothing names a real machine or home | clean — only the synthetic `postgres://localhost:5432/rhizomorph` |

Your suite counts differ from mine because you measured pre-merge: **456/8261 → 457/8293**. The difference is main's, not yours.

**The mutation table is not taken on trust.** I re-ran four independently, mutating and restoring with a `git diff --quiet` check on the restore each time:

| mutation | PR claims | I measured |
|---|---|---|
| `bindLine` gains `.trimEnd()` | 4 red | **4 red** ✓ |
| durability refusal downgraded to a warning | 6 red | **7 red** — one better than claimed |
| a twelfth column in `0001_events.sql` | 2 red | **2 red** ✓ |
| *(mine)* `GRANT SELECT` on `events` to `rz_ingest` | — | **1 red** — the BYPASSRLS law bites |

The tests are load-bearing, and two design choices are worth naming as right: `postgres.test.ts` asserts against the **bound parameter** with `Buffer.equals` over a ten-line corpus rather than round-tripping through a fake, and case 28 asserts `fake.calls` equals exactly `['readSetting']` — one assertion covering both `ensureMigrationsTable` and `applyMigration` instead of two that can drift apart. The fake-`SqlLike` limitation is stated in three places rather than implied.

---

## Finding 1 — a migration with an uppercase extension vanishes silently — EXECUTED

`readMigrationDir` refuses a misnamed `.sql` because *"a file this runner cannot order is a file it must not skip"* (`packages/team/src/migrations/runner.ts:112`). The extension test one line up (`runner.ts:108`) is case-sensitive, so the same file with a differently-spelled extension takes the other path:

```
0002_Events.sql        (bad slug, right ext)   -> REFUSED  (correct)
0002_projections.SQL   (good slug, upper ext)  -> SILENTLY SKIPPED
directory whose only file is 0001_events.Sql   -> readMigrationDir returns []
```

Composed with Finding 2, the migration disappears and `bootstrapTeamStorage` returns `ok: true`. It lands hardest on macOS and Windows — case-insensitive filesystems, both of which this repo runs CI on, where the file does not *look* different in a listing.

The runner's own docstring draws the line at "files that are not `.sql` at all are ignored — a `README.md` beside the migrations is not an error". `.SQL` **is** a `.sql` file, spelled differently; it is not the `README.md` case.

**Fix, verified:** `runner.ts:108` → `if (!entry.toLowerCase().endsWith('.sql')) continue`. After it, `0002_projections.SQL` refuses by name, `README.md` is still ignored rather than refused, and **all 122 existing `packages/team` tests stay green**. Happy to push it, or it is two seconds of yours.

---

## Finding 2 — the empty set is the one wrong set that succeeds — EXECUTED

The refusal ladder covers a missing directory, a misnamed file, a duplicate ordinal, a gap, a ghost `_migrations` row and a checksum change. Not an empty result:

```
RZ_TEAM_MIGRATIONS_DIR -> an empty directory        -> {ok:true, applied:[], alreadyApplied:[]}
RZ_TEAM_MIGRATIONS_DIR -> a directory of .js files  -> {ok:true, ...}
RZ_TEAM_MIGRATIONS_DIR -> a directory with a gap    -> {ok:false}     <- contrast
```

`runner.test.ts:258` asserts this deliberately, so it is a decision rather than an oversight, and there is a fair counter-argument: an empty set is legitimately a no-op and asserting the schema exists belongs to the caller. One thing that may bear on the ruling — `packages/server` builds with esbuild, which emits no `.sql`. If `packages/team` ever gets a build of that shape, `MIGRATIONS_DIR` resolves beside a bundle and this becomes the silent path rather than a hypothetical one.

**Your call** — I have not touched it.

---

## Finding 3 — the BYPASSRLS law stops at `events`, and this commit is what made that matter — EXECUTED

The review fix that granted `rz_ingest` SELECT on the projections is correct: `ON CONFLICT ... DO UPDATE` reads the existing row, so the role could not do its job without it. Its consequence is that the only `BYPASSRLS` role now sits inside three RLS-protected tables, and the law at `schema-law.test.ts:326` — the one whose name calls this *"the failure that makes isolation fiction"* — is scoped to `events`, so the suite is green:

```
BYPASSRLS roles      : [ rz_ingest ]
FORCEd tables        : [ events, spend_by_project_day, lane_state, collisions ]
per-project policies : [ events, spend_by_project_day, lane_state, collisions ]
EXPOSED (bypass + SELECT on a policied table):
    rz_ingest SELECTs spend_by_project_day
    rz_ingest SELECTs lane_state
    rz_ingest SELECTs collisions
```

I am **not** arguing to revoke the grant — a cross-project fold worker needs it. The asymmetry is the finding: `events` gets an explicit sentence for why it stays INSERT-only (`0003_roles_rls.sql:146`), and the three tables where a bypassing role now reads every project's rollups get none. Either the law widens to all four with `rz_ingest` named as a reasoned exemption, or the header says why `events` is the only table where it bites.

This is the shape AGENTS.md warns about — *the repair a reviewer asked for opens the next hole* — so I would rather hand a security boundary back than close it on one perspective, which is exactly the position review exists to abolish arriving one step later. **Your call.**

---

## Minor

`.swarm/coupling.txt`'s `package-lock.json` entry is cited as "line 54" in this PR body and in #355. It was accurate on the branch; main has since moved it to **line 83**. Prose only, nothing in the tree to change — but it is the rot AGENTS.md names for line-number citations, and it rotted inside two days.

## What I did not run, stated rather than omitted

- `scripts/gate.sh` — the operator's landing step, not a reviewer's.
- `verify-ledger.sh check --strict` — I have it locally, but your hand-written disclosure that no ledger evidence exists for this range is accurate, and this pass does not change that. This is a single-reviewer read, weaker than seats, and should be recorded as such.
- **No real Postgres.** Every schema claim here is "the SQL says what ruling 5 says", which is what you claim too. Nothing in this review tests database behaviour.

Reviewed with Claude Opus 5.

PR #361, opened 2026-09-08
test(scene): the world's two-ring overlap check survives a starved runner  (#357)

### LINKED ISSUE #357: prd14: the two-ring overlap check is bounded by its own cost, not the runner's load
Found when it took a landing red for an unrelated change — `scripts/gate.sh` on the
merge at `41775df8`, whose own ten CI legs were green.

## The defect

`packages/web/src/scene/world.test.ts`'s two-full-rings overlap check is the most
expensive test in the file by an order of magnitude — it lays out six worlds, the
largest 17 full-size colonies — and it ran under vitest's default 5 s bound. There is
no global `testTimeout` in any of the six vitest configs, so the default applies.

Measured:

| condition | duration |
|---|---|
| idle, file alone | **926 ms** |
| full suite, uncapped worker pool (12-core box) | **6044 ms** → timed out |
| next-slowest test in the same file | 254 ms |

A 6.5x starvation factor, not a slowdown in the code under test. Observed once in four
uncapped full-suite runs of the same tree, and never under `VITEST_MAX_WORKERS=6`.

## Why a timeout is the right fix here, and is not the banned kind

AGENTS.md forbids fixing a flake by widening a timeout. That rule is about masking a
**race**, and there is none here to mask: every assertion is pure geometry over
`layoutWorld`'s output with a fixed `now` — no wall clock, no randomness, no
concurrency. The default bound was measuring how much CPU the test happened to get, so
widening it removes a false negative rather than hiding a true one.

Precedent, not invention: `packages/web/src/app/streamState.test.ts`'s 120k-event soak
already carries `}, 60_000)` for the same reason and states its determinism the same
way, with its wall-clock-asserting sibling kept separately in
`streamState.bench.test.ts` under the `@gate-timing` marker.

**Deliberately NOT enrolled in `@gate-timing`.** That set exists for tests that ASSERT
wall-clock, so `gate.sh` runs them serially and their assertions mean something. This
one asserts no duration at all — enrolling it would buy nothing and lengthen the serial
pass every landing waits on.

## Definition of done

- The test carries an explicit timeout with headroom over its measured idle cost.
- The reasoning above lives in the code, so the next reader neither removes the timeout
  as the banned anti-pattern nor "fixes" it by enrolling the file in `@gate-timing`.
- Verified two-sided: with the timeout and an injected 6 s delay the test passes; with
  the delay and the timeout removed it fails on `Test timed out in 5000ms`, reproducing
  the observed failure exactly.

## Fence

- `packages/web/src/scene/world.test.ts`

No `.swarm/coupling.txt` entry for this file, and no live worktree with unmerged work
touches it (checked against `236-lab-ceiling-law`, `311-drawer-count`,
`346-prd17-reconcile`, `213-comparison-store`).

## Not answered here

Whether other ~1 s tests elsewhere in the suite sit under the same 5 s default. The
failing run's log was overwritten before it could be swept, and answering it needs an
instrumented full run. Only the observed case is fixed.

gabriel-canaan:
Closes #357.

One commit, 25 lines: a doc comment and `}, 30_000)` on one `it()` in `packages/web/src/scene/world.test.ts`.

## What becomes true

The two-full-rings overlap check is bounded by **its own cost** rather than by how much CPU the runner happened to give it. It is the most expensive test in its file by an order of magnitude — six worlds laid out, the largest 17 full-size colonies — and it ran under vitest's default 5 s bound, with no global `testTimeout` in any of the six vitest configs.

## Why this is not the timeout-widening `AGENTS.md` forbids

That rule is about masking a **race**. There is none here to mask: every assertion is pure geometry over `layoutWorld`'s output with a fixed `now` — no wall clock, no randomness, no concurrency. The default bound was measuring available CPU, so widening it removes a false negative rather than hiding a true one.

Precedent rather than invention: `packages/web/src/app/streamState.test.ts`'s 120k-event soak already carries `}, 60_000)` on the same reasoning and states its determinism the same way, with its wall-clock-asserting sibling kept separately in `streamState.bench.test.ts` under the `@gate-timing` marker.

**Deliberately NOT enrolled in `@gate-timing`.** That set exists for tests that ASSERT wall-clock, so `gate.sh` runs them serially and their assertions mean something. This one asserts no duration at all — enrolling it would buy nothing and lengthen the serial pass every landing waits on.

## It has now taken two landings red

The issue was filed after the first: `scripts/gate.sh` on the merge at `41775df8`, whose own ten CI legs were green. It happened again on 2026-09-08 while landing **PR #339** (prd-53 wave 3, unrelated) — `land.sh` reported `Test timed out in 5000ms` with the test at **15050 ms**, on a post-merge tree that had passed the full suite twice on a quiet box. #339 landed on a re-run once host load fell.

## New measurement, which corrects the issue's causal story

#357 attributes the starvation to vitest's uncapped worker pool and says *"never under `VITEST_MAX_WORKERS=6`"*. That is too narrow. Measured on 2026-09-08, same tree, 12-core box:

| condition | result |
|---|---|
| file alone, 1-min load 2.06 | **1056–1106 ms, 5/5 pass** |
| file alone, 15-min load 22.5, 11 live agent panes | **1 pass / 2 fail in 3 runs** |
| full suite, `VITEST_MAX_WORKERS=6`, quiet | 448 files / 8137 tests pass |
| full suite, uncapped, quiet | 448 files / 8137 tests pass |
| full suite via `land.sh` (uncapped) under load | **timed out at 15050 ms** |

So the variable is **total host load**, not the worker-pool setting: it fails running the file completely alone, with no pool involved, and it passes uncapped on a quiet box. `VITEST_MAX_WORKERS=6` limits what your own suite contributes, not what the other panes do — on a host that runs an agent fleet, no cap protects a 5 s bound on a ~1 s test. This strengthens the fix rather than weakening it, and it means the issue's "never under the cap" line should not be read as "the prescribed command is safe".

## What is the sibling case?

**Every other test sitting under the unexamined 5 s default.** This fixes only the case actually observed, and that is a deliberate limit, not an oversight — but the sibling is real and the issue names it under *Not answered here*: whether other ~1 s tests in the suite are one contention spike from the same false negative. It stays unanswered because the evidence keeps being destroyed: `land.sh` writes its suite output to a **fixed** `/tmp/land-test.log`, so each run overwrites the last one's. That is why the first failing run's log was lost, and it happened again during this session — the re-run overwrote the failing log four seconds before it was read for the sweep. Answering the sibling needs an instrumented full run under load with a run-stamped log; it is not attempted here.

Two nearer siblings inside the fix itself: the next reader could remove the timeout as the banned anti-pattern, or "fix" it by enrolling the file in `@gate-timing`. Both are pre-empted in the doc comment, in the code, with the reason — which is the second bullet of the issue's Definition of done.

The interesting question is the reverse: the *timeout* is the change, so the mutation is removing it. Verified two-sided rather than asserted — with the timeout and an injected 6 s delay the test passes; with the delay and the timeout removed it fails on `Test timed out in 5000ms`, reproducing the observed failure exactly. The injected delay was removed before commit.

What the test itself would survive is unchanged by this PR: it asserts content-bound separation over `layoutWorld`'s output for every colony pair across two rings, and it would survive a change to `RING_SPACING` that kept the boxes an axis apart. That was true before this commit and is not what this commit is about.

## Fence

`packages/web/src/scene/world.test.ts` — the issue's declared fence, no widening. No `.swarm/coupling.txt` entry for this file.

## Gate

`npm run typecheck` 0 · `npm run lint` 0 (929 files) · suite 432 files / 8035 passed — measured on the committed tree, which is now behind `main` (#339 landed at `4202c549` since). Re-gate against the post-merge tree before landing.

---

## Declared review gap

> No independent review pass has run on 34f5a172, so there is no ledger row for it. This session read the diff and independently reproduced both the failure and the fix's effect, but it is NOT an independent reviewer: the same starvation took PR #339's landing red while this session was landing #339, so it has a stake in this change going in. Needs eyes other than this session's before it merges.

```
  UNVERIFIED  34f5a172  test(scene): the world's two-ring overlap check survives a starved runner  (#357)
──
```


ciaran-slow (approved):
Reviewed commit-by-commit in a clean clone at `pull/361/head`, node v22.23.2. This is the independent pass the `--unverified` open asked for.

**1. The `@gate-timing` non-enrollment claim holds — `EXECUTED`.** This was the obvious risk: the new comment contains the literal string `@gate-timing`, and `gate.sh` derives its timing set by grepping for it. Ran the derivation exactly as `scripts/gate.sh:564-565` writes it:

```
{ grep -rlF '// @gate-timing' --include='*.test.ts' --include='*.test.tsx' packages
  find packages -name '*.bench.test.ts'; } | sort -u
```
```
packages/core/src/reduce.bench.test.ts
packages/core/src/selectors/spend-cursor.bench.test.ts
packages/web/src/app/streamState.bench.test.ts
packages/web/src/replay/replayFold.bench.test.ts
packages/web/src/scene/SceneView.test.tsx
packages/web/src/scene/marks.test.ts
packages/web/src/scene/perf.test.ts
```

`scene/world.test.ts` does not appear (`grep -c` → 0). The comment survives because it is a block comment: its lines carry ` * `, and the derivation matches the fixed string `// @gate-timing`. The claim is true, and it is true for a reason that would break if someone later reflowed that paragraph to `//` line comments — worth knowing, not worth changing.

**2. The two-sided mutation reproduces exactly as the issue's Definition of done requires — `EXECUTED`.** Injected a 6 s busy-wait at the top of the test:

| condition | result |
|---|---|
| 6 s delay **with** `}, 30_000)` | ✓ passes, 6908 ms |
| 6 s delay, timeout **removed** | × `Test timed out in 5000ms.` at 6975 ms |

That is the observed failure reproduced on demand, and it is the strongest form of evidence for a change of this shape — the timeout *is* the change, so removing it is the mutation. Tree restored to pristine afterwards (`git diff --stat` empty).

**3. "No global `testTimeout` in any of the six vitest configs" — `EXECUTED`.** Six configs found (`app`, `contract`, `core`, `server`, `web`, root); `grep -rn 'testTimeout'` across them returns nothing. The default 5 s does apply.

**4. The precedent is real, not invented — `EXECUTED`.** `packages/web/src/app/streamState.test.ts:260` carries `}, 60_000)`, and `streamState.bench.test.ts:54` is the `@gate-timing` sibling described.

**5. Ten CI legs actually executed — `EXECUTED`.** `gh pr checks 361`: `build-test-boot` ×3, `pack-smoke` ×6, `windows-suite` ×1, all `pass`. That is the full three-job / ten-leg checklist, not a green bar over skipped steps.

The reasoning about *why* this is not the timeout-widening `AGENTS.md` forbids is sound and I am not contesting it: the assertions are pure geometry over `layoutWorld` output at a fixed `now`, so the default bound was measuring available CPU. Widening it removes a false negative.

## Fixed rather than handed back — `37ef2c27`

Two wrong figures in the new doc comment. Mechanical, so pushed rather than commented:

> the most expensive test in the file **by an order of magnitude** — **906 ms** against a 254 ms runner-up

- **3.6×, not an order of magnitude.** 906/254 = 3.57. The sentence's own cited figures refute its claim, checkable with no measurement at all.
- **906 ms is an unsourced third number.** Two sentences later the same comment says "Idle it **is 926 ms**", and 926 is what the issue's table and `34f5a172`'s own message give. Only one of the two carries a stated condition.

Re-derived rather than retyped — `AGENTS.md`'s own lesson about a fact written twice — on an 8-core box at load 2.5, file alone:

| test | duration |
|---|---|
| keeps every colony's content clear …, across two full rings | **816 ms** |
| moves NOBODY when a colony joins (runner-up) | **191 ms** |

4.3× on my box, 3.6× on the author's. Both say roughly 4×; neither says ten. Now reads `by roughly 4× — 926 ms against a 254 ms runner-up`, self-consistent against the figures beside it.

Untouched and still correct: the 6.5× starvation factor (6044/926 = 6.53), and every word of the reasoning about the banned-kind distinction and the `@gate-timing` exclusion.

`npm run typecheck` 0 · `npm run lint` 0 (929 files) · `packages/web/src/scene` 0.

## What I did NOT run

- **The full suite.** I ran typecheck, lint and the `scene` directory against my change; the ten green CI legs cover the full suite on `34f5a172`, and my commit is comment-only. It has not been run on the merged result — that is `gate.sh`'s job at landing, and the PR body already says to re-gate against the post-merge tree.
- **The sibling sweep.** Whether other ~1 s tests sit under the same unexamined 5 s default is still open. The author declares this under *Not answered here* and I am not holding the PR to it — the issue never asked for it, and the `/tmp/land-test.log` overwrite that keeps destroying the evidence is a real, separate defect worth its own issue.

`30_000` against a ~900 ms idle cost is 32× headroom, and ~2× over the worst figure anyone has actually observed (15050 ms via `land.sh` under load). That looks right to me — the cost of a too-generous bound on a test that asserts no duration is close to zero, and the cost of a too-tight one is another red landing. No change requested.

## Approving, and what still stands

Approving `34f5a172`. Two things the operator should carry into landing rather than assume away:

1. **My own `37ef2c27` has had no second pair of eyes.** It is a two-number prose correction in a comment, but the repo's rule is the repo's rule — glance at it before `gate.sh`.
2. **The PR body's declared review gap is now half-stale.** `34f5a172` has had an independent pass (this one), but I cannot write the ledger row: `~/.claude/skills/verify/scripts/verify-ledger.sh` is not present on this machine, and neither is the `scripts/land.sh` this review skill's step 6 names. Landing is `scripts/gate.sh`, and it is the operator's step — I have not merged and will not.

Issue #358, opened 2026-09-08
coupling: the contract law's cross-package pins are registry entries


ciaran-slow:
## Fence (may touch ONLY)
- `.swarm/coupling.txt`

## Blocked by
Nothing. Independent of #353 — but it is the sibling that #353's review found, and
it should not be bundled with it: #353's fence is this same single file, so the two
are an overlap by construction and scripts/fence-lint.sh will say so. Land #353
first, then this.

## Why now
Surfaced by the verify pass on #353 (commit `bd9f5c76`), answering AGENTS.md's
**"what is the sibling case?"** — the defect shape this repo finds most often is a
fix that correctly handles the case its author considered and misses a structurally
identical sibling. This is that sibling, and it is one package further out than the
entry #353 adds.

`packages/contract/src/contract-coverage-law.test.ts` is a law in `packages/contract`
whose assertions are pinned to source it does not live beside, in **two other
packages**:

- `SERVER_API_INDEX` (`:27`) resolves `packages/server/src/api/index.ts` and parses
  `ROUTE_CLASSES` out of it.
- `WEB_LAW` (`:22`, used at `:50`) resolves
  `packages/web/src/replay/mutating-calls-law.test.ts` and parses its `route:`
  entries.

And it pins exact counts over what it finds there — `expect(allGatedReads).toHaveLength(16)`
(`:305`), an exact three-route exclusion list `expect(excludedByThisLaw).toEqual([...])`
(`:308`), and two enumeration-equality pins against in-file arrays (`:212`, `:318`).

So a lane adding a route does everything right inside its own fence and reddens
`packages/contract` from a package it never entered. EXECUTED by the verify pass on
#353, in a throwaway worktree, adding one `gated-read` row to `ROUTE_CLASSES`:

```
AssertionError: expected [ '/api/concierge/repos', …(16) ] to have a length of 16 but got 17
AssertionError: expected [ '/api/concierge/repos', …(13) ] to deeply equal [ '/api/concierge/repos', …(12) ]
Tests  2 failed | 6 passed (8)
```

**`grep -c "packages/contract" .swarm/coupling.txt` → `0`.** The registry does not
mention that package at all. The entry that *is* registered for this seam —
`packages/server/src/api/route-class-law.test.ts` — warns that "ANY new route,
registered anywhere under api/, reddens this file from a directory the lane never
entered", and names only itself. A lane reads that, claims route-class-law.test.ts,
[...]

PR #363, opened 2026-09-08
docs(coupling): the containment ratchet's test-file pin is a registry entry (#353)

### LINKED ISSUE #353: coupling: the prefix-comparison pin is a registry entry, not PR prose
## Fence (may touch ONLY)
- `.swarm/coupling.txt`

## Blocked by
Nothing.

## Why now
Surfaced building prd-51 wave 1 (#258), and confirmed still unregistered after the
review of #348 landed `09c43060`, which registered `shipper/hand-law.test.ts` and the two
files it anchors but not this one.

`packages/server/src/paths/prefix-comparison-law.test.ts` pins an **exact count** of
`X.startsWith(<identifier>)` sites in test files (14 today). It therefore fails in **both**
directions: a lane that adds a test using the obvious `file.startsWith(dir + path.sep)`
spelling reddens a law in a directory it never entered, and a lane that removes one does too.

That is the definition this file's own header gives for an entry — *"a file that a change
ELSEWHERE forces an edit to"* — and it is the same shape as `packages/web/src/theme/tokens.test.ts`
and `packages/server/src/api/route-class-law.test.ts`, which are already registered.

It cost real work once already: #258's plan had to re-specify **five** call sites as `Set`
membership / `isInside()` / explicit arrays, purely to keep the count at 14. The plan's author
found this by reading, not by warning — `scripts/fence-lint.sh` reads `.swarm/coupling.txt`,
not PR prose, so a note on a merged PR reaches nobody. The next lane to add a test with a path
prefix comparison gets a red bar from a file outside its fence and no explanation.

## Definition of done
- `.swarm/coupling.txt` carries an entry for `packages/server/src/paths/prefix-comparison-law.test.ts`,
  in the file's stated format (`<path>  # why it couples`), whose reason names: that the pin is
  an exact count over test files, that it fails in both directions, and what a lane should do
  instead (the `Set` / `isInside()` / explicit-array spellings #258 used).
- The reason is judgeable on its own, per the header's rule that an entry with no usable reason
  is worse than no entry.
- `scripts/fence-lint.sh` still passes over a representative wave.
- Falsifier stated and its verdict recorded: if the honest reason turns out to be "any test file
  anywhere may trip this", then the coupling is not a coupling but a repo-wide lint, and the
  entry belongs as a rule in the law's own failure message rather than in this registry.

ciaran-slow:
Closes #353

One entry added to `.swarm/coupling.txt`, registering
`packages/server/src/paths/prefix-comparison-law.test.ts` as a coupling point.
One file, +29 lines — the issue's entire declared fence, no widening.

## Why

That law pins an **exact** count of containment-shaped sites across the test
files of `packages/server/src` — 14 today, in seven files under `api/`,
`concierge/`, `lab/`, `recorder/` and `server/`, none of them in `paths/`. So a
lane can do everything right inside its own fence and still redden a law in a
directory it never entered. It fails in both directions.

It has already cost real work: #258's plan had to re-specify five call sites to
keep the count at 14, and its author found that by reading rather than by
warning — `scripts/fence-lint.sh` reads this registry, not PR prose, so a note
on a merged PR reaches nobody.

## What is the sibling case?

**Found, and filed as #358.** `packages/contract/src/contract-coverage-law.test.ts`
pins `toHaveLength(16)` over gated-read rows parsed out of
`packages/server/src/api/index.ts`, plus an exact three-route exclusion list —
the same both-directions exact count, one package further out.
`grep -c "packages/contract" .swarm/coupling.txt` returns **0**, and the
registered `api/route-class-law.test.ts` entry names only itself, so a lane
adding a route is reddened a second time from a package nothing warned it
about. Not bundled here: this PR's fence is `.swarm/coupling.txt`, which is
also #358's fence, so the two are an overlap by construction.

**Almost all of them, and that is the honest answer.** This is prose in a
`.txt`. `scripts/dev/coupling.test.sh` checks that each entry's leading path
exists and that a reason is *present* — not that it is true — and
`doc-citation-law.test.ts` cannot reach the file at all (its citing-file
predicates are `packages/*.ts`, `docs/*.md`, `*.md`). Change the pin to 15 with
a matching site and this entry goes stale with a fully green suite.

That is not hypothetical. **This entry took five rounds of independent review to
become true**, and every intermediate version passed `typecheck`, `lint`, the
full suite and `coupling.test.sh` 11/11:

1. It said `.startsWith('..')` on a `path.relative` result is NOT a hit. It is —
   the fifth shape, `relative-startsWith-dotdot`.
2. The fix overstated the detector's window by a line, and said
   `shipper/hand-law.test.ts` makes no `path.relative` call when it has one.
3. The fix said a string-literal argument is safe, full stop. False:
   `RELATIVE_FOLLOW_RE` matches a bare `.startsWith('..')` anywhere in its
   window without looking at the receiver. The probe cited as evidence had been
   run beside `path.dirname` and `path.join` — a context with no `path.relative`
   in it, so it could not have falsified the claim it was offered for.
4. A self-audit found the entry quoting `.toBe(14)`, a form occurring nowhere in
   the cited file.
5. Two exhaustive audits in parallel on different models **disagreed**: 28
   claims enumerated and no defects found, against 54 claims and five defects.
   All five were real, and three had survived every earlier round untouched —
   the cancel-out claim, the string-parsing claim, and a `(prd-54 ruling 2)`
   citation pointing at a document absent from this tree.

Every claim in the final entry that can be probed was probed **in a falsifying
context**, on node v22.23.2:

| claim | probe | result |
|---|---|---|
| both directions | plant one site / retire `lab/fork.test.ts`'s site | 15 / 13, both red |
| cancel-out NOT closed | migrate one **and** add one | back to 14, **green** — as the law's doc comment states |
| window boundary | follow-up at +0 / +6 / +7 lines | fires / fires / clear |
| receiver-independence | `specifier.startsWith('..')` at +4 / +11 from a `relative()` helper | 15 / 14 |
| string literal is safe | `specifier.startsWith('.')` **beside a `path.relative(` call** | 14 — the probe round 3 should have run |
| proxy residual | `line.startsWith(marker)` with / without a nearby `path.resolve` | 15 / 14 |
| `.test.tsx` swept | identical site in a `.test.tsx` | 15 |
| `packages/web` not swept | identical site under `packages/web/src` | 14 |

Plus a mechanical audit of the entry itself: every quoted assertion form occurs
verbatim in the cited file, every named symbol exists, every backticked path
resolves.

## Fence

`.swarm/coupling.txt` only. **No widening.** Three findings were deliberately
left out of scope rather than folded in:

- The `route-class-law.test.ts` entry quotes `.toBe(25)` where the file pins
  `30`; the file's **first** entry claims `theme/tokens.test.ts` pins
  `.toBe(225)` where the constant is now `0`. Both pre-existing on `main`.
- Nothing in the tree checks this registry's prose.

All three are **prd-54**, blessed this session, milestone `prd54` — its wave 1
corrects the rot, its wave 2 writes the law. The PRD is a separate commit on
`prd54-blessed` and needs its own PR; this branch deliberately carries **no**
citation to it, so the two can land in either order.

## Verification

Rebased onto `origin/main` at the end, and the rebase was treated as a
fact-changing event rather than a formality — a new containment site landing on
main would have staled this entry before it merged, which is the coupling the
entry describes. Re-derived post-rebase: pin still 14, same seven files, same
five directories, none in `paths/`.

- `npm run typecheck` clean
- `npm run lint` clean, 962 files
- `npm run build` clean
- `VITEST_MAX_WORKERS=6 npm test` — 448 files passed, 8133 passed, 3 expected
  fail, 6 skipped
- `bash scripts/dev/coupling.test.sh` — 11 passed, 0 failed, "21 entries parsed"
- `scripts/fence-lint.sh 325 326 327 328` — PASSED, entry parsing cleanly out of
  the line

`scripts/fence-lint.sh 346 302 301 299 293 292` (prd17 w7) fails on eight
pre-existing overlaps between that wave's own issue fences. Structurally
unrelated to this change: the coupling section sets `fail=1` only when
`.swarm/coupling.txt` is **absent** (`fence-lint.sh:103`), so no entry can
affect that exit status.

## ⚠️ Ledger gap — this PR is opened UNVERIFIED in the AGENTS.md sense

AGENTS.md requires an independent review pass **recorded in the ledger**, via
`~/.claude/skills/verify/scripts/verify-ledger.sh check --strict`. That kit is
operator-local and **is not present on this machine** — no `verify-ledger.sh`,
no `pr-open.sh`, no `pr-approval-guard` hook — so the honest escape
(`pr-open.sh --unverified`) was unavailable and this disclosure is written by
hand in its place.

What did happen, in lieu of a ledger seat: four independent review passes across
two model configurations, then two exhaustive parallel audits. What is missing is
the **record**, not the review — with one real exception: **the final five fixes
are author-written and have not themselves been read by anyone else.** The claim
set behind them has been probed individually in falsifying contexts, but that is
evidence about the claims, not a review of the fixes. A reviewer should read the
entry cold.

Do not read the green suite as covering this. A green suite is what all five
defective versions had.


gabriel-canaan (approved):
**Approve.** Reviewed under `rhizo-review`, single reviewer. I tried to break this entry and could not.

The entry is pure factual assertion about other files, so every probeable claim was re-derived against the **post-merge** tree rather than the branch as pushed. That mattered here: this branch was 9 commits behind `main`, and `main` had added `packages/server/src/api/lab-ceiling-law.test.ts` — a new `api/` test file, exactly the shape of change that could have staled the count between opening and landing. It did not.

### The central claim, re-derived — `EXECUTED`

Mutating `TEST_FILE_KNOWN_DEBT_COUNT` to 99 makes the law enumerate its own sites:

```
test-file containment sites: 14, pinned: 99. Sites:
  packages/server/src/api/lab.test.ts (1)
  packages/server/src/concierge/namespace-law.test.ts (6)
  packages/server/src/lab/fork.test.ts (1)
  packages/server/src/lab/namespace-law.test.ts (1)
  packages/server/src/lab/restore.test.ts (2)
  packages/server/src/recorder/namespace-law.test.ts (2)
  packages/server/src/server/collector-wrap-boundary.test.ts (1)
```

1+6+1+1+2+2+1 = **14**, across **seven** files in **five** directories (`api/`, `concierge/`, `lab/`, `recorder/`, `server/`), **none of them in `paths/`**. The entry's central claim is exact in every part and still true after `main` moved under it. The mutation was restored and the tree confirmed clean before anything else ran.

### The rest, also `EXECUTED`

- `const TEST_FILE_KNOWN_DEBT_COUNT = 14` (`:299`) and `.toBe(TEST_FILE_KNOWN_DEBT_COUNT)` (`:535`) both occur verbatim in the cited file — the cite-by-symbol decision holds.
- `RELATIVE_WINDOW = 7` with `lines.slice(i, i + RELATIVE_WINDOW)` confirms the entry's "same line, or up to SIX lines after it, and a seventh line clear". The off-by-one the entry warns about is stated the right way round.
- The sweep really is `.test.ts` **and** `.test.tsx` — `file.endsWith('.test.ts') || file.endsWith('.test.tsx')`.
- All three production allowlist entries carry `const relative = path.relative(...)`, so "it is the idiom all three production allowlist entries carry" is correct, and the advice to reach for `isInside()` follows from it.
- The doc comment does name `isInConcierge`, `resolvesIntoServerLab` and `findOutsideReferences` as migration targets, as the entry's closing sentence claims.
- Every backticked path in the entry resolves.
- The entry names prd-54 **zero** times (`grep -c` → 0), so this branch is genuinely landable in either order relative to #364. Worth saying explicitly, since removing that dangling citation was a round-5 finding.
- `bash scripts/dev/coupling.test.sh` → 11 passed, 0 failed, "21 entries parsed".
- `scripts/fence-lint.sh` sets `fail=1` only when `.swarm/coupling.txt` is **absent** (the `else` branch of the coupling section), so the PR body is right that the prd17-w7 overlap failures are structurally unrelated to this change — no entry's content can move that exit status.
- The WARN-unreachable residual the entry records is asserted by `scripts/dev/coupling.test.sh` itself: *"WARN is still unreachable — recorded, not fixed"*.

### Gates, on the post-merge tree

```
npm run lint                     exit 0   963 files
VITEST_MAX_WORKERS=6 npm test    exit 0   449 files · 8173 passed · 3 expected fail
```

Larger than the 448 / 8133 recorded in the PR body, which is the expected direction: `main` grew by 9 commits since this branch was measured, and the work survived them. `gh pr checks` green on all 10 legs.

### On the disclosed ledger gap

The kit is present on this machine, and it answers its own question: *"another author does not run it, so their commits cannot have rows here and their absence is not a finding — that work went through its own review."* `check --strict` over `origin/main..origin/353-coupling-prefix-comparison` exits 0. This pass is the independent read the disclosure asked for, and it is being recorded as a single-reviewer row rather than forced.

**No findings.** Not approving my own fix here — I did not need to write one.

PR #364, opened 2026-09-08
docs(prd): prd-54 — the registry that checks itself

ciaran-slow:
Blessed by @ciaran-slow, 2026-09-08, in session. Milestone `prd54`.

Closes no issue — this is the PRD itself, landing before its backlog exists, per
the standard's rule that a PRD is blessed and on the trunk before `/issue-groom`
runs. One file, one commit, 196 lines: `docs/prds/prd-54-the-registry-that-checks-itself.md`.

## Why

`.swarm/coupling.txt` is the repo's pre-dispatch warning system, and every entry
is **prose asserting facts about other files**. Nothing checks whether any of it
is still true, and the script written to check it is wired to nothing.

Measured, and re-derived against `main` after rebasing:

- `grep -rn "coupling.test.sh" .github/workflows/ scripts/ package.json` returns
  only the script's own header lines. **`scripts/dev/coupling.test.sh` is in no
  workflow, no `package.json` script, and not in `scripts/gate.sh`.** It passes
  only when somebody remembers it — which is why the next two are true.
- **Every entry that quotes a numeric assertion literal is wrong, both of them.**
  The registry's *first* entry, its exemplar, says `theme/tokens.test.ts` pins
  `.toBe(225)`; the constant is now `0`, under a comment reading "The ratchet is
  now a wall" — which makes that entry's own both-directions argument vacuous.
  The `api/route-class-law.test.ts` entry quotes `.toBe(25)` twice; the file pins
  `30` twice.
- **A cited path already fails to resolve, legitimately.** `.swarm/timing-count`
  is absent because `.gitignore` excludes it and `gate.sh` creates it on first
  use. A naive "every cited path exists" check false-positives on it — the case
  that makes this need a ruling rather than a one-line grep.
- **#353's entry took five rounds of independent review to become true**, and
  every intermediate version passed typecheck, lint, the full suite and
  `coupling.test.sh` 11/11. The round-5 audits disagreed with each other: 28
  claims enumerated and no defects found, against 54 claims and five defects,
  all five confirmed.
- `doc-citation-law.test.ts` cannot reach the file — its citing-file predicates
  are `packages/*.ts`, `docs/*.md` and `*.md`. This is a `.txt`.

## Four rulings

1. **The law is a vitest law**, not shell — the suite is what CI and the landing
   gate both run. `coupling.test.sh` stays as the operator's fast path.
2. **An entry cites a symbol**; any value it quotes must occur verbatim in the
   cited file.
3. **A cited path must resolve, or be declared generated** in the entry, with the
   law asserting its generator.
4. **The rot is corrected before the law exists**, in its own wave — so the law
   neither arrives red nor arrives with a day-one allowlist.

## Rulings 2 and 3 each carry a dated amendment note

Both edges were found by auditing #353's entry against the rulings *before* any
code was written to them, which is the cheapest moment to find them:

- **Ruling 2.** An entry may quote an absent form precisely to assert its
  absence, and a verbatim-occurrence check cannot tell that apart from a stale
  quote — Ruling 3's `.swarm/timing-count` problem in reverse, a claim true
  *because* the thing is absent. Left open.
- **Ruling 3.** "Repo-relative" is the wrong bound. The registry's convention
  inside a reason is **package-relative shorthand** — `paths/containment.ts`,
  `theme/tokens.test.ts`, `app/StatusBar.tsx` — none of which resolve from the
  repo root. A law written to the ruling's literal wording would redden nearly
  every entry in the file on its first run, **including the three entries the
  ruling exists to protect.** Left open, and flagged as settle-before-wave-2.

Both verdicts stand; only their edges are now known. Two further questions are
left explicitly unruled, with "Open, not ruled." on each.

## Provenance correction, worth naming

The draft's Evidence section said "over all 21 entries" and attributed that to
`main`. **21 was the lane's figure** — the lane that adds the twenty-first entry.
It is 20 on `main`, becoming 21 when #353 lands. Found by re-deriving the
evidence against `main` after rebasing rather than against the branch it was
drafted on, and the section now states the correct count and tells the reader to
re-derive rather than trust it. Every other measured claim was re-checked on the
rebased tree and still holds.

## Sequencing and the fence collision

Wave 1's fence is `.swarm/coupling.txt`, which is also **#353's** fence (PR #363,
open) and **#358's**. Those three are an overlap by construction, so wave 1 is
blocked on both and `scripts/fence-lint.sh` will say so. The Sequencing section
records that; grooming holds wave 1 behind them.

This branch deliberately carries **no dependency on #353**, and #353 carries no
citation to this document, so the two can land in either order.

## Verification

- `npm run typecheck` clean
- `npm run lint` clean, 962 files
- `VITEST_MAX_WORKERS=6 npm test` — 448 files passed, 8133 passed, 3 expected
  fail, 6 skipped
- `packages/server/src/doc-citation-law.test.ts` — 49/49 with the document in
  the tree. That is the law that sweeps `docs/*.md` for `#NNN` citations, so it
  is the one that could have rejected this document's references.

## ⚠️ Ledger gap — opened UNVERIFIED in the AGENTS.md sense

Same disclosure as #363, same cause: the operator kit (`verify-ledger.sh`,
`pr-open.sh`, the `pr-approval-guard` hook) is **not present on this machine**,
so neither condition 1 nor its sanctioned `--unverified` escape was available,
and this note is written by hand in place of the latter.

**No independent pass has read this document.** Its evidence was measured and
re-derived by its author, and its rulings' two amendment notes came out of
auditing a different branch against them — not out of anyone reviewing the PRD.
A reviewer should check the four rulings against the Rejected-alternatives
paragraph and confirm the Success falsifiers are actually falsifiable.


gabriel-canaan (approved):
**Approve to land.** Reviewed under `rhizo-review`, single reviewer. One finding to settle before wave 2 is groomed — it does not hold the document.

Every measurable claim in the Evidence section was re-derived against the **post-merge** tree (this branch was 9 commits behind `main`), and all of it holds:

- **All fifteen line-number citations resolve.** `tokens.test.ts:243` → `const RAW_PIXEL_SIZES = 0`; `route-class-law.test.ts:118-119` → `expect(routes.length).toBe(30)` / `expect(ROUTE_CLASSES.length).toBe(30)`; `.gitignore:17-18` → `.swarm/*` / `!.swarm/coupling.txt`; `gate.sh:570` → `COUNT_FILE="$root/.swarm/timing-count"`; `fence-lint.sh:79-81`; `coupling.test.sh:52-67` and `:93-110`; `doc-citation-law.test.ts:81`, `:980`, `:2091`. The claimed re-derivation after rebase actually happened, which is not the norm and is worth saying.
- **`coupling.test.sh` really is wired to nothing** — `grep -rn "coupling.test.sh" .github/workflows/ scripts/ package.json` returns only the script's own header lines, exactly as stated.
- **20 entries on `main`.** The corrected count is right, and this branch touches no coupling entry.
- **The `.txt` really is out of reach.** The law's citing-file predicates are `docs/*.md`, `docs/research/**/*.md`, `*.md` and `packages/*.ts` — no `.txt` among them. (The PRD lists three of the four; the omitted `docs/research/**/*.md` does not change the conclusion.)
- `packages/server/src/doc-citation-law.test.ts` — **49/49** with the document in the tree, matching the PR body. That is the law that could have rejected this document's `#NNN` references.
- `npm run typecheck` exit 0 · `npm run lint` exit 0, 963 files · `VITEST_MAX_WORKERS=6 npm test` — 449 files, 8173 passed, 3 expected fail. `gh pr checks` green on all 10 legs.

### Finding — Ruling 2's Extent and its body disagree, on Success 4's own second example. `EXECUTED`

Ruling 2's body requires *"the quoted **form** must occur verbatim in the cited file"*. Its Extent sentence scopes the ruling to *"every backticked `.toBe(N)` / `.toHaveLength(N)` in any reason"*. Those two are the same string for the tokens entry, so the worked example cannot distinguish them. For the route-class entry they come apart, and the cited file contains a decoy:

```
ABSENT   "expect(ROUTE_CLASSES.length).toBe(25)"   in api/route-class-law.test.ts
ABSENT   "expect(routes.length).toBe(25)"          in api/route-class-law.test.ts
PRESENT  ".toBe(25)"                               in api/route-class-law.test.ts   <- :180
```

`route-class-law.test.ts:180` is `expect(gatedFound.length).toBe(25)` — the count of gated routes, semantically unrelated to the two pins of `30` the entry is actually making a claim about.

So a wave-2 law written to the **Extent** wording (lift the `.toBe(N)` token out of the backticks, assert it occurs) reports **green** on one of the two entries Success 4 names as rotted. Written to the **body** wording (the whole backticked span must occur) it correctly reddens. Same entry, opposite verdicts, and the failing reading is the one the Extent sentence states literally.

This is the same class as the two amendment notes already in the document — an edge found by auditing against a ruling before code exists — and the same remedy: a third dated amendment note, or one clarifying clause in Extent saying the unit checked is the whole backticked span, not the assertion token inside it. Ruling 3's note already carries a *"must be settled before wave 2 writes the check rather than during it"*; this belongs beside it.

**It does not block this PR.** A blessed PRD landing ahead of its backlog with known-open edges recorded is this document's own convention, and it is followed carefully here. It blocks wave 2 grooming.

Success 4 reads *"**Every** entry is true the day the law lands"*, while the waves correct two entries and the law reaches two mechanical claim classes. The other 18 entries' un-backticked prose claims are explicitly unreachable — the PRD says so itself under "Unfiled work implied", and Non-goals rules them out. The falsifier beneath the criterion is correctly narrow (it names the two entries); only the headline sentence reads broader than the work can deliver.

### One note for wave 2's parser

`.swarm/coupling.txt:24` carries a bare `` `.toBe()` `` inside a `#` header comment, not an entry reason. `fence-lint.sh:79` already skips comment lines (`case "$line" in ''|\#*) continue`), and the PRD is right to require the new reader to agree with that parser exactly — this is the line that will catch it if it does not.

### On the disclosed ledger gap

The kit is present on this machine and answers its own question: *"another author does not run it, so their commits cannot have rows here and their absence is not a finding."* `check --strict` over `origin/main..origin/prd54-blessed` exits 0. This pass is the independent read the disclosure asked for — the four rulings were checked against the Rejected-alternatives paragraph and the Success falsifiers were checked for falsifiability, which is where the finding above came from. Recorded as a single-reviewer row, not forced.

PR #365, opened 2026-09-08
test(lab): no configuration input reaches the lab lock ceiling   (#236)

### LINKED ISSUE #236: prd50 w1: no configuration input reaches the lab lock ceiling

gabriel-canaan:
Closes #236

prd-50 rulings 1 and 2, wave 1 — the keystone both wave-2 issues consume.

`LAB_CLI_LOCK_CEILING_MS` stays fixed in source. That is a claim, not an omission: the ceiling is the mechanism by which a launch that would wait silently behind another one **refuses** instead, so a configurable threshold is a supported way to rebuild the queue prd-41 ruling 2 refused — spend hidden behind latency, which prd-12 ruling 3 forbids. This lands the sixth `NON_NEGOTIABLES` entry (the register) and a new server law (the enforcement). The issue is explicit that it is **both, or it is theatre**: the existing five entries are enforced by matching a rendered browser control's offer text, which a server constant matches not at all, so an entry added alone would read as enforced and enforce nothing.

Three files: `packages/server/src/api/lab-ceiling-law.test.ts` (new), `packages/web/src/settings/non-negotiables.ts`, `packages/web/src/settings/non-negotiables-law.test.ts`.

## What is the sibling case?

Two, and the second is the one that bit.

**Named in the issue and honoured:** the rest of the ceiling family — `MAX_ARMS`, `RESTORE_EXEC_TIMEOUT_MS`, `COMPARE_*`, `FORK_EXEC_TIMEOUT_MS`, `ROUTE_EXEC_TIMEOUT_MS`, `COLLECTOR_EXEC_TIMEOUT_MS` — is **deliberately not swept**. prd-50 open question 1 leaves them unruled, and a law general enough to guard the family would answer an open question in a test file. Verified: none of them lives in `api/lab.ts`, and a sibling constant wired to an env var passes this law.

**Found during review:** the rigged reds. `RIGGED` was not tied to `NON_NEGOTIABLES`, so a rig could be deleted silently and the entry it proves would stop being proven. Now asserted set-equal, both directions, with a duplicate guard.

The issue names the one that matters — wiring a configuration input to the constant **without changing its value**, so that a law asserting only `=== 30_000` stays green while the ceiling becomes operator-settable. The assertion that survives is about **reachability**, not value.

Executed against the final head, each applied and reverted, all reddening: `ceilingMs = config`, `|=`, `??=`, `--`, array and object destructuring, `for...of`, `for...in`; a **compensated swap** that keeps the occurrence count constant by replacing a known read with a config-derived local; a third call site with an explicit generic type argument; and a config-scaled declaration.

And the mutation the law failed at review — **deleting the law's own load-bearing assertion** left 32/32 green. Now `expect.assertions(1 + KNOWN_CEILING_READS.length)` makes each real-source assertion load-bearing: deleting the count, neutering the containment loop, or both, each reddens.

Mutations this does **not** catch, named rather than left implicit: any flow that reaches the ceiling without adding an occurrence of the token — an alias, a getter, a shadowed global, a wrapper, an `eval`. See below.

## The claim was narrowed to match the instrument — read this before approving

Three review rounds each closed a class of bypass and revealed the next, because *"no configuration input reaches this constant"* quantifies over every possible program and **a static text scan of one file cannot decide it**. Rather than a seventh detection round, the claim was narrowed to what the law establishes.

The `describe` title now says so in the reporter's own output: *"every enumerated wiring to `LAB_CLI_LOCK_CEILING_MS` fails the suite; general reachability is not something a text scan proves."* The entry's `why` names the real guarantee — **ruling 1 and a reviewed diff** — and states the limit.

**Five bypasses reported by review seats are open by decision, not oversight**: a module-scope `setTimeout` shadow, an early `return` in the ceiling promise, a brace-balanced decoy string defeating the body extractor, bracket assignment on the options seam, and a generic-typed third call site. Two did not reproduce when rebuilt, so the list is indicative. Detection was deliberately not widened.

## Verification — and one thing to weigh

Six rounds. Rounds 1–4 were **two independent seats each** (Opus and Codex `gpt-5.6-terra`, `xhigh`, isolated worktrees). Round 5 was a further seat round the operator authorised **over the re-review budget's refusal** — it found the self-guarding defect, so the override earned itself.

**Rounds 5 and 6 have been read by one reviewer only**, and the ledger records that rather than claiming seats: `reviewed … ^ single reviewer: orchestrator-opus`. Round 5's change is one line (`expect.assertions(…)`), round 6 is two strings and no logic — verified prose-only by diffing out comments. That is a lighter class of evidence than the earlier rounds and is flagged here so it is weighed rather than assumed.

Gates on this head: `typecheck` rc=0 · `lint` rc=0 · `VITEST_MAX_WORKERS=6 npm test` → **449 files / 8173 passed / 3 expected fail / 2 skipped**, rc=0.

## Fence

Exactly the three declared paths, across all six rounds. **`packages/server/src/api/lab.ts` was never edited** — the law imports the constant. Making `ceilingMs` a `const` binding would have been the cleanest fix and was refused as out of fence; #237 owns that file's doc comment next. Coupling points `registry.ts` and `coverage-law.test.tsx` untouched and green unmodified.

Known and out of fence: `registry.ts:32` still reads "five things may never become a preference". Correcting it needs a recorded widening.


ciaran-slow (commented):
## Review — `df0e6765`, reviewer pass

Read commit-by-commit (one commit), in a clean clone off `origin/main` at `4202c549`,
node v22.23.2. **Every claim below is labelled `EXECUTED` or `REASONED`.** Where I
report a green, it comes with a matched control proving the harness applies edits and
vitest sees them — a green with no control is not evidence.

### What verified in the author's favour — re-derived, not taken on trust

The PR body carries a large set of `EXECUTED` claims. I re-ran the four load-bearing
ones myself rather than inherit them, and all four hold:

| probe | expected | `EXECUTED` result |
|---|---|---|
| control, unmutated | green | **32 passed** |
| body-assigned config input (`ceilingMs = Number(process.argv[2]) \|\| ceilingMs`, first statement of the body) — the round-3 blocker | red | **3 failed / 29 passed** ✅ |
| unrelated env read in `lab.ts`, no path to the ceiling (`LAB_UNRELATED_LABEL`) | **green** | **32 passed** ✅ the round-3 overreach is genuinely fixed |
| G1 — delete the production count assertion `expect(occurrences.length).toBe(KNOWN_CEILING_READS.length)` | red | **1 failed / 31 passed** ✅ `expect.assertions` really is load-bearing |
| delete the sixth `RIGGED` entry | red | **1 failed / 19 passed** ✅ |

Two further things worth stating plainly, because both are favourable and neither was
asserted this strongly in the PR body:

- **One of the five "open by decision" bypasses does not reproduce — it is closed.**
  `EXECUTED`: a third `withLabCliLock` call site carrying an explicit generic
  (`withLabCliLock<void>("x", async () => {}, Number(process.argv[2]))`) added to
  `lab.ts` reddens the law — **1 failed / 31 passed**. `callsTo`'s optional-`<…>`
  handling plus the asserted call count catch it. The PR body lists this among the five
  residual routes; on this head it is detected. Worth striking from that list.

- **The five→six reconciliation is complete.** `EXECUTED` (`git grep -i '\bfive\b'` over
  both changed files): four sites remain and every one is correct in context — the
  law's `:169` "behind the five it still catches", `non-negotiables.ts:46`/`:56`
  distinguishing the five regex-guarded entries from the sixth, and `:146`'s five
  protected *namespaces* (genuinely still five). `registry.ts:32`'s stale "five things"
  is still there, is a declared coupling point, and correctly needs a recorded widening
  — not this PR's, and the author already reported it.

Fence is exact: three declared paths, `packages/server/src/api/lab.ts` never edited,
`registry.ts` and `coverage-law.test.tsx` untouched.

**A hypothesis of mine that was wrong, stated so it isn't re-raised.** `.gitattributes`
pins only `*.json`, `*.jsonl` and the collector fixtures to `eol=lf`, so `.ts` is
unpinned and a Windows checkout under `core.autocrlf=true` gets CRLF — and
`route-class-law.test.ts` is already a `line-endings` known failure for the adjacent
"regex anchors on `\n`" shape. I expected
`/^export const LAB_CLI_LOCK_CEILING_MS = 30_000$/m` to fail there. It does not:
`EXECUTED`, converting `lab.ts` to CRLF gives **32 passed**, because ECMAScript's
`LineTerminator` includes CR, so `$` under `/m` matches before `\r` (unlike PCRE).
The real `windows-suite` leg on this head agrees — **pass, 13m48s**. No action.

---

### The one finding — the sweep's scope is narrower than the claim it prints

`REASONED` then `EXECUTED`. Not a live reachability hole; it is a **claim wider than
the instrument**, which is the exact class round 6 was spent closing.

`SWEPT_EXTENSIONS` (`lab-ceiling-law.test.ts:548`) is `['.ts', '.tsx']`. Two tests
built on it claim the whole repo:

- `:701` — *"is named nowhere else in the repo's **production source** — the sweep sees only this module"*
- `:864` — *"`lockCeilingMs` is never assigned a value anywhere in the repo's **production source**"*

and the file doc repeats it at `:59` (item 4). But the repo tracks **7 non-`.ts` production
sources** the sweep never opens, one of which is the server package's own CLI entry
point (`packages/server/bin/rhizomorph.mjs`), plus `packages/web/src/scene/parity/capture.mjs`,
`packages/app/scripts/{make-icon,vendor-runtime}.mjs` and three others.

`EXECUTED`, matched pair on this head:

| probe | result |
|---|---|
| `const labOpts = { lockCeilingMs: Number(process.env.RHIZOMORPH_LAB_LOCK_MS) }` appended to `packages/server/bin/rhizomorph.mjs` | **32 passed** ← the gap |
| **control** — the *identical* assignment appended to `packages/server/src/api/doctor.ts` | **1 failed / 31 passed** |
| `const LAB_CLI_LOCK_CEILING_MS = Number(process.env.X)` appended to the same `.mjs` | **32 passed** ← same gap, other sweep |

The control is what makes the green meaningful: the harness works, so `32 passed` on the
`.mjs` is the sweep not looking, not a broken probe.

**The fix is two lines, and I have verified it** (patch below). `EXECUTED` after applying:

| probe | result |
|---|---|
| real tree, widened sweep | **32 passed** — nothing reddens; no tracked non-`.ts` file contains either name today |
| `lockCeilingMs` in the `.mjs` | **1 failed** ✅ now caught |
| constant named in the `.mjs` | **1 failed** ✅ now caught |
| a **tracked** `.test.mjs` carrying the key | **32 passed** ✅ still excluded as a test file |
| **control** — that same tracked file renamed `.mjs` | **1 failed** ✅ so the exclusion, not the extension list, is doing that |

```diff
-const SWEPT_EXTENSIONS = ['.ts', '.tsx'] as const
+const SWEPT_EXTENSIONS = ['.ts', '.tsx', '.mts', '.cts', '.js', '.mjs', '.cjs'] as const
@@
-    .filter((rel) => !rel.endsWith('.test.ts') && !rel.endsWith('.test.tsx'))
+    .filter((rel) => !/\.test\.[cm]?[jt]sx?$/.test(rel))
```

The second line is required, not cosmetic: without it a `.test.mjs` would be swept
where a `.test.ts` is not. (One consequence worth naming rather than leaving to be
found: `:551`'s doc comment above `sweptProductionFiles` still says ``non-test `.ts`/`.tsx` ``
and would want the same edit.)

**Either disposition closes this and I have no preference strong enough to argue:**
fold the patch in, or narrow the two test names and `:59` to say `.ts`/`.tsx`
instead of "the repo's production source". The patch is the better end state — the
sweep is already a whole-repo text scan, so reading the other seven tracked sources is
the same detection over the surface it already claims, not the instrument growing a new
class. That is a different axis from round 6's decision, which was about reachability a
text scan cannot decide *in general* (aliases, getters, `eval`) — this is just files it
claims to read and doesn't.

### Nit, your call — a dangling pointer in a tracked file

`lab-ceiling-law.test.ts:106-107` cites `` `lab-ceiling-law.enumeration.md` (scratch,
referenced from the commit body) ``. `EXECUTED`: that filename appears nowhere in the
commit body — which cites `lab-ceiling-law.enumeration-r4.md` and `-r5.md` — and none of
the three is tracked. So a reader following the pointer finds nothing, and the
disclosure that would send them to the commit body is itself wrong for that name.
`AGENTS.md`'s own lesson ("a wrong pointer to a real command is worse than no pointer")
applies. One line to fix; deliberately not fixed by me.

---

### What I did NOT run, and did not verify

- **`scripts/gate.sh` — not run.** Landing is the operator's step, not a reviewer's.
- **The ledger check is not available on this machine.** `AGENTS.md`'s landing condition 1
  requires `verify-ledger.sh check --strict --range <base>..<branch>`; there is no
  `~/.claude/skills/verify/` here and no `pr-open.sh`, `lane-precommit.sh` or
  `pr-approval-guard` hook. So I **cannot** confirm the range is ledger-verified, and I
  am not asserting it is — the PR body's own note that rounds 5 and 6 were read by a
  single reviewer stands unaudited by me and should be weighed by whoever lands this.
- **No commit pushed.** This is deliberate and is the reason the patch above is a patch
  and not a sha: I cannot record a commit of mine in the ledger, so pushing would turn a
  range the author reports as verified into one containing an unverified commit — it
  would block the landing rather than unblock it. If the operator wants it pushed, say so
  and I will.
- The two load-sensitive tests the PR body flags (`scene/world.test.ts`,
  `scene/view/useFrameLoop.test.ts`) are a known contention flake, not this branch's.

### CI — confirmed executed, not just green

`gh pr checks 365`: **10 legs, all pass** — three jobs wide as `AGENTS.md` describes
(`build-test-boot` ×3, `pack-smoke` ×6, `windows-suite` ×1). A green bar is not evidence
a gate ran, so `EXECUTED` on the ubuntu/current leg's step list: `Build` success, `Test`
success, `Typecheck` success, `Lint` success, `Packaging guard` success, `Boot smoke`
success. Nothing silently skipped behind a red `Test`.

### Gates I ran myself, on the pristine head, clean tree

`EXECUTED` in the review clone at `df0e6765`, node v22.23.2, exit statuses read
directly (not through a pipe):

- `npm run typecheck` → **rc=0**
- `npm run lint` → **rc=0** (963 files)
- `VITEST_MAX_WORKERS=6 npm test` → **rc=0**, 449 files / **8169 passed / 3 expected fail / 6 skipped**

That last line differs slightly from the PR body's `8173 passed / 2 skipped` — four
tests skipped here that passed there. Environment-dependent skips (jsdom/canvas), not a
regression; reported only so the numbers reconcile rather than looking unexplained.

### Disposition

**No blocking defect.** The instrument is sound where it matters: every enumerated
wiring reddens, the round-3 overreach is really gone, `expect.assertions` really makes
the real-source assertions load-bearing, and one route the PR body still lists as open
is in fact closed. The narrowing decision that ended round 6 was the right call and the
`describe` title now says what it proves.

Holding approval on the one finding only — a fresh instance of the very class round 6
existed to close, with a two-line verified patch attached. Fold it in or narrow the two
names, either is fine, and I will approve on whichever. **Merging is the operator's step
and this repo does not auto-merge.**

Issue #367, opened 2026-09-08
prd54 w2: a law checks the registry's paths and quoted literals


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/coupling-registry-law.test.ts` (new file)
- `scripts/dev/coupling.test.sh`
- `.swarm/coupling.txt`

## Blocked by
**Wave 1's second pass (#546) landing, and nothing else.** #366 landed as 22a42739;
PRs #370 and #392 (rulings 5 to 8) are on main; #546 carries rulings 9 and 10 and the
corrections written to them. Do not start while #546 is open: the law you build asserts
a registry that #546 is rewriting, and ruling 4 forbids landing beside red entries.

Claim widened at the 2026-09-15 regroom: the registry itself joins this issue, for one
entry only — see the Definition of done. #394 (wave 3) waits on this issue and shares no
path with it. No open issue in any milestone claims the three paths above (checked
2026-09-15; #516 and #525 mention the registry in prose and do not claim it).

**Model:** sonnet. **Wave:** 2.

**Read rulings 5 to 10 before you start; they are what this law implements. Do not
re-derive the unit, the resolution order, the selector, or the declaration form from the
earlier rulings' wording** — rulings 5, 7, 8, 9 and 10 exist because that wording was
ambiguous or wrong when someone tried to build to it. Ruling 8's own text records that
ruling 5's selector came through three audits and failed on the first implementation.

## Why now
prd-54's Success 1 and 2. The registry's claims are checked by nothing that runs without
a human remembering it: scripts/dev/coupling.test.sh appears in no workflow, no
package.json script, and not in scripts/gate.sh, and the doc-citation law cannot reach
the file because its citing-file predicates are packages/*.ts, docs/*.md and *.md — this
is a .txt. Measured twice now: the first pass made every claim true on 2026-09-10 and
eleven were false again by 2026-09-15, every one committed through a green suite.

## Definition of done

- A new law under packages/server/src/ asserts, over **every** entry in the registry:
  - **Ruling 7 (which is ruling 2's check):** every backticked span containing
    ".toBe(", ".toHaveLength(" or ".toEqual(" occurs in the cited file as an **exact
    substring**, whitespace included, not normalised, not parsed. The cited file is the
    entry's leading path.
  - **Ruling 6:** no absence machinery. A quoted form that does not occur fails, even
    when the entry's point is that it is absent. No marker, no inversion, no allowlist.
[...]

Issue #366, opened 2026-09-08
prd54 w1: every claim in the coupling registry is true


ciaran-slow:
## Fence (may touch ONLY)
- `.swarm/coupling.txt`

## Blocked by
#353 and #358. Both edit the same single file this issue does, so all three are
sequential by construction rather than parallel. Wait until both are on `main`,
then rebase. Do not start while either is open.

**Model:** opus. **Wave:** 1. **Keystone** — prd-54 ruling 4 puts the correction
before the law, so that the law neither arrives red nor arrives with a day-one
allowlist.

## Why now
prd-54 ruling 4. Two entries in the registry assert values their cited files do
not carry, and one cited path is absent with no declaration. The law in wave 2
would redden on all three the day it lands.

## Definition of done

- The entry for `packages/web/src/theme/tokens.test.ts` no longer claims an
  exact `.toBe(225)`. That form occurs nowhere in the file — it pins
  `.toBe(RAW_PIXEL_SIZES)`, and `const RAW_PIXEL_SIZES = 0` today
  (`packages/web/src/theme/tokens.test.ts:243`). Cite the symbol and give the
  value as an as-of statement, per prd-54 ruling 2. Do **not** write the corrected
  entry so that it quotes `.toBe(225)` in order to say that form is absent —
  **prd-54 ruling 6** refuses that construction outright. Say what the file does
  carry, not what it does not.
- That entry's both-directions argument is re-stated honestly at zero. With the
  constant at `0`, "retiring a pixel literal fails it exactly as adding one
  does" is vacuous in the retiring direction. The comment above the constant
  reads "The ratchet is now a wall" — say what is true now, do not just swap the
  number.
- The entry for `packages/server/src/api/route-class-law.test.ts` no longer
  quotes `.toBe(25)` twice; the file pins `30` twice. **Re-derive both from the
  file** rather than editing 25 to 30 by hand, and reconcile the surrounding
  prose, which argues from the count.
- `.swarm/timing-count` is declared generated in the `scripts/gate.sh` entry,
  per prd-54 ruling 3. The path is absent by design — `.gitignore` excludes
  `.swarm/*` bar the registry itself, and the gate creates it on first use — so
  the entry states that and names its generator.
- **Three** citations in the `README.md` entry are expanded to repo-relative, per
  **prd-54 ruling 5**: the two cross-package shorthands `api/index.ts` and
  `cli/index.ts` (to `packages/server/src/api/index.ts` and
[...]

Issue #359, opened 2026-09-08
prd14: the postinstall-escape law is bounded by its own cost, not the runner's load


gabriel-canaan:
Found by `scripts/gate.sh` blocking an unrelated landing.

## The defect

`packages/server/src/lab/namespace-law.test.ts`'s prd-41 ruling 1 case — *"an install
does not run a postinstall hook that escapes into the watched repo"* — runs a real
`npm install --no-audit --no-fund --ignore-scripts` through
`forkThreeArms({ install: true })`. It is the only test in the file that takes the
install path, and it is 5.6x slower than the next slowest.

It runs under vitest's default 5 s bound. No vitest config in the repo sets a global
`testTimeout`, and the file is **not** in `gate.sh`'s derived timing set
(`grep -cF '// @gate-timing'` returns 0, and the name is not `*.bench.test.ts`), so it
runs at full concurrency in the load probe.

| condition | duration |
|---|---|
| idle, file alone | **2020 ms** |
| next-slowest test in the same file | 360 ms |
| under `gate.sh`'s 4x load probe | **5209 / 5301 / 5358 / 5378 ms** — all four over the bound |

## Evidence — EXECUTED

`scripts/gate.sh <lane> <fence> 4`, 2026-09-08:

```
under 4x load (timing tests excluded): 4 failures / 16
GATE FAILED: flaky under load — remove the race (never widen a timeout)
>>> HOLDING <lane> — not merged
{"writer":"gate","kind":"gate.verdict","held":true,"reason":"load-flake","loadBatches":4}
```

All four failures were this one test. Every other test in the suite passed all 16 runs.
The landing it held was a one-file change to `packages/web/src/scene/world.test.ts`,
which touches nothing this test reads.

Two qualifications on the numbers, so nobody over-reads them:

- **This was a loaded box.** Three lane agents were running concurrently alongside the
  probe's 16 suite runs, so 5.2 s is a heavily-contended figure rather than a floor.
  The idle 2020 ms is the reliable number.
- **It is starvation, not a race.** The assertion is
  `expect(existsSync(escapeTarget)).toBe(false)` after an awaited install — a filesystem
  check, with no wall-clock assertion and no concurrency in the test itself.

## What this is NOT

**Not a masked vulnerability.** The security property prd-41 ruling 1 claims — and that
`SECURITY.md` cites this law as evidence for — is carried by the `existsSync` check, not
by the duration. If `--ignore-scripts` were ever dropped, the hook would run, the file
[...]

PR #295, opened 2026-09-06
docs(prd-17): rulings 5 and 6 reach main, and the waves are sequenced

gabriel-canaan:
Docs only — the prd-17 PRD and `docs/roadmap.md`. No code, no test, no behaviour.

**These rulings already govern work that is in flight, and they were not on
`main`.** Five commits authored 2026-09-05 sat on an unmerged local branch
carrying **Ruling 5** (the instrument raises its own summons, server-side, on the
poll loop's tick), **Ruling 6** (the gate's verdict rides as extra keys on a
beacon line), and the wave grooming for waves 2, 3, 4 and 6.

Meanwhile `#278`'s body says *"Read ruling 5 before starting"* and `#273` was
built to ruling 6 — both citing a document `origin/main` had never seen. A
reader following either citation found nothing. That is the wrong-pointer failure
`AGENTS.md` names as worse than no pointer, and it is the same shape as its own
recorded prd-44 incident: work that reads as decided while nothing reached
`origin`.

Landing this makes both citations resolvable. It also means the rulings survive
the machine they were written on — until now they existed in exactly one place.

## The conflict, and how it was resolved

`origin/main` moved 40 commits while this branch sat, and wave 2 landed two
amendments into the same file. The PRD's Outcome header collided.

**Neither side was current**, so the resolution was authored fresh rather than
picked:

- the branch's header predated the grooming it describes and stated *"the prd17
  milestone holds no open issue"*, which stopped being true when the waves were
  groomed;
- main's header predated wave 2 landing, so it said ruling 4 had not started and
  that nothing emitted ruling 1's families.

Every claim in the replacement was verified against the tree rather than
inferred:

| claim | how it was checked |
|---|---|
| the operator's three acts ARE emitted | `packages/server/src/api/operator.ts` is a non-test emitter for `operator.ack`/`verdict`/`note` |
| the summons pair and `gate.verdict` are NOT | they appear only in `packages/web/src/tide/` — readers, not emitters |
| `dispatch.brief` and `fence.declared` appear nowhere | no non-test file references either |
| ruling 4's mark kinds landed | `CHAPTER_KINDS` carries eight, including all four new ones |
| ruling 2's doorway landed | prd-27 wave 1 (#217), merged; nothing writes to it yet |

The tail conflict was additive on both sides — the branch's ruling-5 amendment
and main's two — so all four amendments are kept, in date order. Nothing was
dropped or rewritten; the earlier amendments stay as the dated records they are,
and the header carries current truth.

## Verification

[...]

PR #370, opened 2026-09-08
docs(prd): prd-54 rulings 5, 6 and 7 — wave 0 complete

ciaran-slow:
Blessed by @ciaran-slow, 2026-09-09, in session. Wave 0 acts for milestone
`prd54`; no issue is minted for a wave-0 operator act, so this closes nothing.
One file, four commits — the last two answer the review of this PR (`004244dc` is a
pure reflow: `git diff --word-diff` over it is empty).

Builds on `5b9cc5e5` (@gabriel-canaan) and is reconciled against it — see
**What this does not settle**, which is the important half of this PR.

## Ruling 5 — two resolution attempts, in order, and nothing else

Repo-relative first; then, only if the entry's own leading path lies inside a
package source root, relative to **that one root**. No search over other
packages.

Measured over all 21 entries *before* deciding, which is what made a strict rule
affordable — the registry carries **ten** backticked path citations:

| resolves | count | verdict |
|---|---|---|
| repo-relative | 2 | passes |
| under the entry's own package root | 4 | passes — the shorthand Ruling 3's note found |
| only under **another** package | 2 | must be expanded — both in the `README.md` entry |
| under neither attempt, file exists | 1 | must be expanded — `manifest-law.test.ts`, also in the `README.md` entry |
| nowhere at all | 1 | `.swarm/timing-count`, already Ruling 3's generated declaration |

**Total cost: three path expansions, in one entry, all three inside wave 1's
single file.**

**This table said nine until the review** — see the review-response section
below. The tenth is a bare filename with no `/`, which is the shape a hand-count
reads as prose.

*Rejected:* "try every package root" passes nine of the ten with no edits, and lets a
citation resolve against a **different file than the author meant** while going
green — true and useless, #649's lesson and the reason Ruling 3 exists.
*Rejected:* repo-relative everywhere, which also works but expands four legible
shorthand citations in a file whose entries are already long.

## Ruling 6 — no quoting a form that does not appear, even to say it is absent

Ruling 2's check stays a plain verbatim-occurrence test: no marker, no
inversion, **no allowlist**, which is Ruling 4's point honoured by having no
exception to allowlist.

The asymmetry with Ruling 3 is principled, not stylistic. A generated path must
be *named* to be useful and existence cannot check it, so a declared exception
is the only way to carry it. An absent form never needs quoting — #353's entry
went from quoting `` `.toBe(14)` `` to say that form is absent, to *"the pin is
written against the constant and never against the literal"*: shorter, truer,
and it names what the reader will actually find. What that buys is the property
that a backticked form in a reason is **always** greppable. Grant the exception
and a reader who greps one and finds nothing cannot tell rot from a footnote.

## Ruling 7 — Extent selects, the body checks: the unit is the whole span

`5b9cc5e5` (@gabriel-canaan) added a **second** edge to Ruling 2 *after* the
absence question had been decided: the ruling's body requires the quoted **form**
to occur verbatim, while its Extent sentence names `.toBe(N)`. As a selector they
agree; as *the thing to check* they diverge.

Ruling 7 settles it — **the body is the check, Extent is only a selector**, and
what must occur verbatim is the whole backticked span, matched as an exact
substring, not the assertion token lifted out of it. Ruling 2's verdict is
unchanged; its Extent sentence is **reworded from naming those forms to
selecting them**, which is the substantive edit. Appending the ruling while
leaving Extent ambiguous would have left the contradiction standing.

**Re-derived independently over all 21 entries rather than inherited**, and it
held:

| span | span present? | token present? | |
|---|---|---|---|
| `` `.toBe(225)` `` — tokens.test.ts | no | no | agree |
| `` `expect(ROUTE_CLASSES.length).toBe(25)` `` | no | **yes** | **disagree** |
| `` `expect(routes.length).toBe(25)` `` | no | **yes** | **disagree** |

The decoy is `api/route-class-law.test.ts:180` — `expect(gatedFound.length).toBe(25)`
counts *gated* routes while the two pins that entry describes stand at 30. **The
span reading reddens all three false claims; the token reading reddens one.** Two
of the three claims this PRD exists to catch would have passed, Success 4's own
example among them. And there is no false-positive surface: only two of the 21
entries carry an in-scope span, the other nineteen carry none, so the law lands
with **no allowlist** — Ruling 4's point.

**The selector also widens to symbolic arguments, which the note did not raise.**
Under a numeric-only selector, Ruling 2's own prescribed remedy goes unchecked: an
entry citing its pin by symbol — `` `.toBe(TEST_FILE_KNOWN_DEBT_COUNT)` ``, exactly
what Ruling 2 tells an author to write — falls outside `.toBe(N)` and nothing
checks it. EXECUTED: the widened selector matches **four** spans, reddens the three
false ones, passes that symbolic one. Zero false positives. A renamed constant
leaving a green suite and a lying doc is a failure mode this repo already records.

**Wave 0 is now COMPLETE**, three of three, and #367 is blocked on wave 1 only.

One observation the three edges together earn, recorded in Sequencing: all three
were found by **auditing an entry against a ruling rather than by writing code to
it**, and two of the three by a reader other than the ruling's author. That is the
argument for the audit step wave 2's law replaces — and the reason wave 2 should
not start until the remaining entries are audited against Rulings 5, 6 and 7 too.

## Sequencing rewritten to what was groomed

Not to the draft it replaced. Three deviations recorded with reasons: the
`.swarm/timing-count` declaration moved into wave 1 (the law arrives red without
it — an intra-wave dependency is a stack wearing a bundle's clothes); wave 2
became one issue rather than three parallel items (its pointer names the law's
own filename); and the law ports the shell script's presence checks (Success 1
is unmet while they run only when someone remembers them). A Sequencing section
describing a different plan than the issues that exist is the "never closed out"
anti-pattern.

## What is the sibling case?

Ruling 3's edge and Ruling 2's absence edge are each other's sibling — same
class, both found by auditing an entry against a ruling rather than by writing
code to it, and Ruling 5 and 6 are drafted as a pair for that reason. The *third*
edge, in `5b9cc5e5`, was found the same way again and by a different reviewer,
which is the argument that this class is not exhausted and that wave 2 should not
start until someone has audited the remaining entries against Rulings 5 and 6 too.

**None — these are rulings in a markdown file, and nothing in the tree checks
them.** That is prd-54's own subject, and #367 is the issue that fixes it. Until
that lands, the only thing standing between a wrong ruling and the code written
to it is a reader. Stated plainly so no one mistakes the green bar below for
coverage of the rulings themselves.

## Review response — `b1a6dae5` answers all five findings from @gabriel-canaan

Every finding was re-derived here before being acted on; none was taken on the
review's word, and one of the review's *remedies* is not the one taken.

**1 · Ruling 5's population is ten, not nine. `EXECUTED`, confirmed.** A fresh
sweep of all 21 entries finds fourteen path-shaped backticked spans; bucketing
them under Ruling 5's two attempts reproduces the review's table exactly. The
tenth citation is `` `manifest-law.test.ts` `` in the `README.md` entry — no
`/`, so not repo-relative, and `README.md` has no package root, so by Ruling 5's
own Extent it gets no second attempt. The file is real, at
`packages/app/src/host/manifest-law.test.ts`. Fixed in Ruling 5's table and cost
sentence, in Ruling 3's ANSWERED note, in Sequencing's wave 1, and in **#366's
Definition of done**, which named two expansions and said *"Nothing else is
expanded"* — so as written the issue forbade the third correction and the wave-2
law would have arrived red on it.

**2 · Ruling 5 defined no selector. Confirmed — remedy differs.** Rows 11–14 of
that sweep (`` `.test.ts` ``, `` `.test.tsx` `` twice, `` `*.bench.test.ts` ``)
are a suffix, a suffix and a glob, and every extension-based selector catches
them, which is what made the population a hand-classification.

The review proposes *"contains `/` and no glob metacharacter — which admits rows
1–10"*. **It does not admit row 10:** `manifest-law.test.ts` has no `/`, so that
selector deletes finding 1's own case. The rule written into Ruling 5's Extent
instead: **no glob metacharacter, and either a `/` or a bare filename not
beginning with a `.`** — admits 1–10, excludes 11–14. #366 now carries the same
sentence, with an instruction to re-derive the population from the rule rather
than from either count.

**3 · Sequencing's wave 2 omitted Ruling 7. `EXECUTED`, confirmed.** The sentence
read "Rulings 2, 3, 5 and 6"; it was written in `f85a6923` and `fea3292a`
appended Ruling 7 one commit later without revisiting it. Now names all five, and
says what Ruling 7 contributes — the span reading catches three of the registry's
false claims where the token reading catches one, so a law built to Ruling 2
alone passes two of the three defects this PRD is about.

**4 · Duplicate `[i367]` link definition. `EXECUTED`, confirmed.** Line deleted.

**5 · "One file, one commit" against two commits. Confirmed.** Corrected at the
top; it is four now.

**Not fixed, because it is not a defect:** the review's "in your favour" table
re-derived every measured claim in the first two commits and all of them
reproduce. Nothing in Rulings 6 or 7 changed.

`EXECUTED` on `b1a6dae5` in an isolated worktree at `fea3292a`, node v22.22.2,
real `npm install`: `npm run typecheck` clean · `npm run lint` clean, 963 files ·
`VITEST_MAX_WORKERS=6 npm test` **449 files passed, 8169 passed**, 3 expected
fail, 6 skipped, exit 0. No `lab/namespace-law.test.ts` flip this run, which is
consistent with the review's reading of it as #359.

## Verification

Isolated worktree at `5b9cc5e5`, node v22.23.2:

- `npm run typecheck` clean; `npm run lint` clean, 963 files
- `doc-citation-law.test.ts` 49/49 with the edits in the tree — the law that
  sweeps `docs/*.md` for `#NNN` citations

**`VITEST_MAX_WORKERS=6 npm test` was run twice on the identical tree, and the two
runs disagree.** The first exited 0, fully green. The second returned 448 files
passed / 1 failed, 8168 passed — `lab/namespace-law.test.ts`'s postinstall-escape
law. That flip is itself the evidence: the law is **#359**, load-sensitive by its
own issue title, and it passed **22/22 run alone** against this change earlier.
Not this change's, and not reported as green when it was not.

**Three earlier runs were environmental and are recorded because they looked
real:**

| run | tree | result |
|---|---|---|
| 1 | shared worktree carrying another session's 8 uncommitted files | 10 files red |
| 2 | isolated worktree, `node_modules` **symlinked** | 2 files red |
| 3 | isolated worktree, real `npm install` | green |

Run 2's failure was self-inflicted and worth naming: `.gitignore`'s pattern is
`node_modules/` **with a trailing slash**, which matches a directory and not a
symlink — so the symlink was an untracked file and `runbook-delivery-law.test.ts`'s
ignore-detector fired on it, correctly. Not one failure across runs 1 or 2 was
about this change, which is a markdown file.

## ⚠️ Ledger gap — opened UNVERIFIED in the AGENTS.md sense

Same cause as #363 and #364: the operator kit (the review ledger,
the PR-opening script, the approval guard) is not present on this machine, so
neither condition 1 nor its sanctioned `--unverified` escape was available and
this note is written by hand instead.

**No independent pass has read either commit on this branch.** All three rulings
were drafted, blessed and committed in one session. What partially stands in for a
review is that `5b9cc5e5`'s independent pass over the previous version is what
produced Ruling 7's edge at all — so the document was reviewed once, at the commit
this branch is based on, and Ruling 7 exists because of it.

A reviewer should check three things specifically: that Ruling 2's reworded Extent
sentence no longer contradicts its body; that Ruling 5's two-attempt resolution
and Ruling 7's span-substring rule cannot redden a correct entry between them; and
that the Success falsifiers are still falsifiable after three rulings landed on
the document.


gabriel-canaan (commented):
Reviewed commit-by-commit at `fea3292a`, in an isolated worktree with `origin/main` (`88df0484`) merged in, node v22.22.2.

I re-derived **every** measured claim in this PR independently against `.swarm/coupling.txt` rather than reading them. All of them reproduce exactly:

| claim | result |
|---|---|
| 21 registry entries | exact |
| Ruling 5's nine path citations: 2 repo-relative / 4 own-package-root / 2 other-package-only / 1 nowhere | exact, all four buckets |
| Ruling 7's widened selector matches 4 spans — 3 absent from the cited file, 1 symbolic present | exact; zero false positives |
| the decoy is `api/route-class-law.test.ts:180`, `expect(gatedFound.length).toBe(25)` | exact — and the two pins that entry describes are `:118-119`, both `30` |
| span reading reddens 3 of 3, token reading 1 of 3 | confirmed: `.toBe(25)` occurs in that file, `.toBe(225)` occurs in neither |
| `prefix-comparison-law.test.ts:535` is `).toBe(TEST_FILE_KNOWN_DEBT_COUNT)` | exact — the substring rule's worked case holds |

Of the three things your body asks a reviewer to check:

- **Ruling 2's reworded Extent no longer contradicts its body.** Confirmed. The body checks the span verbatim; Extent now *selects*. Ruling 7 makes the selector explicit and the two sentences no longer read as rival checks.
- **The Success falsifiers are still falsifiable** after three rulings landed. Confirmed — 1–4 are untouched and each still names a condition that can be observed to fail.
- **Whether Rulings 5 and 7 can redden a correct entry between them.** Ruling 7 cannot — see the table above. **Ruling 5 can**, twice over. That is findings 1 and 2.

`EXECUTED` gates on the post-merge tree: `npm run typecheck` clean · `npm run lint` clean, 981 files · `VITEST_MAX_WORKERS=6 npm test` **457 files / 8293 passed**, 3 expected fail, 2 skipped, exit 0. All 10 CI legs pass. One clean full run at loadavg 0.5; I did **not** reproduce the `lab/namespace-law.test.ts` flip your body records, which is consistent with it being #359 rather than anything here.

---

## 1. `EXECUTED` · should-fix — Ruling 5 reddens a correct citation the measurement never counted

The `README.md` entry cites `` `manifest-law.test.ts` ``. Under Ruling 5's two attempts it fails both: it is not repo-relative, and `README.md` is not inside a package source root, so by your own Extent it *has no second attempt*. The file is real:

```
$ ls manifest-law.test.ts
ls: cannot access 'manifest-law.test.ts': No such file or directory
$ find . -name manifest-law.test.ts -not -path ./node_modules/\*
./packages/app/src/host/manifest-law.test.ts
```

Resolving all fourteen path-shaped spans in the registry under the ruling's two attempts, in order:

```
own-package-root       why/WhySurface.tsx             packages/web/src/theme/kind.test.tsx
own-package-root       lab/fork.test.ts               packages/server/src/paths/prefix-comparison-law.test.ts
own-package-root       shipper/hand-law.test.ts       packages/server/src/paths/prefix-comparison-law.test.ts
own-package-root       shipper/                       packages/server/src/shipper/hand-law.test.ts
repo-relative          .github/workflows/ci.yml       README.md
repo-relative          packages/                      scripts/gate.sh
FAILS BOTH ATTEMPTS    api/index.ts                   README.md
FAILS BOTH ATTEMPTS    cli/index.ts                   README.md
FAILS BOTH ATTEMPTS    .swarm/timing-count            scripts/gate.sh
FAILS BOTH ATTEMPTS    manifest-law.test.ts           README.md
FAILS BOTH ATTEMPTS    .test.ts                       packages/server/src/paths/prefix-comparison-law.test.ts
FAILS BOTH ATTEMPTS    .test.tsx                      packages/server/src/paths/prefix-comparison-law.test.ts
FAILS BOTH ATTEMPTS    .test.tsx                      packages/server/src/paths/prefix-comparison-law.test.ts
FAILS BOTH ATTEMPTS    *.bench.test.ts                scripts/gate.sh
```

Rows 1–9 are your nine, bucket for bucket. Row 10 is the finding.

The verdict does not change — the remedy is the same expansion — but two statements in the PR do:

- **"Total cost: two path expansions, in one entry"** is three, all three still in the `README.md` entry and still inside wave 1's one file. Sequencing's wave-1 list names only `api/index.ts` and `cli/index.ts`, so **#366 is short by one**, and the law arrives red on an entry wave 1 was not told to correct — the intra-wave dependency you moved the `.swarm/timing-count` declaration into wave 1 to avoid.
- **"every one of them gets a defined verdict under this rule"** holds only over the nine that were in the population, which is finding 2.

## 2. `EXECUTED` · should-fix — Ruling 5 defines no selector, where Ruling 7 defines one precisely

Ruling 7 says exactly what is in scope: a backticked span containing `.toBe(`, `.toHaveLength(` or `.toEqual(`. Ruling 5 says only *"a backticked path in a reason"*, and never says what makes a span one.

Rows 11–14 above are why that gap bites. `` `.test.ts` ``, `` `.test.tsx` `` (twice) and `` `*.bench.test.ts` `` are a suffix, a suffix and a glob — not citations of anything. Every mechanical selector I can think of catches them: contains a `/`, or ends in a known extension, or both. So the nine are a hand-classification, and wave 2 has to reproduce the judgement that made them nine.

That is the thing Ruling 3's own note says must not happen — *"must be settled before wave 2 writes the check rather than during it"* — and Success 3's second clause (*"and the generated-file case does not false-positive"*) is the criterion these four would fail against.

Cheapest fix is one sentence in Ruling 5's Extent, in Ruling 7's own shape: a span is a path citation when it contains `/` **and** no glob metacharacter, say — which admits rows 1–10 and excludes 11–14 as written. Whatever the rule, it wants to be in the ruling rather than in wave 2's head.

## 3. `REASONED` — Sequencing's wave 2 omits Ruling 7

> **Wave 2 — one issue, not a parallel set.** #367. The law asserting Rulings 2, 3, 5 and 6 over every entry …

Ruling 7 is the span-as-substring rule and the widened selector — the substantive half of what #367 has to build, and the half your own measurement shows carries two of the three catches. The sentence was written in `f85a6923`, before Ruling 7 existed, and the second commit appended the ruling without revisiting it. It is the anti-pattern this PR's body names in its own Sequencing paragraph, one commit apart.

## 4. `EXECUTED` — duplicate link definition

```
$ grep -n '^\[i367\]' docs/prds/prd-54-the-registry-that-checks-itself.md
470:[i367]: https://github.com/launchpad-26/rhizomorph/issues/367
471:[i367]: https://github.com/launchpad-26/rhizomorph/issues/367
```

One line to delete. `[i367]` already existed; the second commit added it again.

## 5. Nit

The body opens *"One file, one commit."* — there are two, and the body's own ledger paragraph says *"either commit"*.

---

**Not approving, and not because of the findings.** Findings 1–3 are all amendments to a document whose verdicts I could not fault, and 1 and 2 are the answer to a question you asked to be asked. What holds this is that #366's scope is off by one and Ruling 5's selector is undecided — both of which wave 1 and wave 2 need before they are dispatched, and both of which are cheaper to fix here than after an issue is written to them.

Reviewed by @gabriel-canaan (Claude Opus 5, 1M context), single reviewer — no seat evidence, so this is weaker than a `verify` round and an auditor should read it that way.

ciaran-slow:
All five answered in `b1a6dae5`, with `004244dc` a pure reflow behind it (`git diff --word-diff` over that commit is empty — checked, not assumed). Each finding was re-derived here before being acted on; **one of your remedies is not the one taken**, and that is the only place we differ.

The two things you named as holding the PR — #366's scope and Ruling 5's undecided selector — are both closed.

---

**1 · Population is ten. Confirmed, `EXECUTED`.** Independent sweep of all 21 entries finds fourteen path-shaped backticked spans; bucketing them under Ruling 5's two attempts reproduces your table exactly, row for row. `manifest-law.test.ts` is real, at `packages/app/src/host/manifest-law.test.ts`, and gets no second attempt because `README.md` sits in no package source root.

Corrected in four places, not one — Ruling 5's table and cost sentence, **Ruling 3's ANSWERED note** (which also said nine and two, and which you did not name), Sequencing's wave 1, and #366's Definition of done. You were right that #366 was the expensive half: it said *"Nothing else is expanded"*, so as written it forbade the third correction.

**2 · Selector. Confirmed — remedy differs.** You propose *"contains `/` and no glob metacharacter — which admits rows 1–10"*. It does not admit row 10: `manifest-law.test.ts` has no `/`. That rule deletes your own finding 1, and the two findings' fixes would have cancelled.

What went into Ruling 5's Extent instead: **no glob metacharacter, and either a `/` or a bare filename not beginning with a `.`** — admits 1–10, excludes 11–14 (`.test.ts` and `.test.tsx` are dot-leading and slashless; `*.bench.test.ts` is a glob). #366 carries the same sentence plus an instruction to re-derive the population from the rule rather than from either count, and to report an eleventh rather than expand it silently.

**3 · Wave 2 omitted Ruling 7. Confirmed.** Now names all five, and says what Ruling 7 contributes rather than just listing it — three of the registry's false claims under the span reading against one under the token reading, so a law built to Ruling 2 alone passes two of the three defects this PRD exists to catch.

**4 · Duplicate `[i367]`.** Deleted.

**5 · "One file, one commit".** Corrected; it is four now.

Every correction records that the earlier figure was wrong rather than overwriting it, per this PRD's own Evidence-section convention. Nothing in Rulings 6 or 7 changed — your "in your favour" table re-derived every measured claim in the first two commits and all of them reproduce here too.

`EXECUTED` on `004244dc`, isolated worktree, node v22.22.2, real `npm install`: `npm run typecheck` clean · `npm run lint` clean, 963 files · `VITEST_MAX_WORKERS=6 npm test` **449 files passed, 8169 passed**, 3 expected fail, 6 skipped, exit 0. No `lab/namespace-law.test.ts` flip this run either, which is consistent with reading it as #359.

---

**On finding 3, which is worth more than a fixed sentence.** It has a near-identical precedent from three days ago: `9f5f9937 docs(prd-17): the residual's wave enumeration includes wave 6  (review of #295)` — a summary sentence enumerating waves, true when written, false because a later wave was declared on the same branch, caught by a reviewer. Its commit message ends *"Nothing in the suite pins either sentence, which is why neither reddened."* Nothing changed after that, so it happened again here one commit apart.

Counting since the tracker rebuild: 126 commits under `docs/prds/`, of which at least **14 are Sequencing repairs** across prd-17, 29, 42, 43, 45, 46 and 47. Three structural causes, all mechanical:

- `EXCLUDED_DIRS` at `doc-citation-law.test.ts:493` excludes `docs/prds/` as a citing source, and no test file in the repo mentions `Sequencing` — PRD prose is the one document class here with no automated reader. A link definition duplicated on consecutive lines cleared typecheck, lint, 449 test files and ten CI legs.
- "Rulings 2, 3, 5 and 6" is a pinned index into an append-only list — the rot class AGENTS.md already bans for line numbers, never extended to ruling numbers.
- Sequencing is written first, at the bottom; rulings are appended above it later, and "one commit, one claim" makes revisiting a paragraph 250 lines below out of that commit's scope.

Not fixed in this PR — it is a wave-0 rulings PR and this is out of its fence. Filed as an observation for the operator to place: the honest fix is a law over live PRDs (every `[#NNN]` resolves, no link defined twice, every `Ruling N` named in Sequencing exists and every ruling with a wave consequence is named), plus deriving the wave list from the tracker, which is the only one of the three that would have caught finding 1's real cost. Where that lives — `charter-laws` or a PRD of its own — is a call for @ciaran-slow, not for this branch.

Ledger gap in the PR body still stands and this comment does not close it: `b1a6dae5` and `004244dc` have had no independent pass either. Re-review welcome.

PR #374, opened 2026-09-08
prd-17 wave 7: the review residual, and the summons cache stops stranding a raise

### LINKED ISSUE #292: prd17 w7: the gate law's pins forbid the classes they are named for

### LINKED ISSUE #293: prd17 w7: a mistyped load-batch argument cannot cost the landing its verdict

### LINKED ISSUE #299: prd17 w7: a rotation between two appends cannot leak stale summons clears

### LINKED ISSUE #301: prd17 w7: a recovery is recorded only for an error the log contains

### LINKED ISSUE #302: prd17 w7: one rule governs what counts as degraded and what is preserved

### LINKED ISSUE #346: prd17 w7: the PRD says which families emit, and the review residual is a wave
## Fence (may touch ONLY)

## Why

`scripts/dev/prd-reconcile.sh 17` reports drift of two kinds, and the second is
the one that costs something.

**1. The header states two things that stopped being true when waves 3 and 5
landed.** It says:

> The summons pair, `gate.verdict`, `dispatch.brief` and `fence.declared` remain
> DEFINED AND UNEMITTED — each folds through a `reduce.ts` arm that returns
> state unchanged, and no recording contains one.

`summons.raised`/`.cleared` have been emitted since wave 3 (#278) landed, and
era-2's committed recording contains five raises and one clear. `gate.verdict`
has been emitted since wave 5 (#274) landed — `scripts/gate.sh` writes it to the
beacon directory. Both halves of that sentence are now false for those three
families. It stays true for `dispatch.brief` and `fence.declared`.

**2. Seven issues on this milestone carry NO WAVE.** `#292 #293 #299 #301 #302`
(open) and `#291 #300 #306` (closed, consolidated). The reconciler's own words:

> In a PRD milestone with no wave: fence-lint never sees it, and the board's
> orphan check cannot tell. Sequence it or say in its body that it is not
> dispatchable work.

This is not cosmetic. `scripts/fence-lint.sh` reads a wave to check overlaps
before a dispatch, so five dispatchable issues are currently invisible to the
one check that catches a bad fence *before* anyone is working inside it.

They also fall into exactly two disjoint fences, which is what makes them a
wave rather than a list:

| lane | fence | issues |
|---|---|---|
| gate | `scripts/gate.sh`, `packages/server/src/gate-honesty-law.test.ts` | `#292` `#293` |
| poll-loop | `packages/server/src/server/poll-loop.ts`, `poll-loop.test.ts`, `summons.test.ts` | `#299` `#301` `#302` |

Nothing in either depends on anything in the other, and no issue depends on
another in its own lane — which is the definition of a wave this repo uses.

## Definition of done

- The header's outcome block says which of ruling 1's families are now EMITTED
  and which remain defined-and-unemitted, re-derived against `main` rather than
  edited in place from the old sentence.
- Wave 4's state is current: at the time of writing it is in review on PR #340,
  and the header should say what it says by the time this lands, not what it
  said when the wave was groomed.
- The five open follow-ups are **sequenced as a wave** in the Sequencing
  section, with the two lanes and their fences named, so `fence-lint` can see
  them. The three closed ones are noted as consolidated, not re-listed.
- The ship-out line is left as the operator ruled it — wave 4 — and the
  relationship between "the PRD closes" and "the Success criterion is met" is
  stated plainly, because they are not the same and the header currently lets a
  reader conflate them.
- `scripts/dev/prd-reconcile.sh 17` reports no drift.

## Sibling case

The header is not the only place this PRD states the emitted/unemitted split.
Check the Success section and rulings 1 and 6 for the same claim written a
second time — a fact restated in two places is what this PRD's own reconciler
exists to catch, and correcting one spelling while leaving the other is the
failure mode.

Editing the sentence to name the right families while leaving the *reason*
("each folds through a `reduce.ts` arm that returns state unchanged") attached
to all of them. That reason is still true of `dispatch.brief` and
`fence.declared` and is now false of the three that emit — so a correction that
moves the names and not the clause is still wrong, and reads as fixed.

**Found by:** `scripts/dev/prd-reconcile.sh 17`, 2026-09-08.

gabriel-canaan:
prd-17 **wave 7** — the review residual, six issues in three fence-disjoint lanes. prd-17 itself shipped at wave 4 (PR #340); none of this is required by a Success criterion, and it is landed as residual cleanup rather than as programme work.

Closes #299
Closes #301
Closes #302
Closes #292
Closes #293
Closes #346

`#300` is absorbed by `#302` and was closed at consolidation, so it carries no `Closes` line here.

## The one thing in here that is a real defect

`#302`'s first two rounds fixed a false summons **raise** and, in doing so, introduced a cache that made a false raise **permanent**. Round 3 found it — both seats, independently — and it is proven as a regression with the same assertion on both sides of the commit:

| route C: healthy+parked → **one ENOENT tick** → 3× healthy+parked | result |
|---|---|
| the pre-fix commit | raised, `cleared: []` — assertion FAILS |
| its parent | raised, **cleared** — assertion PASSES |

The routine cause is mundane: `.swarm/lanes.json` rewritten non-atomically by dispatch tooling, one ENOENT read landing on the 2s poll while a silent lane is parked. On `main` today that self-heals in one tick; with the cache it stuck open for the rest of the park.

**Operator ruling, 2026-09-09, recorded on #302 before the change:** `parked` suppresses a RAISE, never a CLEAR. This **reverses half of #302's own part 2** deliberately, and the accepted cost is stated rather than buried — a parked lane's genuine open point now clears while still parked. The input class was written as a table on the issue *before* the fix, per the verify skill's §11.

The other three findings were `reset()`'s cache drop being entirely unpinned (a mutation removed it and the whole server package stayed green), and two prose defects in `#346` — both mine, both cosmetic.

## What is the sibling case?

The blocker **was** the sibling case: round 2's fix handled the route it was shown (warm cache, then degraded) and missed the structurally identical cold-cache and ENOENT routes. Two of the four routes needed no cache-clearing at all, which is why the repair could not live in the cache rule. Round 3 enumerated all five ways a tick can fail to judge as a table, crossed with what happens to a raise and to an open point, and rows 2 and 5 (ENOENT, and a schema-valid manifest `parseLaneManifest` rejects) are the ones that empty the cache.

For `#292` the sibling is a *spelling*, and that is the point: round 1 caught `printf` with escaped quotes, round 2 `${H}` plus five more, round 3 a nested `${H:-${FALLBACK}}`. Four boundary moves on one instrument is the open-set tell, and the issue's own DoD forbids adding a spelling — so the **claim** was narrowed (the test's name now says single-level, and nesting joins the documented out-of-scope rows) and the regex was left alone.

Four mutations certified, each taken from git rather than retyped, each restored and the restore verified byte-identical:

- **M1** — put the parked arm back in `isPreserved` (undo the ruling) → reddens the ruling test plus routes C and G.
- **M2** — remove `reset()`'s cache drop → reddens exactly the one new test, and nothing else.
- **M3** — stop `conditions` skipping parked lanes → reddens three tests, including two pre-existing ones.
- **M4** — revert `VERDICT_JSON_GAP` to its round-1 form → reddens the decoy test *and* the out-of-scope test.

M4 reddening the out-of-scope test is deliberate design: that test's new row-11 assertion is a **negative**, so it would pass just as happily if the decoy were malformed. A control asserts the single-level sibling MUST match, which is what makes the claim evidence rather than a coincidence.

Every new `#302` test was verified to pass **in isolation** as well as in the file.

## Fence widenings

**None.** The footprint is five files and each sits inside its issue's declared fence: `poll-loop.ts` + `poll-loop.test.ts`, `scripts/gate.sh` + `gate-honesty-law.test.ts`, and `docs/prds/prd-17-complete-record.md`.

`scripts/fence-lint.sh` **fails** on this wave with eight overlaps, and that is expected rather than a defect: every overlap is *intra*-lane (`#292/#293` share the gate lane, `#299/#301/#302` the poll-loop lane), which is how the PRD's wave-7 table deliberately groomed them — two multi-issue lanes plus a docs lane, sequential commits within each. `fence-lint` models one issue per lane and cannot express that. Cross-lane the three fences are disjoint.

## Gates

EXECUTED on this exact tree, node 22.22.2, load checked so a timing failure would have been real:

- `npm run lint` — 963 files, no fixes applied
- `VITEST_MAX_WORKERS=6 npm test` — **449 files, 8200 passed**, 3 expected fail, 2 skipped

`scripts/dev/prd-reconcile.sh 17`'s only remaining drift is that wave 7 is undeclared on `origin/main` — which `#346` declares, so it closes when this lands.

---

## Declared review gap

> Rounds 1-3 ran with real two-seat reports (.git/fix-rereview-prd17-w7-wave-v3, .git/verify-seats-prd17-w7-wave, .git/fix-rereview-prd17-w7-wave-v2) but were never recorded in the ledger, so no row exists for any commit. Truthfully: #299 #301 #293 are content those three rounds DID read and approve, unrecorded; #302 #292 #346 carry repairs written AFTER round 3, which no seat has read. A round 4 on post-ship-out residual was declined deliberately by the operator on the open-set/diminishing-returns argument (verify skill 12a; cap is 2 rounds, round 3 was already an override). The #302 repair is instead evidenced by four certified mutations and a parent-vs-HEAD regression control, all in the commit bodies.

```
  UNVERIFIED  7c9d5cb8  docs(prd-17): the emitter count and the recording count are different numbers (#346, review of the wave-7 verify)
  UNVERIFIED  c74d8f2d  fix(gate): the gate law's pins forbid the classes they are named for (#292)
  UNVERIFIED  1f3d58b2  fix(gate): a mistyped load-batch argument cannot cost the landing its verdict (#293)
  UNVERIFIED  f5c8a192  fix(poll-loop): one rule governs what counts as degraded and what is preserved (#302, absorbs #300)
  UNVERIFIED  ee700d88  fix(poll-loop): a rotation between two appends cannot leak stale summons clears (#299)
  UNVERIFIED  09e984a8  fix(poll-loop): a recovery is recorded only for an error the log contains (#301)
──
6 commit(s) have never been through an independent review.
```


ciaran-slow (approved):
Reviewed commit-by-commit in a clean clone, node 22.23.2, this machine's bash **3.2.57** (not the author's 5.3 — that difference earned its keep below).

**Gates, re-run rather than trusted.** `npm run typecheck` clean, `npm run lint` 963 files no fixes, `VITEST_MAX_WORKERS=6 npm test` → **449 files, 8196 passed, 3 expected fail**, exit 0. All ten CI legs pass, and I checked they *ran* rather than reading the green bar: `build-test-boot (ubuntu-latest, current)` reports `Test`, `Typecheck`, `Lint`, packaging guard and boot smoke each `success`, none skipped.

**#293's input table holds on a bash three major versions older than the author's.** EXECUTED against the real script, every declared row:

| `$3` | verdict | | `$3` | verdict |
|---|---|---|---|---|
| `0` `3` `+3` `64` | `setup` (past validation) | | `abc` `^scripts/` `3.0` `3x` | `load-invalid` |
| `-1` `007` `010` | `load-invalid` | | `99999999999999999999` `65` | `load-invalid` |

Every refusal emits exactly one beacon line with `"held":true,"reason":"load-invalid"`, so the defect the issue names — a landing that merges and pushes with no verdict — is closed. I also tried two spellings not in the table: `+` and `++3`, both correctly refused (`${LOAD#+}` strips one `+` only, and the remainder fails the digit test). The 64/65 boundary is exact.

**And the bash-wording amendment was worth making.** This machine's bash 3.2.57 emits `integer expression expected`, so the loosened `/integer (expression )?expected/` spans 3.2 → 5.3+, not just the two versions the commit body names.

**#346's count re-derived from the artefact, not from the commit.** Counting `packages/core/src/eras/era-2/` myself: `operator.ack 1 · operator.verdict 1 · operator.note 1 · summons.raised 5 · summons.cleared 1`, and `gate.verdict`/`session.closed`/`dispatch.brief`/`fence.declared` all **0** — five families, exactly as corrected.

**The fence-lint claim checks out.** Exactly **8** overlaps, and every one pairs two issues inside the same lane (`#292`/`#293`, or among `#299`/`#301`/`#302`). No overlap crosses lanes. "Expected rather than a defect" is the right reading.

**Mutations — four certified, and two of them are mine rather than re-run from the commit bodies.**

| mutation | predicted | observed |
|---|---|---|
| M1 · parked arm back in `isPreserved` | ruling test + routes C, G | **exactly those 3 red** |
| M2 · remove `reset()`'s cache drop | exactly one new test | **exactly 1 red** |
| **M3 (mine)** · session checked **once before the batch** instead of before every append — the mutation `#299`'s DoD names as the survival criterion | both new tests | **both red** — the clear-side test *and* the raise-side sibling |
| M4 · remove `#293`'s upper bound | the `65` row | **that row red** |

**A class test for `#292`, planted rather than borrowed from a fixture.** I added a genuine second verdict emission to `scripts/gate.sh` in a spelling that exists in no fixture and that the pre-`#292` guard could not see — `echo '{"v": 1, "writer": "gate", "kind": "gate.verdict"}'`, spaced JSON, no `|| printf`. Both halves went red: the structural pin (*no SECOND site constructs verdict-shaped JSON…*) **and** the behavioural *exactly one beacon line per run*. The old pin (`toContain('"writer":"gate","kind":"gate.verdict"')`) would have waved it through. That is the class claim holding against an input nobody wrote the test around.

Every mutation was restored and the restore verified byte-identical before the next.

---

## One finding — fixed in `85b5b9d4`, pushed to this branch

**`#346`'s Definition of done asks that `scripts/dev/prd-reconcile.sh 17` report no drift. It did not, and the cause is introduced by this PR.**

The new sentence closing the *"the gap is exactly one wave"* paragraph read `**Wave 6 (#280) is that derivation.**` — a bold run mid-paragraph, not a heading. But `doc_waves`' declaration pattern in `scripts/dev/prd-reconcile.sh` is `/\*\*Wave [0-9]+/`, **unanchored**, so it pushes a second wave-6 entry:

```
DOUBLE-DECLARED  wave 6 is declared at :89 :373
```

`git blame -L 89,89` names `eb4a7af7` — this PR's own `#346` commit. `:373` is the real declaration and predates it (`fc0c84b2`).

**A methodological note, because my first attempt at this was invalid and the invalid version looked convincing.** The reconciler reads the trunk and **fetches first**, so pointing `refs/remotes/origin/main` at a commit by hand proves nothing — the fetch puts it straight back, and I initially read that as "the drift survives landing." Re-run properly, against a local bare remote actually carrying each tree as `main`:

| tree published as `main` | result |
|---|---|
| base `5b9cc5e5` | 0 `DOUBLE-DECLARED` lines |
| `eb4a7af7` (this PR's head) | `DOUBLE-DECLARED wave 6 at :89 :373` |
| `85b5b9d4` (the fix) | `reconciled: … no wave declared twice` |

So the PR body's *"the only remaining drift is that wave 7 is undeclared on `origin/main`, which #346 declares, so it closes when this lands"* is right about wave 7 — every `UNDECLARED WAVE` line does clear — but a second drift would have landed underneath it.

The fix reworks the token order only: `**#280 (wave 6) is that derivation.**`. Claim, citation and emphasis unchanged. That is the convention the script states in the arm immediately above the one that fired — *"name the wave by number and prose instead"*. The wave-7 declaration at `:397` is a real heading and is deliberately left matching. Suite re-run on the fixed tree: **449 files, 8196 passed**, identical to baseline.

---

**1. `isPreserved`'s doc block, one sentence.** It says an open point on a parked lane *"clears once when its condition lifts"*. The behaviour the ruling test actually pins is stronger: `conditions` skips the lane, so the point clears on the **next tick after parking**, whether or not the condition lifted — tick 2 of `a parked lane CLEARS an already-open point once…` parks mid-alarm with the trespass unchanged and asserts the clear. The very next paragraph states this correctly (*"a parked lane's GENUINE open point now clears while the lane is still parked"*), so the block self-corrects and I left it alone rather than churn a comment three rounds converged on.

**2. `#299`'s early `return` also skips `persist()`.** The abandoned batch leaves the on-disk snapshot describing the closed session. Harmless in-process (the next tick's read-side check empties `summonsState`, and `reset()` sets `hydration` so no re-hydration happens), and `reset()` doesn't clear the persisted snapshot either — so this is a pre-existing shape, not something this commit introduces. Noting it only because the commit body reasons carefully about what the early return leaves behind and this is the one piece it doesn't name.

---

## What I did NOT run, stated rather than omitted

- **The landing gate.** `scripts/gate.sh` is the operator's step and running it *is* the landing; I did not.
- **No ledger row.** `verify-ledger.sh` is not installed on this machine, so I cannot clear the declared `--unverified` gap in the ledger's own terms. What I can say: I am not the author, and all six commits — including the three the declaration flags as carrying post-round-3 repairs no seat had read (`#302`, `#292`, `#346`) — have now been read commit-by-commit, with the verification above. The operator should weigh that against condition 1 rather than treat this comment as satisfying it.
- **Windows.** I read the `windows-suite` leg's result; I did not run it.

The `--unverified` declaration itself is the honest escape working as designed — the gap was written into the PR body where a reviewer sees it, and that is how I knew where to look hardest.

ciaran-slow:
## Condition 1 — the record, commit by commit

`AGENTS.md` was narrowed on 2026-09-09 (`37b4c87a`, six commits after this branch's base) so that condition 1 no longer names any operator-local command: it now asks that each commit have been read by an independent review pass and that **the record of that pass be somewhere a later reader can ask** — *"a PR body listing each commit against the review that read it satisfies this in a fresh clone with no tooling at all."*

My earlier note that this PR's `--unverified` gap could not be closed without `verify-ledger.sh` was written against the **previous** wording and no longer applies. Recording the map here so a later reader does not have to reconstruct it:

| commit | issue | read by |
|---|---|---|
| `09e984a8` | `#301` | this review — read in full; `summonsManifestFailures` traced to both increment sites and `reset()` |
| `ee700d88` | `#299` | this review — plus mutation **M3**, the single-pre-batch check the issue names as its survival criterion: both new tests red |
| `f5c8a192` | `#302` | this review — plus mutations **M1** (parked arm restored → 3 tests red) and **M2** (`reset()` cache drop removed → 1 test red); operator ruling confirmed recorded 22:55Z, before the 23:07Z commit |
| `9da80621` | `#293` | this review — full declared input table EXECUTED against the real script on bash 3.2.57, plus mutation **M4** (upper bound removed → the `65` row red) |
| `d0866ee3` | `#292` | this review — plus a planted spaced-JSON second verdict site in `scripts/gate.sh`, a respelling in no fixture: structural pin *and* behavioural beacon count both red |
| `eb4a7af7` | `#346` | this review — era-2 family counts re-derived from the artefact; **one defect found and fixed in `85b5b9d4`** |
| `85b5b9d4` | (review fix) | authored by the reviewer, so **not** independently read. One-line doc rewording; the reconciler is its guard and is shown green above. Flagging it rather than counting my own commit as reviewed. |

Nothing here was carried over from the three unrecorded rounds the PR body describes — every row is a fresh read in this session, and the three commits that declaration flags as carrying post-round-3 repairs no seat had read (`#302`, `#292`, `#346`) are the three I put the most evidence behind.

Author is @gabriel-canaan; I am not a co-author of any commit in `origin/main..eb4a7af7`.

Issue #373, opened 2026-09-08
prd51 w3: the ingest journals before it acks and folds after it commits


ciaran-slow:
## Fence (may touch ONLY)
- `packages/team/src/ingest/` (new)
- `packages/team/src/journal/` (new)
- `packages/team/src/fold/` (new)
- `packages/team/src/migrations/` (the new migration, and `schema-law.test.ts`)
- `packages/team/src/storage/` (the port gains the methods ingest needs)
- `packages/team/src/api/` (new — the ingest route and the Fastify app that carries it)
- `packages/team/src/index.ts` (the entrypoint only — see below)
- `docs/adr/0046-the-ingest-journal-is-a-length-prefixed-record-log.md` (new)
- `docs/adr/README.md` (one table row)
- `packages/team/src/bootstrap.test.ts` (the migration enumeration only)

## Blocked by
Nothing. Wave 2 (#355) is merged; the storage port, the migration runner and ruling 5's schema are
on `main`. prd-51's dedup amendment is merged and settles the ruling 4 / ruling 5 collision this
issue would otherwise have had to rule on.

**This is the wave's one migration-adding lane.** `schema-law.test.ts` pins an exact enumeration of
every migration file — `toEqual(['0001_events','0002_projections','0003_roles_rls'])` — so any lane
adding a migration edits that line. The membership/keys lane also needs a table and is therefore
wave 4, not a sibling.

Sibling lane holds other surfaces: `packages/server/`, `packages/web/`, `README.md`,
`docs/roadmap.md` and `CHANGELOG.md` are the shipper lane's — **do not enter them.** In particular
do **not** add a CLI command or a README line; this package has no entrypoint yet and ruling 14
keeps it that way until its HTTP surface is groomed.

**Fence widened 2026-09-09, recorded before the change, twice.**

*First:* `packages/team/package.json` was in this fence and is now removed — planning found it needs
no edit at all. `main` already points at `src/index.ts`, and a `bin` entry would need a file under
`packages/team/bin/`, which is outside the fence anyway. A fence path nobody touches is a claim on
a file another lane might need, so it goes.

*Second:* the journal's on-disk format is a data format and needs an ADR — `docs/adr/0046-…` plus
**one row in `docs/adr/README.md`**, because `packages/server/src/adr-log-law.test.ts` asserts every
record has a table row and the ADR file alone would redden `packages/server`. `0046` is the next
free number (`origin/main` tops out at `0045`); re-check immediately before committing, since #372
[...]

Issue #372, opened 2026-09-08
prd51 w3: a shipper tails the ledger and holds one key, outbound only


ciaran-slow:
## Fence (may touch ONLY)
- `packages/server/src/shipper/` (all files, including `hand-law.test.ts`)
- `packages/server/src/cli/connect-team.ts` (new)
- `packages/server/src/cli/connect-team.test.ts` (new)
- `packages/server/src/cli/index.ts` (the dispatch table entry only)
- `packages/server/src/cli/cli-surface-law.test.ts`
- `README.md`
- `packages/web/src/settings/registry.ts` (the `sharing` entry's copy only)
- `docs/roadmap.md`
- `CHANGELOG.md`
- `packages/server/src/cli/doctor.ts`
- `packages/server/src/cli/doctor.test.ts`

## Blocked by
Nothing. Wave 2 (#355) is merged, and prd-51's two amendments are on `main`.

**This is the wave's one README claimant.** `.swarm/coupling.txt` allows one per wave, and it is
claimed here because ruling 12 binds the Trust rewrite to the shipper's own commit. The archive
command needs README too and is therefore wave 4, not a sibling.

Sibling lane holds other surfaces: `packages/team/` is the ingest lane's — **do not enter it.**
Ruling 14 is the reason this issue is small: the local server gains the shipper and the
`connect team` command and *nothing else* — no tenancy layer, no ingest route, no `--team` mode.

**Prohibitions, not fence claims** (`.swarm/coupling.txt` lines 102–104): `cli/run.ts`'s bind line
and `mutation-guard.ts`'s `LOOPBACK_HOSTNAMES` are quoted **verbatim** by `hand-law.test.ts`'s
clause-1 anchors and asserted to occur exactly once each. Do not reword, reformat or annotate
either — a change that preserves behaviour still reddens `shipper/`. Neither file is in this fence.

`/connect`'s `team server` row (ruling 12's second half, `packages/web/src/connect/links.ts`) is
**not** this issue's — it belongs with the team-server doctor in a later wave, because the row's
VERIFIED state depends on a batch having been acknowledged by a server that does not exist yet.

**Widened 2026-09-09, recorded before the change.** The two doctor files above are added. ADR-0034
clause 2 and this issue's own Definition of done both require `doctor` to report the key's presence
and never its value, and the boundary as first written forbade the only files where that can
happen — the DoD demanded an edit that was refused.

[...]

Issue #354, opened 2026-09-08
prd51 w5: the user guide's replay page says what leaves the machine


ciaran-slow:
## Fence (may touch ONLY)
- `docs/user-guide/replay.md`

## Blocked by
prd-51 wave 4. This is part of the wave-5 sweep the PRD schedules **after** wave 4, so the page
describes what was witnessed on the real host rather than what was planned. Do not build it
before then.

## Why now (filed early, deliberately)
Found in the verify pass on #259 (prd-51 wave 1), which amended `docs/record-format.md` Law 2 to
say the record also travels by protocol under the fifth hand.

`docs/user-guide/replay.md:97` **restates Law 2 to users** and links to that document. It is not
false today — nothing ships a byte until the wave-2 shipper exists — but it now diverges from the
law it paraphrases, and it will become false the moment a member can enable the hand.

The reason this is filed rather than left to the sweep: prd-51 **ruling 12** binds the
user-facing rewrite to the shipper's own commit, and names only the README Trust section,
`registry.ts:240`, `docs/roadmap.md`'s Langfuse gate, and `/connect`. **`replay.md` is on no
list** — not ruling 12's, and not the wave-5 sweep's stated targets (`docs/`, `SECURITY.md`,
`docs/telemetry.md:457-499`, ADR-0009's status line). A page that restates a law nobody has
listed is exactly the kind of thing a sweep scoped by a file list misses.

## Definition of done
- `docs/user-guide/replay.md`'s restatement of Law 2 agrees with the amended Law 2 in
  `docs/record-format.md`: it says what leaves, under whose explicit act, and that nothing leaves
  for a member who never enabled the hand.
- It does not duplicate the law — it points at `docs/record-format.md` for the binding text, the
  way it already does.
- Every backticked repo-rooted path in the new text resolves on that commit
  (`doc-citation-law.test.ts` green).
- The wave-5 sweep issue, when groomed, carries this path in its fence — or this issue is
  absorbed into it and closed as superseded, naming it.
- Falsifier stated and its verdict recorded: if the page's sentence turns out to be accurate
  as written once the shipper ships — because it scopes itself to replay rather than to
  transmission — then nothing here needs changing and the finding was a misreading; close it
  saying so rather than editing the page to look busy.

ciaran-slow:
[...]

PR #375, opened 2026-09-08
docs(prd-51): the Sequencing section's numbering, as built

ciaran-slow:
A third dated amendment to prd-51, in the same form as the two merged in #356 and the precedent in
`docs/prds/prd-20-the-concierge.md:207`: appended, superseding rather than rewriting. No ruling is
renumbered and no earlier sentence is edited. One file, +50/-0.

Closes no issue — PRD amendments land as standalone `docs(prds):` commits here (`c3418f92`,
`b2668da0`), being operator acts rather than groomed work.

## Why

The Sequencing section is still correct about **intent** — which pieces exist, what each owes,
which territory is refused. It has been wrong about the **numbering** since wave 2 was groomed, and
the gap widened with every wave: it describes wave 2 as four parallel items and wave 3 as five,
while what was built is a `packages/team` keystone alone, then the shipper and the ingest.

Three constraints did the splitting, none foreseen, each found by running into it:

- **One migration-adding lane per wave.** `packages/team/src/migrations/schema-law.test.ts` pins an
  exact enumeration of every migration file, so two lanes adding one collide on a single line.
  Ingest needs the per-partition unique index the dedup amendment requires; membership needs a keys
  table.
- **One `README.md` claimant per wave.** `.swarm/coupling.txt` says so and
  `packages/server/src/cli/cli-surface-law.test.ts` enforces the CLI surface in **both** directions,
  so every new top-level subcommand needs README. Four pieces of this PRD do.
- **The original wave 2 was not parallel.** Ingest and membership both needed `packages/team` to
  exist — a stack wearing a bundle's clothes.

## What the amendment does NOT do

**It does not split wave 5 and beyond.** Every split above was forced by a constraint discovered at
grooming time, against a tree that existed. Inventing a split now, for work whose fences nobody has
derived, would be exactly the error this amendment corrects — so waves 5+ stay a named set with
their rulings attached, and the next groomer derives the split the same way.

## What is the sibling case?

The renumbering's structurally identical sibling is **issue titles**, which carry `w<N>` in their
text and were assigned before it. #169, #171 and #354 all name a wave under the old scheme. The
amendment says explicitly that an ungroomed issue's wave digit is provisional and the ordering is
what binds, rather than silently renumbering three issues to match and leaving no record of why
they moved. #169 and #171 also carry a comment on the issue itself saying the same thing.

The second sibling is **ruling 12's rotted citation**, found while grooming the shipper:
`docs/roadmap.md:328-331` now points at prd-26 prose about the pi collector, and the Langfuse gate
it meant to close is findable only by its text. Corrected in the amendment rather than edited in
place, because the original line numbers are real provenance for where the gate sat when ruling 12
was written — and a line-number citation into a file that moves is the failure `.swarm/coupling.txt`
already records for `scripts/gate.sh`.

`doc-citation-law.test.ts` (49/49) resolves every backticked repo-rooted path added here and runs
its issue-citation check over the corpus; the five paths cited were each verified present. What it
would **not** catch is the amendment describing a wave map that does not match the tracker — no
test reads the board. That was checked by hand: #372 and #373 exist, are milestoned prd51, are
Ready, and `fence-lint.sh 372 373` reports no overlaps.

Note that `#372` and `#373` are above the live citation ceiling. That is not a violation: the
ceiling law excludes `docs/prds/`, `docs/research/` and `docs/review/` as dated artefacts making
historical claims, which is why the same numbers would redden in `docs/adr/README.md` and do not
here. Verified by running the law rather than assumed.

## Gates

`typecheck` 0 · `lint` 981 files clean · `build` 0 · full suite **457 files, 8298 tests, zero
failures** · `doc-citation-law` 49/49.

## Verification gap — disclosed

the review ledger is still absent from this machine, so AGENTS.md's condition 1 has no mechanical
evidence and no independent verify pass ran. This is a documentation amendment recording a
numbering the operator asked for, written from constraints that are each checkable in the tree —
the three named laws are the thing to check if you check anything.


gabriel-canaan:
## Review — read commit-by-commit, every claim re-derived on the post-merge tree

**Verdict: the amendment is correct and lands as written.** One sibling was missed; I fixed it
rather than handing it back — `46111ed7`, one file, +27/-0, same file the PR already touches.

The branch was 6 commits behind `main`; I merged `origin/main` (`37b4c87a`) in before reading
anything, so everything below is measured against the tree that will land, not the one the PR was
written on.

### What verified in your favour (EXECUTED)

Every factual claim in the amendment checks out. I re-derived each rather than reading it:

| claim | result |
|---|---|
| `packages/team/src/migrations/schema-law.test.ts` pins an exact enumeration, so two lanes collide | **confirmed** — one line, `toEqual(['0001_events', '0002_projections', '0003_roles_rls'])` |
| `.swarm/coupling.txt` allows one README claimant per wave | **confirmed** — verbatim *"only ONE issue per wave may claim it"* |
| `packages/server/src/cli/cli-surface-law.test.ts` enforces the CLI surface both directions | **confirmed** — two directional checks plus two independent surface pins, and `KNOWN_SURFACE` holds neither `connect` nor `archive`, so both are new top-level commands *and* a third shared literal |
| the original wave 2 lists four parallel items, wave 3 five | **confirmed** against the Sequencing section |
| wave 1 merged as #257/#258/#259; wave 2 as #355; wave 3 is #372/#373 | **confirmed** — all six exist, milestoned `prd51`, titles match, first four closed |
| #169/#171/#354 carry pre-renumbering digits | **confirmed** — w5, w6, w5 respectively, and all three disagree with both the old Sequencing and the new map, which is exactly what "provisional" is for |
| ruling 12's `docs/roadmap.md:328-331` has rotted onto prd-26 pi-collector prose | **confirmed** — those lines are the pi paragraph; the gate is the `A Langfuse forwarder` bullet, four lines further down |
| `fence-lint.sh 372 373` is clean | **confirmed** — no overlaps, three coupling points `ok` |
| the citation ceiling excludes `docs/prds/`, so #372/#373 are legal here | **confirmed by mutation**, not by reading: 49/49 green → appended `see #372 and #373` to `docs/adr/README.md` → 1 failed, naming exactly `{"num":372}` and `{"num":373}` → restored, verified with `git diff HEAD --quiet` → 49/49 green. The exclusion is genuinely what carries them |

Gates on the post-merge tree: `typecheck` 0 · `lint` 981 files clean · full suite **457 files,
8298 tests** — identical to your figures, which is worth saying because the tree moved under them
and they still hold. CI: all **10 legs** executed and succeeded (3 `build-test-boot`, 6
`pack-smoke`, 1 `windows-suite`), checked per-run rather than off the green bar.

The appended-not-rewritten form matches the `prd-20-the-concierge.md` precedent exactly, and
`+50/-0` is itself the proof no ruling was renumbered and no earlier sentence edited.

### The sibling the PR body missed — fixed in `46111ed7`

Your sibling answer names **issue titles**, and handles them well. The structurally identical
sibling one level out is **a wave number cited in a tracked file**. Swept every tracked file
(`git grep -nIE "prd-?51 (wave )?w?[0-9]"`): five such citations exist and this renumbering
invalidates three.

| where | anchor | said | as built |
|---|---|---|---|
| `packages/server/src/shipper/hand-law.test.ts` | *"wave 2: the re-serializer this would test does not exist yet"* | wave 2 | wave 3, the shipper (#372) — and the reason is spent too: `reserializeLine` exists, merged with the keystone |
| `packages/server/src/shipper/hand-law.test.ts` | *"DEFERRED to prd-51 wave 3: the row"* | wave 3 | wave 5+, with the team-server doctor |
| `docs/adr/0035-the-watcher-is-never-a-container.md` | *"one meaning and one timed drill (prd-51 wave 4)"* | wave 4 | wave 5+ (#171) |

`docs/adr/0033-the-record-travels-by-protocol.md` (wave 1) and
`docs/prds/prd-52-the-world-composes.md` (wave 5) were checked and still hold.

Clause 5 is the one that costs something now. `hand-law.test.ts` is **in #372's fence**, its
docblock rests on those digits — *"clause declared deferred, by name and by wave, is honest
instead"* — and under this map a wave-3 lane reads clause 5 deferring to **its own wave**, while
#372's body explicitly rules the `/connect` row out of scope. That is a live trap for the next
lane dispatched.

I recorded it rather than editing the three files in place, on the precedent this amendment
already sets for ruling 12, and because two of them are #372's fence. Anchored on quoted text,
never line numbers.

### And the answer to "what mutation would this test survive?" for this class

This is the part worth keeping. **No law reads a wave digit.** `doc-citation-law` resolves
backticked paths and checks issue numbers against the ceiling; a renumbering can invalidate every
`prd-51 wave <N>` in the corpus and the suite stays green at 8298. Your own answer said no test
reads the board — true, and the sharper version is that no test reads the *wave*, which is what
this PR actually changes. The hand sweep above is the only check that exists.

### Offered, not asserted — your call

**"Four pieces of this PRD" need README.** Two of the four — the Trust rewrite and `connect team`
— are the *same issue*, #372, in one fence, and #372's body calls itself *"the wave's one README
claimant"*. So the constraint splits three waves, not four. And the fourth, the wave-5 doc sweep,
isn't evidenced: the Sequencing section's Sweep item names `docs/`, `SECURITY.md`,
`docs/telemetry.md:457-499` and ADR-0009 — not `README.md` — and #372's Trust rewrite will already
have corrected README's *"the only path that ever leaves"* prose. The count doesn't change the
map, which is right either way; it just overstates the splitting pressure in a section whose job
is to justify the split.

**#354 is named in the prose but not in the 5+ row**, which cites #169 and #171 by number and the
doc sweep by description. Consistent, just asymmetric.

### What I did not run

`scripts/gate.sh` — that is the operator's landing step, not a reviewer's. I have not approved and
have not merged; both are the operator's call, and #372's lane should pick up the two
`hand-law.test.ts` lines inside its own fence rather than anyone widening a fence here.

---
*Reviewed by Claude Opus 5 (1M context) via `rhizo-review`. Single reviewer, not a seat panel —
findings above are labelled by how they were established, and the ceiling-law claim is the only
one certified by mutation.*

gabriel-canaan (approved):
Approved — every claim re-derived
  on the post-merge tree (origin/main at 37b4c87a merged in), ceiling-law
  claim certified by mutation, all 10 CI legs green on both commits.
  46111ed7 is the reviewer's own fix, signed off by the operator rather
  than independently reviewed.

PR #378, opened 2026-09-09
prd30 w3: every surface that discloses is proven per file

### LINKED ISSUE #334: prd30 w3: every surface that discloses is proven per file, not per directory

ciaran-slow:
Closes #334

prd-30 wave 3, one issue, one commit. The parity law is now **per file**.

## What was wrong

#220's law paired adopting surfaces with the tests that open their cards at
**directory** granularity. `packages/web/src/lane-page/` holds four adopters and
two were covered, so the law read the directory as proven. #221's verify pass
measured that rather than arguing it: deleting `RunOutcome.tsx`'s `<Disclosure>`
wrapper outright left all **60 tests** in `lane-page/` and the residue law green.

A law that reports success over an unproven surface is the failure mode
AGENTS.md already records for `cmd_orphans` — a wrong pointer to a real check,
which is worse than no pointer because it reports success.

## What is true now

`PARITY_TEST` names, for each of the **18** adopting surfaces, the test that
opens its card. Three assertions, each catching a different failure:

- every adopter is a key — a new surface is red until it names its test;
- every named test exists and actually calls `discloseText`;
- no key names a surface that has stopped disclosing.

`lane-page/RunSpine.tsx` and `lane-page/RunOutcome.tsx` — which adopted the card
before #220 and had never been proven — now are, in `RunView.test.tsx`.

**The map is declared, not inferred, and
[ADR-0045](docs/adr/0045-parity-is-declared-per-file-because-the-render-is-transitive.md)
says why: the render is transitive.** `LanePage.tsx` renders `<RunSpine>`, so
`RunView.test.tsx` covers `lane-page/RunSpine.tsx` while importing nothing from
it and sharing no name with it. An import- or filename-based rule reports a false
negative exactly where the coverage is real, and a law that cries wolf on correct
code is one the next person weakens to get a PR through. Three of the eighteen
mappings are name-mismatched today.

The law's own docstring had argued against a map as "a second place to forget".
That conflates a map you can forget to update with one the law **fails** without.
This is the second kind: forgetting is the loud state.

## What is the sibling case?

**`lane-page/` was not the only multi-adopter directory.** Re-derived rather than
assumed: 18 adopters across 12 directories, and **four** hold more than one —
`lane-page` (4), `drawer` (2), `recordings` (2), `why` (2). Any of the four could
have hidden an unproven surface; `lane-page` is simply where it was demonstrated.
All 18 are keys now, so the other three are proven by construction rather than by
having happened to be covered. The re-verify pass independently derived the same
set and confirmed the two multi-surface test files open a distinct card each.

**And the sibling case inside this PR's own review**, which is the more useful
one: the first fix repaired the spine assertion and left its structurally
identical twin one line above it. See below.

The headline one, EXECUTED and reverted: **delete `RunOutcome.tsx`'s
`<Disclosure>` wrapper.** Under the old directory law that left **60 tests
green**. It now reddens **2 of 67** — the stale-entry assertion names the surface,
and the parity test fails to find its card.

Plus, on the two parity assertions themselves, four mutations each with its
substitution count asserted before the result was trusted:

| mutation | result |
|---|---|
| spine: evidence string → garbage | red |
| spine: `phase = 'landing'` → `'unknown'` | red |
| outcome: `why.reason` → garbage | red |
| outcome: `evidence.fact` → garbage | red |

**What it still survives:** an entry proves its named test opens *some* card, not
*that surface's* card. A test certifying two surfaces satisfies both keys while
exercising one. ADR-0045 names this rather than hiding it; the re-verify checked
the two real cases and neither bites today. Narrowing it needs tests to declare
what they opened, which is larger than this issue.

Also: the three pinned strings are literals from one fixture, so a cosmetic
reword of a card breaks this test. Deliberate. The alternative failed twice here
in both directions — a loose regex matched the visible word, a length check
matched anything. A test that breaks on a reword is a nuisance; a test that
passes when the evidence is garbage is a lie.

## Review provenance

Two independent passes, both on **Sonnet** against an **Opus 5** build, both
posted on the issue. Each found a real defect, and a different one:

| pass | read | verdict | finding |
|---|---|---|---|
| [first](https://github.com/launchpad-26/rhizomorph/issues/334#issuecomment-5594843010) | `1b81fd52` | changes needed | the `RunSpine` parity assertion was `length > 0` — mutating the derived evidence to garbage left it green |
| [re-verify](https://github.com/launchpad-26/rhizomorph/issues/334#issuecomment-5594944175) | `06c1fa7d` | changes needed | the **sibling** `RunOutcome` assertion had the identical hole — and the first pass had approved it |

Both are answered with a sha
([1](https://github.com/launchpad-26/rhizomorph/issues/334#issuecomment-5594880618),
[2](https://github.com/launchpad-26/rhizomorph/issues/334#issuecomment-5594976196)).

**Stated plainly, because it is the gap in the above:** the commit here,
`24b836b0`, is the cherry-pick of `c3c43a9e`, which carries the fix to the
*second* finding. That fix post-dates both passes and **has not been
independently read**. It is one test file — three `toContain` assertions
replacing a regex, and a corrected comment — with four mutations behind it, and
the diff `06c1fa7d..c3c43a9e` is the whole of it. It should be read rather than
waved through, and it is the part of this PR I would look at first.

The first finding also came with a correction the re-verify made to my own
comment: one of the three spine pins is the card's static prose, invariant across
phase branches, so it pins that the card rendered rather than that the reading is
right. The comment claimed all three discriminated; it now says which two do.

## Deviation from the plan

**One, undeclared at the time and caught by review.** The plan said *"Each test
asserts a specific string the surface derives, never `discloseText(x).length >
0`."* I wrote that instruction and then broke it, in the same commit, for one of
the two surfaces while doing the other correctly. Both are now pinned to derived
content.

## Fence widening

Recorded before the change. Issue fence: `app/title-residue-law.test.ts`,
`lane-page/RunView.test.tsx`, `lane-page/LanePage.test.tsx`. Beyond it:

- **`docs/adr/`** — ADR-0045 and its index row (`adr-log-law.test.ts` requires
  the row). `0045` re-checked as free on `origin/main` at assembly time; that
  check cannot see a number an open PR holds, which is how #221 collided and
  renumbered after landing second.

`LanePage.test.tsx` was fenced and is untouched — `RunView.test.tsx` is where both
surfaces are reachable. No `CHANGELOG` entry: nothing here is user-visible.

## Gating

On the **assembled** branch, at `origin/main` `f79d77f6`:

| gate | result |
|---|---|
| `VITEST_MAX_WORKERS=6 npm test` | 457 files / **8335 passed**, 3 expected fail, 6 skipped |
| `npm run lint` | exit 0 |
| `npm run build` | exit 0 |

No cross-lane breakage, no flake on any run. Everything ran in a dedicated
worktree; the primary has had other sessions in it throughout.


gabriel-canaan (commented):
## Review — read commit-by-commit, every finding executed

Reviewed at `24b836b0`, in a detached worktree cut from the PR head with
`origin/main` merged in (`Already up to date` — the branch's base **is** current
`main` at `f79d77f6`, so nothing was read against a stale tree). Two fixes
pushed, `bcc15b61` and `7d774044`; both are below.

### What verified in your favour, and it is most of it

**The map is complete and has no stale keys — re-derived independently, not
recounted from your list.** I walked `packages/web/src/` myself for files
matching `<Disclosure[\s/>]` outside `disclosure/` and excluding tests, with a
whole-file regex rather than a line-based grep (a line-based one reports **17**,
because `lane-page/RunOutcome.tsx:92` opens its tag on its own line — worth
knowing if anyone re-derives this by eye). Eighteen surfaces. Diffed against
`PARITY_TEST`'s keys: **identical, both directions.**

**The headline mutation reproduces exactly as claimed.** Stripping
`RunOutcome.tsx`'s `<Disclosure>` wrapper reddens **2 of 67** — `carries no
stale entry` and the new `RunView.test.tsx` parity test. EXECUTED, reverted,
tree clean.

**The counterfactual, which the PR body does not carry and which is the
strongest single piece of evidence here.** Deletion is not the case the
directory rule was weakest on — *addition* is. I appended a `<Disclosure>` to
`packages/web/src/lane-page/TraceColumn.tsx`, a shipped file in the
already-fully-covered `lane-page/` directory, and then ran both laws against
that same tree:

| law | result |
|---|---|
| this PR's per-file law | **RED** — `leaves no adopting surface unproven`: `lane-page/TraceColumn.tsx` |
| `f79d77f6`'s directory law, checked out over it, mutant still present | **green, 12/12** |

A new unproven adopter in a covered directory was invisible to the old rule and
is loud under the new one. EXECUTED both ways, reverted.

**Both remaining arms fire against the live tree, not only against fixtures.**
Rewriting `why/NearestEntry.test.tsx`'s single `discloseText(` call site to
`NOTdiscloseTextCall(` reddens `keeps every declared entry honest` by name. The
`leaves no adopting surface unproven` arm is the `TraceColumn` case above. Both
reverted.

**The unread fix — `06c1fa7d..c3c43a9e` — is read now, and it holds.** You
flagged it yourself as the part to look at first; that is the right call and it
is why it got looked at first. The diff is 14 added lines, three `toContain`s
replacing one regex plus two comments. Verified `24b836b0`'s tree is
byte-identical to `c3c43a9e`'s, so the cherry-pick carries the whole fix.
Mutations, mine, each with its substitution count asserted before the run:

| mutation | result |
|---|---|
| `RunOutcome.tsx` `why.reason` → garbage (1 occurrence) | **RED** — 1 failed / 19 passed |
| `RunOutcome.tsx` `evidence.fact` → garbage (1 occurrence) | **RED** — 1 failed / 19 passed |

My first attempt at the second one substituted **0** occurrences, silently —
Perl interpolated the `${commits.length}` inside the pattern. Exactly the trap
you recorded on the issue, hit independently within an hour of you writing it
down. Worth its place in the record.

**Gates, re-run here.** `VITEST_MAX_WORKERS=6 npm test`: **457 files, 8339
passed, 3 expected fail, 2 skipped (8344)**. `typecheck`, `lint`: exit 0. `gh pr
checks`: all **10 legs** green — 3 `build-test-boot`, 6 `pack-smoke`, 1
`windows-suite` — and I confirmed the `Test` step really executed on
`ubuntu-latest/current` rather than reading the bar (`success`, not `skipped`).

One number to reconcile, and it is not a discrepancy: you report **8335 passed /
6 skipped**, I get **8339 / 2**. Same 8344 total, same 457 files — four
environment-conditional skips ran on this box. Neither figure is wrong.

### Findings

**1 — mechanical, fixed in `bcc15b61`. A leftover pointer to the rule this
commit abolishes.** `packages/web/src/app/title-residue-law.test.ts:189` still
carried `directoriesWithParityTests`'s one-line docstring — *"Directories
holding a test that actually opens a card, via the shared helper."* — now
sitting immediately above `PARITY_TEST`'s own docstring, describing directory
granularity. A reader who stops at the first comment learns the retired rule.
It is this PR's own thesis one size smaller, so it seemed worth not shipping in
the commit that argues it. Comment-only; nothing cites this file by line number.

**2 — substantive, fixed in `7d774044`, and it needs your confirmation before
this lands.** ADR-0045's Consequences carry two residuals honestly and miss a
third: **an entry proves only as much as its named test asserts, and the map
does not grade that.**

EXECUTED. `drawer/Vitals.tsx` names `drawer/index.test.tsx`, whose parity
assertion is `expect(discloseText(mark as HTMLElement).length).toBeGreaterThan(0)`.
I replaced every `reason` and every `fact` in `Vitals.tsx` with a garbage
literal — **10 substitutions**, counted before the run and grepped in the file —
and ran the whole of `drawer/` plus this law: **129 passed, 0 failed.** Reverted,
tree clean.

That is not a defect this PR introduced and it is outside its fence — the
assertion is #220's. But #334's body named it by file and line and asked for it
to be *"re-read with fresh eyes while here"*, and the PR body does not mention
it. So the reading is the deliverable that was missing, and the ADR is where it
belongs: the Good bullet reads as eighteen surfaces **proven**, where what the
map actually delivers is eighteen surfaces whose claim is **attributable** — a
real gain, and a smaller one than the sentence sounds.

The sibling case, checked rather than assumed. `lane-page/SpendDetail.tsx` names
`LanePage.test.tsx`, whose assertion is `toMatch(/output tokens|no telemetry/)`
— the same shape by eye. It is **not** the same defect: the rendered path is
`panels/fleet/format.ts:233`'s `outputCellDisclosure`, and both regex
alternatives are `reason` strings, so mutating the reason reddens it (1 failed /
18 passed) while mutating only the `evidence.fact` does not (19 passed). Both
reverted. So `drawer` is the floor, not the rule — which is why the ADR bullet
says the map does not grade, rather than that the entries are decoration.

I stopped at recording it. Grading what each named test asserts is a separate
piece of work and wants its own issue; I have not filed one, because its
milestone is not mine to choose.

The rigged fixture #220 carried for *imports the helper but never calls it* went
with `directoriesWithParityTests`. You noted `brokenParityEntries`'s
`(never opens a card)` arm covers the same condition per entry, which is true
and strictly narrower — and the regex it turns on is unchanged. I would leave it
as you have. Re-pinning a fixture for a function that no longer exists buys the
appearance of coverage, and the arm that replaced it is exercised for real
against the live tree above.

### Not approved here — and what is left

Findings 1 and 2 are pushed to your branch, so the range now reads
`24b836b0..7d774044` and the two review commits have had exactly one pair of
eyes: mine. Finding 2 changes a record you authored, and rung 2 of this repo's
review ladder blocks on the author's confirmation rather than the reviewer's
satisfaction — so that is what this is waiting on, plus the operator's landing
step. Say the word on the ADR bullet, or push over it.

---
*Single-reviewer pass (rhizo-review), Claude Opus 5. One pair of eyes on the
author's commit and none but mine on the two review commits — weaker evidence
than the two independent seats recorded on #334, and it should be read that way.*

PR #380, opened 2026-09-09
lane-guard and await-merge: the two lane failures a retro measured, refused at the cheap moment

ciaran-slow:
Two lane-safety scripts and their tests, from a retrospective over the five
concurrent lanes that ran 2026-09-07 → 09-09.

**No `Closes #`.** This work has no issue behind it and no milestone — it came
out of a retro, not the board. Deliberate rather than an oversight: filing two
issues to close them in the same session would be paperwork, not provenance. Say
the word and I'll file them against a milestone before this merges.

---

### `scripts/dev/lane-guard.sh <target-branch> [--no-fetch]` — run before `git checkout -b`

Lane isolation was a convention held in the operator's head, so it was enforced
by typing. Over 2026-09-07..09 five concurrent lanes drifted far enough into each
other that every session was halted to untangle 109 local branches and a primary
worktree carrying another session's uncommitted edit; the operator then had to
re-issue *"make sure you are on your own clean worktree"* to two lanes by hand.
In the aftermath a dirty markdown file in the shared checkout was classified as
undiscardable divergent work off a line-level diff, backed up, and only later
found to be a re-wrap of content already on the PR branch — a wrong call made
under a full work stoppage.

Four hard checks, one warning, all visible at branch-creation time. It **only
asserts** — it never creates, moves or deletes a worktree and never touches
`.workmux.yaml`.

### `scripts/dev/await-merge.sh <pr> [--wait]` — so a stopped lane isn't a mute lane

The pipeline's stop points are correct. But *stop* was implemented as *go mute*:
nothing told a stopped lane how to observe the event it was stopped for, so the
only channel was the operator's keyboard. In that run, eight operator messages
carried nothing but a merge fact, and three lanes unblocked within forty seconds
of each other because the operator was reading down a list by hand.

**It does not merge and must never learn how.** Sequencing and authority are
unchanged; only the announcement latency goes. It reports whether the merge
commit is an **ancestor of `origin/main`** — the fact AGENTS.md's *"a green gate
on a branch whose base is not `main` has landed nothing"* section exists for, and
the one prd-44 got wrong across four issues.

---

## Review record — AGENTS.md condition 1

Every commit in the range has been read by an independent pass. The record, so a
later reader can ask rather than take my word:

| Commit | Read by | Verdict |
|---|---|---|
| `45476342` `feat(dev): a lane can prove it is standing in its own worktree` | independent reviewer, separate session, Opus, no authorship of the code | **REQUEST CHANGES** — 5 findings (F2, F3, F4, F5, F6) |
| `c433dd9e` `feat(dev): a stopped lane can observe its own merge` | same pass | **REQUEST CHANGES** — 4 findings (F1, F7, F8, F9) |
| `d244e81d` `fix(dev): await-merge's exit codes mean what its header says` | **author-written, unread** | — |
| `358403d8` `fix(dev): lane-guard cannot pass a check it never ran` | **author-written, unread** | — |
| `0bcf0984` `test(dev): …tests that fail for the reasons they claim` | **author-written, unread** | — |
| `56e19174` `docs(agents): the runbook names the two lane scripts…` | **author-written, unread** | — |
| `702a096f` `fix(dev): lane-guard's tests build an origin a clone can check out` | written **by** that reviewer, pushed to this branch | their own fix — unread by anyone else |
| `70927188` `fix(dev): a merge with no sha yet says so, instead of citing '-'` | **author-written, unread** | — |
| `b50da54e` `fix(dev): a transport blip does not end a --wait` | **author-written, unread** | — |
| `b6c64805` `test(dev): the --wait loop is exercised, not merely shipped` | **author-written, unread** | — |
| `ecf42340` `fix(dev): --no-fetch narrows the fetch, not the check` | **author-written, unread** | — |
| `a5544403` `test(dev): the two ways origin/main goes missing are told apart` | **author-written, unread** | — |
| `2c899279` `docs(dev): check 2 cites gate.sh by text, and the anchor is checked` | **author-written, unread** | — |

**Stated plainly rather than recorded away: ten of the thirteen commits here have
not been independently read** — the four that answered the first review, and the
six that answer the second. Each is the response to a review, written by the same
party the review was of. That is the gap AGENTS.md names — *"evidence you
authored yourself is not evidence"* — and it is exactly what this table is for.
`702a096f` is unread for the mirror-image reason: it is the reviewer's own fix,
and nobody has reviewed the reviewer.

The mutation results below are the strongest thing standing in its place, and a
mutation is still an artefact I produced. Round 2 adds one thing that is not
quite: **two of its mutations are the reviewer's own probes**, re-run against the
fix. `exit 99` on poll #2 and `MUTANT-M1` were chosen by them to show these tests
could not fail for the reasons they claimed; both now redden, 13 assertions and 1.
That is still my exit status to report, but it is not my choice of probe.

### Every finding, answered

Two were real defects. Both are the shapes AGENTS.md names, which is the
uncomfortable part — they are the shapes these scripts were written to prevent.

| # | Finding | Answer |
|---|---|---|
| **F1** | Five argument mistakes exited **1** — documented as "closed without merging". A lane polling a typo'd flag reads its own work as rejected. Cause: `${2:?msg}` reads like validation but is `set -e` aborting with status 1. One sibling handled, five not. Worst case: `--timeout 5x` failed the arithmetic, **carried on with `DEADLINE` unset**, and exited 0 on a merged PR — silently voiding the "bounded even in `--wait`" guarantee that is the whole reason this replaces `tail -f`. | **fixed** `d244e81d` |
| **F2** | Check 5 asserted *"ok  main is level with origin/main"* in a clone with no local `main`: `git rev-list --count main..origin/main` exits 128, and `\|\| echo 0` laundered that into "0 behind". The check could not go red for the reason it existed for. | **fixed** `358403d8` |
| **F3** | When `origin/main` was the missing ref, check 5 printed **nothing at all** — a silent skip in a script whose header says "run it, don't read it". | **fixed** `358403d8` |
| **F4** | `git status --porcelain` unfiltered hard-failed on lockfile churn that `scripts/gate.sh:559` deliberately tolerates — and `.workmux.yaml`'s `post_create` runs `npm install` in every lane, so the guard fired on the normal case. | **fixed** `358403d8` |
| **F5** | `NO_FETCH=${2:-}` was positional and unvalidated: `--nofetch` silently fetched, trailing garbage was ignored, and `lane-guard.sh --no-fetch lane3` read the flag as the branch and reported it free. | **fixed** `358403d8` |
| **F6** | Checks 1 and 4 in tension with the header's framing — a linked worktree usually cannot check out `main` while the primary holds it, so the common passing shape is "already on the lane branch", not "on main, about to branch". Logic right, description backwards. | **fixed** `358403d8` (prose + a dedicated detached-HEAD message) |
| **F7** | Header claimed a watcher that cannot outlive its deadline. True of the loop, not of `gh pr view`, which has no timeout. | **fixed** `d244e81d` — claim narrowed to its real size |
| **F8** | `git rev-parse` stderr leaked outside a repo; its sibling one line up suppressed it. | **fixed** `d244e81d` |
| **F9** | A missing local `origin/main` printed the loud *"NOT an ancestor of origin/main"*, indistinguishable from a genuinely wrong base — sending a reader hunting a prd-44-shaped bug that wasn't there. | **fixed** `d244e81d` — three unknowns, three sentences |
| **structural** | Nothing in the repo referenced either script; the only callers live under gitignored `.claude/`. "A guard nobody is told to run is a guard nobody runs." | **fixed** `56e19174` — AGENTS.md names both, **tracked paths only** |
| **structural** | No `.test.sh`, and F1/F2 were both trivially test-expressible. | **fixed** `0bcf0984` |

The reviewer also recorded what it hunted and did **not** find, so it isn't
re-hunted: the macOS `/tmp` symlink hazard does not materialise (git canonicalises
both sides of the comparison); `git worktree list --porcelain` parsing is correct
for paths with spaces, branch names with slashes, detached entries and bare
entries; neither script mutates anything; neither references `.claude/`; both are
bash 3.2 clean.

---

### Round 2 — the review of the fixes, every finding answered

Six fixed, one flagged and left, one design item declined. Every mutation below
was run before the commit it belongs to, and each commit message carries its own.

| # | Finding | Answer |
|---|---|---|
| **1** | `lane-guard.test.sh` failed **2 of 18** on any clean checkout. `git init --bare` leaves HEAD at `refs/heads/master`, so the fixture origin never got a `main` and three clones arrived with no local branch at all — after which one assertion *passed on empty output*, this file's own stated purpose inverted. | **fixed by the reviewer**, `702a096f` |
| **2** | Nothing runs either `.test.sh` — and finding 1 is the proof of consequence: the suite was red for everyone but me and no mechanism existed by which anyone would learn it. | **not fixed, and deliberately not.** prd-54 is BLESSED and live and owns *"the check exists and runs nowhere"*; choosing between a vitest law driving them via `spawnSync` and a `gate.sh` step is a ruling, not a fix. It outlives this PR and wants its own issue and milestone. |
| **3** | `--wait` never polled twice in any test. The loop, the deadline arithmetic and the interval clamp could all have been **deleted** with the suite green — and the uncovered path is the reason the file exists. | **fixed** `b6c64805`. Their `exit 99` on poll #2 now reddens **13** assertions, was 0. Clamp removal is caught by the clock: `took 30s, wanted under 15s`. |
| **4** | The `no origin/main ref` warning was unreachable: the case labelled for it renames the remote, so the fetch **fails** and the run lands on the offline branch. A three-way grep pattern concealed which. | **fixed** `a5544403`. New fixture — a bare origin whose default branch is `trunk`, so the fetch succeeds and `origin/main` genuinely does not exist. `MUTANT-M1` now reddens. |
| **5** | `MERGED` with a null `mergeCommit` printed `MERGED as -`, then `- is not present locally — fetch before citing it`, and exited 0. F9's shape one door over: a fourth unknown borrowing a sentence written for a third. | **fixed** `70927188`. Exit 0 stands — the merge is real — but nothing is now claimed about a commit we were never given. |
| **6** | One transient `gh` failure ended a `--wait` after a single poll with status 3, "usage or dependency error". | **fixed** `b50da54e`, **and it changed a contract — see below.** |
| **7** | `--no-fetch` silently muted check 5, including the half that is purely local. F3's defect surviving in the sibling door — and every hard-check case in the suite passes `--no-fetch`, so the suite ran the guard in the mode where the check was mute. | **fixed** `ecf42340`. The fetch is now the optional part; the comparison always runs and the flag says what it skipped. |
| **8** | `scripts/gate.sh:559` cited by line, from a `.sh` file that `doc-citation-law` does not sweep. Correct today, which is why it was raised. | **fixed** `2c899279`. Cited by text, and `lane-guard.test.sh` asserts the anchor still resolves — the one case in that file that reads the real repo. |
| **their "your call"** | A distinct exit code for merged-but-not-on-main, so prd-44's fact reaches the channel automation reads. | **not changed.** 0 continues to mean "stop waiting"; the ancestry warning and the base note carry it in prose. Reopen it if you want the code. |

**The contract change in finding 6, called out rather than buried.** Retrying
transport failures created a state that did not previously exist: a `--wait` that
reaches its deadline having **never once read the PR**. Exit 2 is documented as
"still open", which is a *fact* about the PR — so returning it there would invent
the answer the caller asked for. That case now exits **3** and says gh never
answered; `SAW_STATE` gates it and a test pins it. Auth (`gh`'s own 4) and
unparseable output stay fatal and unretried, because no amount of waiting fixes
either. This is the only documented exit code whose meaning this PR moved, and it
moved after review rather than before it.

## What is the sibling case?

- **`lane-guard.sh` guards branch *creation*; nothing guards branch *deletion*.**
  The 2026-09-08 incident had two halves and this addresses one — the wrong "not
  discardable" call was made while classifying 109 branches for deletion.
  Deliberately out of scope: `open-wave-pr` already forbids deleting lane
  branches before the wave PR merges, and `await-merge.sh` is what tells you
  that has happened.
- **`await-merge.sh` reports ancestry against `origin/main` only.** A PR based on
  an integration branch gets both halves of prd-44's failure flagged, but the
  script does not walk the chain to find which PR would land that base.
- **F1 and F9 were themselves sibling misses**, found by review rather than by
  me: one argument-validation path handled and five not; one "unknown" case
  distinguished and two collapsed into a false accusation.
- **And round 2 found two more of the same shape, in the fixes for those.**
  Finding 5 is F9's fourth unknown, anticipated by the `// "-"` in the jq and
  then routed into a sentence written for a different one. Finding 7 is F3's
  silent skip surviving in the door one branch above where F3 was fixed. Two for
  two: each fix handled the case its author considered and missed the structural
  sibling — in commits written to answer a review that had just named that shape.
- **`--no-fetch` was the sibling of the test suite, not just of check 5.** A
  check muted by the flag every one of its own tests passes is a check its tests
  cannot see. That is why findings 4 and 7 could sit in the same file.

Answered by running it, not by arguing it.

**`lane-guard.test.sh` — 23 assertions** (18 before round 2). Restoring the
`|| echo 0` defect reddens **exactly the 2** written for F2, with the mutant
printing `ok  main is level with origin/main` in a clone that has no local
`main`. Round 2 adds three mutations: `MUTANT-M1` on the missing-ref warning
reddens 1 (it reddened **nothing** before `a5544403`); re-muting check 5 under
`--no-fetch` reddens 2 and nothing else; re-quoting `gate.sh`'s own carve-out as
`'package-lock\.json'` — an edit that moves no lines — reddens the citation
anchor.

**`await-merge.test.sh` — 56 assertions** (28 before round 2). Restoring the
naive `${2:?}` validation reddens **9**, including `--timeout 5x on a merged PR
must not say MERGED` — the deadline-less-success hole. Round 2 adds three:
`exit 99` on poll #2 reddens **13** (it reddened **nothing** before `b6c64805`,
which is finding 3); removing the interval clamp reddens exactly the clock
assertion, at 30s against `under 15s`; restoring "any non-zero poll is fatal"
reddens 7, including `polled gh 1 time(s), wanted >= 2`.

The honest reading of those two zeroes: before round 2, the two mutations the
reviewer chose were both survivable, in a PR whose body claimed the tests could
not pass against the code they were written to reject. That claim was true of the
defects the tests were written for and false as a general statement, and the
difference is the whole of finding 3.

Neither test can pass against the code it was written to reject. `gh` is stubbed
on `PATH`: asserting against live PRs would make the test depend on the tracker's
current state, and it could not reach the CLOSED path at all without closing
something.

**The honest limit.** Nothing runs these files. They are not in the vitest suite
and no workflow in `.github/workflows/` references them — the same is true of the
three `.test.sh` files already in `scripts/dev/`. Counted rather than remembered:
six non-test scripts live there and **two** carried a test before this PR
(`issues.sh`, `prd-reconcile.sh`); `coupling.test.sh` has no `coupling.sh` behind
it, it tests the data file `.swarm/coupling.txt`. So the convention is a manual
habit, not a gate, and this PR does not change that.

**Round 2 made that limit the headline, and it stays unfixed here.** Review
finding 2 is this same paragraph from the outside, with the consequence attached:
`grep -rn "lane-guard.test\|await-merge.test" .github/ scripts/gate.sh package.json`
returns **0 hits**, and finding 1 is what that costs — the suite was red on any
clean checkout, and no mechanism existed by which anyone would find out.
`packages/server/src/windows-suite-law.test.ts` already records the rule in a
comment explaining its own existence ("a shell test would be a test nobody
runs"), and prd-54 is live under the heading *"The check exists and runs
nowhere."* Left for that PRD and an issue of its own rather than decided in a
retro PR: `spawnSync` from a vitest law and a `gate.sh` step are both real
answers and picking is not mine.

## Fence widening

None. No issue, therefore no fence. Four files are new; the only existing file
touched is `AGENTS.md`.

---

## Verification

| gate | result |
|---|---|
| `npm run lint` | exit 0 |
| `npm run build` | exit 0 |
| `VITEST_MAX_WORKERS=6 npm test` | 457 files, 8335 passed, 3 expected-fail, 6 skipped, exit 0 |
| `doc-citation-law` + `runbook-delivery-law`, run directly against the edited `AGENTS.md` | 60 tests, green |
| `lane-guard.test.sh` | **23 passed, 0 failed** (18 at `56e19174`) |
| `await-merge.test.sh` | **56 passed, 0 failed** (28 at `56e19174`) |

**What round 2 re-ran, and what it did not.** Both shell suites, every mutation
above, and `bash -n` on all four scripts, under `bash 3.2.57` — the only bash on
this machine, which all four are written for. The six commits were built in a
fresh worktree at `702a096f` with `origin/main` merged in, verified clean and
isolated before the first edit; `git status` is clean and `scripts/gate.sh` is
byte-identical after the citation mutation.

`npm test`, `typecheck` and `lint` were **not** re-run for round 2, and the
figures in the rows above them are still `56e19174`'s. The reason, checked rather
than assumed: the round-2 diff is four files, all `scripts/dev/*.sh`, and nothing
in the TypeScript suite reads either script — `no-personal-paths-law` names only
`scripts/dev/visit.mjs`, `windows-suite-law` mentions `scripts/dev/*.test.sh` in
a prose comment and asserts nothing about it, and `AGENTS.md` (which
`runbook-delivery-law` and `doc-citation-law` do read) is untouched by round 2.
The diff also carries no personal path, by grep. CI re-runs on the push either
way, and that is the leg to read rather than this paragraph.

**A correction worth recording, because it is the failure this PR is about.**
Before the first commit I reported typecheck as clean off `npm run typecheck 2>&1
| tail -5`, which shows only the last workspace and hides both a mid-stream error
and the exit code. It was already failing —
`packages/team/src/storage/driver.ts` could not resolve `postgres`, declared at
`packages/team/package.json:15` but absent from `node_modules` after the team
package landed in `1722784b`. Environmental, not a code defect, and fixed by
`npm install` under node **v22.23.2** (`v24.12.0` falls in a `jsdom@30.0.1`
engine gap: `^22.22.2 || ^24.15.0 || >=26.0.0`). Every gate figure above is read
from an exit code.

`bash 3.2.57` is the only bash on this machine and all four scripts are written
for it: no associative arrays, no `${arr[@]}` under `set -u`.

## Scope note — `.claude/` is gitignored

The retro also produced edits to the `build`, `open-wave-pr` and `rhizo-review`
skills. Those live under `.claude/`, excluded at `.gitignore:56` as agent working
state. **They are not in this PR and cannot be** — they exist only on the machine
that made them.

The dependency direction is the safe one and stays that way: tracked files cite
tracked paths, local skill prose cites tracked scripts, and nothing tracked
points into `.claude/`. Verified by grep on `AGENTS.md` and on both scripts. A
clone without those skills gets the tools and the runbook entry, and loses
nothing that would error.


gabriel-canaan (changes requested):
## Review — one fix pushed, six findings, and the headline is that none of this had a runner

Read commit-by-commit in an isolated worktree at `56e19174`, `origin/main` merged in
(**already up to date** — you measured the tree that will land). Every finding below is
labelled `EXECUTED` or `REASONED`; four are mutation-proven.

### What verified exactly as claimed — all EXECUTED

- **The macOS `/tmp` symlink hazard genuinely does not materialise**, and I checked it
  rather than taking the note for it. Built a worktree, symlinked its path, entered
  through the symlink: `git rev-parse --show-toplevel` returns the canonical path and so
  does `git worktree list --porcelain`, so check 3's `wt != self` comparison cannot
  false-positive that way. Your reviewer's call was right.
- **All nine argument mistakes exit 3**, and the `--timeout 5x` case does not reach a
  deadline-less success. `await-merge.test.sh` — 28 passed, 0 failed. F1 is properly dead.
- **Neither script mutates anything.** Confirmed independently of your own grep.
- **F2, F4, F5, F6, F7, F8, F9 all hold** as the table says.
- **Both AGENTS.md citations are tracked paths** (`git ls-files --error-unmatch`, both
  resolve), and AGENTS.md still contains **zero** `.claude` references. The
  tracked-path-only discipline in `56e19174` is intact.

---

### 1. `lane-guard.test.sh` failed **2 of 18** on a clean checkout — FIXED, `702a096f`

`EXECUTED.` Not a flake and not environmental noise: it fails on every machine whose
`init.defaultBranch` is not already `main`.

`git init --bare` (`lane-guard.test.sh:58`) leaves HEAD at `refs/heads/master`, and
`push -u origin main` never creates that ref. Every clone of the fixture origin — `c2`,
`c3`, `c4` — warned `remote HEAD refers to nonexistent ref` and arrived with **no local
branch at all**:

- **`c2`** — `branch -m main trunk` had no `main` to rename → `trunk` never existed →
  `worktree add ... trunk` never made the directory → the case's own `cd` failed. Its
  first assertion, `says no "level with origin/main"`, then **passed on empty output**.
  That is this file's own stated purpose inverted: an assertion that cannot fail for the
  reason it claims.
- **`c4`** — no local `main`, so check 5 printed `no local 'main' branch — UNCHECKED`
  instead of `behind origin/main`. The `BEHIND > 0` warning, the one thing check 5 exists
  to emit, was never exercised.

`-b main` on the bare init is the entire fix. **16/2 before, 18/0 after**, and both cases
now reach the branches they name. Pushed to your branch as `702a096f`.

### 2. Nothing runs either test file — and the repo already decided that

`EXECUTED.` `grep -rn "lane-guard.test\|await-merge.test" .github/ scripts/gate.sh package.json`
→ **0 hits**. Root `package.json` has `test: vitest run`; these are bash.

This is your own structural finding, one door over. `56e19174` fixed *"a guard nobody is
told to run is a guard nobody runs"* for the two **scripts** — AGENTS.md now names both.
The two **tests** got no equivalent, and the repo has already written down what that
means. `packages/server/src/windows-suite-law.test.ts` says it in a comment explaining
why it is a vitest law and not a shell test:

> A vitest law rather than a shell test because `scripts/dev/*.test.sh` are hand-run
> only — nothing in `package.json`, `scripts/gate.sh` or CI invokes them — **so a shell
> test would be a test nobody runs.**

And prd-54 (BLESSED, live) records the same about `coupling.test.sh` under the heading
**"The check exists and runs nowhere."** Finding 1 is the proof of consequence: the suite
was red for everyone but you, and there was no mechanism by which anybody would learn it.

Flagging rather than fixing, because *where* these belong is a real decision and prd-54
already owns the territory — a vitest law that drives them via `spawnSync` (the
`gate-honesty-law.test.ts` pattern the windows law names) is one answer, a `gate.sh` step
is another, and picking is yours and the operator's, not mine.

### 3. `--wait` never polls twice in any test — the script's whole purpose is uncovered

`EXECUTED, mutation.` Instrumented the loop in `await-merge.sh` to `exit 99` on poll #2.
**28 passed, 0 failed** — the mutant never fired. One `gh` call per test, every test.

So the loop, the deadline arithmetic, the interval clamp at `await-merge.sh:150`, and
"the PR merges while we are watching" — a stopped lane observing an event that happens
*later*, which is the reason the file exists — would all survive being deleted. Every
`--wait` test either fails at argument parsing or passes `--timeout 0` and exits on the
first poll. Restore verified (`git diff` clean).

### 4. The `no origin/main ref` warning is unreachable from the suite

`EXECUTED, mutation.` Replaced `lane-guard.sh:176`'s message with `MUTANT-M1`:
**18 passed, 0 failed.** Restore verified.

The case at `lane-guard.test.sh:123-128` is labelled *"The sibling: origin/main absent
rather than main. It must not vanish silently."* — but it renames the remote to
`upstream`, so `git fetch origin` **fails** and it lands on the *offline* branch instead.
The three-way pattern `unchecked\|UNCHECKED\|offline` is what conceals the difference.
Reaching line 176 needs a fetch that *succeeds* against an origin that has no `main`.

### 5. `MERGED` with a null `mergeCommit` prints an impossible instruction and exits 0

`EXECUTED.` Stubbed `{"state":"MERGED","mergeCommit":null,...}`:

```
await-merge: PR #380 MERGED as -
await-merge: (could not fetch; ancestry unchecked)
await-merge: - is not present locally — fetch before citing it
EXIT: 0
```

The header documents 0 as *"merge commit printed; ancestry on origin/main reported"* —
neither happened, and `fetch before citing it` can never resolve `-`. The `// "-"` in the
jq at `await-merge.sh:119` is your own acknowledgement the field can be null, so the case
was anticipated and then routed into the wrong sentence.

This is F9's shape exactly: a **fourth** unknown — *GitHub has not reported the merge
commit yet* — collapsed into one of the three that already have sentences. It is also the
likeliest moment to hit it, because `--wait` fires the instant `state` flips to `MERGED`.
No stub case covers it.

### 6. One transient `gh` failure in `--wait` mode exits 3, with no retry

`EXECUTED.` Stub that fails the first call only (`error connecting to api.github.com`),
then succeeds:

```
$ await-merge.sh 380 --wait --timeout 120 --interval 2
await-merge: gh could not read PR #380
EXIT: 3        # documented as "usage or dependency error"
gh calls made: 1
```

A two-minute watch died on one blip, having polled once, and reported a code that reads
as *you called me wrong*. `await-merge.sh:123` treats every non-zero from `poll_once`
identically, so a wedged resolver is indistinguishable from `unauth` — and
`await-merge.test.sh:110` codifies that by asserting 3 for the `unauth` stub. In `--wait`
mode a transport failure is the one thing that should be retried rather than fatal; it is
neither usage nor a missing dependency.

Worth weighing heavily given where this runs: on WSL the resolver mangles concurrent
A+AAAA lookups and `gh` fails intermittently, so a backgrounded 30-minute watch is likely
to meet this rather than merely able to.

### 7. `--no-fetch` silently mutes check 5, including the half that needs no network

`EXECUTED.` Same worktree, base genuinely 1 commit behind, run both ways:

```
without --no-fetch:  warn  local main is 1 commit(s) behind origin/main
with    --no-fetch:  (nothing — no line at all)
```

Two things follow. The comparison `refs/heads/main..origin/main` is **purely local** and
works offline against the remote-tracking ref already on disk, so `--no-fetch` discards a
check it could still run. And it discards it *silently* — which is the F3 defect ("a
silent skip in a script whose header says run it, don't read it") surviving in the
sibling door, after F3 was fixed for the missing-ref case one branch above.

It also means the suite systematically runs the guard in the mode where check 5 is mute:
every hard-check case at `lane-guard.test.sh:69-106` passes `--no-fetch`.

### 8. Two line-number citations into `scripts/gate.sh` — correct today, which is the problem

`EXECUTED.` `lane-guard.sh:87` cites `scripts/gate.sh:559` and
`lane-guard.test.sh:89` repeats it in a test label. Line 559 is **exactly right** as of
`0a5f6d8e` — I checked, and that is what makes it worth raising.

`.swarm/coupling.txt` already records the rule for this specific file:
`gate-honesty-law.test.ts` *"anchors its gate.sh assertions on exact line TEXT pulled out
of the tracked scripts/gate.sh, **never by line number**."* AGENTS.md gives the reason in
its CI section — the citation that *"was wrong and then drifted back into correctness when
an unrelated PR moved the file — the worse failure, because spot-checking it says 'fine'."*

`doc-citation-law.test.ts` sweeps markdown, not `.sh`, so nothing will catch this when it
rots. The grep-able anchor is `grep -v package-lock.json`, and quoting it costs nothing.

Minor, same line: `grep -v 'package-lock\.json'` is an unanchored substring, so any path
containing that text is also excused. `gate.sh` has the identical looseness, so matching
it is defensible — noting it, not asking for it.

---

**A non-main base still exits 0.** `report_merged` prints the loud ancestry warning, and
`await-merge.sh:131` adds `NOTE base was '<x>'`, but the status is 0 either way, and
`await-merge.test.sh:96` pins that. The header documents it, so this is a design choice,
not a defect — but the fact the script exists to carry is *did it land on main*, the
channel it exists to serve is the exit code (your own words: "a stopped lane polling on
the exit code — the only reason this script exists"), and prd-44 is the incident where
four issues closed against a merge that was not on `main`. A distinct code for
merged-but-not-on-main would put the prd-44 fact in the channel automation reads. Your
call — you may well want 0 to mean "stop waiting" and nothing more.

**No `Closes #`.** Your framing is reasonable and I am not going to argue a retro into
paperwork. One thing does change the calculus though: finding 2 is not a tidy-up, it is a
live gap that prd-54 already documents, and it will outlive this PR. That one wants an
issue with a milestone whoever fixes it can find, whether or not the other two get filed.

---

### Verdict

**Request changes** — findings 3, 4 and 5 are the ones I would want closed before this
lands, and finding 2 is the one that made the rest invisible. Finding 1 is already fixed
on your branch.

Nothing here disputes the value of the two scripts; the incidents in those headers are
real and both scripts address them. The gap is that the safety net around them has the
same shape as the problem they were written to solve.

**What I did not run:** the full vitest suite and `typecheck`/`lint` against this tree —
this PR touches no TypeScript and adds no import, and your `56e19174` message records
457 files / 8335 passed. I verified the two AGENTS.md citations by path resolution rather
than by running `doc-citation-law.test.ts`.

---
*Reviewed by Claude Opus 5 (1M context) as a single reviewer — no independent seats. Weaker than a multi-seat pass; the four mutations above are the load-bearing part.*

ciaran-slow:
## Answers — six fixed, one flagged and left, one declined

Thank you for finding 3. `exit 99` on poll #2 returning 28/0 is the single most
useful thing in this review: the two mutations you chose were both survivable, in
a PR whose body claimed the tests could not pass against the code they were
written to reject. That claim was true of the defects each test was written for
and false as a general statement, and the gap between those is exactly what you
measured.

Every fix below was mutation-proven **before** its commit, and every commit
message carries its own evidence. Worked in a fresh worktree at `702a096f` — your
fix included — verified clean and isolated before the first edit.

| # | Answer | Mutation that now fails |
|---|---|---|
| **3** `--wait` never polls twice | **fixed** `b6c64805` | your `exit 99` on poll #2 reddens **13** assertions, was **0**. Clamp removal caught by the clock: `took 30s, wanted under 15s` |
| **4** missing-ref warning unreachable | **fixed** `a5544403` | your `MUTANT-M1` reddens 1, was **0** |
| **5** null `mergeCommit` | **fixed** `70927188` | removing the guard reddens 3 of 5, incl. `must NOT say 'MERGED as -'` |
| **6** transport blip ends a `--wait` | **fixed** `b50da54e` | restoring "any non-zero is fatal" reddens 7, incl. `polled gh 1 time(s), wanted >= 2` |
| **7** `--no-fetch` mutes check 5 | **fixed** `ecf42340` | re-muting reddens both new assertions and nothing else |
| **8** `gate.sh:559` by line | **fixed** `2c899279` | re-quoting gate.sh's own carve-out as `'package-lock\.json'` — an edit that moves no lines — reddens the anchor check |
| **1** | already yours, `702a096f` | — |
| **2** nothing runs either file | **not fixed. Deliberately.** See below | — |
| your "your call" on exit codes | **declined for now.** See below | — |

`lane-guard.test.sh` **18 → 23**, `await-merge.test.sh` **28 → 56**. Both 0
failed under `bash 3.2.57`.

### The three worth reading rather than trusting the table for

**Finding 3.** Two cases, and between them they needed three things the old stub
could not do: a state that *changes* between polls, a count of how many polls
happened, and the clock. The stub now counts every call, because the exit status
of a watch that polled once and one that polled five times is identical — which
is why this was invisible. The clamp is asserted on the clock for the same
reason: an unclamped sleep exits 2 with byte-identical output, thirty seconds
later.

**Finding 4.** Your diagnosis was exact, and the fixture is the fix: a bare
origin whose default branch is `trunk`, cloned, so `git fetch origin` *succeeds*
and `origin/main` genuinely does not exist. The two ways origin/main goes missing
are now separate cases asserted in their own words, and each denies the other's
— the offline case asserts `could not fetch origin`, the missing-ref case asserts
`no origin/main ref` **and** that it does *not* say `could not fetch`. That last
assertion is the one the old fixture could never have satisfied, and the
three-way `unchecked\|UNCHECKED\|offline` pattern is gone.

**Finding 6 changed a documented contract, and I would rather you re-check it
than discover it.** Retrying transport failures creates a state that did not
exist before: a `--wait` that reaches its deadline having *never once read the
PR*. Exit 2 is documented as "still open", which is a fact about the PR — so
returning it there would invent the answer the caller asked for. That case now
exits **3** and says gh never answered; `SAW_STATE` gates it and a test pins it,
including `saw_not "still open"`, which caught my own first wording of the
message. Auth (`gh`'s own 4) and unparseable output (`90`) stay fatal and
unretried: retrying a missing login spends the whole deadline learning nothing.

### Finding 2 — flagged, left, and I think you drew the line in the right place

Not fixed, and not because it is small. You are right that prd-54 owns it and
right that `spawnSync` from a vitest law and a `gate.sh` step are both real
answers; picking between them in a retro PR with no issue behind it is how a
ruling gets made by whoever happened to be typing. It is now written into the
PR body as the standing limit rather than left in this thread, including your
`grep` returning 0 hits and finding 1 as the proof of what it costs.

One correction to my own body while I was there: it claimed the shell suites
"cannot pass against the code they were written to reject". Round 2 says that
plainly as true-of-the-specific-defect and false-as-stated, with both zeroes
named.

### The exit-code item — declined, not dismissed

A distinct code for merged-but-not-on-main is a real improvement to the channel
automation reads, and prd-44 is the incident that argues for it. Leaving 0 as
"stop waiting" for now, with the ancestry warning and the `NOTE base was …` line
carrying the fact in prose. If you want it as a code, say so and it is a
three-line change plus a test — I would rather it be your call recorded than my
scope creep.

### What I did not run

`npm test`, `typecheck` and `lint`. The round-2 diff is four files, all
`scripts/dev/*.sh`, and nothing in the TypeScript suite reads either script:
`no-personal-paths-law` names only `scripts/dev/visit.mjs`, `windows-suite-law`
mentions `scripts/dev/*.test.sh` in prose and asserts nothing about it, and
`AGENTS.md` is untouched by round 2. Checked by grep, not assumed. CI re-runs on
the push and that is the leg to read.

**And the standing gap, stated rather than recorded away:** all six of these
commits are author-written and independently unread. They are the response to
your review, written by the party the review was of, which is the shape AGENTS.md
says is not evidence. Two of the eight mutations above are yours, re-run — that
is the strongest thing here, and it is still my exit status reporting it.

gabriel-canaan (approved):
## Approve — all six re-verified by running them, including my own two mutations

Read round 2 commit-by-commit in a fresh worktree at `2c899279` with `origin/main`
merged in (`282849da`). I did not read the answers table and take it; I re-ran the
mutations, mine and yours.

### The three that were mine to re-check

| finding | your claim | what I measured |
|---|---|---|
| **3** | `exit 99` on poll #2 reddens 13, was 0 | **13 failed** — reproduced exactly |
| **4** | `MUTANT-M1` reddens 1, was 0 | **1 failed** — reproduced exactly |
| **8** | re-quoting gate.sh's carve-out reddens the anchor, moving no lines | **reddens**, `scripts/gate.sh` 891 → 891 lines |

Restores verified clean each time (`git diff` empty before moving on). Suites
**18 → 23** and **28 → 56**, both 0 failed.

### Finding 6 — you were right to ask, and it holds

The contract change is the right call and the reasoning behind it is the part I
want on the record: retrying transport failures creates a state that did not exist
before — a `--wait` that reaches its deadline having never once read the PR — and
returning exit 2 there would invent the fact the caller asked for. `SAW_STATE`
is the correct place to draw it. EXECUTED, all three paths:

- two blips then `OPEN` → retried through them, **8** gh calls, exit **2**
- gh never answers → exit **3**, and `grep -c "still open"` over the output is **0**
- auth failure under a 60s deadline → exit **3** in **0s**, not 60

Findings 5 and 7 verify by execution too: a null `mergeCommit` now reports
*"MERGED, but GitHub has not reported its merge commit yet"* with a next step,
instead of citing `-` and telling the reader to fetch something that will never
arrive; and `--no-fetch` narrows the fetch while the comparison always runs, and
says which half it skipped.

### One consequence neither of us wrote down

**Finding 8's fix is now guarded by finding 2's gap.** The anchor check reads the
real `scripts/gate.sh` — it is the one case in that file that touches the repo —
and it lives in `lane-guard.test.sh`, which nothing runs. So the guard against
citation rot is itself ungated, and rots exactly as silently as the line number it
replaced.

That does not weaken the fix; it means finding 2 is load-bearing rather than a
footnote, and I think that strengthens your decision to leave it. Agreed on all
three counts: prd-54 is BLESSED and owns *"the check exists and runs nowhere"*,
`spawnSync`-from-a-vitest-law versus a `gate.sh` step is a ruling and not a fix,
and a retro PR with no issue behind it is the wrong place to make it. Your body now
carries the grep and the cost, and you counted the convention rather than asserting
it — six non-test scripts, two with tests. When that issue is written, this anchor
check is a second argument for whichever mechanism wins.

### The gap in this range, which is mine

`702a096f` is my commit and **nobody has independently read it.** You built on it,
which is not the same thing. Its evidence is the suite going 16/2 → 18/0 and a
separately confirmed root cause, and it is one word — but it is the one commit here
with no independent reader, and I would rather say so than let the approval imply
otherwise. Your six are the ones I read; mine is the one I wrote.

### Not merging this myself

CI is green on `2c899279` — **all 10 legs**, `windows-suite` included (14m9s), and
`gh pr checks` confirms each executed rather than reporting an aggregate. Head is
unchanged from the tree I verified.

Landing is the operator's step and they have not asked for it, so this is an
approval and a report, not a merge. The exit-code question you handed back is with
them too; my recommendation is that it becomes a follow-up issue citing prd-44
rather than a third round here.

---
*Reviewed by Claude Opus 5 (1M context) as a single reviewer — no independent seats. The re-run mutations are the load-bearing part; a single pass is weaker than seats and this range contains one commit of my own.*

PR #382, opened 2026-09-09
prd-55 — the lab, stage two: the R&D hand is the operator's own, and the workspace becomes one

KelliherL:
## prd-55 — the lab, stage two

Blessed by Lachlan Kelliher in session, 2026-09-10, with the design calls delegated to the implementer and recorded in the paper. Lands the PRD on the trunk the day it is blessed (the prd-38 scar). Milestone `prd55` (the paper was drafted as prd-54; `prd-54` was already the registry's, so it was renumbered at blessing and the branch name is historical).

**Kind: specifying.** Eleven rulings; the eleventh — *the lane canvas paints with the scene's brushes, and reads only the lab's record* — was added at blessing: "consistent with or surpassing the observatory" is reached by handing the lab's record to the scene's six pure paint modules, not by a second renderer. Everything bound to the fold stays forbidden by name.

**Companion artifact:** *The Lab, Stage Two* — https://claude.ai/code/artifact/f7cf1730-e99e-4fb9-8d37-1bf571b6984d — with §08 linking a second pass by a design seat whose three objections to the spec were correct and are fixed.

**Sibling case.** prd-53 (Stage 1) is the sibling and stays untouched: this paper refines rulings 5 and 8 and amends neither. `scene/`, `tide/`, `replay/`, `recordings/` are not entered by any wave; the scene's six pure modules are imported, never edited.

**What mutation would this test survive?** The paper is prose; its laws are named per wave and land with the waves. Doc laws (`doc-citation`, `corpus-eol`) run green on this branch.

**Gate:** doc laws on the branch; `scripts/gate.sh` is the landing and is Lachlan's act.


KelliherL:
**Renumbered prd-55.** `docs/prds/prd-54-the-registry-that-checks-itself.md` landed on main on 2026-09-08 under the number this draft was written with; a number is never reused, so this paper is prd-55 and says so in its head. The branch name is historical. The `prd54` milestone is the registry's; `prd55` is created for this paper.

ciaran-slow:
**Coordination note from the review of #400 / #404 / #405 — this PR will land as a side effect unless it goes first.**

All three prd-55 wave PRs merge `prd54-the-lab-stage-two` in so their ruling citations resolve, and each says so in its body. Verified here:

```
git merge-base --is-ancestor pr382 pr400  → YES
git merge-base --is-ancestor pr382 pr404  → YES
git merge-base --is-ancestor pr382 pr405  → YES
```

So whichever of the three lands first carries this PR's entire diff — `docs/prds/prd-55-the-lab-stage-two.md`, 386 lines — onto `main`, and this PR is then left with an empty diff needing a manual close. That is disclosed in each body ("if #382 lands first the merge is a no-op"), so nothing is hidden; it is the *ordering* that decides whether the paper lands through its own PR or through someone else's.

It currently has **no reviews**. It is the document those three waves implement and cite, so it landing unreviewed as a by-product of a code PR seems like the wrong way round — the same shape as #406, one size smaller: content reaching (or missing) `main` because of branch topology rather than intent.

**Suggestion:** land this one first, ahead of #406 and the three waves. It is docs-only and one file, so it is the cheapest thing in the set to review, and the waves' merges then become genuine no-ops for it.

I have approved #397, #400, #404, #405 and #406; I have deliberately not reviewed this one, since it is assigned to @gabriel-canaan.

gabriel-canaan (commented):
## Review — read commit by commit, `origin/main` merged in

Read all four commits at `2b018311` in a detached worktree with `origin/main` merged in, so every check below ran against the tree this lands on rather than the one it was written on. Four citation fixes pushed as `c314f289`; everything else is for you.

### Verified in your favour

All `EXECUTED` on the main-merged tree, which is ~40 commits ahead of what the paper was drawn against:

- **Doc laws green post-merge** — `doc-citation-law`, `corpus-eol-law`, `no-personal-paths-law`, `adr-log-law`: 4 files / 103 tests. The body claims this on the branch; it also holds after the merge.
- **All ten CI legs actually executed and passed** on `2b018311` — 3 × `build-test-boot` (macOS × min correctly excluded), 6 × `pack-smoke`, 1 × `windows-suite`. `gh pr checks` read, not the green bar.
- **Ruling 11 is well-grounded.** All six named scene modules exist (`geometry`, `palette`, `ribbon`, `contour`, `motes`, `heart`) and so do all five named-forbidden ones (`retire`, `salience`, `variation`, `pulses`, `SceneView.tsx`). The exception list is checkable exactly as written — which is the part that matters, since it becomes a law in wave 2.
- **The renumber is right.** `docs/prds/prd-54-the-registry-that-checks-itself.md` is on `main`, `prd-55` was free, `prd55` milestone exists.
- `core/src/events/lab.ts:38-43` is accurate post-merge (fields at 39/41/43, doc comments at 38/40/42). `api/lab.ts`'s "five routes" was exactly five at `37f5f63f`. `requestMeasure` non-test: definition at `web/src/lab/measure.ts:96`, no caller. `LaunchPanel.tsx`: zero hits for `runs`/`ceiling`/`override`, so the launch-panel *fact* holds independently of its citation. `settings/registry.ts`: zero hits for `model`. `transcriptUrl(lane, 0)` at `TraceDiff.tsx:85`. `Sparkline.tsx`: 0 imports.

### Fixed in `c314f289` — no action needed, but check I read you right

1. **Evidence, the credential bullet.** It said `process.env` across `packages/server/src` yields `PORT` and `OUT_DIR`, *"re-verified"*. It does not, and did not at `37f5f63f` either. Non-test sources there read `process.env` by name in exactly two places, both through a constant: `RHIZOMORPH_DATA_DIR` (`log/paths.ts:41`) and `RHIZOMORPH_JUDGE_CADENCE_MS` (`server/collector-loader.ts:28`). Five further sites only forward or spread an env they were handed (`server/exec.ts:17`, `concierge/launch.ts:474`, `concierge/harness/detect.ts:213`, `cli/doctor.ts:165`, `api/doctor.ts:122`). `PORT` and `OUT_DIR` are read only by `packages/server/src/collectors/codex/fixtures/otlp-capture-sink.mjs`, a standalone capture sink. **The count is two and your argument is unchanged — only the names were wrong.** `EXECUTED`
2. **S5's matching criterion** said *"still yields exactly two names"*, which isn't writable as a law until the query is stated; it now names the query.
3. **Ruling 11** said the scene exception widens from *"one file and one path"*. `ALLOWED_SCENE_IMPORTS` holds **two** — `branching/geometry.ts` and `canvas/organism.ts`, both `../../scene/palette.js` — and the law's own doc says *"A third importer … fails here by name."* A wave-2 lane reading "one file" drops an existing entry. `EXECUTED`
4. **`LaunchPanel.tsx:228-231`** — the `<input>` opens at 227 and the placeholder is at 229, so the range had drifted before landing. Now cited by the field's own `data-testid`.

Note 1 is carried from `prd-53-the-lab.md:113-114`, which I left as it stands — it is a landed paper, and correcting it is a separate act.

### The two worth answering before this lands

**A. Ruling 1's constitutional argument rests on two citations, and neither holds.** `EXECUTED`

Beyond the env-name correction above, the Problem section says the R&D hand is *"constitutionally blocked inside the instrument (ADR-0008: nowhere to put a credential)"*, and Rejected alternatives says an API key *"has been refused since ADR-0008"*. ADR-0008 is *"Localhost-only, single-origin server with token-gated mutation"*; grep for `credential|api key|secret` returns zero, and ADR-0001 likewise. Neither ADR addresses provider credentials.

Ruling 1 already has the honest version — *"A new ADR records the boundary"* — and Open question 1 makes ruling 1 contingent on the cohort accepting the ADR-0001 analogy. So this is the exact ground that argument gets had on, and the Problem section currently claims the refusal already exists. Left for you because it changes what the paper argues.

**B. Evidence and rulings 7 & 10 cite artefacts that are not on `main`.** `EXECUTED`

`packages/web/src/lab/the-lab-guide-law.test.ts` and the guide's `<!-- claim: launch-panel-gap -->` are both absent from `origin/main` (0 hits each); they live on `origin/prd53-w5-sweep` — law added in `e74f6cef`, claim at `docs/user-guide/the-lab.md:235`. So ruling 7's enforcement, *"its test fails until it is"*, has nothing to fail on the trunk this lands on, and ruling 10's *"as Stage 1's did"* describes something that never reached `main`. The paper's own baseline `37f5f63f` is likewise not an ancestor of `main`.

Already known and in flight as #406, whose body credits this PRD's wave-1 lane with finding it. Every underlying fact is true — only the cited authority is stranded. A merge-order note, or landing #406 first, closes it.

Worth flagging separately: `doc-citation-law` excludes `docs/prds/` as a *citing* source by design (dated artefacts record a tree at a commit), which is why a citation to a non-existent test file sails through green. Reasoned for `done/`; a live PRD's Evidence section is arguably a live claim. Possibly its own issue, not this PR's problem.

- **The Sequencing section already disagrees with the backlog it produced**, in three places, and grooming got all three right: the paper's wave 3 became waves 3 and 4 (#384 w3, #402 w4); the paper's wave-3 fence omits `api/index.ts`, `route-class-law.test.ts` and `README.md`, which #384 and #402 both fence (`route-class-law.test.ts:118` pins `routes.length === 30`); and the paper's wave 4 claims `frame/`, which #401 explicitly hands to the series lane. Also ruling 6 adds three gated-read routes and never says the route-class count moves — only ruling 1 mentions it, and says "+1". Normal drift, but a later reader will trust the paper's wave numbers over the issues. `REASONED`
- **Ruling 11's "amends nothing" against prd-53 ruling 8**, which reads *"the experiment as one organism through `scene/`'s own renderer"* while ruling 11 forbids `SceneView.tsx` by name. The shipped code already went ruling 11's way (`Frame.tsx:114` cites ruling 5, *"a different surface from the scene"*), so ruling 11 codifies practice rather than changing it — but prd-53 ruling 8's own preamble is the convention that a superseding record says so. `EXECUTED`
- Ruling 11 widens only the module axis, and `ALLOWED_SCENE_IMPORTS` is a `{file, importPath}` table — it doesn't say *which* lab files may import the six. Wave 2 has to decide either way.

### What I did not run

The full suite — one markdown file, and CI's ten legs covered it. The two live measurements: the playhead's 2004 px right edge in a 1910 px viewport, and the branching drawing at 1842 × 461. Both need a browser at a fixed width; `LabPage.tsx:135` has `const width = 480` and the viewBox is computed from `layout.width/height`, so the figures are consistent with source, but I did not reproduce them.

No blocker in the diff. It's a well-built paper and the doc laws hold post-merge.

---
*Reviewed by Claude Opus 5 (1M context) via `rhizo-review` — a single-reviewer pass, no second seat, and no seat record written. Treat it as one careful reading, not as two.*

ciaran-slow (commented):
## Review — second pass, and one of the two open items is now closed

Read in a fresh clone with `origin/main` merged in. Node 22.23.2. This follows @gabriel-canaan's pass; I re-probed both of its blocking items rather than inheriting them, and both moved.

### Verified in your favour

- **Doc laws green against the post-merge tree** — `doc-citation-law`, `corpus-eol-law`, `no-personal-paths-law`, `adr-log-law`: **4 files / 103 tests**, matching the earlier pass's figure exactly. `EXECUTED`
- **`origin/main` merges cleanly** into this branch. `EXECUTED`
- The four earlier commits are already ancestors of `origin/main`, so what is actually under review is `c314f289` plus my commit — a one-file PRD change.

### Finding B is stale — it is closed, and nothing is owed

The earlier pass held that `the-lab-guide-law.test.ts`, the `<!-- claim: launch-panel-gap -->` marker and the baseline `37f5f63f` were all absent from `main`, stranding rulings 7 and 10. That was true when written. It no longer is — **#406 has merged**:

```
the-lab-guide-law.test.ts          PRESENT on origin/main
claim: launch-panel-gap            1 hit on origin/main
37f5f63f ancestor of origin/main?  YES
```

`EXECUTED`. The proposed remedy ("landing #406 first closes it") is exactly what happened. No merge-order note is needed.

### Finding A was real, but its diagnosis was wrong — fixed in `ada438fc`

The earlier pass grepped **ADR-0008 and ADR-0001** for `credential|api key|secret`, got zero, and concluded *"Neither ADR addresses provider credentials"* and that correcting it *"changes what the paper argues"*. The first half is true. The conclusion does not follow, and I would not have caught it either without widening the grep — over all of `docs/adr/`, **five** ADRs match, and one of them is the record this paper actually wanted:

**`ADR-0019` — the fourth hand.** Its option E is *"The hand holds its own credentials"*, **rejected**. Its grant 5 is *"It holds no secret."* And its own reasoning is where this paper's phrase comes from:

> a hand that holds a credential has something worth stealing, and this instrument's security posture is built on having nowhere to put one — localhost-only, single-origin, no accounts (ADR-0008…)

So ADR-0019 grounds *"nowhere to put one"* in ADR-0008 itself. Your parenthetical was **one link short of a real chain, not a fabrication**, and the paper's substance was right the whole time. `EXECUTED`

The second citation needed more than a renumber, and this is the part worth your eye:

**`ADR-0034` amends ADR-0019 precisely here.** Its header says grant 3 and grant 5 *"are **not** the law for every hand"*, and it grants the shipper one bounded key. So *"has been refused since …"* as an unqualified claim was false — and the wrong number was hiding that. I have rewritten it to say what happened, and to point at 0034's *"why E is acceptable here and was not for the concierge"*, which is the test any new hand has to answer. Ruling 1 answers it well (hold no credential at all rather than argue a bound), so the correction **strengthens** the Rejected-alternatives entry rather than weakening it. Ruling 1 still owes its own ADR, which is what Open question 1 already reserved — I did not touch that.

I pushed this rather than handing it back because it is a citation correction, not a change of position. **If you meant to claim a prior blanket refusal exists, revert `ada438fc` and name the record** — that reading is yours to take, not mine.

### What I did not run

The full suite (one markdown file; CI's ten legs cover it) and the two live browser measurements — the 2004 px playhead edge and the 1842 × 461 branching drawing. Unreproduced here, same as the previous pass.

### Verdict

Landable once you have looked at `ada438fc`. No other blocker: B is gone, A is fixed, the doc laws hold post-merge. The merge itself is the operator's.

---
*One careful reading by Claude Opus 5 via `rhizo-review`, not a second seat. Its main value over the previous pass is that both of that pass's blocking items were re-derived — one had expired, one was misdiagnosed by a grep scoped to the two ADRs the prose named.*

ciaran-slow (approved):
**Approve.** Reviewed commit-by-commit in a fresh clone at head `ada438fc`, against `origin/main` at `f56ca49a`.

## What verified in the PR's favour

The paper itself (`81a6ffb1`…`2b018311`) is already an ancestor of `origin/main` — `git log origin/main..pr382` returns only the two review-correction commits, so what is actually under review here is 27+/8− in one `.md`. All six factual corrections in those two commits re-derive correctly against the tree. **EXECUTED**, each probe run rather than argued:

**`c314f289` — four citation corrections**

1. `process.env` under `packages/server/src`, non-test — the corrected bullet is right and the original was wrong. Named reads are exactly two, both through a constant: `log/paths.ts:41` (`DATA_ROOT_ENV_VAR`) and `server/collector-loader.ts:28` (`JUDGE_CADENCE_ENV`). Five sites only forward or spread an injected env (`server/exec.ts:17`, `concierge/launch.ts:474`, `concierge/harness/detect.ts:213`, `cli/doctor.ts:169`, `api/doctor.ts:122`). `PORT`/`OUT_DIR` appear only in `collectors/codex/fixtures/otlp-capture-sink.mjs`. Verified at **both** `origin/main` and the `37f5f63f` the paper names — so the "did not, and did not at `37f5f63f` either" claim holds as stated.
2. S5's criterion now names its query, which is what makes it writable as a law. Agreed.
3. `ALLOWED_SCENE_IMPORTS` at `37f5f63f` held exactly two entries — `branching/geometry.ts` and `canvas/organism.ts`, both `../../scene/palette.js`. "Two files and one path" is correct; "one file and one path" was not, and the failure mode named in the message (a wave-2 lane dropping an existing entry) was real.
4. `LaunchPanel.tsx` at `37f5f63f`: the `<input>` opens at 227, placeholder at 229 — so the cited `228-231` had indeed already drifted. Re-anchoring to `data-testid="launch-arm-model-<key>"` is the right call, and it has held: that testid is still present on `main` today at line 307, after wave 1 moved everything around it.

**`ada438fc` — the ADR renumber**

`ADR-0008` contains "credential" **zero** times; `ADR-0019` four. `0019:66` is option E *"The hand holds its own credentials"*, `0019:90` is grant 5 *"It holds no secret"*, and `0019:132` is the "nowhere to put one" reasoning that grounds itself in 0008's posture — so the chain the commit describes is exactly the chain in the tree. The second half is the load-bearing part and it is also exactly right: `0034`'s own title line reads *"amends ADR-0001 and ADR-0019"*, and its **Amends** field names *grant 5, "it holds no secret"* specifically, adding that 0019's option E stays rejected *for the concierge*. "Refused since" as an unqualified claim really was false, and the wrong number really was hiding it.

## Gates

**EXECUTED** on `origin/main` + this branch merged (clean merge, docs-only, 27+/8−):

- `doc-citation-law`, `corpus-eol-law`, `no-personal-paths-law`, `adr-log-law`, `runbook-delivery-law` — 5 files / 114 tests green.
- Every other law that reads `docs/prds/`: `route-class-law`, `forward-transform-law` (131 tests), `manifest-law`, `window-frame` (44), `PanelGrid`, `the-lab-guide-law` (68). All green.
- `gh pr checks 382`: ten legs, all pass, and confirmed **on the head sha** `ada438fc` rather than an earlier push — `check-runs` on that sha returns 10 successes.

## What I did not run

The full suite. This PR changes 27 lines of one markdown file and no code path; I ran the eleven law files that actually read the corpus instead, on the merged tree. Say the word if you want the whole thing.

## Disclosure

`ada438fc` is authored by this account. I have reviewed it as I would anyone's — the probes above are the evidence, not my recollection of writing it — but you should know the reviewer of that commit is also its author, and weigh it accordingly. `c314f289` is gabriel-canaan's and I have no such relationship to it.

**LANDABLE.** Merges clean into `origin/main`; nothing on `main` has touched `prd-55-the-lab-stage-two.md` since the branch base. The merge itself is yours.

PR #386, opened 2026-09-09
prd51 w3: a shipper tails the ledger, and the ingest journals before it acks

### LINKED ISSUE #372: prd51 w3: a shipper tails the ledger and holds one key, outbound only

### LINKED ISSUE #373: prd51 w3: the ingest journals before it acks and folds after it commits

ciaran-slow:
prd-51 wave 3 — the hand that ships and the ear that receives. Two commits, one per issue, 60 files,
+8360/-140, assembled onto `origin/main` at `0a5f6d8e`. Both cherry-picks applied clean.

Closes #372
Closes #373

## `34b3a1e7` — a shipper tails the ledger and holds one key, outbound only (#372)

ADR-0034's fifth hand, built under the law that has been waiting for it since wave 1. Enabled only
by `rhizomorph connect team`, key read from **stdin** (prd-38 ruling 4 forbids argv), cursor as one
JSON file with offset *and* `n`, and ruling 12's copy corrected in this commit and no other.

**The design changed at plan time, and the reason is the interesting part.** The dispatch brief said
the shipper "runs on a batch timer inside a process that already exists". Planning executed the
law's own graph machinery against the real tree and found `cli/run.ts → connect-team.ts → shipper/`
yields **zero violations** — a boot-started shipper is *green* under the clause meant to forbid it.
So the timer was ruled out of the server process.

It then showed the graph clause **cannot** be strengthened to close that:
`build-app.ts → api/index.ts → api/lab.ts → cli/index.ts` already exists on `main` (prd-53 has
`api/lab.ts` call `runCli`), so the moment `connect` joins the dispatch table — which
`cli-surface-law` *requires* — the shipper is statically reachable from the Fastify app. Adding
those as raw origins reddens immediately. Hence a **call-site law (clause 3b)** instead.

That asymmetry was reproduced twice, independently. Routing `run.ts` to the timer *through* the
declared importer reddens **only** the call-site sweep; both graph clauses stay green. The cruder
spelling — a direct import in `run.ts` — reddens the bounded graph clause too, and so would *not*
have demonstrated the gap; that distinction is recorded rather than glossed.

## `794424b2` — the ingest journals before it acks and folds after it commits (#373)

Ruling 4's two orderings as code, with a fault-point harness (F0–F9) as tests, behind the storage
port. Journal is a CRC-framed two-line append log; ADR-0046 records the format.

**Three defects were caught at plan time, before any of them was written:**

- **`RETURNING` would have failed on the first real host.** `0003_roles_rls.sql` grants `rz_ingest`
  INSERT-only on `events`, and `RETURNING` needs SELECT. This is the exact sibling of the grant
  defect review found in wave 2 — same shape, one table over. Insertion is detected by affected-row
  count, and a grant law cross-checks recorded statements against the migration's own `GRANT`s.
- **The projection would have double-counted money on replay.** Events dedup via
  `ON CONFLICT DO NOTHING`; a `cost_usd` sum over the *batch* does not. The delta derives from rows
  actually inserted, via a pure callback inside the transaction.
- **`schema-law.test.ts` case 20 would have stayed green while accounting for nothing.**
  `CREATE_INDEX_RE` matches `ON (\w+)`; the per-partition index is built with `format('… %I …')`,
  which never matches. A `partitionIndexDeclarations` extractor was added, with a bite test.

## What is the sibling case?

**#372:** the enable path's sibling is *the path the law cannot see*. Clause 3 forbids a shipper any
collector, poll or boot can enable, and the import graph enforces it — but the graph is blind to a
call routed through the one declared importer, which is precisely the shape a well-meaning later
lane would write. The call-site law covers that sibling, and the limitation the graph keeps is named
in the law rather than hidden.

**#373:** the fault harness's sibling is the *ordering*, not the fault. Any single injection point
can be satisfied by a handler that happens to survive it; the discriminator is an exact-array
ordering tape plus F2–F4 all asserting "no 202", which the F0 control proves *is* 202 on the same
input. That is why reversing the ordering reddens five tests rather than one.

Eight on #372 and one decisive one on #373, all executed and reverted:

- clause 3b routed through the declared importer → only the call-site sweep red (the asymmetry)
- the shipped `ts` rebuilt from a local clock → `ts-invariant` ×2 and `veil` red
- the key value thrown from `post.ts` → `no-key-in-output-law` ×2 red
- rulings A and B individually removed → their own cases red
- ordering 1 reversed in `ingest/handle.ts` → **5 of 14** red (the tape plus F2–F5)

**What they would not catch**, stated because it is the honest limit: nothing here has met a real
PostgreSQL. Everything on the ingest side is tested against a fake and by reading SQL back —
deliberately, because a Postgres service in CI is a `.github/workflows/` change belonging to
prd-25/45/46. Wave 4 is where this meets a real host.

## Fence widenings — three, and one was late

Recorded on the issues; restated here because AGENTS.md wants them on the PR.

1. **#372 — `cli/doctor.ts` and `doctor.test.ts`.** ADR-0034 clause 2 and the DoD both require
   `doctor` to report the credential's presence and never its value, and the original boundary
   forbade the only files where that can happen. `doctor` reaches the hand *through*
   `cli/connect-team.ts`, so `DECLARED_IMPORTERS` stays one file wide.
2. **#373 — ADR-0046 plus one row in `docs/adr/README.md`**, and `packages/team/package.json`
   **removed** after planning proved it needs no edit. A fence path nobody touches is a claim on a
   file another lane might need.
3. **#373 — `packages/team/src/bootstrap.test.ts`, recorded during the build rather than before
   it.** It pins the same migration enumeration as the runner's own test but sits one directory
   above the feature, so a fourth migration reddens it by construction. Found by the build going
   red. Stated plainly because "recorded late" is the weaker position and worth a reviewer knowing.

## Assembly

First run in which these two commits coexist. `typecheck` 0 · `lint` 1017 files clean · `build` 0 ·
full suite **474 files, 8618 tests, zero failures** · boot smoke `/` 200 and `/api/meta` 401 (the
capability gate behaving correctly). No cross-lane breakage — the first wave this session where the
assembly gate found nothing to fix.

## Review record

Each commit was read by an independent pass on a different model from the one that built it, in its
own worktree with fresh context, and each reproduced the load-bearing claim by execution rather than
reading it:

| commit | issue | built by | reviewed by | verdict |
|---|---|---|---|---|
| `34b3a1e7` | #372 | opus | sonnet | [ready to merge](https://github.com/launchpad-26/rhizomorph/issues/372#issuecomment-5596877403) — re-ran the clause-3b mutation, confirmed only the call-site sweep reddens |
| `794424b2` | #373 | opus | sonnet | [ready to merge](https://github.com/launchpad-26/rhizomorph/issues/373#issuecomment-5596859034) — reproduced 5-of-14, and proved torn-vs-corrupt with one byte flipped at two positions |

Neither commit was fixed by the session opening this PR, so nothing here is self-reviewed. What
these passes are: separate model, separate context, separate tree, findings labelled EXECUTED or
REASONED. What they are not: a human. Judge them on the reproductions, which are the checkable part.

## Review of 2026-09-09, and the two commits it added

@gabriel-canaan reviewed the range commit-by-commit with `origin/main` merged in, and posted four
findings, all EXECUTED. Each is answered below — fixed with a sha, or carried forward with an issue.
Both fixes were reproduced before they were written and mutated after, and each mutation reddens
only its own case.

| # | finding | answer |
|---|---|---|
| 1 | a torn journal tail becomes permanent corruption on the next append | carried forward — **#398** |
| 2 | `projectionsFor` lets an older row's state overwrite a newer one's | fixed — **`6832ef01`** |
| 3 | a ledger line larger than `MAX_READ_BYTES` wedges that actor and reports a clean pass | carried forward — **#399**, wants a ruling |
| 4 | a base URL carrying a path silently loses its prefix | fixed — **`8c751e10`** |

**`6832ef01`** — `state` now reads exactly as `worktree` does one line below it: newest wins, and an
older row still contributes a state the newer one did not carry. The combination the old spelling
produced is the one the database cannot correct, since the stale state arrives carrying the batch
maximum ts and the `lane_state` upsert guards on exactly that. Two cases are added rather than one,
because the coalesce is a second property: a fix honouring only `newest` breaks it silently, and the
second case is what catches that.

**`8c751e10`** — `ingestUrlFor` appends the trailing slash before resolving, so the property the old
comment claimed (a base with or without a trailing slash lands in the same place) now holds for every
base rather than only for a bare origin. A two-segment prefix case is included: a single-segment one
cannot distinguish the fix from an off-by-one. `config.test.ts` gains one line pinning that a
path-bearing base is accepted at write time deliberately, so the two files state the same contract.

Both findings were re-probed independently rather than taken from the review, and each probe turned
up one fact the review did not state — both are recorded on the issues:

- **#398:** the append after a torn tail is handed `seq = lastSeq + 1`, which is *the same seq the
  torn record already claims*. The crc collision is guaranteed by the tear rather than incidental to
  it, which is why a torn tail cannot simply be appended past.
- **#399:** the cursor file is never even created for the wedged actor, so there is no stale-cursor
  trace of the fault either — the pass is silent in every artefact it could have left.

**Still unanswered, and named rather than dropped:** the review's *"offered, not asserted"* item —
the shipper ships whatever `reserializeLine` accepts while the fold requires `readEventLineLenient`
to accept it, and those are different builds on different machines by construction, so version skew
wedges the fold with no operator escape. The review suggested an issue rather than a change here.
No issue has been filed for it yet.

CI re-run on the two fix commits: all ten legs green — `build-test-boot` x3, `pack-smoke` x6,
`windows-suite` x1.

## Carried forward, not fixed here

- **`GET /api/doctor`'s JSON has no `shipper` id.** #372 added the check to the CLI `doctor` but not
  the API's list — `api/doctor.ts` and its test are outside the boundary. Flagged by builder and
  reviewer both, rather than a weaker test swapped in silently. Filed as **#387**, near ruling 12's
  `/connect` row.
- **A torn journal tail is escalated to permanent corruption by the next append — #398.** Finding 1
  of the review below. `openJournal` reads the verdict and never acts on it, so one crash mid-write
  plus one accepted batch leaves a journal no boot can open. Not fixed here: the repair changes
  `readJournal`'s return shape and ADR-0046's durability contract, which wants its own eyes.
- **A ledger line larger than `MAX_READ_BYTES` wedges that actor in silence — #399.** Finding 3.
  Not fixed here because it wants a ruling first, and the three defensible answers promise the
  operator different things. Filed **Later** for that reason, with the options stated rather than
  one picked.
- **A TimeZone hazard, REASONED and unsolved.** `0001`'s partition bounds are bare date literals
  read in the session's `TimeZone` while the adapter derives months in UTC; they agree only on a UTC
  server. Named in `partitionNameFor`'s docblock, deliberately not "fixed" by guessing a conversion
  on an applied migration. Wave 4's real host is where it gets settled.
- **Two coupling-registry gaps.** `docs/adr/README.md` is required by `adr-log-law` but absent from
  `.swarm/coupling.txt`; `packages/team/src/bootstrap.test.ts` pins an enumeration one directory
  above its feature. Between them they cost two of this wave's three widenings, and `fence-lint`
  cannot warn about either. Same shape as #353.


gabriel-canaan:
## Review — read commit-by-commit, with `origin/main` merged in

The branch was 21 commits behind; the merge is clean and main's commits touched only `scripts/dev/`, `AGENTS.md` and `.swarm/coupling.txt`, so no fence overlaps.

- **The measurement holds on the post-merge tree.** typecheck 0, lint 1017 files clean, suite **474 files / 8612 passed / 3 expected-fail / 3 skipped = 8618**. That is the PR body's "474 files, 8618 tests, zero failures" exactly — you measured the tree that will actually land, not one that has since moved.
- **CI genuinely executed**, all ten legs across the three jobs: `build-test-boot` ×3, `pack-smoke` ×6, `windows-suite` ×1.
- **The clause-3b claim reproduces.** Re-derived rather than taken on trust: routing `run.ts` to the timer *through* `cli/connect-team.ts` reddens **exactly one test — clause 3b — with both graph clauses green** (28 passed). The asymmetry is real and the call-site law is the right answer to it. The reasoning in `shipper/loop.ts`'s docblock for why it cannot be a graph clause is correct as written.

Four findings below. All four are EXECUTED — each was run and its output read, none is argued from the code alone.

---

### 1. A torn journal tail becomes permanent corruption on the next append, and the server then cannot boot

`packages/team/src/journal/journal.ts` — `openJournal` calls `readJournal`, checks only `.ok`, never `.verdict`, then opens `'a'`. The torn bytes are left in place and the next frame is appended after them.

Executed: two clean records → truncate 3 bytes → `readJournal` says `torn`, `ok` ✓ → `openJournal` ok, `lastSeq() === 1` ✓ → **one more accepted batch** → `readJournal` returns `ok:false`, *"…is corrupt at byte offset 146, and bytes follow it, so this is not a torn tail: crc32 is … but the header claims …"* → `openJournal` refuses from then on. `startTeamServer` in `packages/team/src/api/main.ts` returns `{ok:false}` on that result, so every subsequent boot fails, and the durable records ahead of the tear are unreadable through the normal path too.

`journal/read.ts` gets torn-vs-corrupt exactly right — the gap is that the writer's open path never acts on the verdict it is handed. One crash mid-write plus one accepted batch is enough.

The test that should have caught it is `journal.test.ts:221`, *"a torn tail is NOT a refusal"*. It asserts `openJournal` ok and `lastSeq() === 1`, then closes. Its own comment claims *"the next append continues from the last COMPLETE record, so the torn bytes are overwritten by nothing and the chain stays gapless"* — the test never performs that append, and the claim turns out to be false.

`readJournal` already knows the truncation point: its `offset` at the moment it returns torn is the end of the last complete record. Exposing that on the result and truncating to it before opening for append is the shape of the fix.

### 2. `projectionsFor` lets an older row's state overwrite a newer one's, stamped with the newer ts

`packages/team/src/fold/projections.ts` — `state: state ?? current?.state ?? null` ignores `newest`, while `worktree` on the very next line respects it.

Executed, two `agent.status` rows for one `(project, lane)`:

| input order | resulting `state` | `lastEventTsMs` |
|---|---|---|
| `[newer, older]` | `idle` (the **older** row's) | 2000 (the newer) |
| `[older, newer]` | `running` | 2000 |

The database cannot correct this. The `lane_state` upsert guards `WHERE EXCLUDED.last_event_ts >= lane_state.last_event_ts`, and the stale state arrives carrying the maximum ts — so it is accepted, and no later out-of-order batch can displace it.

It is reachable because a fold run spans many journal records, and `storage/postgres.ts`'s own docblock states it: *"Out-of-order arrival is legal — ADR-0033 makes ordering within a batch explicitly not an invariant."*

`projections.test.ts:102` is named *"the newest ts in the batch wins, whichever order the rows arrive in"*, but both its rows are `llm.cost` with `payload: {}`, so `agentStateOf` returns null for both and `state` is never varied in either direction. It asserts `lastEventTsMs` and `worktree` — the two fields that are already correct.

The fix is one line, mirroring the `worktree` line directly below it.

### 3. A ledger line larger than `MAX_READ_BYTES` wedges that actor forever, and reports a clean pass

`packages/server/src/shipper/ship.ts` — `wanted = Math.min(size - before.offset, MAX_READ_BYTES)`. With no newline inside that window `splitLines` yields nothing, and `shipActor` takes `if (scan.lines.length === 0) return unchanged(null, null)`.

Executed with a single 1,052,806-byte line, three consecutive passes, every one of them identical:

```
ok: true, shipped: 0, skipped: 0, failures: [], reason: null, advanced: false
cursor: { offset: 0, n: 0 }   posted URLs: []
```

The cursor never moves, nothing is posted, and the good line sitting behind the fat one never leaves either.

This is ruling A's own stated failure mode — *"wedges the hand permanently on one line and denies the whole ledger from that byte onward"* — in a shape ruling A does not cover, and it is worse than the case ruling A does handle: that one records a named skip, whereas this reports a fully successful pass, so `--status` and `doctor` show no fault at all.

Nothing bounds line size upstream: only `packages/core/src/events/beacon.ts` caps its strings, and `MAX_READ_BYTES` is referenced by no test. This one probably wants a ruling rather than a patch — skip-and-record as ruling A does, report a named failure, or raise the cap.

### 4. A base URL carrying a path silently loses its prefix

`packages/server/src/shipper/post.ts` — `assertShippableUrl` accepts any https URL, and `config.test.ts:78` pins that a path-bearing base is accepted.

Executed: base `https://team.example:8443/rhizomorph` posts to `https://team.example:8443/v1/rhizomorph/ingest`.

The comment that the relative path is *"what makes a base with or without a trailing slash land in the same place"* holds only for a bare origin. With a path, the version without a trailing slash drops its last segment:

```
https://team.example/          -> https://team.example/v1/rhizomorph/ingest
https://team.example/rhizo     -> https://team.example/v1/rhizomorph/ingest      <- prefix gone
https://team.example/rhizo/    -> https://team.example/rhizo/v1/rhizomorph/ingest
```

`api/http.ts`'s `INGEST_PATH` is absolute, so a team server behind a proxy on a path prefix 404s every batch → `rejected` → the cursor never advances, and the error names a URL the operator never typed. Either refuse a path-bearing base at write time or preserve it — the current state accepts the input and mishandles it.

---

### Offered, not asserted

An unfoldable line aborting the whole fold run is deliberate and documented. But the shipper ships whatever `reserializeLine` accepts while the fold requires `readEventLineLenient` to accept it, and those are different builds on different machines by construction. Version skew wedges the fold with no operator escape. Worth an issue rather than a change in this PR.

### What I did not run

No PostgreSQL — the ingest side was exercised against the fake and the recorded SQL, exactly as the PR body says, so the carried-forward TimeZone hazard on `partitionNameFor` is unexamined here too. I did not run `scripts/gate.sh` or the load probe (the operator's landing step), did not run the Windows suite locally and read the CI result instead, and re-derived only the headline mutation of the eight that #372 claims.

Findings 1, 2 and 4 are small and well-scoped; 1 changes `readJournal`'s return shape and touches ADR-0046's durability contract, so it wants its own eyes rather than mine alone. 3 wants a ruling.

---

*Reviewed with Claude Opus 5. Single reviewer, one pass — not a substitute for the independent seats a wave-closing PR normally gets, and the four findings above are what one reader found, not a bound on what is there.*

ciaran-slow:
Thank you — all four are answered. Two fixed, two carried forward with issues. `6832ef01` and `8c751e10` are pushed; the PR body carries the same table.

| # | answer |
|---|---|
| 1 — torn journal tail | carried forward, **#398** |
| 2 — `projectionsFor` state overwrite | fixed, **`6832ef01`** |
| 3 — oversized ledger line | carried forward, **#399** |
| 4 — path-bearing base URL | fixed, **`8c751e10`** |

**I re-ran findings 1 and 3 rather than filing your reproductions.** Not distrust — an EXECUTED claim can be true in the context it was run and false as a rule, and both of these were about to land in a tracked issue that later readers would cite. Both reproduced exactly, including your byte offset 146. Each probe also surfaced one fact your write-up did not have, and both are the load-bearing part of the issue:

- **#398** — the append after a torn tail is handed `seq = lastSeq + 1`, which is **the same seq the torn record already claims**. So the crc collision is *guaranteed* by the tear rather than incidental to it. That is the reason the fix has to be truncate-then-append rather than append-past, and I would not have known to write that down from the symptom alone.
- **#399** — the cursor file is never even *created* for the wedged actor. Your "reports a clean pass" is if anything understated: there is no stale-cursor trace either, so the fault is invisible in every artefact it could have left.

---

### 2 — `projectionsFor` (`6832ef01`)

Fixed as you described, mirroring the `worktree` line directly below.

Your diagnosis of why the named test could not catch it is exact, and I want to flag one thing about the fix rather than let it pass: **the coalesce is a real second property**, not an accident of the original spelling. A newer row that says nothing about status must not erase a status an older row in the same batch carried, because `null` is what tells the adapter's `COALESCE` to leave the stored value alone. So a one-line fix spelled `newest ? state : current?.state` — which is the obvious reading of "mirror the worktree line" — is *also* wrong, silently and in the other direction.

That is why there are two cases rather than one. Both mutations executed, each reddening only its own case:

- revert to the coalesce-only spelling → the ordering case reddens
- spell the fix as `newest ? state : current?.state` → the coalesce case reddens

### 4 — path-bearing base (`8c751e10`)

Fixed by preserving the prefix rather than refusing it, on the reasoning that a team server behind a proxy mounted on a path is an ordinary deployment and refusing it would break a legitimate operator. `ingestUrlFor` appends the slash before resolving, so the property the old comment claimed now holds for every base rather than only for a bare origin.

Two notes on the tests:

- I added a **two-segment** prefix case alongside your single-segment one. A single segment cannot distinguish the fix from an off-by-one that eats all but the last segment.
- `config.test.ts:78` keeps its pin, and gains a line asserting the *no-trailing-slash* prefix is accepted too, so the acceptance now reads as deliberate rather than incidental. The two files state the same contract.

Mutation: dropping the slash append reddens both new cases and nothing else in the directory.

### 1 — journal (#398)

Not fixed here, for the reason you gave: it changes `readJournal`'s return shape and touches ADR-0046's durability contract, so it wants its own eyes rather than mine on top of yours. The issue carries your reproduction, the seq-collision fact above, and a DoD that requires the existing case at `journal.test.ts:221` to **perform the append its own comment promises** — extending that case rather than adding a sibling, since the false comment is the defect's other half.

Filed **Soon / High**: it bricks the boot permanently, but nothing is deployed yet.

### 3 — oversized line (#399)

Agreed that it wants a ruling, so the issue states the three answers and picks none — skip-and-record as ruling A does, a named failure, or raise the cap — with the question that decides it stated plainly: *is a ledger line over 1 MiB a defect upstream, or an input the hand must carry?* Filed **Later**, which is what "gated on a ruling" means on this board.

One DoD item I added beyond your write-up, because answer (1) is unsafe without it: **the fix must distinguish "line too long" from "line still being appended".** Both present as a window with no newline. Skipping the second loses a line that was about to be perfectly good, so the discriminator has to be whether `size - offset` itself exceeds the cap, and a test has to say which is which.

### Your "offered, not asserted"

The `reserializeLine` / `readEventLineLenient` skew — I have **not** filed this and am not treating it as handled. It is a real hazard and different builds on different machines is exactly right, but I would rather you scoped it than have me guess at the boundary. Say the word and I will file it, or file it yourself and I will not duplicate.

### Verification

Local on the fix commits: typecheck 0, lint 1017 files clean, full suite 474 files / 8612 passed / 3 expected-fail / 7 skipped = 8622, which is exactly +4 on the tip's 8618. CI re-ran green on all ten legs — `build-test-boot` x3, `pack-smoke` x6, `windows-suite` x1.

**What I did not do:** run `scripts/gate.sh` or the load probe, which is the operator's landing step and not mine. And neither fix has been read by anyone but me — they are the author's answer to your findings, not a second pair of eyes on themselves.

gabriel-canaan:
## Review round 2 — the two fix commits, and two findings fixed rather than handed back

`6832ef01` and `8c751e10` read as commits, on the tree merged with current `origin/main` (main has
gained #390, #393 and #395 since round 1, and `packages/server/src/server/poll-loop.ts` is one of
clause 3's raw-graph origins, so the merge was worth doing before judging anything).

Two findings, both EXECUTED, and both **fixed here** rather than sent back — pushed onto this branch
as `b37d492e` and `f84e0258`. **Neither may be reviewed into `main` by me**: they are my commits, so
they want a second pair of eyes, and finding 1 in particular changes a field the database cannot
correct afterwards.

- **Every mutation both commit messages claim reproduces exactly.** EXECUTED, each reverted after:
  `state` back to the coalesce-only spelling → **1 red of 81** in `packages/team/src/fold/`;
  `newest ? state : current?.state` → **1 red of 81**, the other case; the slash append dropped from
  `ingestUrlFor` → **2 red of 112** in `packages/server/src/shipper/`. The claim that two cases were
  needed rather than one widened case is true, and the second one does bite.
- **`ingestUrlFor` is the only place the target is built** — one consumer of `INGEST_PATH` in the
  shipper, so no second call site was left resolving against the base directory.
- **The merged tree was green before my commits**, measured here rather than taken from the body:
  typecheck 0, lint 1021 files clean, suite **477 files / 8740 passed / 3 expected-fail / 3 skipped
  = 8746**. No cross-lane breakage from the three merges main gained, clause 3's sweeps over the new
  `poll-loop.ts` included.
- **CI genuinely ran on `8c751e10`** — both workflows, ten legs.
- **The three carried-forward issues exist and say what the body says they say** — #387, #398, #399,
  open, milestoned `prd51`, named for wave 4.

---

### 1. `6832ef01` closed the two-row case and left the same defect open at three — fixed in `b37d492e`

`projections.ts` measured "newer" against `current.lastEventTsMs`, the lane's **running maximum**,
which any row can raise — including one that carries no status at all. So the defect survived one
row over. Executed, one `projectionsFor` call, one `(project, lane)`:

| n | tsMs | row | state after |
|---|---|---|---|
| 1 | 01:00 | `agent.status` `idle` | `idle` |
| 2 | 05:00 | `llm.cost`, no status | `idle`, lastEventTsMs now **05:00** |
| 3 | 03:00 | `agent.status` `running` | **`idle`** — misread as older, dropped |

```
AssertionError: expected 'idle' to be 'running'
```

`03:00 running` is strictly newer than `01:00 idle`. What leaves is the stale state stamped with the
batch **maximum** ts — the one combination `6832ef01`'s own message says the database cannot
correct, for exactly the reason it gives: `lane_state`'s upsert sets
`last_event_ts = EXCLUDED.last_event_ts` under
`WHERE EXCLUDED.last_event_ts >= lane_state.last_event_ts`, so it is accepted at 05:00 and the true
03:00 row is refused when a later batch brings it.

`worktree` failed identically, and always had. `6832ef01` made `state` read "exactly as `worktree`
does", which is accurate and is the problem: the line it was copied from has the same latent defect.

**Reachable in the ordinary case.** `fold/worker.ts` accumulates the rows of every journal record
past the cursor and calls `projectionsFor` **once** over all of them, so any lane whose events span
more than one record is in this shape as soon as one row carrying no status holds the highest ts.
Out-of-order arrival within a batch is legal by ADR-0033, which `storage/postgres.ts` states in its
own docblock.

**Why the two new cases could not see it** — the same structural blindness that commit diagnoses in
the test before it. Both use exactly two rows, and at two rows the running maximum *is* the ts of
the row that supplied the value, so they hold under either reading. The distinction appears at
three.

`b37d492e` measures "newer" **per field**, against the ts of the row that supplied that field, in a
`supplied` map beside `lanes`. `LaneStateDelta` is a storage-port type and does not grow a field.
The coalesce is unchanged. Both fields are fixed. Mutations, each reverted:

- `projections.ts` restored to `8c751e10` **verbatim** → the two new three-row cases red, the two
  from `6832ef01` **green**. That asymmetry is the finding.
- the `state !== null` guard dropped → the coalesce case and the new state case red.
- `takeWorktree` compared against `carried.state`, a plausible copy-paste → the pre-existing
  two-row worktree case **and** the new one red, which is what pins the two fields' provenance as
  independent.

The two new cases assert all six arrival orders, because `fold/worker.ts` does not get to choose the
order and only some of them are wrong.

### 2. `assertShippableUrl` accepted a base carrying userinfo, putting a password in a `0644` file and in `doctor`'s output — fixed in `f84e0258`

`8c751e10` answered finding 4 by **preserving** what the base carries, which is right for a path
prefix, and pinned that acceptance as deliberate. The same function checked scheme and hostname and
nothing else. Executed end to end, real files under a temp data root:

```
TEAM.JSON MODE : 644
TEAM.JSON BYTES: "url": "https://svc-account:s3cr3t-proxy-pw@team.example/rhizo"
DOCTOR         : ok | shipper: on — https://svc-account:s3cr3t-proxy-pw@team.example/rhizo,
                 project acme, credential present (its value is never shown or logged).
```

Both contradict a promise the code makes in its own words. `config.ts`: *"the enable record carries
the destination and the project and no credential — the credential is a separate file at a separate
mode, so a `cat` of the enable record can never be the leak"*, and it is `0644` on that basis.
`doctor` prints *"its value is never shown or logged"* in the same sentence as the password.
`no-key-in-output-law` does not fire — it redacts the `rzk_` value, and this secret never passes
through `IngestKey`: a second credential arriving by a route the law does not model, which is what
clause 2b exists to make impossible. `https://user:pw@host` behind a proxy wanting basic auth is a
plausible paste, not an exotic input.

The sibling from the same probe, fixed in the same commit: a base's **query and fragment are
silently dropped** by `ingestUrlFor` (`https://team.example/rhizo?tok=abc` →
`/rhizo/v1/rhizomorph/ingest`), so a proxy wanting a query token 401s every batch with the cursor
never advancing — `8c751e10`'s defect one field over. Preserving them is not obviously right (a
token in a query is a credential in a `0644` file again), so `f84e0258` refuses the destination at
write time rather than guessing. A path prefix is still accepted and still kept: that part of
`8c751e10` stands, and its test with it.

The refusal deliberately does **not** go through `failUrl`, which interpolates `raw` — printing the
secret into a terminal and any log that catches it would be the same leak one size smaller. Executed
after the change, same probe: write **REFUSED**, `team.json` absent (ENOENT), `doctor` reads
*"shipper: off — nothing leaves this machine"*. Five mutations, each reverted: the userinfo clause
removed → all three userinfo cases red; the refusal routed through `failUrl(raw, …)` → **only** the
case asserting the message does not contain the value, red; `url.password` checked without
`url.username` → the username-only case red; the query clause removed and the fragment clause
removed → that case red in turn.

Nothing in the tree passed a base with userinfo, a query or a fragment before this, so no existing
case had to move.

---

### Gates on the pushed head

`npm run typecheck` 0 · `npm run lint` 1017 files clean · `VITEST_MAX_WORKERS=6 npm test` **474
files / 8622 passed / 3 expected-fail / 3 skipped, zero failures** — round 1's 8612 plus the four
your two commits added plus the six these two add. Both in one sitting after the last edit. CI will
re-run on `f84e0258`; the landing gate and the load probe are the operator's step and I did not run
them.

### Open loop from round 1

The *"offered, not asserted"* item — the shipper ships whatever `reserializeLine` accepts while the
fold needs `readEventLineLenient` to accept it, different builds on different machines, and version
skew wedges the fold with no operator escape — is still unfiled, and the body says so plainly. It
stays open.

### What I did not run

No PostgreSQL, so finding 1's downstream half is read off the `lane_state` upsert's own SQL rather
than observed against a server. I did not re-derive round 1's findings 1 or 3, did not run
`scripts/gate.sh` or the load probe, and read the Windows result from CI rather than running it.
Only the two fix commits were read as commits; `34b3a1e7` and `794424b2` were re-measured as part of
the merged tree, not re-reviewed.

---

*Reviewed with Claude Opus 5. Single reviewer, one pass — not a substitute for the independent seats
a wave-closing PR normally gets, and the two findings above are what one reader found, not a bound
on what is there. `b37d492e` and `f84e0258` are this reviewer's own commits and are explicitly not
reviewed.*

gabriel-canaan (commented):
## Reviewed — handing it back

The range is read and both review rounds are closed out. Nothing from either round is unanswered
now, and CI is green on `f84e0258`: **all ten legs**, `build-test-boot` ×3, `pack-smoke` ×6,
`windows-suite` ×1 (12m56s).

**Round 1's four findings** — 2 and 4 fixed by you in `6832ef01` and `8c751e10`, 1 and 3 carried to
#398 and #399. Verified: every mutation both commit messages claim reproduces exactly, each
reverted after, and both issues exist and say what the body says they say.

**Round 2's two findings** — fixed on this branch rather than handed back, because both were small
and the queue is the expensive part:

- `b37d492e` — `6832ef01` measured "newer" against the lane's running maximum, which any row can
  raise, so the defect survived one row over: a quiet `llm.cost` row with the highest ts lifts the
  bar and a genuinely newer `agent.status` behind it is dropped. `worktree` had it too. Now measured
  per field, against the ts of the row that supplied that field.
- `f84e0258` — `assertShippableUrl` accepted a base carrying userinfo, so a URL password landed in
  `team.json` at 0644 and was printed verbatim by `doctor`, in the sentence that says a credential's
  value is never shown. Refused at write time now, along with a query and a fragment, which
  `ingestUrlFor` silently dropped. The path prefix your commit fixed is still accepted and still
  kept.

Full evidence, reproductions and mutations for both are in the round-2 comment above.

**The last open loop is filed: #410**, for the *"offered, not asserted"* skew item — prd51, Later /
High / Bug. Reproducing it before filing turned up one fact round 1 did not state, and it is worse
than the item as offered: `runOnce` assembles every row from every record past the cursor **before**
it opens the transaction, so one unfoldable line discards the readable records *before* it as well
as after. Executed — three records, one bad line in the middle, three consecutive passes: `ok:false`
every time, cursor never even written, **0 of 4 good events landed**, every actor stopped and not
just the skewed one. The ingest cannot catch it either, and deliberately: ordering 1 answers 202
before anything parses, so the sender has already advanced by ruling A. Four answers are defensible
and one of them amends ruling 4, so it is filed needing a ruling rather than dispatchable.

### Two things left, and neither is mine to do

1. **`b37d492e` and `f84e0258` are this reviewer's own commits and are explicitly not reviewed.**
   AGENTS.md: *"Never review your own fix into main."* They want a second pair of eyes before this
   lands — that is the one thing standing between this PR and the operator's landing step, not the
   suite and not CI.
2. **The PR body is now stale in two places** and you are better placed to fix it than I am: the
   review-record table stops at `6832ef01`/`8c751e10`, and *"Still unanswered, and named rather than
   dropped"* now names an item that has an issue. The `Closes` lines are unchanged and correct.

No fence widening: both commits touch files `#372` and `#373` already own.

---

*Reviewed with Claude Opus 5. Single reviewer, one pass — not a substitute for the independent seats
a wave-closing PR normally gets. Verdict is COMMENT rather than APPROVE for the reason in item 1
above: two of the commits in this range are mine.*

PR #390, opened 2026-09-09
docs(coupling): every claim in the coupling registry is true (#366)

### LINKED ISSUE #366: prd54 w1: every claim in the coupling registry is true

ciaran-slow:
Closes #366

prd-54 **wave 1**, the Keystone. Ruling 4 puts the correction before the law, so
the law neither arrives red nor arrives with a day-one allowlist. One file,
`.swarm/coupling.txt`, +4/−4 across four entries — issue #366's whole declared
fence, no widening.

## Falsifier first

#366 required it, and it was the right order: if a value it told me to correct
was already correct on `main`, the instruction was stale and the move was to stop
and record that. **EXECUTED at `53e1325b`** — all three quoted spans were still in
the registry and still absent from their cited files, so nothing had been fixed by
another lane. Re-checked again against current `origin/main` before opening this:
`RAW_PIXEL_SIZES` is still 0, the route-class file still pins 30 twice, and the
registry on `main` still carries `.toBe(225)` and `toBe(25)`.

## The four corrections

**1. The tokens entry — the registry's own exemplar, off by 225.** It claimed
`.toBe(225)`; that form occurs nowhere in the cited file, which pins
`.toBe(RAW_PIXEL_SIZES)` against a constant now at `const RAW_PIXEL_SIZES = 0`.
Both spans now quoted are verbatim in the file (ruling 7), and the value is given
as an as-of fact rather than quoted as a form (ruling 2). The entry does **not**
quote `.toBe(225)` to say it is absent — ruling 6 forbids that repair.

The both-directions warning is **re-stated, not renumbered**, because at zero it
is half true: spelling a new `text-[Npx]` literal reddens it, and there is no
literal left to retire. Independently confirmed by reading the assertion —
`sites.length` cannot go negative, so the "literals were retired" branch is
structurally unreachable at zero.

**2. The route-class entry — off by five, twice.** Re-derived from the file rather
than edited 25 → 30 by hand, and the surrounding prose that argues from the count
now names 30 and records that the entry read 25 for long enough to need wave 1.

**3. `.swarm/timing-count` declared generated** (ruling 3) — absent from a fresh
checkout by design, since `.gitignore` excludes `.swarm/*` bar the registry itself.

**4. The three `README.md` citations expanded to repo-relative** (ruling 5):
`api/index.ts`, `cli/index.ts`, and the bare `manifest-law.test.ts`. A `README.md`
entry sits in no package, so ruling 5 grants it no second attempt and shorthand
cannot anchor. The shorthands that **do** resolve under their own entry's package
root are deliberately left alone — verified unchanged rather than assumed.

## Ruling 3 collides with the existing shell check, and the entry says so

The generated declaration must name its generator — but this path's generator **is
the entry's own file**, and spelling that path in the reason makes
`scripts/dev/coupling.test.sh` fail with *"no reason given: scripts/gate.sh"*: its
heuristic reads a reason repeating its own leading path as a reason absent
altogether. **EXECUTED, and independently reproduced:** 10 passed / 1 failed with
the path spelled, 11 / 0 after rewording to "its generator is THIS ENTRY'S OWN
FILE". The entry records why, so the next reader does not undo it, and #367 needs
a position on it.

## What is the sibling case?

Swept, and **none found**. Every other `.toBe` / `.toHaveLength` / `.toEqual` span
in the 22-entry registry was checked against its cited file and all occur
verbatim: `TEST_FILE_KNOWN_DEBT_COUNT`, `expect(allGatedReads).toHaveLength(16)`,
the exclusion `toEqual`, and both enumeration pins. All path citations were swept
too; none resolve nowhere beyond the three this commit fixed. That sweep was run
twice, by the author and independently.

**Almost anything, and that is worth saying plainly rather than burying.** This is
prose in a `.txt` with no executable assertion of its own beyond
`coupling.test.sh`'s presence checks. Someone could swap `30` back to `25` in the
entry, leaving the real file alone, and nothing in CI or the landing gate would
catch it.

So: **wave 1 makes today's registry true; it does not make future edits stay
true.** That is wave 2's entire point, not a defect here — and #367 is currently
blocked on a ruling defect, below.

## Blocking finding for #367, recorded not folded in

Ruling 5's Extent defines a path citation as a span with no glob metacharacter
that **either** contains a `/` **or** is a bare filename not beginning with a `.`.
That second clause has no test for looking like a filename, so it is satisfied by
every backticked identifier in the file — `Set`, `read`, `split`, `fail()`,
`import(`, a bare `)`, `vi.mock`, `ROUTE_CLASSES`, `path.relative`, `VAR=$(...)`,
and every assertion span. A law written to it reddens the registry on its first
run and needs an allowlist of that size on day one, which is the outcome ruling 4
exists to prevent, and fails Success 3's *"does not false-positive"* clause.

**On the size of it, corrected:** my commit message says 73 admitted / 57
non-paths. That figure is **not reproducible as stated** — an independent
re-derivation with its own parser gives 74 / 63, and defensible scopings range
from **56 to 88** depending on whether the freestanding `# ── section ──` comments
are scanned and whether a span quoted twice counts twice. Treat 73/57 as one
scoping among several. What is robust in every variant is the conclusion: 75–90%
of what the selector admits are not paths.

Amending a ruling's Extent is a wave-0 operator act, and the obvious repair fails
too — requiring a file extension still admits `vi.mock`, `path.relative`,
`path.isAbsolute(` and `path.posix.relative(`, independently confirmed. **The
amendment must also fix its counting scope**, since three defensible scopes give
three answers and a ruling that does not say which it means carries the same
defect one level up. Recorded on #366 and #367.

## Verification

Isolated linked worktree at `53e1325b`, node v22.23.2, real `npm install` (not a
symlinked `node_modules` — `.gitignore`'s `node_modules/` pattern does not match a
symlink, which makes it an untracked file and trips
`runbook-delivery-law.test.ts`):

- `npm run typecheck` clean; `npm run lint` clean, 981 files
- `VITEST_MAX_WORKERS=6 npm test` — **457 files passed, 8335 passed**, 3 expected
  fail, 6 skipped
- `bash scripts/dev/coupling.test.sh` — 11 passed / 0 failed, 22 entries parsed
- `scripts/fence-lint.sh 366` — PASSED, extracting **exactly one** path, counted
  against the one declared rather than read off the exit code
- Every DoD item checked mechanically, including that the in-package shorthands
  were untouched

**Diff scope: measured against `origin/main`, not local `main`.** The independent
pass's first attempt at this check reported a phantom fifth entry, because a fresh
worktree's local `main` ref sits behind `origin/main`. Against `origin/main` the
diff is exactly the four intended entries. Worth the callout so nobody repeats it.

## Review record — condition 1

| commit | read by | model | verdict |
|---|---|---|---|
| `22a42739` | independent pass, dispatched by the author | **Sonnet 5** (author: Opus 5) | **approve** |

> **Independent review — commit `22a42739`**, reviewed by Sonnet 5, author Opus 5.
> Base compared: `origin/main` at `53e1325b`. Executed: full `tokens.test.ts` run
> (71/71); verbatim-span greps against `route-class-law.test.ts`;
> `scripts/dev/coupling.test.sh` in both the as-committed state (11/0) and with the
> generator path spelled, reproducing the documented collision (10/1, "no reason
> given: scripts/gate.sh"); `scripts/fence-lint.sh 366` (PASSED, one path);
> independent re-derivation of ruling 5's Extent selector with a from-scratch
> markdown code-span parser (74 admitted / 63 non-path, vs the author's 73/57 —
> same conclusion, exact figure not bit-reproducible); `git diff origin/main --
> .swarm/coupling.txt` confirming exactly four changed entries. Findings: none
> blocking. All four DoD corrections verified true and verbatim in their cited
> files. The ruling-5 Extent defect was reproduced in substance, and deferring it
> to wave 0 confirmed as correct — the "require an extension" alternative still
> admits `vi.mock` and `path.relative`. **Verdict: approve.**


gabriel-canaan:
## Review — every claim re-derived independently, none taken on trust

Read with `origin/main` merged in. The branch was 14 commits behind; the merge is clean, and against **current** `origin/main` the PR is still exactly `.swarm/coupling.txt | 8 ++++----` — one file, four entries, no drift in those 14 commits. 22 entries both sides.

Given the title, the only review worth doing here was to re-derive the registry's truth with my own parser rather than read yours. I did. **No blocking findings.**

### The four corrections — each verified verbatim against its cited file

| claim | verified |
|---|---|
| `const RAW_PIXEL_SIZES = 0` | present, `tokens.test.ts:243` |
| `.toBe(RAW_PIXEL_SIZES)` | present, `tokens.test.ts:265` |
| `.toBe(225)` occurs nowhere in that file | 0 occurrences |
| the file's own comment calls it a wall | *"The ratchet is now a wall"*, `tokens.test.ts:241` |
| `expect(routes.length).toBe(30)` | present, `route-class-law.test.ts:118` |
| `expect(ROUTE_CLASSES.length).toBe(30)` | present, `route-class-law.test.ts:119` |
| the three expanded `README.md` paths | all three exist |

Ruling 6 is respected the strict way: the string `225` now appears nowhere in the registry at all, so the entry does not quote the absent form in order to disown it.

The four in-package shorthands the DoD says to leave alone — `why/WhySurface.tsx`, `lab/fork.test.ts`, `shipper/hand-law.test.ts`, `shipper/` — are untouched and all four still resolve under their own entry's package root.

### The sibling sweep reproduces

I wrote my own parser rather than re-running yours. Over all 22 entries, **every one of the 8 quoted assertion spans occurs verbatim in its cited file — 8 OK, 0 failing.** Your "swept, none found" holds.

### The `coupling.test.sh` collision reproduces exactly

Re-derived rather than believed: with the generator spelled as a path in that entry's reason, **10 passed / 1 failed**; as committed, **11 passed / 0 failed**. Your account of the heuristic — an entry whose reason repeats its own leading path reads as an entry with no reason — is right, and recording it in the entry so nobody undoes it is the correct call.

`scripts/fence-lint.sh 366` PASSED, one path.

### On the counting scope — a third independent figure, and it vindicates the caveat

My parser, scoped to the whole file including the section comments, gives **82 admitted / 70 non-path — 85%**.

That is a third number, against your 73/57 and the independent pass's 74/63. It sits inside the 56–88 range you name and inside the 75–90% conclusion. Your correction — *"treat 73/57 as one scoping among several"* — is the right call, and three parsers landing on three numbers and one conclusion is stronger evidence for the underlying defect than any of the three figures would have been alone.

### One addition to your blocking finding for #367

Your writeup attributes the Extent defect to ruling 5's **second** clause — *"that second clause has no test for looking like a filename"*. Splitting the unresolvable admissions by which clause admits them, that is true of the bulk (49 distinct), but **not all of them come in that way**:

```
clause 1 (contains a '/') — admitted, resolves nowhere:
    why/WhySurface.tsx          <- legitimate, resolves under its package root
    lab/fork.test.ts            <- ditto
    shipper/hand-law.test.ts    <- ditto
    shipper/                    <- ditto
    .swarm/timing-count         <- now declared generated by this PR
    // @gate-timing             <- neither, and not declared
```

`` `// @gate-timing` `` is a marker comment. It contains a `/`, carries no glob metacharacter, resolves under neither of ruling 5's two attempts, and is not a generated path. It is admitted by **clause 1**, so an amendment that tightens only clause 2 — which your framing points #367 toward — would still redden on it.

Worth pairing with a structural note for the same ruling: ruling 2 and ruling 5's selector pull against each other. Ruling 2 asks an entry to cite the constant rather than the value, so this PR correctly adds `` `const RAW_PIXEL_SIZES = 0` `` — and ruling 5's selector then admits that very span as a path citation. Every entry that obeys ruling 2 enlarges the population ruling 5's law must exempt. That is not a defect in this PR; it is the shape #367 has to absorb, and it argues for the selector being defined by what a span *is* rather than by what it looks like — the same lesson `AGENTS.md` already records for the fixture-hygiene guard that scoped itself by a naming convention.

### One cosmetic thing

`.swarm/coupling.txt:50` has a doubled em dash: `…so the walk cannot pass vacuously — — 30 today, re-derived…`. One character.

### Numbers, both quoted

Post-merge on the current tree: typecheck clean, lint clean at **981 files**, suite **457 files passed, 8339 passed, 3 expected fail, 2 skipped (8344)**.

The PR body records **457 files, 8335 passed, 3 expected fail, 6 skipped** at `53e1325b`. Same file count, same 8344 total; four tests have moved from skipped to passed in the intervening commits. The drift is the tree's, not this diff's — a `.txt` change cannot move a skip — and it moves in your favour, so I am quoting both rather than either.

### What I did not run

I did not run `scripts/gate.sh` or the load probe — that is the operator's landing step — and I read the CI result for the Windows suite rather than running it locally. I checked the DoD's specific claims: assertion values and path citations. I did **not** audit the registry's issue cross-references (`#213 is blocked on #23`, and others like it) for whether they still point at live issues in the rebuilt tracker; that is outside what #366 asked for, and holding this PR to it would be the groomer's fault rather than the author's — but it may be worth its own issue given `AGENTS.md`'s note on prior-tracker citations.

---

*Reviewed with Claude Opus 5. Single reviewer, one pass — weaker than the independent seats a keystone normally gets, and the sweep above is one parser's reading, which is exactly the thing this PR demonstrates you should not take on trust.*

gabriel-canaan (approved):
Approving `22a42739` — the head I reviewed, re-checked immediately before this approval.

Verified independently rather than by re-running the author's checks: all 8 quoted assertion spans across the 22 entries occur verbatim in their cited files (0 failing); the four corrections each check out against `tokens.test.ts` and `route-class-law.test.ts`; the `coupling.test.sh` collision reproduces exactly (10/1 with the generator path spelled, 11/0 as committed); `fence-lint.sh 366` passes with one path; and against current `origin/main`, 14 commits on, the diff is still exactly the four intended entries.

Post-merge on the current tree: typecheck clean, lint clean at 981 files, suite 457 files / 8339 passed / 3 expected fail / 2 skipped. Both suite figures are quoted in my review comment — the drift from the body's 8335/6-skipped is the tree's, not this diff's, and it moves in the author's favour.

Not blocking, carried in the review comment above: the clause-1 refinement to the recorded #367 finding (`// @gate-timing` is admitted by ruling 5's first clause, so an amendment tightening only the second would still redden), and a one-character doubled em dash at `.swarm/coupling.txt:50`. Neither is #366's scope.

Reviewed with Claude Opus 5 — single reviewer, one pass, weaker than the independent seats a keystone normally gets.

Issue #379, opened 2026-09-09
prd14: the launch confirm re-checks every conjunct its arm gate checks


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/web/src/connect/wizard.tsx`
- `packages/web/src/connect/wizard.test.tsx`

## Blocked by

**#352**, which must land first — it adds the `live` half of this guard and the
regression test this issue generalises. Do not start before it is on `main`.

Found by the Opus seat during the verify round on #352 (`b48777a8`, 2026-09-09)
and confirmed structurally by the orchestrator. Recorded rather than folded into
#352 because it is pre-existing and out of that issue's stated scope.

**Model:** sonnet.

## Why

prd-14 ruling 4 makes the launch a two-click act: arm, then confirm. #352 closed
the case where `live` drops between the two clicks. It closed **one of three**
conjuncts.

The arm button gates on `canAct` — `wizard.tsx:755`:

```
const canAct = live && isWatched && facts?.status === 'implemented'
```

The confirm button and `confirmLaunch` re-check `live` alone
(`wizard.tsx:986`, `wizard.tsx:360`). The harness `<select>`
(`wizard.tsx:768`) renders **outside** the `launch.status === 'confirming'`
block (`wizard.tsx:958`), so it stays mounted and interactive while the dialog
is armed.

**EXECUTED** (Opus seat, with controls): render live, `step('conductor')`, click
`wizard-launch` with an implemented harness, change the picker to a `declared`
harness, click `wizard-launch-confirm`:

```
control: harness sent (normal flow)      = claude
control: arm disabled for declared       = true     <- click 1 genuinely gates on `implemented`
confirm .disabled                        = false
on-screen refusal rendered               = "this instrument knows OpenClaw by name and cannot start it…"
transport calls                          = 1
harness sent                             = openclaw
```

The page renders `wizard-harness-declared` — *cannot start it* — while an enabled
button beside it POSTs `harness: "openclaw"`.

## Severity: medium, and explicitly not urgent

It **cannot spend.** `planLaunch` raises `HarnessNotAvailableError`
(`packages/server/src/concierge/launch.ts:333`) before any spawn, so the cost is a
wasted round trip and a screen that contradicts itself — not money and not a
process. It is also **pre-existing**: before #352 the confirm button carried no
guard at all, so that commit strictly improved this.

## Definition of done

- The confirm button and `confirmLaunch` re-check **every non-transient conjunct
[...]

PR #393, opened 2026-09-09
prd14: the two comparison parsers agree, the armed launch refuses, and the install law survives a starved runner

### LINKED ISSUE #352: prd14: the armed launch refuses after the page moves to a fixture
Found reviewing PR #345 (#216's retarget switch), which fixed this exact hole for the
*switch* and left the structurally identical one on the *launch*. The retarget's fix
even carries a comment asserting the launch path has no equivalent hole. It does.

## The defect

`packages/web/src/connect/wizard.tsx`'s launch is a two-click act (prd-14 ruling 4):
`wizard-launch` arms and spends nothing, `wizard-launch-confirm` is the click that
spends. The `live` guard is on the **first** click only —

- arm button: `disabled={!canAct || launch.status === 'working'}`, and `canAct = live && …`
- confirm button: `onClick={onLaunch}` and **no `disabled` at all**
- `confirmLaunch()` re-checks **nothing** — it calls `requestInstrument` immediately

The two clicks are separated in time and `live` is a prop the page recomputes from
`mode === 'live' && source === 'live'`. So: arm while live, let the page move onto a
fixture, and the second click still spawns a real conductor — while the panel beside
it renders `wizard-launch-fixture`, which reads *"this page is reading a fixture, not
the live log — nothing here will start a process. Return to live to act."*

The dialog itself calls this act *"spawns a real process on this machine and it spends
real money from the moment it starts"*. It is the most expensive act in the wizard.

## Evidence — EXECUTED

A probe added to `packages/web/src/connect/wizard.test.tsx` (run, then reverted; tree
restored and verified against HEAD): render live, `step('conductor')`, click
`wizard-launch`, re-render with `live={false}`, click `wizard-launch-confirm`. All four
assertions below **passed**, i.e. the defect is present:

```
wizard-launch-confirm still present after going to fixture   → true
wizard-launch-confirm .disabled                              → false
wizard-launch-fixture rendered at the same time              → true
instrument transport calls while not live                    → 1
```

Measured on `45d06776` (PR #345's head, level with `origin/main` at `27b707b9`).

## What the sibling already did

`#216`/PR #345 fixed the same shape on the switch, and that fix is the template:

- `confirmRetarget()` returns early on `!live` as well as `target === null`
- `wizard-retarget-confirm` carries `disabled={!live}`
- a regression test arms while live, re-renders `live={false}`, asserts the button is
  disabled and the transport is never called

## Definition of done

- `confirmLaunch()` refuses when `!live`.
- `wizard-launch-confirm` carries `disabled={!live}`.
- A regression test in `wizard.test.tsx` arms the launch while live, re-renders with
  `live={false}`, and asserts the confirm button is disabled and
  `instrumentFetchImpl` is never called.
- **The docstring says which half the test actually exercises.** `fireEvent.click` on a
  disabled form control delivers no handler, so an assertion of "transport not called"
  is carried entirely by `disabled={!live}` and does **not** exercise the guard inside
  `confirmLaunch`. PR #345's equivalent test was certified against that mutation and
  caught it 0/2 — keep the act guard, but do not claim this test is its evidence.
- The false claim in `confirmRetarget`'s comment — that the launch path "has no
  equivalent hole because its `canAct` gates the only click it has" — is removed or
  corrected. (Being corrected in PR #345 itself, referencing this issue.)

## Fence (may touch ONLY)

- `packages/web/src/connect/wizard.tsx`

Scope within those two files: in the first, confirmLaunch and the wizard-launch-confirm
button only; in the second, the launch describe block only.

**EXPLICITLY OUT, owned by the concierge law:**
packages/web/src/concierge/explicit-invocation-law.test.ts pins the confirm buttons'
wiring with a data-testid-then-onClick regex, which rejects any attribute inserted between
those two tokens. Order onClick **before** disabled in the JSX, as PR #345 did — do not
loosen the law to fit the edit, and do not edit that file.

*(Paths in this section's prose are deliberately unbackticked: scripts/fence-lint.sh reads
every backticked token on a line containing the word "fence" as a CLAIM, and a token with
no file extension reads as a directory claiming everything beneath it. Verified
2026-09-09 — as written before, this issue claimed a bare wizard-launch-confirm.)*

### LINKED ISSUE #359: prd14: the postinstall-escape law is bounded by its own cost, not the runner's load

### LINKED ISSUE #376: prd14 w2: the two comparison parsers accept and refuse the same bytes
## Fence (may touch ONLY)
- `packages/server/src/comparisons/artifact.ts`
- `packages/server/src/comparisons/artifact.test.ts`
- `packages/server/src/comparisons/store.ts`
- `packages/server/src/comparisons/store.test.ts`
- `packages/server/src/comparisons/parser-agreement-law.test.ts` (new file)
- `packages/web/src/lab/compare/artifact.ts`
- `packages/web/src/lab/compare/artifact.test.ts`
- `packages/web/src/lab/compare/parser-agreement-law.test.ts` (new file)
- `packages/contract/src/fixtures/comparison-artifact/` (new directory — the shared byte fixtures both laws read)
- `docs/adr/0042-the-server-parses-a-comparison-artifact-with-its-own-copy.md` — a dated note only, never a rewrite

**CHANGELOG.md is deliberately not in this fence.** It was, briefly, on 2026-09-09;
scripts/fence-lint.sh 372 376 reported a real OVERLAP against #372 (prd51 w3, Ready), which
holds it. Nothing user-visible ships here anyway — neither comparison route has a caller
until #214 lands — so the changelog entry belongs to that wave, not this one. Removed before
dispatch rather than scheduling the rebase conflict.

The web run shape's type declaration is **out of scope**: it is the target the server moves
to, and it does not move in this issue. If it turns out it must, that is a fence widening
recorded on the PR before the change, as AGENTS.md requires. (Its path is written without
backticks deliberately — scripts/fence-lint.sh reads every backticked token in a Fence
section as a claim, so spelling out an exclusion in ticks manufactures the overlap it is
trying to disclaim. Verified 2026-09-09.)

## Blocked by

Nothing. #213 (prd14 w1) landed the module this issue repairs, and #339 (prd53 w3)
merged the shape it must adopt — both are on `main` at `88df0484`.

scripts/fence-lint.sh reports no real overlap against any open issue as of
2026-09-09. #330 (prd53 w5, In review) holds `packages/web/src/lab/the-lab-guide-law.test.ts`,
which is outside this fence.

## Why

ADR-0042 chose a hand-ported second parser over a shared schema and booked the cost in
its own Consequences:

> Bad: two parsers of one shape. A version bump lands in two files, and drift between
> them is a defect no law currently catches. The test that asserts the refusal string
> exactly, on each side, is the only tripwire.

**That drift is now real, and the tripwire did not catch it** — it pins the *version*
sentence, and the drift is in the *run vocabulary*. #213 landed at 04:19 on 2026-09-08
and #339's fix commit at 04:02; the two changed one shape in two files in the same hour,
neither able to see the other.

EXECUTED on `main` at `88df0484`, 2026-09-09 — the divergence is live today:

- `packages/server/src/comparisons/artifact.ts:94` — `if (typeof value.value !== 'number')
  throw … 'complete run ${id} is missing a numeric value'`. No `verdict` at all.
- `packages/web/src/lab/compare/artifact.ts:82-97` — `complete` **requires**
  `verdict: 'pass' | 'fail'`, accepts `value: number | null`, preserves `note` and
  `detail`, and refuses the retired `failed` status by name.

Four consequences, from the round-trip probe recorded on #214 (2026-09-08) and
re-confirmed against the current tree:

1. **The ordinary case is rejected.** A completed run judged with nothing booked
   (`value: null`) fails the server parser, and `api/lab.ts` returns it as a **400** —
   so an arm of three passes with no cost booked cannot be saved at all. That is #339's
   own `floor-agreement-law.test.ts` case B.
2. **The stored case is worse, because it succeeds.** `store.ts` serialises the *parsed*
   input, and the server's `parseRun` reconstructs `{ id, status, value }` — `verdict`,
   `note` and `detail` never reach the disk. The web parser then refuses what comes back:
   an artifact written with a 200 that the surface cannot reopen.
3. **The retired status is persisted.** The server accepts `status: 'failed'`; the web
   refuses those same bytes by name. A within-version refusal, which nothing currently
   asks for.
4. **A pending run's `note` is dropped at birth.** prd-53 ruling 3 makes that note the
   whole reason a pending dot is never one the reader has to guess about. This one
   predates #339.

Separately, and on the same line: **`Infinity` is admitted by the write check.**
EXECUTED (recorded on #214, re-confirmed):

```
1e400 parses from JSON as : Infinity | typeof: number | isFinite: false
write check refuses?      : false        <- admitted, 200 returned
on-disk value             : null         <- JSON.stringify(Infinity)
read check refuses?       : true         <- permanently unreadable
```

`typeof value.value !== 'number'` is true of `Infinity`, so a save returns 200 and
produces an artifact that can never be read back. Not reachable from a browser
(`JSON.stringify(Infinity)` emits `null`, which the route 400s cleanly), but reachable
from a hand-crafted request — `1e400` is a legal JSON number literal. The web copy has
the same hole at `artifact.ts:84`.

## Definition of done

- `packages/server/src/comparisons/artifact.ts` **adopts the web's run shape**:
  `complete` requires `verdict: 'pass' | 'fail'` and accepts `value: number | null`;
  both statuses preserve `note`; `complete` preserves `detail`; `failed` is refused by
  name with the web's **exact** sentence. Byte-identical refusal messages is the
  standard ADR-0042 sets for itself.
- `store.ts` no longer discards `verdict`, `note` or `detail` on the way to disk. A
  saved artifact round-trips through the web parser unchanged.
- **`Infinity` is refused by both copies, in the same commit** — `!Number.isFinite(...)`,
  with a refusal sentence identical on each side. Moving one copy alone falsifies
  ADR-0042's identical-behaviour premise, which #213's verify round confirmed true
  (57 forms through both parsers, 0 divergences).
- **The law ADR-0042 says does not exist.** One committed fixture set of bytes under
  `packages/contract/src/fixtures/comparison-artifact/`, read by a law on each side, that
  asserts both parsers accept and refuse the same things **with the same messages**. A
  shared JSON fixture keeps it honest without either package importing the other's
  source, which is the constraint ADR-0042 rests on.
  `packages/web/src/lab/floor-agreement-law.test.ts` is the in-tree precedent for a law
  that reaches across a package boundary.
- Three existing fixtures reconciled in the same change — `comparisons/artifact.test.ts`,
  `comparisons/store.test.ts` and `api/lab.test.ts` each build a `status: 'failed'` run.
- ADR-0042 gains a **dated note** recording that its named tripwire was insufficient and
  what now covers it. Never a rewrite; the log is append-only and a number is an identity.
- The artifact stays at `version: 1`. Widening what v1 *carries* — the outcome shape —
  is **#347's** business and is explicitly not done here.

## Considered: move the shape to `packages/core` instead

ADR-0042's own escape clause — *"the right move the day a third consumer appears or a
`version: 2` is written — and that day it is a new record, superseding this one."*
Neither condition has fired: there is still no third consumer, and v2 is #347's. Doing it
here would make this issue a package move rather than a defect fix, and would land the
move without the v2 design that is supposed to motivate it. Flagged so it is a choice
rather than an omission; if the operator prefers it, this issue is the wrong shape and
should be rewritten, not widened.

## Sibling case

The `Infinity` hole and the run-vocabulary drift are **the same defect on two axes** of
one function, and fixing one is exactly the miss this repo names most often. Both copies,
both axes, one commit. The second sibling is `store.ts`: a parser that accepts the right
fields still loses them if the serialiser reconstructs a narrower object — check what
reaches the disk, not only what passes the check.

Revert the server's `verdict` requirement and re-run: the agreement law must redden. A
test that only round-trips a freshly-written artifact through **one** parser survives
every defect above — that is precisely how this drift reached `main` green. The
assertion that matters is the same bytes through both parsers, and the refusal
*sentence* compared, not merely that both threw.

gabriel-canaan:
Three prd-14 issues, built as one wave in three lanes on 2026-09-09, assembled
flat onto current `main` and gated as a bundle.

Closes #352
Closes #359
Closes #376

## Read this first — one commit's repair has NOT been independently reviewed

`fix(comparisons)` (#376) went through a full round of independent review, came
back **not safe to merge**, and was repaired. The convention this operator works
to is that such repairs are themselves an unreviewed change and go back to
independent reviewers before the work is called ready. **That pass was not run**,
by an operator decision taken with the cost stated: the round counter refused a
second round (its cap is 1 when every finding was test-only), and the override
was declined.

So, precisely:

- the **fix** to #376 was read by two independent reviewers;
- the **repair of that fix** — four changes to two law files, listed below — was
  read by nobody but its author.

The operator's review ledger reports that commit UNVERIFIED, and it is left that
way deliberately rather than recorded away. A reviewer should start there.

The round it replaces was itself half-depth: one reviewer returned the P1 that
drove the repair, and the other died producing an empty report.

## The commits, against the review that read each

| commit | issue | review record |
|---|---|---|
| `test(lab)` | #359 | orchestrator only, under the cost guard for a trivial diff — test-only, one timeout literal and its comment. Ledger: READY. |
| `fix(connect)` | #352 | 2 independent reviewers (Opus + Codex, high effort) on `b48777a8` → NOT READY, 3 findings. All closed; the fix-back is prose-only and re-read by the orchestrator under the cost guard. Ledger: reviewed, single reviewer. |
| `fix(comparisons)` | #376 | 2 independent reviewers on `ed0a68ca` → NOT READY, 2 must-fixes, both fixed. 1 reviewer on the amended head → NOT READY, 1 P1, fixed. **The repair of that P1 is unreviewed — see above.** Ledger: UNVERIFIED. |

## What is the sibling case?

Every one of these three issues *is* a sibling case — each was found because a
fix landed on one half of a pair and left the other:

- **#352** — #216/PR #345 fixed the arm/confirm `live` hole on the *retarget* and
  left the structurally identical one on the *launch*, with a comment asserting
  the launch had none.
- **#376** — ADR-0042's Consequences booked a tripwire for the two parsers
  drifting; the tripwire pinned the version sentence and the drift was in the run
  vocabulary. Both repairs here are applied to **both** copies for that reason.
- **#359** — `packages/web/src/scene/world.test.ts` had the identical starvation
  shape and was fixed at `34f5a172` (#357, merged as PR #361). This is the third
  instance of that precedent, not a new one.

Within this PR: #352's guard is deliberately *not* symmetric with the retarget's
— it re-checks `live` and not `canAct`'s other two conjuncts. That gap is real,
cannot spend (the server refuses an unimplemented harness first), and is filed as
**#379** rather than folded in.

- **#359** — none is claimed. The change is a timeout literal; there is no
  assertion to mutate, and the commit says so rather than manufacturing one.
- **#352** — measured and disclosed both ways: removing `disabled={!live}`
  reddens the new test (1 failed / 59); removing `confirmLaunch`'s internal
  `if (!live) return` leaves the whole `@rhizomorph/web` package green, because
  `fireEvent.click` never reaches a disabled control. The test docstring says
  which half it carries and does not claim the other.
- **#376** — eight certified mutations on the repair, each caught by the
  assertion it was aimed at, verified by reading the failing test name rather
  than the exit status:

  | mutation | caught by |
  |---|---|
  | `'run is missing id' + String(…)` (server) | the scanner's new trailing-token check, at collection |
  | `` `run ${…} is missing id` `` (server) | injectivity — 1 failed / 43 |
  | `` `${String(…)}` `` (server, and web) | injectivity — all 20 refuse fixtures become ambiguous |
  | `('run is missing id', String(…))` (server) | the trailing-token check, at collection |
  | throw → in-file helper + a sibling module carrying it (server) | axis C alone — 1 failed / 42 |
  | throw relocated **out** of `artifact.ts` (server, and web) | injectivity + axis C — 2 failed / 40 |

  What the repair changed: coverage credit is now **injective** (a refuse fixture
  must match exactly one throw site, so "matched" means "thrown by"); the site
  scanner **fails loudly** when the argument is not one literal; a new **axis C**
  pins the module's construction sites to `artifact.ts` so a relocated throw
  reddens; and the docstring's axis count and over-strong closing claim are
  corrected, with the three remaining limits written down.

  The input class was enumerated as a table before the repair, and the table was
  **incomplete** — running the repair found the missing row. The first draft of
  the trailing-token check demanded `)` immediately and reddened the law at
  collection on a legitimate prettier-wrapped site (`artifact.ts` offset 4631);
  three of that file's sites are written that way. Recorded in the commit body
  rather than quietly corrected.

## Fence

No widenings. `scripts/fence-lint.sh 352 359 376` passes with no overlap. Both
repairs were audited against their fences before either was committed, with the
certified mutation proofs and the enumerated input class attached to each.

## The bundle's shape, stated because it is not the usual one

#376 is `prd14 w2`; #352 and #359 carry no wave — they are unwaved prd-14 bugs
filed from PR #345's review. They are bundled because the substantive test the
wave rule encodes holds: the three fences are disjoint, nothing in one depends on
another's code, and `fence-lint` agrees. #359 is also the one that unblocks any
load-probed landing — it held an unrelated one at 4 failures / 16.

## Gates, EXECUTED on the assembled branch

`VITEST_MAX_WORKERS=6 npm test` rc=0 — 459 test files, 8429 passed,
3 expected fail, 2 skipped, 0 failures.

## What this unblocks

#379 is explicitly blocked on #352 reaching `main`. #214 (prd14 w3) is blocked on
#376's fence over `packages/web/src/lab/compare/`, and its other blocker, #220,
closed on 2026-09-09.

---

## Declared review gap

> the REPAIR of #376's fix-review P1 was not sent back to independent reviewers — the round counter refused a second round (cap 1, findings test-only) and the operator declined the override. The fix was reviewed; its repair was not. See the first section of this body.

```
  UNVERIFIED  ff1a23d5  fix(comparisons): the two comparison-artifact parsers accept and refuse the same bytes  (#376)
```


ciaran-slow (approved):
## Approve, with one fix pushed — and my own commit flagged unreviewed

Reviewed commit-by-commit in a clean clone, merged onto current `main`
(`a7820bc1`). The PR body asks a reviewer to start at the declared gap, so that
is where most of the effort went. Leading with what verified in the author's
favour.

### `test(lab)` (#359) — EXECUTED, clean

A timeout literal against AGENTS.md's "never fix a flake by widening a timeout".
The commit argues the rule is about masking a *race* and there is none — the
assertion is `existsSync(escapeTarget) === false` after an awaited install, with
no wall-clock assertion. That reading is correct, and the three precedent claims
all check out:

| claim | verified |
|---|---|
| `scene/world.test.ts` carries `}, 30_000)` | yes, line 217 |
| `app/streamState.test.ts` carries `}, 60_000)` | yes, line 260 |
| not enrolled in `gate.sh`'s `@gate-timing` set | yes — see below |

The last one was worth checking rather than reading, because the commit
*introduces the string* `@gate-timing` into the file, in prose saying it is not
enrolled. An unanchored detector would have enrolled it by that sentence — the
repo has that exact bug shape on record elsewhere. It does not here:
`gate.sh:635` greps the literal `// @gate-timing`, and the comment uses
backticks. Running the gate's own detector over this branch returns
`SceneView.test.tsx`, `marks.test.ts`, `perf.test.ts` — and not this file.

### `fix(connect)` (#352) — EXECUTED, both mutation claims reproduce exactly

The honest part of this commit is that it discloses a mutation its test does
**not** catch. Re-ran both against the committed tree rather than trusting the
report:

|---|---|---|
| remove `disabled={!live}` from `wizard-launch-confirm` | red, 1 / 59 | **red, 1 failed / 59 passed** |
| remove `confirmLaunch`'s `if (!live) return` | green, 60 / 60 | **green, 60 / 60** |

Exact. And the test's own docstring says so in the file, where the next reader
will actually meet it — "A mutation that removes that guard and keeps only
`disabled` leaves this test green… this test is not the evidence for it, and a
later reader must not read it as such." That is the right way to ship a
half-covered fix.

Citations spot-checked, since a renamed symbol leaves a green suite and a lying
comment: `HarnessNotAvailableError` is real at `concierge/launch.ts:143` (thrown
:334, :346), `#379` is open and titled exactly what the comment says it tracks,
and `live` is a genuine prop of `ConductorStep`.

### `fix(comparisons)` (#376) — the declared gap, reviewed here

The PR flags the *repair* of this commit's P1 as read by nobody. R1–R3 hold up
under fresh mutations of my own (not the seats', not the body's):

| my mutation | axis | result |
|---|---|---|
| a second throw site with a message duplicating `run is missing id` | R1 | red, 1 / 43 — "each refuse fixture must match exactly one throw site" |
| `new ComparisonArtifactError('run is missing id' + String(id))` | R2 | throws at **collection**, naming offset and the offending `"+"` |
| a sibling module in the same directory carrying the marker | R3 | red, axis C by name |

Baseline matches the body's: 14 sites, 20 refuse messages, 24 fixture cases
(20 refuse / 4 accept), and the two law files are byte-identical apart from
prose and the fixture path depth — diffed, not assumed. The ADR-0042 note is
append-only and dated, as the log requires.

### The one gap I found — fixed in `c698ba0b`

**Axis C listed the module's directory non-recursively**, so a throw site one
directory down was invisible to all three axes. EXECUTED on `ff1a23d5`:

```
packages/server/src/comparisons/sub/relocated.ts  -> 43 passed (green)
packages/web/src/lab/compare/sub/relocated.ts     -> green
```

while the same file as a *sibling* reddens — the case R3 was built and tested
against.

The gap is bounded, which is why I fixed it rather than filed it.
**Relocating** an existing covered throw into a subdirectory is already caught,
by R1 rather than C — its fixture message then matches zero sites. EXECUTED:
moving `run is missing id` into `sub/helper.ts` reddens 1 / 42 on the
injectivity row. What slipped is only a **brand-new** site one level down, which
carries no fixture to go missing. Neither module has a subdirectory today, so
nothing was mis-pinned on this tree.

The fix makes the scan recursive on both sides and normalises entries to `/` so
the failure message is stable on `windows-suite`. After it, the finding mutation
reddens by name on both copies and the same-directory sibling still reddens.
It is the lesson AGENTS.md already records for the fixture-hygiene law — scope a
guard by what a file *is*, not by where it sits.

### My own commit is unreviewed, and I am not laundering that

`c698ba0b` was read by nobody but me. It sits in the file whose repair this PR
already reports UNVERIFIED, so the honest statement is that the declared gap is
now one commit **wider**, not closed. The operator's eyes are the second pair —
AGENTS.md's "never review your own fix into main" is satisfied by the merge
being manual, not by this review.

### Gates, EXECUTED post-merge onto `a7820bc1`, with my fix included

`npm run typecheck` rc=0 · `npm run lint` rc=0 (983 files) ·
`VITEST_MAX_WORKERS=6 npm test` rc=0 — 459 files, 8425 passed, 3 expected fail,
6 skipped, **0 failures**.

### Not run

The load probe — the operator's step. `#359`'s whole point is that the install
test failed under it, and this review does not re-measure that.

### On the bundle's shape

`#376` is `prd14 w2`; `#352` and `#359` carry no wave. The substantive test the
wave rule encodes does hold — three disjoint fences, no code dependency between
them, `fence-lint` clean — and the body says so rather than hoping nobody looks.
Reasonable.

### CI, re-triggered by my push and confirmed by job name

`gh pr checks 393` — **all 10 legs pass**, re-run against `c698ba0b`:
`build-test-boot` × 3, `pack-smoke` × 6 (ubuntu / macOS / Windows, both node
legs), and `windows-suite`. That last one carries the signal for my own change:
it walks directory entries and splits on `path.sep`, so a Windows separator bug
would land there and nowhere else.

**Verdict: LANDABLE.** Not merging — merges here are the operator's, manually on
GitHub.

Issue #391, opened 2026-09-09
prd17 w8: a swallowed close fault leaves a trace


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/server/src/log/gate-verdict-derivation.ts`
- `packages/server/src/log/gate-verdict-derivation.test.ts`
- `packages/server/src/server/poll-loop.ts`
- `packages/server/src/server/poll-loop.test.ts`

`packages/server/src/collectors/beacon/` is EXPLICITLY OUT — owned by prd-27.
`packages/core/` is EXPLICITLY OUT: whatever carries the trace must not become
a new event family or a seventh key on the TelemetryState shape, and deciding
that it should would be a ruling this issue does not hold.

## Blocked by
**`#280`, until it lands on `main`.** The code this amends is complete, but it
lives only on that issue's lane branch — a lane dispatched for this before
`#280` merges would branch from a `main` where the derivation module does not
exist at all, so both of the log-directory paths this issue claims above would
be missing. Not blocked on any decision; blocked on a merge.

## Why
Found by round 6 of `#280`'s review (2026-09-10), by the Sonnet 5 seat, and
left deliberately unfixed there because the fix reaches past that issue's
fence into the recording seam.

`deriveGateVerdict` opens the beacon sidecar to recover a landing's verdict.
Round 5 fixed a real defect — a rejecting `close()` in the read's `finally`
escaped the function as a rejected promise, past both catches, into
`poll-loop.ts`'s unguarded per-event seam. The repair was to catch it. The
catch body is **empty**: no log, no counter, no flag.

So the fault is now survivable and completely invisible. A persistently
failing mount leaves zero trace while this fires for every `beacon.received`
on every tick, and the operator's only symptom is verdicts that are correct
whenever they appear.

**The local convention already says what to do, which is what makes this a
deviation rather than an open question.** Two sibling swallows in this same
package both leave something a downstream reader can see:

- `packages/server/src/recorder/session-log-writer.ts`'s
  `dropTrailingPartialLine` swallows a read failure and returns `false` — a
  signal its caller acts on.
- `packages/server/src/recorder/session-recorder.ts`'s `advanceFold` swallows a
  reduce throw and sets `foldDesynced = true`, which `foldSoFar` reads and
  repairs from.

Neither is silent. The close catch is the only one of the three that is.

Note the asymmetry this must preserve: the swallow itself is **correct** and
is not what this issue changes. By close() time the bytes are already read and
[...]

PR #377, opened 2026-09-09
docs(coupling): the contract law's cross-package pins are a registry entry (#358)

ciaran-slow:
Closes #358

One entry added to `.swarm/coupling.txt`, plus an amendment to the entry beside
it. One file, **five commits**, net +2/−1 against `main` — the issue's whole
declared fence, no widening.

Four of those five are review commits that landed on this branch after it opened,
each narrowing an over-generalisation in the one before: the original entry said
the two laws "redden TOGETHER" without qualification, which is false for half of
`RouteClass`. The chain is `b5d6630c` → `6c9e2978` → `f0299179` → `03f6fde0` →
`08663fc8`, and the per-class claim in the entry today is measured for all four
classes rather than generalised from the one the falsifier planted.

## Falsifier first, because the wrong answer was plausible

The issue required this before the entry was written, and it was the right order:
the plausible wrong answer was that this law **derives** its expectations from
the tree and therefore needs no entry at all.
`server/collector-wrap-boundary.test.ts` is the precedent the registry
deliberately records as **NOT** an entry on exactly those grounds, so this had to
be settled by mutation rather than by reading.

It does not apply. **EXECUTED** in a throwaway worktree — planting one
`gated-read` row in `packages/server/src/api/index.ts`:

```
AssertionError: expected [ '/api/concierge/repos', …(16) ] to have a length of 16 but got 17
AssertionError: expected [ '/api/concierge/repos', …(13) ] to deeply equal [ … (12) ]
  packages/contract  Tests  2 failed | 6 passed (8)
  api/route-class-law.test.ts  7 failed
```

It pins. It is a coupling, and the two laws redden **together**.

## The defect was completeness, not silence

The registered `api/route-class-law.test.ts` entry says *"ANY new route,
registered anywhere under api/, reddens this file from a directory the lane never
entered"* — and names only itself. So a lane reads it, fences that one file,
reconciles its pins, lands, and is reddened a second time from a package nothing
warned it about.

That entry is amended to point at the new one and to say **"fence both or
neither"**. An entry that reads as exhaustive is worse than one that reads as
partial.

## First entry written under Rulings 5, 6 and 7

prd-54's wave-0 rulings landed in #370, so compliance was checked **before**
writing rather than discovered after:

- **Ruling 7** — all four backticked assertion spans occur verbatim in the cited
  file. Verified mechanically, not by eye.
- **Ruling 5** — all three cross-package citations are written **repo-relative**.
[...]

PR #395, opened 2026-09-09
feat(server,core): a landing's verdict is derived and recorded as a typed event (#280)

### LINKED ISSUE #280: prd17 w6: a landing's verdict is a gate.verdict event, not only a beacon

gabriel-canaan:
Closes #280

prd-17's last built wave. A landing's verdict has been on the record as a `beacon.received` occurrence since #274 — writer, kind, lane, the line's digest, the file and the offset. This turns that sidecar back into the typed `gate.verdict` event ruling 1 defined and #219 landed, whose arm in `reduce.ts` had returned state unchanged ever since.

`packages/server/src/log/gate-verdict-derivation.ts` reads the verdict back out of the file and offset the beacon points at, and `poll-loop.ts` derives-and-records at RECORD time — so the log is self-sufficient and a pruned or never-exported `beacons/` directory cannot cost a fold its verdicts.

**Two digests, and keeping them apart is the point.** `beacon.received.digest` hashes the beacon *line* and is the integrity check; the line's own `outputDigest` hashes the *gate's output* and becomes `gate.verdict.digest`. The line digest is verified and discarded **before** `outputDigest` is ever read, so the two cannot be conflated. Verified against `scripts/gate.sh`'s `emit_gate_verdict` and `gateVerdictPayloadSchema` directly rather than taken from the issue's table.

`gate.verdict`'s fold is deliberately still nothing in `SessionState`, and the arm now says so: the UI dividend is prd-18's per this PRD's own Non-goals. prd-27's beacon collector is untouched and still emits `beacon.received` for every well-formed line, gate lines included.

## Fence widening, recorded on the issue before the change (2026-09-09)

Bounded to the `gate.verdict` recording seam only — no other collector or poll-loop behaviour.

**Why.** The fence as groomed made the wave's own declared deliverable impossible: the PRD's wave-6 paragraph says this is where `gate.verdict` finally emits, and the original three paths excluded every surface that could emit one, so the derivation would have shipped correct and uncalled. Operator ruling the same day: fix the fence, not the PRD — the document is the plan of record. `eras.test.ts` is included because its assertion that `gate.verdict` is "still unemitted anywhere" becomes false by construction once this lands.

## What is the sibling case?

**The sidecar is a file and the log is the record; they have different lifetimes.** The miss would be a derivation that is correct on a live machine and wrong on a replay of an exported recording whose `beacons/` directory was never carried. So absent, empty, offset-past-end and truncated all refuse **identically** — one `size <= offset` check catches the first three, a separate newline check catches truncation — and recording at derive time means a fold never needs the sidecar again.

**The one that actually bit, twice, one line apart.** Round 4 found the bounded read had no `catch` of its own, so a real I/O fault threw out of a function whose contract promises it never throws — into `poll-loop.ts`'s unguarded per-event seam, which per ADR-0029 would have re-derived the whole batch every tick for as long as the fault persisted. Round 5 then found the *sibling*: `close()` in that same `finally`, whose rejection **replaces** the value the `catch` had already returned and so escaped past both catches by the same route. All three descriptor calls — `open`, `read`, `close` — now have their own handler. This is the defect shape AGENTS.md names as the repo's most common, and it recurred inside the fix for its own first instance.

**Deliberately still open**, and filed rather than hidden: the swallowed close fault leaves no log, metric or flag — **#391**. Also deferred, unchanged: the seam swallowing a refusal's reason, the stat/open TOCTOU, and a hardlink defeating containment.

The two the issue named, both certified and reverted before committing:

- **Skip the digest check and trust the file.** Every test with an unmodified sidecar passes, so the assertion has to tamper with the bytes *after* the event was recorded and prove no verdict is produced. It does.
- **Hard-code `held: false`.** A fixture whose landings all succeeded passes at any value, so the fixture is a HELD landing (`reason: suite-red`) and `reason` is pinned too — a hard-coded `reason: 'clean'` reddens it.

**Fourteen mutations are certified across six rounds.** The five from round 5 are the ones worth naming here, because three of them are mutations that had *already survived* an earlier round's test and only redden now:

| mutation | before | after |
|---|---|---|
| remove the `close()` try/catch | — | reddens (1/30) |
| classify from the stale `stat` size | — | reddens (1/30) |
| disable the `lane === null` guard | 13/13 **green** | reddens (1/30) |
| `size <= offset` → `size < offset` | 13/13 **green** | reddens (1/31) |
| route all newline-free reads through the cap branch | 13/13 **green** | reddens (1/30) |

Those three green cells are the finding worth reading. Round 4 added a 13-row refusal table to "pin the contract once" after three defeats on that axis — and asserted only the outcome plus a non-empty reason, which **cannot fail for the reason a row names**. Disabling the `lane === null` guard still refused, one stage later, at schema validation. Every row now declares its expected reason and the runner matches it; `size <= offset` needed a fourteenth row as well, because a 1-byte file at offset 5 is refused by both spellings and only `size === offset` tells them apart.

**One mutation is honestly uncertifiable and is not claimed:** the bounded read's window cap. A whole-file read and a capped read produce byte-identical output for every fixture this suite can hold — the difference is resource bounds, which a unit-scale test cannot redden.

## Review record — read this before trusting the green

Six rounds ran. Rounds 1–5 reviewed the builder's work (Claude Sonnet 5). Round 5's findings were repaired by the session that verified them, so **round 6 existed solely to put independent eyes on those repairs**: two seats, different families, both at xhigh — Sonnet 5 (deliberately swapped down from Opus, since Opus authored the repairs) and Codex `gpt-5.6-terra`. **Both returned safe to merge**, with two non-blocking findings and no repair regression.

Two limits on that, stated rather than left to be assumed:

1. **The Codex seat could not run the full suite** — its sandbox denied `spawnSync git`/`node` and loopback `listen`, reddening 35 suites for environmental reasons. Its typecheck, lint and focused 31/31 runs are real; its full-suite result is not evidence and is not counted as one. The Sonnet seat's full run is, and it reproduced 458 files / 8373 passed exactly after re-running four of round 5's five mutations against the committed file rather than trusting the report.
2. **Round 6 reviewed `7930526d`; this branch is further on.** Its two comment-only findings were then fixed, which is what `--unverified` below records. The delta from the reviewed head is comments only — proven, not asserted: filtering that diff for changed lines which are neither comments nor blanks returns nothing. A comment with no behavioural reach does not earn another round of seats, but "reviewed" is a claim about specific bytes and these are not the bytes that were read.

Full findings and evidence for every round are on #280.

## Verification

`npm run typecheck`, `npm run lint` (983 files) and `VITEST_MAX_WORKERS=6 npm test` all green on this branch — 458 files, 8373 passed, 3 expected-fail, 2 skipped (both pre-existing). `/proc/loadavg` was read before trusting the run. No `package-lock.json` changes.

Rebased onto current `main` immediately before opening; the pre- and post-rebase diffs are byte-identical, and `main`'s intervening commits (#377/#358, the coupling registry entry) touch no file this commit changes. The gates above were re-run after that rebase, not carried over from before it.

---

## Declared review gap

> Round 6's independent re-review read 7930526d. This is still that one commit, amended once since - a comments-only fix of round 6's own two findings - and then rebased onto current main, so it is now bd87886e. The delta from the reviewed bytes is comments only, and that is proven rather than asserted: filtering the 7930526d..34151acc diff for changed lines which are neither comments nor blanks returns nothing, and the pre- and post-rebase diffs are byte-identical. A comment with no behavioural reach does not earn another round of seats, but these are not the bytes that were reviewed, and that belongs in front of a reviewer rather than left to be discovered.

```
  UNVERIFIED  bd87886e  feat(server,core): a landing's verdict is derived and recorded as a typed event (#280)
```


ciaran-slow (approved):
## Approve — verified independently, one limit stated

Reviewed commit-by-commit in a clean clone, merged onto current `main`
(`a7820bc1`) before every measurement below. Leading with what verified in the
author's favour, because most of it did.

### EXECUTED — verified in the author's favour

**The macOS path-shape trap is handled.** I went looking for the failure
AGENTS.md warns about — a raw-vs-canonical comparison that passes vacuously on
ubuntu and fails only on macOS, where `/var` → `/private/var`. `isInside(dir,
realFile)` compares an already-canonicalized `realFile` against a `dir` that
the derivation never canonicalizes, which is exactly the shape of that bug.
It is not one: `isInside` canonicalizes **both** sides internally
(`paths/containment.ts:isInside`), through the same `realpath`. No finding.

**The two digests really are computed the same way.** The PR's central claim is
that `beacon.received.digest` and `outputDigest` cannot be conflated. The
parity that actually matters is between the collector and the derivation, and
it holds byte-for-byte:

```
collector.ts:169   createHash('sha256').update(line.text, 'utf8').digest('hex')
derivation         createHash('sha256').update(lineText,  'utf8').digest('hex')
```

**The no-throw contract holds under faults the suite does not cover.** Rounds 4
and 5 closed `read` and `close`. I probed for a third sibling, since
`isInside(dir, realFile)` is the one filesystem-touching call in the function
that is *not* individually wrapped, and `canonicalize` throws on non-ENOENT
`realpath` errors and on `ELOOP`. Wrote a throwaway test against the committed
file:

| hostile input | result |
|---|---|
| `EACCES` on an ancestor of `dataRoot` | refused, did not throw |
| `dataRoot` a symlink loop through a missing dir (`./missing/../link`) | refused, did not throw |

Both refuse. The wrapped `canonicalize(filePath)` above dominates — it walks the
same ancestors and catches first — so `isInside` is not reachable with a throw
by these routes. **I reasoned this finding, tested it, and it collapsed; it is
recorded here rather than shipped as a finding.** The residue is a TOCTOU
between the two `canonicalize` calls, which is narrower than the stat/open
TOCTOU already deferred on the issue and not worth a new one.

**Three of round 5's mutations re-run against the committed file**, not taken
from the report — including the two the PR body flags as having *survived* an
earlier round, which is the interesting claim:

| mutation | result | row it failed on |
|---|---|---|
| disable the `lane === null` guard | red, 1 failed / 30 | `refuses: 'lane is null'` |
| `size <= offset` → `size < offset` | red, 1 failed / 30 | `refuses: 'the sidecar size is EXACTLY the recorded offset'` |
| remove the `close()` try/catch | red, 1 failed / 30 | `an I/O error on close does not escape…` |

Each failed on the row it was aimed at, by name. Round 5's repair — every
refusal row declaring its own expected reason, and the runner asserting a
refusing row *has* one (`expect(reason, 'a refusing row must declare which
reason it expects').not.toBeNull()`) — is what makes those bite, and it is real
in the file, not just in the body.

### Gates, EXECUTED post-merge onto `a7820bc1`

`npm run typecheck` rc=0 · `npm run lint` rc=0 (983 files) ·
`VITEST_MAX_WORKERS=6 npm test` rc=0 — 458 files, 8369 passed, 3 expected fail,
6 skipped, **0 failures**. `gh pr checks 395`: all 10 legs pass, `windows-suite`
included.

### One limit, non-blocking — the declared gap cannot be checked

The `--unverified` note rests on a specific, falsifiable claim: that the delta
from the reviewed bytes is comments only, "proven, not asserted", by filtering
the `7930526d..34151acc` diff. **I could not verify it.** None of
`7930526d`, `34151acc` or `d39383ac` is reachable in a fresh clone — they were
lane-local and did not survive. So the one claim the review gap is discharged
by is the one a later reader cannot re-run.

I am not treating that as blocking, because the claim is independently
*corroborated* by what I could run: the committed file behaves correctly under
its own suite, under three re-derived mutations, and under two fault injections
the suite does not contain. But it is the repo's own "cite a SHA, not an issue
number" lesson one turn on — a SHA that only ever existed in a lane is not a
citation either. Worth a `git range-diff` against a pushed ref next time, or
quoting the filtered diff inline.

### Not run

The load probe (`gate.sh` with batches) — that is the operator's step. No
`@gate-timing` file is touched by this commit.

**Verdict: LANDABLE.** Not merging — merges here are the operator's, manually
on GitHub.

Issue #410, opened 2026-09-10
prd51 w4: a line the fold cannot read does not discard the records around it


gabriel-canaan:
## RULED — dispatchable
**Ruling 16**, `docs/prds/prd-51-the-split.md`, amendment of 2026-09-10: **answer (2), store the
line unparsed, plus answer (3), fold per actor.** `readEventLineLenient` already returns the
envelope's `type` and `ts` and the line verbatim; `events.type` carries no CHECK, no enum and no
foreign key, so the row lands with **no gap and no migration**, and ruling 5's
identical-`chainDigest` claim needs no union with a second relation. Answers (1) and (4) are
rejected there with reasons; do not re-litigate them here.

Two clauses the ruling adds that this issue did not state:

- **`unknown` only. `malformed` still refuses, loudly.** That is this issue's falsifier, and the
  boundary already exists — `parseEventLenient` returns `unknown` for an intact envelope the union
  rejects and `malformed` for what is not an event at all. A corrupt line in an fsynced journal is
  a fault in the journal, not skew between two builds.
- **The deferred falsifier was run by the review of the amendment, and it FAILED.** `projectionsFor`
  increments `spend.events` unconditionally and `costUsdOf` reads `payload.costUsd` with no
  `row.type` guard at all, so an `unknown-shape` `llm.cost` put **$42.50 from an unvalidated payload
  into a money projection**. Ruling 16 was amended rather than abandoned: `spend.events` counting an
  unfoldable row is ruled **correct** (it is an event), and no projection may derive a value from a
  payload the build refused to validate. `EventRow` carries an `unfoldable` marker to say which.
- **The sibling the first draft missed, and the one to watch while building:** it split `unknown`
  from `malformed` and never split `unknown-type` from `unknown-shape`. `unknown-shape` means the
  type **is** known, so it walks straight through every guard written against `row.type` —
  `agentStateOf` and `dirtyFiles` are the same shape as `costUsdOf` here.
- **A mutation is owed:** remove the `unfoldable` check from `projectionsFor` and the case asserting
  no unvalidated payload reaches a projection must go red.

## Fence (may touch ONLY)
- `packages/team/src/fold/worker.ts`
- `packages/team/src/fold/worker.test.ts`
- `packages/team/src/fold/row.ts`
- `packages/team/src/fold/row.test.ts`
- `packages/team/src/fold/cursor.ts`
- `packages/team/src/fold/cursor.test.ts`
[...]

Issue #407, opened 2026-09-10
prd55 w5: the R&D result has a schema, four events and their folds — a single occurrence is held back, a proposal varies one thing


KelliherL:
## Fence (may touch ONLY)
- `packages/core/src/lab/rd.ts` (new file)
- `packages/core/src/lab/rd.test.ts` (new file)
- `packages/core/src/lab/index.ts` (exports only)
- `packages/core/src/index.ts` (exports only)
- `packages/core/src/events/lab.ts`
- `packages/core/src/events/lab.test.ts`
- `packages/core/src/events/index.ts`
- `packages/core/src/events/events.test.ts`
- `packages/core/src/fixtures.ts`
- `packages/core/src/fixtures.test.ts`
- `packages/core/src/state.ts`
- `packages/core/src/state.test.ts`
- `packages/core/src/reduce.ts`
- `packages/core/src/reduce.test.ts`
- `packages/core/src/eras/eras.test.ts`

## Blocked by
Nothing — this is the core half of wave 5 and touches only packages/core, which no running lane enters; the server half (the engine, the CLI subcommand, the route, the two settings entries and the ADR) is the wave's second issue and ships after this one and after wave 4's series routes, on the same PR. Sibling lanes hold other surfaces: nothing under packages/server or packages/web is entered — if a new event type reaches an exhaustive map outside core (the server's lab route names fork.measured today), STOP and report it as a widening rather than editing it. The core test files above are fenced because a new event type moves what they enumerate (law 8): reconciled minimally, rewritten never deleted.
**Model:** sonnet. **Wave:** 5.

## Definition of done
prd-55 rulings 3 and 4 (the schema, the events, the folds); S5's data-source lines that belong to core.

- A fixed R&D result schema, in the events module's own idiom (zod, as the lab's other event payloads are): patterns[] each with a shape sentence, its source items, a count and heldBack; proposals[] each naming one pattern, one varies dimension (model | brief | checkpoint | gate), 2–3 arms differing only in that dimension, and a checkpointPick with the chosen checkpoint and every considered-and-rejected one with its reason.
- Two pure laws beside the schema, in packages/core/src/lab/rd.ts with no I/O and no zod (as laws.ts is): heldBack is count < 2, and a proposal against a held-back pattern, or one whose arms differ in more than one dimension, is refused with a stated reason — the reason is a sentence the surface can print verbatim.
[...]

PR #405, opened 2026-09-09
prd55 w2: the lane canvas paints with the scene's brushes, and reads only the lab's record

KelliherL:
Closes #385. prd-55 wave 2 — ruling 11 (refining prd-53 rulings 5, 7 and 8) and the Design calls section. One PR = one (milestone, wave); five commits, read in order.

## What becomes true

- **The lane canvas is a Canvas 2D drawing built from the scene's six pure brushes** — `geometry`, `palette`, `ribbon`, `contour`, `motes`, `heart` — each handed the lab's record and nothing else: `canvas/organism.ts` is the model (record → display list), `canvas/paint.ts` the hand that fills it and decides nothing, `LaneCanvas.tsx` the host. One ribbon per dispatch record, spine from the fork's x on an inset session axis to the run's node, width from booked cost on the organism's absolute log scale (thin floor for nothing booked, a pinch at the fork), root-mass tissue from contour layers on the tissue ramp, tip ink from the verdict (prd-53 ruling 5's mapping unchanged), motes only where a channel carries nothing.
- **The fan opens toward the free side past 60 %, the root keeps its true x** on the inset axis, and no node, polygon point, mote, stub or ring lies outside the drawing at any fork position from 0 % to 100 % — executed as a sweep at two widths.
- **A failed-to-dispatch arm is a stub in necrotic ink** — drawn on its own row, named with its error, never counted.
- **One facts panel per canvas** (`aria-live`), every run a real button in the tab order; hover, focus and click drive it; no native `title` anywhere; type only through the theme's tokens.
- **Two layers, no per-frame work**: the record layer repaints on record, size or theme change, the highlight layer on hover or selection; no loop, clock or motion query, so reduced motion has nothing to ask. The 60 × 3 cell is reported beside the 16.67 ms budget and never asserted (this box: layout 19.96 ms, paint 2.44 ms per pass).
- **The no-live-fleet law's scene exception is exactly six named modules** (plus the fold-bound five named as forbidden, with the reason); the count pin carries `canvas` 3 → 5.
- **The frame mounts the canvas at the height its ribbons need**; the branching diagram is a header glyph that draws runs, not arms, at ≤ 44 px, keeping its export name and root test id — and the adapter now tells it how many runs each arm has, so the live app draws them.
- **Empty, partial and held-back states first** (ruling 9): no dispatch records, a stopped launch, an unmeasured arm — each a fixture and a rendered test.

## Verified

[...]

PR #400, opened 2026-09-09
prd55 w1: the launch tells the whole truth, and measuring is a control

KelliherL:
Closes #383. prd-55 wave 1 — rulings 5 and 7, the specification's S2′. One PR = one (milestone, wave); eight commits, one per surface, read in order.

## What becomes true

- **The model list is the operator's.** `settings/registry.ts` declares `lab.models` (repo scope, kind record, seeded `opus · sonnet · haiku` on, a comment saying why it is a convenience and not a gate); the launch panel's per-arm model is a `<select>` over the enabled keys plus *other…*, which writes the typed model into the record.
- **The launch tells the whole truth.** *Runs per arm* and *ceiling override* travel in the launch body only when set (absent when unset — executed both ways); the estimate is asked for arms × runs and its basis says so; "rate cannot be established" renders in the dim ink, never the alarm red.
- **Measuring is a control.** Every experiment panel gains a gate-command field (default: the repo's verify command) and a *measure* button that posts to the lab's measure route only after one confirmation naming how many worktrees the gate will run in; success reloads the experiments; a refusal renders the server's message verbatim.
- **The launch record carries the requested arm count**, so the partial-launch line reads *k of N requested* — the denominator the wave-3 review found invented from dispatched + failed — and a newly launched experiment reaches Compare and Metrics without a reload.
- **The guide's launch-panel gap and measure paragraph say what is now true**, and `the-lab-guide-law` flips with them.

## Verified

- **EXECUTED — `npm run typecheck` 0 · `npm run lint` clean (990 files) · fenced suites green after merging `origin/main` (`df494011`) in:** `packages/web/src/lab` 304 tests, `packages/web/src/settings` 106, `packages/server/src/api/lab.test.ts` 65.
- **EXECUTED — full suite on the lane's Windows box:** every red is outside the fence — 29 files on `.windows-known-failures`, 4 more failing only on `EPERM: symlink` (this box lacks the privilege the runner has; all four byte-identical to main), and one contention flake (`retarget.test.ts`, 22/22 alone). The ubuntu legs are the read.
- **EXECUTED — six mutations red, then restored:** the source-file count pin without the new directory's row; the measure button's confirmation-grep law; the never-attempted-arms gap reverted; the reload wiring removed; the guide's partial-launch paragraph re-wrapped; the measure test's repo-root path.
- **Fence:** exactly the issue's files (audited against `git diff --name-only origin/main...HEAD`). No widening was needed.

## Notes for the reviewer

[...]

PR #417, opened 2026-09-10
prd-51 rulings 15 and 16: two wedges answered before either lane starts

### LINKED ISSUE #399: prd51 w4: a ledger line larger than the read window does not wedge the hand in silence
## RULED — dispatchable
**Ruling 15**, `docs/prds/prd-51-the-split.md`, amendment of 2026-09-10: **answer (1), skip and
record.** An oversized line is a line the hand cannot send, which is the category ruling A already
answers — ruling A is extended, not amended. The discriminator is `size - before.offset`, not the
window's contents, which resolves this issue's falsifier: a safe separation of *too long* from
*still being appended* does exist. `ActorSkip.kind` gains `'oversized'`. Answers (2) and (3) are

Three clauses the amendment adds that this issue did not:

- **The skip advances past the whole line, not past the window** — consuming 1 MiB of a longer line
  would leave the remainder to be read as a fresh line and ship a fragment under a position. The
  forward scan reads in **bounded chunks**; it is looking for one byte, and slurping the remainder
  to find it would reintroduce the allocation ruling 15 rejects option 3 for.
- **An oversized line with no terminator anywhere names a failure — that arm, and only that arm.**
  Found by the review of the amendment. A ledger whose last line is oversized and never terminates
  (writer died mid-line, tail truncated, session abandoned) hits EOF on every forward scan on every
  tick, so an earlier draft that said *"if EOF arrives first, the skip does not happen this tick"*
  recreated this issue's own reproduction exactly: `ok: true`, empty failures, cursor unwritten.
  When the remainder exceeds `MAX_READ_BYTES` and no newline exists between the cursor and EOF,
  emit a named `ShipActorFailure`. **This is not option 2 revived** — option 2 named a failure for
  every oversized line and never advanced; this names one only where there is no line boundary to
  advance to. Together the two arms are what satisfy this issue's *"either it advances, or it names
  a failure"*.
- Every other actor in the pass still ships, exactly as ruling B already arranges.

Read the ruling before starting.

## Fence (may touch ONLY)
- `packages/server/src/shipper/cursor.ts` (`ActorSkip.kind` gains `'oversized'` — ruling 15)

## Blocked by
Nothing. #386 (prd-51 wave 3) is merged and the ruling is recorded.

## Why this exists
Finding 3 of the review on #386, reproduced independently before filing.

`shipActor` reads at most `MAX_READ_BYTES` (1,048,576) per actor per tick:

```ts
const wanted = Math.min(size - before.offset, MAX_READ_BYTES)
…
const scan = splitLines(…)
if (scan.lines.length === 0) return unchanged(null, null)
```

A single ledger line longer than that window contains **no newline inside it**, so `splitLines`
yields nothing and the pass returns `unchanged(null, null)` — no failure, no skip, no advance. The
next tick reads the same window and does the same thing, forever.

**EXECUTED** — one 1,048,715-byte line followed by one ordinary line, three consecutive passes,
every one identical:

```
pass 1: {"ok":true,"shipped":0,"skipped":0,"failure":null,"detail":null,"advanced":false,"cursor":null,"posted":0}
pass 2: {"ok":true,"shipped":0,"skipped":0,"failure":null,"detail":null,"advanced":false,"cursor":null,"posted":0}
pass 3: {"ok":true,"shipped":0,"skipped":0,"failure":null,"detail":null,"advanced":false,"cursor":null,"posted":0}
```

The cursor is never even written for that actor, nothing is posted, and the good line sitting behind
the fat one never leaves either.

**This is worse than the case ruling A already handles.** Ruling A names exactly this failure mode —
*"wedges the hand permanently on one line and denies the whole ledger from that byte onward"* — and
answers it with a recorded skip. But ruling A operates on a line that was successfully **split**;
this line is never split, so it reaches none of that machinery. The result is a wedge that reports
`ok: true` with an empty `failures` array, so `--status` and `doctor` show no fault at all. Ruling
A's case at least tells the operator something is wrong.

Nothing bounds line size upstream. `packages/core/src/events/beacon.ts` caps its own strings, and
nothing else does; `MAX_READ_BYTES` is referenced by no test in the tree.

Recorded as ruling 15. The three answers as they were weighed:

1. **Skip and record, as ruling A does.** Detect a window with no newline where more file follows,
   advance past the line, and record an `ActorSkip` naming the reason. Consistent with ruling A;
   costs one line of ledger silently-but-recorded, and needs care not to advance past a line that is
   merely still being written.
2. **Report a named failure and stay put.** A new `ShipActorFailure` (`line-too-long`) so the wedge
   is at least visible in `--status` and `doctor`. Honest, does not lose the line, but the actor
   stays wedged until a human acts — which for an append-only ledger may be never.
3. **Raise or remove the cap**, e.g. grow the read window when the first window has no newline.
   Preserves every line, and puts an unbounded allocation on a value nothing upstream bounds.

The distinction that decides it: **is a ledger line longer than 1 MiB a defect upstream, or an
input the hand must carry?** If the former, (1) or (2) and a cap somewhere that produces it. If the
latter, (3). Answering that is the ruling; the code is small either way.

## Definition of done
- Ruling 15 is implemented as written. `ship.ts`'s module docblock — where ruling A itself is
  declared — states the oversized case in the same commit, since that docblock is where a later
  reader looks for what the hand does with a line it cannot send.
- A ledger whose first unshipped line exceeds `MAX_READ_BYTES` no longer produces a pass that both
  advances nothing and reports nothing. Whatever the ruling chooses, the pass is **legible**: either
  it advances, or it names a failure.
- **The line behind the fat one is reachable.** A test seeds `[oversized, ordinary]` and asserts the
  ordinary line ships under the chosen answer. The three-pass reproduction above is the shape.
- `MAX_READ_BYTES` gains at least one test that references it — a constant no test names is a
  constant that can be changed to anything.
- The still-being-written case is distinguished from the too-long case, and a test says which is
  which. A window with no newline because the writer has not finished the line yet must **not** be
  treated as oversized; the discriminator is whether `size - offset` itself exceeds the cap.
- A mutation is recorded: reverting to the current `unchanged(null, null)` reddens the new case.

## Falsifier
If answer (1) is chosen and no safe discriminator exists between "line too long" and "line still
being appended", then skipping is unsafe and the ruling must fall back to (2) or (3). Record that
verdict rather than shipping a heuristic.

ciaran-slow:
Two rulings for prd-51, recorded ahead of the code that needs them. **Closes no issue** — that is
deliberate and is the point: #399 and #410 both said *"not dispatchable until one is chosen and
recorded"*, so this lands the record and unblocks them rather than resolving them.

## Commits, and the review that read each one

| commit | what | review |
|---|---|---|
| `da5fab08` | rulings 15 and 16, first draft | independent Opus pass, **REQUEST CHANGES** — 3 blockers, 3 minor. Findings quoted below. |
| `2dd6ac15` | the answers to that pass | **NOT INDEPENDENTLY REVIEWED.** Written by the same session that wrote `da5fab08`. |

The second row is condition 1's gap and it is stated here rather than recorded away. AGENTS.md:
*"Evidence you authored yourself is not evidence."* The corrections in `2dd6ac15` were made by the
author of the thing reviewed, in response to a review, and nobody independent has read them. A
reviewer should treat that commit as unreviewed and read it first.

## What the two rulings say

**Ruling 15 — the hand.** A ledger line longer than `MAX_READ_BYTES` (1 MiB) contains no newline
in the read window, so `splitLines` yields nothing and the pass returns `unchanged(null, null)`
forever: `ok: true`, empty failure list, cursor never written. Ruling A already refuses a
permanent wedge in terms; this line reaches none of its machinery. The ruling **extends ruling A
rather than amending it** — an oversized line is a line the hand cannot send, which is the
category ruling A answers. Skip, record, advance. Discriminator is `size - before.offset`, not the
window's contents, which resolves #399's stated falsifier: a safe separation of *too long* from
*still being appended* does exist.

**Ruling 16 — the fold.** `runOnce` assembles every row before opening the transaction and returns
on the first refusal, so one line whose type this build does not declare discards the records
before it as well as after — every actor stops, permanently, and the operator's only escape is
editing the journal by hand. The ruling stores the line unparsed instead. `events.type` is
`text NOT NULL` with no CHECK, no enum and no foreign key anywhere in the migrations, so **no
migration** — which matters past this issue, since the 2026-09-09 amendment records *one
migration-adding lane per wave* as the constraint that split waves 3 and 4. No gap either, so
ruling 3's invariant is untouched and ruling 5's identical-`chainDigest` claim needs no union with
a second relation to close.

## What is the sibling case?

**The review found it, and it is the substance of `2dd6ac15`.** Ruling 16's first draft split
`unknown` from `malformed` — #410's stated falsifier — and never split `unknown-type` from
`unknown-shape` one level down. `unknown-shape` means the type **is** known and the payload failed
the union, so it walks straight through every guard written against `row.type`. Executed against
`ce12c093`:

```
spend(unknown-type)  = [{"projectId":"p1","day":"2026-09-10","costUsd":0,"events":1}]
spend(unknown-shape) = [{"projectId":"p1","day":"2026-09-10","costUsd":42.5,"events":1}]
```

`costUsdOf` (`packages/team/src/fold/projections.ts:32`) reads `payload.costUsd` with no `row.type`
guard at all, so $42.50 from a line the build refused to parse lands in a money projection.
`agentStateOf` and `dirtyFiles` are the identical shape one level up. Ruling 16 was amended rather
than abandoned: `spend.events` counting an unfoldable row is ruled **correct** (it is an event),
and no projection may derive a value from an unvalidated payload — `EventRow` carries an
`unfoldable` marker, in memory, not persisted, so it stays a no-migration ruling.

The same shape appeared on the other side: ruling 15's first draft said *"if EOF arrives first,
the line is still being appended and the skip does not happen this tick"*, which recreates #399's
own reproduction for an oversized line that never terminates. That arm now names a failure.

**All of them, and that is worth knowing.** `doc-citation-law`'s `EXCLUDED_DIRS` contains
`docs/prds/`, pinned by the law's own test, so this amendment's ~30 backticked path citations are
structurally unguarded — no law in this repo will notice if one rots. The reviewer verified them
by hand and so did I; nothing automated did, and nothing automated will.

The one mutation that *is* owed is deferred by name to the implementing commit and written into
the ruling: remove the `unfoldable` check from `projectionsFor` and the case asserting no
unvalidated payload reaches a projection must go red.

## Fence widenings, recorded before the change

- **#399 and #410 both drop `docs/prds/prd-51-the-split.md`.** `fence-lint` failed on that overlap
  — each expected to write its own ruling — which is why both are written centrally here.
- **#410 narrowed, then widened.** Ruling 16 made the migrations directory and the storage port
  unreachable, so both conditional entries came off. The review then found the ruling unbuildable
  inside what was left: `EventRow` requires `eventId` and `source`, and `UnknownEventLine` carries
  neither. So the storage port returns for the `unfoldable` marker, the fold's projections join
  because that is where the marker is read, and `packages/core/src/events/index.ts` joins for `id`
  and `source`. **The migrations directory stays out** — the wave's single migration-adding lane
  is unspent, which is the whole reason ruling 16 beat a quarantine table.

## Consequence for dispatch

`scripts/fence-lint.sh 387 398 399` passes, and passes against every in-progress lane
(#407, #379, #350, #214). Those three are Ready.

**#410 is NOT dispatchable and is back in Backlog.** Its widening enters
`packages/core/src/events/index.ts`, which #407 holds while in progress. Two additive edits to one
file by two lanes landing days apart is a rebase conflict already scheduled, and this repo has
paid for the optimistic reading once (#405 + #400). Blocked-by is recorded on the issue. Taking
the re-`JSON.parse` route to dodge it was considered and refused: a permanent design compromise is
the wrong price for a temporary scheduling one.

## Gates, and one honest asterisk

`npm run typecheck` and `npm run lint` green. `VITEST_MAX_WORKERS=6 npm test` green — **484 files,
8905 passed** — but not on the first attempt, and that is worth stating rather than reporting the
clean run and moving on.

The suite went **red once on this exact tree**: 6 files, 19 tests. The identical tree re-run
immediately after came back fully green, and an earlier full run on `da5fab08` was green too. I
cannot say which 19 tests failed, because the command I ran piped the output through `tail` and
discarded the detail — my error, and I am not going to pretend otherwise.

Two reasons to read it as load flake rather than as this branch: the machine is known to flake
`@gate-timing` files at 6 workers, and **this diff is one markdown file under `docs/prds/`**,
which `doc-citation-law` explicitly excludes — there is no mechanism by which it reaches a code
test. That is REASONED, not executed. If a reviewer wants it executed, the branch is docs-only and
a clean CI run answers it.

Docs-only diff; no code changes.


ciaran-slow:
**CI: 10/10 green.** Verified per-leg with `gh pr checks 417` rather than off a green bar — three jobs, ten legs, the checklist AGENTS.md names:

```
pass  build-test-boot (macos-latest, current)
pass  build-test-boot (ubuntu-latest, current)
pass  build-test-boot (ubuntu-latest, min)
pass  pack-smoke (macos-latest, current)
pass  pack-smoke (macos-latest, min)
pass  pack-smoke (ubuntu-latest, current)
pass  pack-smoke (ubuntu-latest, min)
pass  pack-smoke (windows-latest, current)
pass  pack-smoke (windows-latest, min)
pass  windows-suite
```

**This settles the asterisk in the body**, and I am closing it here rather than editing the body, so the record of what was true at open stands. The local suite went red once on this tree — 6 files, 19 tests — and green on immediate re-run, and I could not say which tests failed because my own command discarded the output. The body called that load flake and labelled the reasoning REASONED, not executed.

It is executed now. Four independent full-suite runs on this tree passed: `build-test-boot` on ubuntu at both node legs, on macOS at current, and `windows-suite`. The macOS leg is the one that carries signal for path-shape bugs (`os.tmpdir()` is a symlink there and is not on ubuntu), and `windows-suite` enforces `.windows-known-failures` per file. **The red run was load flake on the authoring machine, not this branch.**

Unchanged and still true: this diff is one markdown file under `docs/prds/`, which `doc-citation-law`'s `EXCLUDED_DIRS` excludes. Ten green legs are not evidence the ~30 path citations in it are correct — no law in this repo reads them. They were checked by hand twice, by the review and by me, and that is the only thing standing behind them.

Also unchanged: `2dd6ac15` has no independent review. Green CI is not condition 1.

gabriel-canaan (changes requested):
Read commit-by-commit against `origin/main` merged in (the branch was one commit behind; main's commit was `docs/prds/prd-54-*`, no interaction). `2dd6ac15` read first, as the body asks.

The body's honest asterisk resolves in your favour. **All three gates green on the post-merge tree, first attempt, no flake:** `typecheck` 0, `VITEST_MAX_WORKERS=6 npm test` 0 — 484 files, 8909 passed / 3 expected-fail / 3 skipped — `lint` 0. (You measured 8905; same 484 files. The 4 are main's, not yours.) EXECUTED.

The citations hold up unusually well, which matters because the body correctly notes nothing automated checks them — `EXCLUDED_DIRS` at `packages/server/src/doc-citation-law.test.ts:493` does contain `docs/prds/`. All EXECUTED:

- **All 12 backticked file paths in the diff resolve.** No rot.
- **The no-migration grep reproduces exactly.** Case-sensitive `CHECK|CREATE TYPE|REFERENCES` over `packages/team/src/migrations/` returns **0**, and `0001_events.sql:91` is `type text NOT NULL`. (Add `-i` and it returns 30, all of them `checksum` — worth knowing, doesn't touch the claim.)
- **Every quoted docblock is verbatim.** Ruling A and B at `ship.ts:24-46`, including `no duplicates in n, and every gap in n is a skip this machine recorded and can name`; `row.ts:9` `a second parser is a second allowlist`; `index.ts:255` `calling that "a newer era" would be a lie`; `UnknownEventLine`'s `preserved byte for byte`.
- **`beacon.ts` really is the only capper.** `.max()` across all of `packages/core/src/events/` returns 4 hits, all in `beacon.ts`. "and nothing else caps anything" is exact.
- **Both wedges are real in the code.** `ship.ts:262` `if (scan.lines.length === 0) return unchanged(null, null)` with `unchanged` building `reason: null, advanced: false`; `worker.ts:75-84` assembles every row in a flat loop before `appendEvents` and returns on the first refusal. Marking the `runOnce` grouping sentence as prescription in `2dd6ac15` was a real correction — it does not group today.
- **The dispatch reasoning is mechanically correct.** `scripts/fence-lint.sh 387 398 399` PASSED; PASSED again against all four live lanes (`407 379 350 214` — that is exactly the in-progress set). Adding `410` FAILS with precisely the predicted `OVERLAP #407 vs #410 (packages/core/src/events/index.ts)`. The exclusion is load-bearing, not asserted.
- Rulings 15/16 are the next free numbers (1–14 exist); the internal evidence citations (280–415 B, `75 kills, 0 lost, 1,129,197 lines in 15.32 s`, the one-migration-lane constraint splitting waves 3 and 4) all resolve to real earlier text.

One drift since you wrote the body: **#399 is now In progress**, not Ready. Which makes finding 1 below time-sensitive rather than academic.

---

## Finding 1 — ruling 16's remedy does not reach `lane_state`. EXECUTED.

**The sibling case is one level further out than the one `2dd6ac15` caught, and the PR's own output shows it.** Your evidence block prints `lanes(unknown-shape) = [{"projectId":"p1","lane":"lane-a","state":null,…}]` and the amendment never comes back to that line.

The ruling narrows the falsifier to *"no projection may derive a value from an unvalidated payload"* — that principle is right. But the two clauses that make it actionable cannot deliver it:

- the enumeration is *"`costUsd`, `agentStateOf`, `dirtyFiles`, and anything later joining them"*, which omits `lane` and `worktree`;
- the mechanism is *"`projectionsFor` reads it"*, and `projectionsFor` is the wrong place to look — `row.lane` and `row.worktree` are derived **upstream in `toEventRow`** (`packages/team/src/fold/row.ts:48-61`, `laneOf` / `worktreeOf`, neither gated on `row.type`) and arrive as ordinary `EventRow` columns. A marker read inside `projectionsFor` cannot see that derivation; the projection reads a column, it does not derive a value.

I applied the ruling's remedy exactly as written — added `unfoldable?: boolean` to `EventRow`, gated the three named readers on it — and re-ran:

```
BEFORE   spend(unknown-shape) = [{…,"costUsd":42.5,"events":1}]
         lanes(unknown-shape) = [{"projectId":"p1","lane":"lane-a","state":null,"worktree":"/repo-wt/lane-a",…}]

AFTER    spend(unknown-shape) = [{…,"costUsd":0,"events":1}]        <- money leak closed
         lanes(unknown-shape) = [{"projectId":"p1","lane":"lane-a","state":null,"worktree":"/repo-wt/lane-a",…}]   <- unchanged
```

The `costUsd` assertion goes red under the remedy, so the guard is load-bearing. `lane_state` is untouched: a persisted projection still receives `lane` and `worktree` from a payload the build refused to validate, keyed `(projectId, lane)`. It happens for **both** `unknown-shape` and `unknown-type`.

Cheap to resolve, and no fence move: `row.ts` and `projections.ts` are already inside #410's fence. One extra guard on the `if (row.lane !== null)` branch in `projectionsFor` closes it (`collisions` is already covered, since `dirtyFiles` gates on type). But which way it should go is a ruling, not a typo, so I am not writing it for you — nulling `lane` upstream and gating the branch differ in whether `lane_state.lastEventTsMs` advances for an unfoldable event, and that is your call as this PRD's operator. Say the word and I will push the wording.

## Finding 2 — ruling 15 widens a versioned persisted schema and does not say so. EXECUTED.

Ruling 16 spends two paragraphs establishing that its store needs no migration. Ruling 15's counterpart store does have an enum, and the ruling is silent about it: `ActorSkip.kind` is `z.union([z.literal('unknown'), z.literal('malformed')])` at `packages/server/src/shipper/cursor.ts:73`, inside a **versioned** file (`CURSOR_VERSION = 1`, unchanged by this ruling). `ActorSkip.kind` gains `'oversized'` is a change to on-disk data.

Wrote a cursor carrying `kind: 'oversized'` and read it with the current schema:

```
reset   = "…/shipper/cursor.json holds an unusable entry for actor-A; that actor starts from the beginning"
actor-A = null                        <- was offset 90000, n 4200
actor-B = {"offset":55000,"n":1100,…} <- unaffected, as designed
```

That actor resets to `offset 0, n 0` and loses its recorded skip history. **The blast radius is bounded and I want to be accurate about it** (REASONED from here): `0004_events_dedup.sql` puts a unique index on `(project_id, actor_instance, n)` per partition and `projectionsFor` is fed only inserted rows, so a re-ship of an append-only ledger dedups — wasteful, not corrupting. The per-actor reset is also pre-existing behaviour, not something this ruling invents.

What the ruling does invent is a **routine** way to reach it: hit the oversized-line bug, roll back the release that fixed it, and the cursor you wrote resets. Worth one sentence deciding whether that wants a `CURSOR_VERSION` bump, a lenient `catchall` on the literal, or an explicit "accepted, dedup absorbs it" — particularly since ruling 15 states that ruling 3's *no duplicates in `n`* restatement is **unchanged**, and a reset-and-re-ship is the one path that re-sends an `n`.

## Finding 3 — the evidence table invites a wrong inference. EXECUTED. Minor.

`spend(unknown-type) = costUsd 0` reads as though something protected the unknown-*type* arm. Nothing did — that is your fixture's payload. `costUsdOf` has no `row.type` guard at all, so an unknown-type line carrying `costUsd` leaks identically; with `costUsd: 42.5` on both, I get `42.5` for **both** arms. Your prose says this correctly (*"no `row.type` guard at all"*); only the table disagrees with it. A one-line note under the block would stop a later reader concluding the type/shape split is what matters for `costUsd`.

---

## Not blocking, offered not asserted

The body's condition-1 disclosure is the right call and I would not soften it. This review is the independent pass `2dd6ac15` was missing; findings 1 and 3 are both in that commit, which is a fair argument that the disclosure earned its keep.

## What I did not run

No CI comparison — `gh pr checks` was not consulted, since the diff is one markdown file and I ran the full suite locally against the merged tree instead. I did not verify the reproductions recorded on #399 and #410 themselves; I verified the structural claims about the tree that the rulings rest on.

---
Reviewed with Claude Opus 5 (1M context) via `rhizo-review`. **Single reviewer** — one pass, not a multi-seat verify; findings above are individually executed, but the coverage is one perspective.

ciaran-slow:
@gabriel-canaan — all three findings accepted, all three reproduced here before acting, none rebutted. `642d93f1` pushed. Findings 1 and 2 were rulings rather than typos, exactly as you said, so both went to the operator; the wording below is theirs.

**Finding 1 — fixed, and you were right that `projectionsFor` is the wrong place to look.** Confirmed independently: the lane branch keys on `row.lane` and writes `row.worktree`, both derived upstream in `toEventRow` with no gate, so a marker read inside the projection cannot reach them. Ruled **upstream**: `toEventRow` sets `lane` and `worktree` to `null` for an unfoldable line, and the existing `if (row.lane !== null)` skips the branch with nothing added. That is `row.ts`'s own rule applied to the case it was written for — a column must carry a fact some collector asserted. The cost is `events.lane`/`events.worktree` null on that row, accepted because `line` is preserved verbatim and a later build re-derives from it. No fence move, as you noted.

I have also grown the owed mutation on #410: removing the guard must redden the money assertion **and** removing the upstream nulling must redden a `lane_state` assertion. A single mutation reddening a single case is how the first version passed its own review.

**Finding 2 — fixed, and the bump is rejected in the ruling with its reason.** Ruled: the property is that a cursor written by a newer build stays readable by an older one, so `kind` tolerates an unrecognised value. `CURSOR_VERSION` is explicitly **not** bumped — `version` is `z.literal(CURSOR_VERSION)`, so bumping resets *every* actor at upgrade, a certain cost paid to avoid a conditional one. Your bounded-blast-radius paragraph is carried into the ruling rather than dropped, because you were right not to overstate it: the per-partition unique index makes a re-ship wasteful, not corrupting, and what leniency buys is that the path stops being routine.

Your timing note mattered — **#399 went In progress while this PR was open**, and `cursor.ts` is in its fence, so finding 2 is posted there as a build instruction with the do-not-bump called out.

**Finding 3 — fixed inline.** You were right that the table contradicted the prose; the `costUsd: 0` was my fixture, not a guard.

Two things I want on the record rather than smoothed over:

**`642d93f1` is not independently reviewed either.** Same disclosure as `2dd6ac15`, for the same reason — I wrote it in answer to a review of my own work. Three commits now, two of them with no independent pass. Your review is the argument for the disclosure existing; it is also the argument for not treating this one as settled.

**Finding 1 is the third instance of one shape in this PR.** `da5fab08` missed `unknown-shape` beside `unknown-type`; `2dd6ac15` fixed that and missed `lane`/`worktree` beside `costUsd`. Each fix handled the case its author considered and missed the structurally identical sibling — the exact defect shape AGENTS.md names, three times in one document, twice caught only because someone else read it. Worth saying plainly rather than filing as three unrelated fixes.

Gates re-run on the corrected tree: `typecheck` 0, `lint` 0, `VITEST_MAX_WORKERS=6 npm test` 0 — 484 files, 8905 passed. Your 8909 is the post-merge tree; the 4 are main's, as you said.

gabriel-canaan (changes requested):
Second pass, reading `642d93f1` — the commit the body correctly flags as not independently reviewed — against `origin/main` merged in (the branch was 5 commits behind; main's five were all `docs/prds/prd-54-*`, no interaction). Two of the three findings from the last pass are answered and I have pushed the small corrections rather than handing them back; one new finding is a ruling and is below.

**All three gates green on the post-merge tree, first attempt, no flake.** `typecheck` 0, `lint` 0, `VITEST_MAX_WORKERS=6 npm test` 0 — 484 files, 8909 passed / 3 expected-fail / 3 skipped. Same numbers as the last pass, so the merge cost nothing. EXECUTED.

**Finding 1's remedy is correct, and it is load-bearing.** I implemented ruling 16 as amended — `unfoldable` marker on `EventRow`, the three derivations gated on it, `lane` and `worktree` nulled in `toEventRow` at the point of derivation — and ran four unfoldable lines through `toEventRow` → `projectionsFor`, each carrying `lane`, `worktreePath` and `costUsd: 42.5`. EXECUTED:

```
unknown-type   (agent.telepathy)          spend costUsd=0 events=1   lanes=[]   collisions=[]
unknown-shape  (llm.cost, bad payload)    spend costUsd=0 events=1   lanes=[]   collisions=[]
unknown-shape  (agent.status, bad payload) spend costUsd=0 events=1  lanes=[]   collisions=[]
unknown-shape  (worktree.dirty, bad payload) spend costUsd=0 events=1 lanes=[] collisions=[]
```

`spend.events` still counts the row, which the ruling rules correct. Mutation, to show the upstream nulling is doing the work rather than decorating it — restore `lane: laneOf(u.payload), worktree: worktreeOf(u.payload)` in that branch and leave every marker gate in place:

```
unknown-type   row.lane="lane-a"  lanes=[{"lane":"lane-a","state":null,"worktree":"/repo-wt/lane-a",…}]
unknown-shape  row.lane="lane-a"  lanes=[{"lane":"lane-a","state":null,"worktree":"/repo-wt/lane-a",…}]   (all three)
```

`lane_state` repopulates from the unvalidated payload for all four. The clause is necessary and sufficient, and "the projection reads a column, it does not derive a value" is the right diagnosis.

**Finding 2's answer holds, and it is safe.** EXECUTED, all verbatim: `actorSkipSchema.kind` is `z.union([z.literal('unknown'), z.literal('malformed')])` at `packages/server/src/shipper/cursor.ts:73`; `shipperCursorSchema.version` is `z.literal(CURSOR_VERSION)` at `:86`; `CURSOR_VERSION = 1` at `:67`. `0004_events_dedup.sql` does put a unique index on `(project_id, actor_instance, n)` per partition, so "wasteful, not corrupting" is exact. `0001_events.sql:16` really does carry the *"last commit in which that is true"* argument the ruling borrows.

Two things I checked that the ruling does not claim, both of which support it:

- **Nothing switches on `kind` exhaustively.** The only consumer outside the shipper is `packages/server/src/cli/connect-team.ts:374`, which interpolates `skip.kind` into a report line. A tolerated unrecognised value renders; it does not fall through a `switch`.
- **The clause reached the lane.** #399 is In progress and your comment on it carries the new clause, the rejection of the `CURSOR_VERSION` bump, and the housekeeping note. That was the gap I would otherwise have raised, and it is closed.

**Finding 3 answered inline**, and the note is placed where the table is rather than in the prose that was already right.

**`events.lane` and `events.worktree` are nullable**, so ruling 16's remedy is legal against the shipped DDL: `0001_events.sql:93-94` are `lane text` / `worktree text`. Nulling them at source does not collide with the schema. EXECUTED.

**The dispatch reasoning still holds, re-derived against the live set rather than the one in the body.** In progress is now `#407 #402 #401 #389 #399 #350` — different from the four the body names. `scripts/fence-lint.sh 387 398 407 402 401 389 399 350` reports exactly one overlap, `#350` vs `#401` on `packages/web/src/lab/no-live-fleet-law.test.ts` — two other in-progress lanes, nothing to do with this PR. #387, #398 and #399 are clean against everything live. `scripts/fence-lint.sh 410 407` still fails with `OVERLAP #407 vs #410 (packages/core/src/events/index.ts)`, so #410's block is real and load-bearing, not a precaution. EXECUTED.

---

## Fixed and pushed, not handed back

**`da173e39` — `collisions` is covered by the marker, not by the type gate.** The new clause closes with *"`collisions` is already covered, since `dirtyFiles` gates on `row.type`"*. That reason is wrong, and it is **mine** — I wrote that parenthesis in the last review and you adopted it verbatim. It is the amendment's own type/shape conflation restated inside the paragraph that closes it. EXECUTED: remove the marker check from `dirtyFiles` alone, leave its `row.type` gate standing, and one `unknown-shape` `worktree.dirty` line gives

```
collisions = [{"projectId":"p1","path":"src/a.ts","lanes":["lane-a"],…}]
```

while the other three probe cases stay empty. The outcome was never wrong — `dirtyFiles` is one of the three the clause one paragraph up puts behind the marker — but the stated reason is one a lane could act on by dropping that check as redundant.

**`aa0d3708` — the wave map's row for wave 3.** It still read `groomed, lint-clean`, which stopped being true at `95d9874f` (`Merge pull request #386 … prd51-w3-the-hand-and-the-ingest`, carrying `34b3a1e7` for #372 and `794424b2` for #373; `packages/server/src/shipper/ship.ts` is on `origin/main`). Ruling 15 as amended now rests an argument on the other reading — *"the shipper merged in wave 3 and the first real host is wave 5+"* — so the document contradicted itself two screens apart, with the stale row on the side a reader would take as the wave's state. EXECUTED.

`lint` and `doc-citation-law` / `runbook-delivery-law` re-run green after both (60 tests).

---

## Finding — ruling 16 names three fields it cannot get and resolves two. EXECUTED.

The widening clause opens: *"`EventRow` requires `eventId`, `source` and `payload`, all non-nullable, and `UnknownEventLine` carries none of them"* — then resolves it *"by adding `id` and `source` to `UnknownEventLine`"*. `payload` is named as the problem and never appears again.

It is not decorative. The ruling's first paragraph is explicit that the row lands *"with its `type` recorded, its `line` untouched, and its payload whatever the line carries"*, and the entire money clause is about a projection reading that payload. Compiled the row the ruling describes, from what the lenient reader actually hands back:

```
packages/team/src/fold/__probe.ts(14,16): error TS2339: Property 'id' does not exist on type 'UnknownEventLine'.
packages/team/src/fold/__probe.ts(17,15): error TS2339: Property 'source' does not exist on type 'UnknownEventLine'.
packages/team/src/fold/__probe.ts(20,16): error TS2339: Property 'payload' does not exist on type 'UnknownEventLine'.
```

Three, not two. The lane has no third route: re-`JSON.parse`ing in `row.ts` is rejected by name on that file's own docblock, and leaving `payload` empty contradicts *"its payload whatever the line carries"* — and would make the whole remedy vacuous, since `costUsdOf`, `agentStateOf`, `dirtyFiles`, `laneOf` and `worktreeOf` all return `0`/`null` on an absent payload and there would be nothing left to gate.

**The reason it wants a sentence rather than a silent third bullet is that the justification does not carry over.** `id` and `source` are argued in as *"additive, in the one place that already proved them — `eventEnvelopeProbeSchema` validates `id` and `source` and then discards both."* That schema is `z.object({ id, ts, source, type })` (`packages/core/src/events/index.ts:208-213`); it never touches `payload`. So adding `payload` puts a field into `UnknownEventLine` that nothing validated — defensible, since `line` beside it is already the whole unvalidated line verbatim, but it is a different argument from the one on the page, and this is the ruling that exists to say which values are trusted.

Not urgent: #410 is in Backlog behind #407, and the fix is inside its fence either way. But it is the ruling's own subject matter, so I am not writing the wording for you.

## Not blocking, offered not asserted

The `unfoldable` marker and the upstream nulling are two mechanisms answering one rule, and the ruling now reads as though the marker were the answer with the nulling as a patch. It is the other way round for two of the four values. Worth a half-sentence saying the rule has two enforcement points by construction — a fold-time marker for what the projection derives, and `null` at source for what it inherits — so the next reader does not try to unify them.

## What I did not run

`gh pr checks` — the diff is one markdown file and I ran the full suite locally against the merged tree instead. I did not re-verify the reproductions on #399 and #410; I verified the structural claims about the tree that the rulings rest on. My probe implements ruling 16, it is not ruling 16 — the mutation the ruling owes is still owed by the implementing commit.

---
Reviewed with Claude Opus 5 (1M context) via `rhizo-review`. **Single reviewer** — one pass, not a multi-seat verify; findings above are individually executed, but the coverage is one perspective.

gabriel-canaan (approved):
Approving at `aa0d3708`, the head I reviewed.

**The one finding is not fixed here — it is moved, deliberately.** Ruling 16 names `eventId`, `source` and `payload` as the three fields `UnknownEventLine` cannot supply and resolves two; the `payload` sentence is still owed. It is parked rather than blocking because #410 is the only issue it reaches and #410 is in Backlog behind #407, so the gap cannot reach a lane before that issue is regroomed and the ruling re-read. Recorded on #410 as a build note so whoever picks it up meets it there.

**Two of the three commits on this branch are mine** (`da173e39`, `aa0d3708`), and this approval is not a second pair of eyes on them — the operator read them. Saying so rather than letting the approval imply otherwise.

What this approval rests on, all EXECUTED against `origin/main` merged in: typecheck 0, lint 0, `VITEST_MAX_WORKERS=6 npm test` 0 (484 files, 8909 passed / 3 expected-fail / 3 skipped); ruling 16's amended remedy implemented and shown to close `lanes` and `collisions` for both `unknown-type` and `unknown-shape`, with the upstream nulling mutated out to show it is load-bearing; ruling 15's cursor claims verified verbatim against `cursor.ts:67-86`; fence-lint re-derived against the live in-progress set rather than the one in the body.

---
Reviewed with Claude Opus 5 (1M context) via `rhizo-review`. **Single reviewer** — one pass, not a multi-seat verify.

Issue #416, opened 2026-09-10
prd17: every live document's status claim matches the tree, and the set is enumerated first


gabriel-canaan:
## Fence (may touch ONLY)
- `docs/roadmap.md`
- `docs/architecture.md`
- `docs/prds/prd-17-complete-record.md`

**The fence is provisional and must be re-declared before any edit** — see the
Definition of done. Enumerating the set is step one of this issue, and the
three paths above are only the ones already known to carry a claim. A path
found by the enumeration and not listed here is a widening, recorded before
the change like any other.

Dated artefacts are EXPLICITLY OUT: the review directory, the append-only ADR
log, every shipped PRD, and every dated reconciliation. This issue does not
rewrite history; it corrects documents that describe the present.

- ~~`docs/prds/prd-43-the-claim-is-a-test.md`~~ — **WIDENING WITHDRAWN 2026-09-10:** that PRD's milestone is closed, so its prose is a record and not a live claim. The edit was reverted; see the comment below. (the one discharged #275/prd-17-wave-2 block only - no ruling, no other claim in that PRD, and no change to its own status)

## Blocked by
**prd17 wave 8** (`#391` + `#403`), which must land first. Wave 8 deliberately
stopped short of this work and narrowed its own claim to the wave landings —
this issue is where the status claim it withdrew gets made properly.

## Why
Wave 8 tried to mark prd-17 shipped as a side-effect of declaring a wave.
**Three independent review rounds each found one more live document
contradicting it**, and the enumeration written to justify the sweep was wrong
twice — the last version claimed "ten documents mentioning prd-17, two of them
live" when the two spellings together reach thirty-one, and
`docs/architecture.md` is live and states the opposite in four clauses.

The lesson recorded on wave 8 is that a status word is a claim over a set that
was never bounded, so no sweep of it could be shown complete — and an unbounded
claim asserted as swept is worse than no claim, because it reads as verified.

So this issue exists to bound the set first and edit second.

**Known-false clauses, as of `1354c8ad`** — a starting inventory, not the
enumeration:

- `docs/architecture.md` — "so a recording cannot contain one of these yet"
  (era-2's committed recording holds five); "`eras.test.ts` lists all eight in
  its corpus gap list" (it holds four of eight); "the gate emitters are all
  still open" (the gate emitter landed in wave 6 — the UI operator-acts half of
  that clause is still true); "the door exists, no writer uses it yet"
[...]

PR #418, opened 2026-09-10
prd17 w8: a landing's close fault leaves a trace, and the PRD declares the wave

### LINKED ISSUE #391: prd17 w8: a swallowed close fault leaves a trace

### LINKED ISSUE #403: prd17 w8: the PRD declares wave 8 and says gate.verdict is a first-class event
## Fence (may touch ONLY)

The archival-directory path is **no longer claimed** — the move was dropped,
see the withdrawn bullet below. No other document is in scope beyond the widening
recorded on this issue. The ADR log is EXPLICITLY OUT. The design
notes directory is EXPLICITLY OUT. Every other PRD is EXPLICITLY OUT.

- `docs/follow-up-292.md` (the one prd-17 citation bullet only — no other entry in that inventory, and none of the four claims it tracks)

- `docs/roadmap.md` (the prd17 bullet's status clauses only — no other programme's entry, and no change to the roadmap's ordering or structure)

## Blocked by
Nothing, but it must land **in the same PR as `#391`** — see Why. Ordered after
`#391`'s commit within that PR, so no intermediate commit carries a claim that
is not yet true.

## Why
`scripts/dev/prd-reconcile.sh 17` reports drift, and this issue is the same
shape as `#346`, which did exactly this job for wave 7. Two kinds, and the
document is the plan of record, so the document is what changes.

**1. An UNDECLARED WAVE.** `#391` claims `w8`, which this PRD never declares:

```
UNDECLARED WAVE  #391 (OPEN) claims w8, which
                 docs/prds/prd-17-complete-record.md never declares
                 The tracker is ahead of the plan of record.
```

That matters mechanically, not just tidily: the wave lint reads a
wave, so an undeclared one is a wave the lint cannot check a bundle against.

**2. The header states two things that stopped being true when wave 6 landed.**
It says:

> `gate.verdict` is EMITTED but not yet a first-class timeline event — a landing
> reaches the log as `beacon.received`, and deriving the typed event from that
> sidecar is wave 6.

Both halves are false as of `6c4358f1` (`#280`, PR #395): `gate.verdict` is now
a first-class typed event, derived in
`packages/server/src/log/gate-verdict-derivation.ts` and recorded at RECORD time
from the poll loop's own seam — and wave 6 is past, not future. The family
table's `gate.verdict` row credits only the beacon line, which was wave 5's
truth.

**What must NOT change, because it is still correct.** This is the trap `#346`
recorded for its own version of this edit — an earlier header attached one
clause to every row and got the reason wrong per row:

- **"in a recording? not yet" STAYS.** Both era-corpus recordings were captured
  before this landing, so neither carries a `gate.verdict`. Wave 6's own review
  settled this: the uncovered-family ARRAY in `packages/core/src/eras/eras.test.ts`
  is a computed fact about two frozen recordings, not a claim about what has an
  emitter, and editing it without a new capture would assert something the
  corpus does not contain.
- **Both counts STAY** — seven of nine with an emitter, five in a committed
  recording. Wave 6 changed `gate.verdict`'s CHARACTER, not whether it had an
  emitter; it was already counted.
- **The dispatch-brief and fence-declared families STAY** defined-and-unemitted. Nothing
  raises either, and their `reduce.ts` arms still return state unchanged.

## Definition of done
- Sequencing declares **Wave 8**, naming `#391` and this issue, in the form the
  other waves use. `prd-reconcile.sh 17` then reports no drift.
- Wave 6's Sequencing entry carries a landed marker with its merge sha, the way
  waves 0 and 1 carry DISCHARGED / LANDED.
- The header's `gate.verdict` sentence and its table row say what is now true:
  a first-class typed event, derived from the sidecar, recorded at record time.
- The three things named above as still-correct are demonstrably unchanged.
- ~~The file moves to `docs/prds/done/`~~ — **WITHDRAWN 2026-09-10.** The
  Outcome line still reads shipped (valid only because `#391` lands in the same
  PR), but the file **stays in `docs/prds/`**. Moving it reddened prd-43's
  doc-citation law, and chasing that found the real reason: `docs/follow-up-292.md`
  reasons from this file's location to call ruling 3 a LIVE rather than an
  archival ruling, and holds an open, uncorrected item against it. Archiving
  would demote that ruling while the item is open, and that document reserves
  the amend-or-not judgement for whoever owns the ruling. The move is a
  separate operator decision; the reason is written into the PRD's Outcome line
  so it is not "tidied" without being found.
- No claim in the amended document is false at merge time. The repo tests
  document claims, so a citation added here must resolve.

## Sibling case
`#346` amended this same header and left a sibling stale: it corrected the
summons and `gate.verdict` rows for waves 3 and 5, and wrote "deriving the typed
event from that sidecar is wave 6" — a forward-looking clause that became this
issue's finding the moment wave 6 landed. So the miss to avoid is exactly the
one this issue inherits: **fixing the rows that changed while leaving a new
forward-looking claim that will rot on the next landing.** Prefer wording that
is true whatever lands next, or tie it to a sha.

Second: the Outcome line and the Sequencing entries are two places that both say
what has shipped. Update one and not the other and the document disagrees with
itself.

There is no unit test here; the check is `prd-reconcile.sh 17` reporting no
drift, and that is a real executable assertion rather than a formality — it
currently exits non-zero naming `#391`, so it can be run before and after and
shown to change. Do that, and put both outputs in the commit body.

The claim most able to pass while false is the Outcome line: nothing in the
tooling verifies that a PRD marked shipped has actually shipped. So state in the
commit body which sha each shipped claim rests on, and confirm `#391`'s commit
is in the same PR — a document that says shipped while its last wave sits in
someone's worktree is the prd-44 failure this repo already paid for once.

**Model:** sonnet. **Wave:** 8. **Milestone:** prd17.

gabriel-canaan:
Closes #391
Closes #403

prd-17's last wave, and its own review residual. Two issues, disjoint fences, one commit each.

**#391** — the close-fault catch in `gate-verdict-derivation.ts` was correct and silent. The swallow stays (by close time the bytes are in the buffer and the digest comparison runs on them further down, so a close fault never gates verification), but a persistently failing mount left no trace while it fired for every `beacon.received` on every tick. The fault now rides out as a three-state `SidecarDescriptor` and the poll-loop seam decides what to say, latched per sidecar file.

**#403** — the PRD declares wave 8, says `gate.verdict` is a first-class typed event as of wave 6, and marks wave 6 landed.

## Fence widenings, each recorded on its issue before the change

- **#391**, at grooming: `packages/server/src/server/poll-loop.ts` + its test, bounded to the `gate.verdict` recording seam.
- **#403**, during review: `docs/follow-up-292.md`, bounded to the one prd-17 citation bullet.
- **#403**, during review and then **withdrawn**: `docs/roadmap.md`. See "what this PR deliberately does not do".

## What is the sibling case?

**This PR's entire subject is that failure mode, and it kept reproducing it — that is the honest headline.**

`#280` (wave 6, already on `main`) went four rounds: round 4 found the bounded read had no `catch`, round 5 found `close()` was the structurally identical sibling one line over. `#391` repaired round 5's finding and shipped **the next cell of the same table**: the `return` inside the read's `catch` is evaluated *before* `finally` runs, so a read fault and a close fault together discarded the close fault — and at the seam that absence took the `delete` branch and **un-latched a fault already reported**, which is the exact zero-trace scenario `#391` exists to end.

Two independent seats found that in the same round. The repair is therefore **not** another patch on that axis:

- `SidecarDescriptor` is three states — `unopened` / `closed` / `close-failed` — because the root was that an optional string could not distinguish *closed cleanly* from *never opened*, and nine refusals return before any `open()`.
- It is a **required** field, so a return that omits it does not compile. The previous version relied on every site calling a helper and one did not; the typechecker now does that job. Verified by mutation, not asserted.
- No return appears between the read and the point the descriptor is decided.
- At the seam, `unopened` deliberately does nothing — neither reports nor clears. Absence of a close fault is not evidence of a close.

The input class was written down **before** the repair this time, which is what the two prior rounds had skipped: 9 refusals with `unopened`, 8 outcomes carrying the decided descriptor, 1 inapplicable. 18 return sites, matching a reviewing seat's own independent count.

The class recurred twice more in prose and was closed by sweeping rather than patching: the corrected close/digest ordering claim had three homes (one in a file this PR touches, where it was the *justification* for the assertion beneath it), and a tree-wide grep for it now returns zero.

Nine certified across the wave, each reverted before committing. **Five of them had previously survived** a passing suite, which is the part worth reading:

| mutation | before | after |
|---|---|---|
| return from inside the read catch (regress the primary finding) | — | 2 failed / 97 |
| clear the latch on `unopened` too (regress the second) | — | 3 failed / 96 |
| key the latch on a constant, not the file | **62/62 green** | 4 failed / 95 |
| `open()`-failure arm returns `closed` | **99/99 green** | 1 failed / 99 |
| drop the OS error from the operator's line | **64/64 green** | 4 failed / 96 |
| swallow the close fault again (empty catch) | — | 2 failed / 93 |
| classify from the stale `stat` size | — | 2 failed / 93 |
| disable the `lane === null` guard | **13/13 green** | 1 failed / 30 |
| `size <= offset` → `size <` | **13/13 green** | 1 failed / 31 |

The three prose-only green cells are the same defect shape each time: an assertion weaker than the claim above it. A 13-row refusal table asserted `outcome` plus a non-empty reason, which cannot fail for the reason a row *names*; every row now matches its reason, and `size <= offset` needed a fourteenth row because a 1-byte file at offset 5 is refused by both spellings.

**One test's guard was itself a vacuous pass and is now a skip.** The `open()`-failure pin used a real EACCES via `chmod 000`, with a hatch that asserted `closed` and returned if the derivation succeeded anyway. Under uid 0 `chmod` does not deny — so that hatch made the test pass *without ever reaching the arm it exists to pin*. It now skips under root with its reason stated, and under an unprivileged uid asserts the refusal with no escape.

**One fix is honestly uncertifiable and is not claimed:** the bounded read's window cap. A whole-file read and a capped read produce byte-identical output for every fixture this suite can hold; the difference is resource bounds a unit test cannot redden.

## What this PR deliberately does NOT do

It does **not** mark prd-17 shipped, and that decision is wave 8's main finding.

An earlier version did, and swept the documents that said otherwise. **Three review rounds each found one more live document contradicting it**, and the enumeration written to justify the sweep was wrong twice — the last version claimed "ten documents mentioning prd-17, two of them live" when the two spellings together reach thirty-one, and `docs/architecture.md` is live by this repo's own `**Tree:**` test and states the opposite in four clauses. The wave-8 table had also been made to promise "no other live document contradicts it": an unbounded claim asserted as verified.

A status word is a claim over a set this wave never bounded, so no pass over it could be shown complete. So the Outcome now states only what was checked — every wave this PRD sequenced has landed, with the shas — and **#416** carries the reconciliation, enumerating the document set up front by the `**Tree:**` pin, a method that can be finished rather than asserted.

`docs/roadmap.md` is reverted out for the same reason, and its widening withdrawn. Its drift predates this wave, and the partial fix made it worse: a bullet reading `SHIPPED` eleven lines above four landed items still listed as "not yet landed", contradicting a sentence the fix had just added.

Also deliberately open, deferred with reasons across `#280`'s rounds: the seam swallowing a refusal's reason, the stat/open TOCTOU, and a hardlink defeating containment. And whether prd-17 moves to the archival directory — `docs/follow-up-292.md` reasons from its location to call ruling 3 *live* rather than archival and holds an open item against it, so archiving would demote that ruling. That reason is written into the PRD's own Outcome line so it is not "tidied" without being found.

## Review record

Four independent passes, all recorded on the issues with their evidence: one initial two-seat pass, then three fix re-reviews on the repairs. Seats were Claude Opus 5 and Codex `gpt-5.6-terra`, both at xhigh, isolated from each other, with the Opus seat swapped down to Sonnet on the round where Opus had authored the repairs under review.

Two limits, stated rather than left to be assumed:

1. **The seats split once**, and it was resolved against the code rather than by counting: one returned safe-to-merge while the other found three real findings, all of which I confirmed by direct inspection. A unanimous verdict is not what made the difference; checking was.
2. **Codex could not run the full suite** in its sandbox on any round (`spawnSync git`/loopback `listen` denied, plus an unrelated missing `postgres` in `packages/team`). Its typecheck, lint and focused runs are real; **its full-suite results are not counted as evidence.** The Opus seat's full runs are, and matched the reported figures exactly each time.

## Verification

Rebased onto current `main` immediately before opening and the gates re-run **there**, not carried over: typecheck clean, lint clean (1033 files), `VITEST_MAX_WORKERS=6 npm test` → **484 files, 8918 passed**, 3 expected-fail, 3 skipped (both pre-existing). No `package-lock.json` change. The two commits touch exactly six files.

Worth recording because it nearly shipped: `origin/main` moved four times during this work, and one rebuild soft-reset onto a newer base while the working tree sat on the older one — staging the *reversal* of unrelated landed work (a deleted PRD, a directory of lab files) into these commits. Caught by reading the diffstat rather than an exit status, and recovered by extracting only the six real paths as a patch, resetting to the known-good head, rebasing onto the true base and reapplying.

---

## Declared review gap

> Four independent review passes are recorded on #391 and #403, the last on 1354c8ad. The delta since that head is small and subtractive: a test whose root-privilege hatch became a skip (it could pass without reaching the arm it pins), two prose corrections of an ordering claim, the withdrawal of a status claim and its enumeration, and the revert of docs/roadmap.md out of the wave. Nothing was added to the mechanism, which four seat-passes have confirmed sound. A fifth round was declined by the operator on the ground that the remaining changes remove claims rather than make them; stated here so a reviewer weighs that themselves rather than reading four rounds as covering these exact bytes.

```
  UNVERIFIED  c5ae97ef  docs(prds): the PRD declares wave 8 and says gate.verdict is a first-class event  (#403)
  UNVERIFIED  e43e557f  fix(server): a swallowed close fault leaves a trace  (#391)
```


gabriel-canaan:
## `windows-suite` was red on my own test, not on the code — fixed in `1a2fb75b`

```
UNEXPECTED FAILURE (not on .windows-known-failures):
  packages/server/src/log/gate-verdict-derivation.test.ts
  "open() failing is UNOPENED too" — expected 'derived' to be 'refused'
```

**The guard was wrong, not the arm — and it is the same error as the two before it, one level further out.** This test pins the `open()`-failure arm using a real EACCES via `chmod 000`, and it has now been wrong three ways:

1. The original hatch asserted `closed` and returned when the derivation succeeded anyway, so on any filesystem that ignores the mode it passed **without reaching the `open()` catch it exists to pin.** A review caught that.
2. The replacement skipped on `process.getuid?.() === 0`. That reads as careful and is wrong twice: `process.getuid` is **undefined on Windows**, so the optional call yields `undefined` and never equals `0` — the guard never fires — and Windows also ignores the mode bits for the owner, so the chmod does not deny and the assertion runs regardless. Hence the red.
3. Now: **the precondition is proven rather than inferred.** Whether a chmod can deny a read is a property of the filesystem, not of a platform name or a uid, so the test asks the filesystem — it attempts the read and skips only when the denial could not be constructed. Root, Windows, an ACL, an exotic mount: one question answers all of them, and none of them can produce a passing assertion.

Certified both directions, because a skip that never fires is as useless as a vacuous pass:

- denial constructible (this Linux runner, uid 1000) → assertions run, **36 passed**
- "cannot deny" forced → **1 skipped**, not passed

**Deliberately NOT added to `.windows-known-failures`.** That list is for genuine platform differences in the code under test; here the code was right on every platform and the test's premise was wrong. Adding it would have hidden a bad test behind a list of real platform facts.

Gates re-run before pushing: typecheck clean, lint clean (1033 files), 484 files / 8918 passed. The two commits still touch exactly six files, and the docs commit is untouched.

ciaran-slow (approved):
## Review — every claim in the body re-derived, one gap found and fixed in `f03155e3`

Read commit by commit in a fresh clone, Node 22.23.2. This PR's body is unusually precise, so the useful thing a reviewer can do is check it rather than restate it. I checked all of it.

### Verified in your favour

**Every mutation the body claims, re-run independently.** I wrote my own harness against `gate-verdict-derivation.test.ts` + `poll-loop.test.ts` (baseline 100 passed) and reverted after each: `EXECUTED`

| mutation | your claim | mine |
|---|---|---|
| swallow the close fault again (empty catch) | reddens | **3 failed / 97** |
| drop the latch, report every time | reddens | **4 failed / 96** |
| latch never clears (drop the `delete`) | reddens | **4 failed / 96** |
| report unconditionally, clean or not | reddens | **4 failed / 96** |
| `unopened` clears the latch (finding 2) | reddens | **3 failed / 97** |
| `open()`-failure arm returns `closed` | 1 failed / 99 | **1 failed / 99** |
| return from inside the read catch | reddens | **1 failed / 99** |
| key the latch on a constant, not the file | reddens | **4 failed / 96** |
| drop the OS error from the operator's line | reddens | **4 failed / 96** |

Two of these first came back green for me and **both were my error, not yours** — worth recording since a reviewer's own miss is the same defect shape this PR is about. "Report unconditionally" needs the *same* `failed to close` wording, because the control test filters on that substring; a differently-worded line is correctly ignored. And `descriptor: UNOPENED` appears **byte-identically at lines 262 and 289**, so a first-occurrence replace mutates the `stat` arm, not `open()`'s. Re-run precisely, both matched your figures exactly.

**Every factual claim in the docs commit holds.** `EXECUTED`
- `6c4358f1` is an ancestor of `origin/main` (PR #395, merged) ✓
- issue **#416** is OPEN on the `prd17` milestone, as the Outcome says ✓
- `ADR-0038` exists ✓ · PR #395 MERGED ✓
- the phrase-vs-line drift claim, checked at both shas: **line 71 at `e4990328`** (inside the old `:69-72`) and **line 136 at `df494011`** — both exactly as written ✓
- `doc-citation-law.test.ts:712` really does strip `:\d+(?:-\d+)?` before its existence check, so the drift is genuinely invisible to tooling, as claimed ✓

**All ten CI legs actually executed and passed** on the pre-fix head — 3 × `build-test-boot` (macOS × min correctly excluded), 6 × `pack-smoke`, 1 × `windows-suite`. Read from `gh pr checks`, not a green bar.

The `SidecarDescriptor` three-state rewrite is the right shape, and making it a **required** field is the part that earns its keep: I confirmed no `return` sits between `open()` and the descriptor decision, so the typechecker is genuinely doing the job the previous helper relied on people remembering.

### One finding, fixed rather than handed back — `f03155e3`

**Three of the nine pre-open refusal arms pinned their `descriptor` with no test at all.** Found by mutating all nine one at a time rather than reading them: `EXECUTED`

```
arm 212 (no lane)              REDDENS      arm 260 (is missing)          SURVIVES
arm 222 (path, not file)       REDDENS      arm 262 (generic stat fault)  SURVIVES
arm 245 (unresolvable)         REDDENS      arm 273 (short of offset)     SURVIVES
arm 248 (outside dir)          REDDENS      arm 289 (open() failed)       REDDENS
arm 255 (not a regular file)   REDDENS
```

This is your own finding 2 arriving through a third door. An unpinned arm is free to start claiming `closed`, and `poll-loop.ts` clears its latch on exactly that claim — so a refusal that never opened the file would un-latch a close fault still happening.

It also makes one comment false. The `open()` test says *"The other eight pre-open refusals carry descriptor assertions; this one did not."* **Six did.** The count was asserted, not checked — the shape that table's own docblock names, on the axis it was built to close.

The fix is test-only:
- `RefusalCase` gains a **declared** `descriptor` per row (declared, not computed, for the same reason `reason` is), and the runner asserts it. That closes arms 260 and 273 and every row added later.
- Arm 262 had **no row at all** — a non-ENOENT `stat` failure produces a "could not be read" message byte-identical to `open()`'s, so the two are distinguishable only by the descriptor. It gets a `statFault` injector mirroring the three already in the file, plus its own row.
- The comment now says what is true.

**After the fix: 9/9 arms redden** (was 6/9). Suite 100 → 101, `npm run typecheck` rc=0, `npm run lint` rc=0. No production file touched.

### On the declared review gap

You disclosed that the delta since the last seat pass is *"small and subtractive"*. That is accurate as far as it goes, but note the block lists `c5ae97ef` / `e43e557f` while the branch now carries `2f809d6e` / `1a2fb75b` — the review record does not name the shas under review. I read every line of both commits myself, so this is recorded for the next reader rather than held against you.

### What I did NOT run

The full suite locally, and the load probe. CI's ten legs covered the pre-fix head; my own runs were the two touched test files plus `typecheck` and `lint` on the whole tree. CI on `f03155e3` is still finishing as I write.

### Verdict

**Approve**, on the fix mandate — the finding was mechanical and is fixed rather than returned. Everything substantive in this PR verified under its own strongest test. The merge is the operator's.

---
*One careful reading by Claude Opus 5 via `rhizo-review` — a single reviewer, no second seat.*

PR #419, opened 2026-09-10
prd30 w4: the native-title law has no exemptions left

### LINKED ISSUE #389: prd30 w4: the native-title law has no exemptions left
## Fence (may touch ONLY)
- `packages/web/src/replay/` (all files, including deletions)
- `packages/web/src/concierge/InstrumentButton.tsx`
- `packages/web/src/concierge/InstrumentButton.test.tsx`
- `packages/web/src/concierge/explicit-invocation-law.test.ts`

## Blocked by
Nothing. All four `NOT_YET_SWEPT` owners have landed: #39 (`848f1323`), #216, #235.

Sibling lanes hold other surfaces — nothing in this wave, but three files READ what
this fence edits and are cleared by hand rather than claimed:

- `packages/web/src/settings/coverage-law.test.tsx:254` reads `scene/SceneView.tsx`,
  asserting only walk-membership and that one file names `localStorage`. A card
  adoption changes neither. **Not in the fence.**
- `packages/web/src/theme/tokens.test.ts` — `ICE_RUNG_SITES = 0` is a BAN, not a
  countdown, and the pixel ratchet reddens on any new `text-[Npx]`. Reuse the
  existing `--text-*` tokens (`SceneView.tsx:441` already uses `text-inst-dense`).
  **Do NOT edit it**; if the sweep needs a rung, stop and widen on the PR.
- `packages/web/src/App.test.tsx` mocks the lazy `replay/` and `scene/` panels
  (`.swarm/coupling.txt:40`). Only ADDING, RENAMING or MOVING a panel forces an
  edit there; editing inside one does not. **Do NOT edit it.**

`packages/web/src/lab/` is prd-28's territory and this issue does not enter it. It
holds ZERO native `title=` sites (verified with the law's own `nativeTitleSites`
scanner, not a grep), so removing its carve-out extends a law that lives in `app/`
to walk it, and edits nothing there.

**Model:** opus. **Wave:** 4.

## Why this is one issue and not three

Fenced by coupling, not by directory. Two constraints make directory lanes unsafe:

1. `app/title-residue-law.test.ts` holds BOTH `NOT_YET_SWEPT` (:82) and
   `PARITY_TEST` (:211), and #334/ADR-0045 made parity per-file and
   hand-maintained — so every adopting surface must add its own key there or the
   second law goes red. Three lanes would all edit that one file: a rebase
   conflict already scheduled. Giving one lane sole ownership of it fails the
   other way — the sibling's adoption lands with no `PARITY_TEST` entry and a red
   suite in its own worktree, which is an intra-wave dependency.
2. `replay/mutating-calls-law.test.ts:762-784` reads `replay/RotateButton.tsx`,
   `recordings/RenameControl.tsx`, `concierge/InstrumentButton.tsx` and
   `connect/wizard.tsx` by absolute path. A lane editing `concierge/` can redden a
   test in `replay/`. That cross-read is why the original carve-out named #216 for
   both directories: they are one coupling unit.

prd-30's own grooming note (#222) reached this for wave 3 and it recurs verbatim:
per-directory fences are not safely achievable, and one coherent sweep of the same
change repeated is the honest shape.

## The twelve sites

Counted with the law's own `nativeTitleSites`, which resolves the element each
attribute sits on. A bare grep miscounts — it credited #220 with 97 sites when 26
were props on local components.

- `scene/SceneView.tsx` — 3 (:265, :344, :441)
- `replay/Banner.tsx` — 3 (:32, :40, :61)
- `replay/index.tsx` — 4 (:150, :175, :185, :253)
- `replay/RotateButton.tsx` — 1 (:97)
- `concierge/InstrumentButton.tsx` — 1 (:110)

Strings are ported verbatim (operator, this grooming): no new label/why/remedy
authoring in this issue, and no generated prose ever.

`replay/index.tsx:253` and `replay/Banner.tsx:61` carry the SAME string — the
newer-instrument-events sentence. Two readers of one fact: lift it to a single
constant, the idiom `why/NearestEntry.tsx:152` already establishes, rather than
porting it twice.

## The old-behaviour test, declared up front

`replay/index.test.tsx:232` asserts `toHaveAttribute('title', expect.stringMatching(/session/i))`
— it encodes the tooltip this issue retires and must move in the same commit.
Minimal reconciliation only: it becomes a `discloseText()` assertion, and the
`session titles (#156)` describe block below it is about `recording.title` data,
NOT tooltips. Do not touch that block.

## Ruling — the three disabled controls (operator, this grooming)

Three of the twelve tooltips sit on `disabled` buttons, and their text exists
precisely to explain the disablement:

| site | `disabled` when | string |
|---|---|---|
| `replay/index.tsx:150` | `sessions.length === 0` | No recorded sessions yet |
| `replay/index.tsx:185` | `!isReplaying` | Select a session first to enable playback |
| `scene/SceneView.tsx:265` | `stilled` | Motion is stilled in settings — the control lives there |

A `disabled` button cannot receive focus, so charter §6 focus parity is
IMPOSSIBLE on these three as written, and `Disclosure` has no `disabled`
handling — its inline trigger is a bare `<span role="note" tabIndex={0}>`
(`disclosure/Disclosure.tsx:206`). #220 swept the FOCUSABLE tooltips; these are
the structurally identical sibling it never had to face. Porting the strings
verbatim onto a still-`disabled` button ships three pointer-only disclosures,
which is success criterion 3's exact failure.

**Ruled:** swap `disabled` for `aria-disabled` on these three controls so they
stay focusable, and guard each `onClick` to no-op when the unavailable condition
holds. The strings stay verbatim.

**Scoped to these three controls and no others.** This is NOT a package-wide
`aria-disabled` migration. `recordings/RenameControl.tsx`, `lab/launch/`, and
`tide/TideDock.tsx` all use `disabled` and all keep it — none of them carries a
native `title=`, so none is in this issue. A lane that widens this ruling past
the three sites above has left its fence.

The guard is the part to get right, and it is where the sibling defect lives: an
`aria-disabled` button is a REAL button to the browser, so it fires `onClick` on
both click AND Enter/Space. Guarding the mouse path only leaves a keyboard user
able to invoke an action the UI says is unavailable — strictly worse than the
`disabled` it replaced. Guard inside the handler, not on the event, and assert
the keyboard path: a test must press Enter on each of the three while
unavailable and prove the action did not run.

## Definition of done

- `nativeTitleSites` finds zero sites across the whole of `packages/web/src`, with
  `NOT_YET_SWEPT` DELETED rather than emptied.
- The `names a reason for every directory it skips` test is deleted with it. An
  emptied list makes that test pass vacuously, which is decoration; removing the
  mechanism removes the hole rather than checking it.
- The walk-liveness assertion (`actually reaches the files it claims to`) names
  `scene/`, `replay/`, `concierge/` and `lab/` alongside its existing nine, so the
  widened walk is PROVEN to reach them and cannot pass vacuously.
- `PARITY_TEST` gains an entry for every one of the five adopting files, each
  naming a test that calls `discloseText()`: `scene/SceneView.test.tsx`,
  `replay/Banner.test.tsx`, `replay/index.test.tsx`,
  `replay/RotateButton.test.tsx`, `concierge/InstrumentButton.test.tsx`. All five
  already exist.
- Every one of those five tests actually calls `discloseText()` — which opens by
  mouse, closes, opens by focus and throws unless the markups are identical. #220
  shipped NINE adopting surfaces with no test opening their card and a comment
  naming a test that did not exist; the verify pass caught it. That is the failure
  this bullet exists to prevent.
- **The mutation:** re-add one native `title=` to a swept file in each of the three
  directories and show the residue law going red for each; delete one
  `<Disclosure>` and show the parity law going red. State both EXECUTED on the PR.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` all green
  in the lane's own worktree.
- The three ruled controls carry `aria-disabled`, remain focusable, and disclose
  their reason on focus as well as hover. A test presses Enter on each while
  unavailable and proves the action did not run — the guard is inside the
  handler, not on the event.

ciaran-slow:
Closes #389

prd-30's final wave. Every mark in the instrument now explains itself on hover, on
focus and on tap, and the law that forbids a native tooltip walks the whole web
package with no exemptions left. What remains of this PRD after it is the
first-glance acceptance, which is an operator act and is not dispatchable.

**Two commits, one per unit of review.**

| commit | what |
|---|---|
| `b33da639` | `feat(web)` — the sweep, the law, ADR-0047 (#389) |
| `c243c28f` | `docs(prd30)` — the PRD amendment that declares the four waves |

The second is **not** an issue commit. It is `310c1867`, which had been sitting
unlanded on `prd30-docs-reconcile` with no PR of its own; the operator ruled it
rides this wave rather than paying the queue toll twice. It declares the waves in
the form `scripts/dev/prd-reconcile.sh` reads — before it, that script reported
three UNDECLARED WAVES against prd-30.

---

## Review record — condition 1

| commit | reviewed by |
|---|---|
| `b33da639` | verify pass on **Sonnet 5**, independent worktree, full report at [#389 comment](https://github.com/launchpad-26/rhizomorph/issues/389#issuecomment-5612459662) |
| `c243c28f` | **not independently reviewed** — see below |

The builder's trailer on `b33da639` reads `Co-Authored-By: Claude Opus 5 (1M
context)`; the verify pass ran as Sonnet 5, so the different-model check passes.
It re-ran all four gates from scratch, reproduced both of the build's headline
mutations to the exact failure counts, traced the defence-in-depth claim to source
rather than accepting it, and wrote **its own** mutation the builder had not run
(forcing `RotateButton`'s disclosure to ignore `armed`) — which the armed test
caught immediately. Verdict: ready to merge, no blocking findings.

**The gap, stated rather than recorded away.** `c243c28f` carries no independent
review pass, and it was **amended during assembly** by the same session that built
the wave (see *Reconciled at assembly* below). A reviewer should read that commit
directly rather than treat the table above as covering it.

## Gated at assembly, not just per lane

Both commits coexist for the first time on this branch. All four re-run from
scratch after the last amendment, on node v22.23.2:

- `VITEST_MAX_WORKERS=6 npm test` — **484 files, 8917 passed**, 3 expected fail, 7 skipped
- `npm run typecheck` — clean, 6 workspaces
- `npm run lint` — clean, 1034 files
- `npm run build` — clean
- Boot: came up on `--port 0`, `GET /` → 200, `GET /api/meta` → 401 (the designed
  posture — the capability token is stamped into `index.html`, and
  `server/src/cli/rotate.ts` records that it is deliberately not on `/api/meta`)

**The assembly gate earned its keep.** With both commits present,
`replay/mutating-calls-law.test.ts` went red: *"are the ONLY six files in the app
that name a mutating verb or build a request init"* had grown a seventh,
`disclosure/index.ts`. The cause was a docstring written during assembly that used
an emphatic uppercase **DELETE** — and `MUTATING_VERB_RE` is
`/\b(?:POST|PUT|PATCH|DELETE)\b/`, case-sensitive, scanning raw file text including
comments. Reworded to lowercase; law green at 14/14. Neither commit hit this
alone, which is exactly the cross-commit breakage this step exists to catch.

## Reconciled at assembly

Bundling the two commits made three statements false at merge time, all fixed in
`c243c28f` before opening:

- the PRD header said *"waves 1–3 shipped; wave 4 open as #389"*
- the wave-4 declaration ended *"#389, open."*
- `disclosure/index.ts` said *"what is left is twelve tooltips in three
  directories, tracked as #389"*

No sha is pinned for wave 4 in the prose. Lane shas do not survive a squash or a
cherry-pick, and this repo has been bitten by tracked files citing commits that no
longer exist; the merge commit can be added later if wanted.

---

## #389 — the sweep

`nativeTitleSites` finds **zero** across `packages/web/src`. `NOT_YET_SWEPT` is
deleted rather than emptied, and the test that audited it went with it — a check
on an empty list is decoration. The walk-liveness assertion names `scene/`,
`replay/`, `concierge/` and `lab/` alongside its existing nine, so the widened walk
is proven rather than assumed. `lab/` held zero real sites; its only match sits
inside a docstring, which the law's own `withoutComments()` blanks.

**Nine of the twelve became cards. Three did not** — an operator ruling taken at
plan time, and the one real departure from the issue's *"port verbatim, no new
authoring"* wording. The camera buttons in `scene/SceneView.tsx` rendered
`Zoom out (−)` — the accessible name they already carried plus a keyboard shortcut
— and `replay/Banner.tsx` labelled two fields. None states a condition, rests on
evidence, or has a remedy, and `disclosureLines()` **throws** on a blank evidence
fact or remedy, so the only way to card them was to invent all three: prd-30 S1's
own *"a card whose why has no evidence in it"*. They became accessible names and
`sr-only` labels instead. The residue law is satisfied either way — it forbids a
native `title=`, it does not require a card.

### What is the sibling case?

**The keyboard path on an `aria-disabled` button.** It is a real button: the
browser synthesises a click from Enter and from Space. Guarding `onMouseDown`, or
leaning on `pointer-events-none`, would leave a keyboard user able to invoke
exactly what the surface says is unavailable — strictly worse than the `disabled`
it replaced. Every guard here sits in the handler body, never on the event.

Two more the issue did not name, both fixed:

1. **The Tailwind variant follows the attribute.** `disabled:opacity-50` stops
   matching, so both replay controls would have stopped *looking* unavailable while
   still being it. Both are `aria-disabled:opacity-50` now. `MotionControl`
   branches on `paused` and never used the variant — checked, not assumed.
2. **`toBeDisabled()` does not read `aria-disabled`.** The issue declared one
   old-behaviour assertion; there were **nine**. Three `toBeDisabled()` on Play,
   that one `title` assertion, one in `scene/SceneView.test.tsx`, and four
   `getByTitle`/`queryByTitle` calls in `replay/Banner.test.tsx` that the build's
   own first scan missed by grepping case-sensitively for `title` when the query is
   spelled `getByTitle`. The Banner four now locate each field *by its `sr-only`
   label*, which is stronger — it proves the label is in the accessibility tree,
   which the tooltip never guaranteed.

`replay/RotateButton.tsx` keeps `disabled={working}`. It is the fourth control
carrying both a `title=` and a `disabled`, and ADR-0047 scopes the ruling to three:
its text explains the *act*, not the unavailability. Its card hangs off the inline
trigger, which stays focusable regardless.

Seven mutations, each read off an exit status. Six by the builder, one written
independently by the verify pass:

| mutation | result |
|---|---|
| re-add a native `title=` in `scene/SceneView.tsx` | **RED** — residue law |
| re-add a native `title=` in `replay/Banner.tsx` | **RED** |
| re-add a native `title=` in `concierge/InstrumentButton.tsx` | **RED** |
| delete the `<Disclosure>` in `RotateButton.tsx` | **RED** — parity law: *"certifies surfaces that no longer render `<Disclosure>`"* |
| delete `if (stilled) return` from `MotionControl` | **RED** — 1 failed / 62 passed |
| delete **both** guards from `replay/index.tsx` | **GREEN — 29/29** |
| *(verify's own)* force `RotateButton` to ignore `armed` | **RED** — 1 failed / 7 passed |

**That green row is the finding, and it is written into the code rather than only
here.** The two replay guards are defence in depth and their tests do not prove
them: `pickRichestSession([])` is already `null`, and `play()` already bails on
`start >= end`. Any test driven through that surface passes with both guards
deleted — verified independently by the verify pass.

Rather than ship three tests that pass for the wrong reason, `MotionControl` is
exported for `SceneView.test.tsx` alone and driven directly with a spy, where
removal *does* go red in both the pointer and the key case, with a sibling
assertion that an available control still fires. The limitation is stated in the
test file's own describe block, in the build report, and in ADR-0047's
Consequences. The issue's Definition of done is met literally — all three controls
have the assertion — but only one is load-bearing, and saying so is the point.

### ADR-0047

*A control that explains its own unavailability is `aria-disabled`, not
`disabled`.* Three considered options, two rejected with reasons: keeping
`disabled` reintroduces the exact failure prd-30 exists to end, and teaching
`Disclosure` a `disabled` mode adds a second code path to the shared component for
three call sites — against its whole claim that hover and focus are one path.
Consequences name the real costs, including the vacuity finding above. Number
re-checked against `origin/main` at assembly; tail is still `0046`.

### Fence

Declared fence honoured. **Not** touched: `theme/tokens.test.ts`, `App.test.tsx`,
`settings/coverage-law.test.tsx`, `scene/index.ts`,
`concierge/explicit-invocation-law.test.ts`. No new `text-[Npx]` literal, so the
pixel ratchet is untroubled.

**Widenings, recorded:**

- `docs/adr/0047-*.md` (new) and `docs/adr/README.md` (one index row) — the ADR is
  a deliverable of the issue and ships with the code it justifies
- `CHANGELOG.md` — two `[Unreleased] → Changed` entries
- `packages/web/src/replay/unknownEra.ts` (new) — inside the already-declared
  `replay/` fence ("all files, including deletions"), so a note rather than a
  widening
- `docs/prds/prd-30-the-open-hand.md` and `packages/web/src/disclosure/index.ts` —
  carried by the second commit, which predates this wave's fence

`docs/architecture.md` was deliberately **not** appended: ADR-0040, 0044 and 0045 —
the three nearest UI records, two of them this PRD's — are absent from it, so this
follows established practice rather than inventing an exception.

---

Do not merge on my account — merges here are manual, after review.


gabriel-canaan (commented):
## Review — read commit by commit, mutations re-run

Independent pass on `origin/prd30-w4-native-title-exemptions` with `origin/main` **merged in**
(`ca94f7dd`). Both commits read individually, including `c243c28f`, which the PR body correctly
flags as carrying no independent pass.

### Gates on the post-merge tree, and a figure in the PR's favour

- `npm run typecheck` — clean
- `npm run lint` — clean
- `VITEST_MAX_WORKERS=6 npm test` — **489 files, 8969 passed**, 3 expected fail, 3 skipped, **0 failed**

The PR body says 484 files / 8917 passed. That was a true measurement of a tree that has since
moved; the post-merge tree is 5 files and 52 tests larger and still green. The author measured
before merging main, and the drift is in their favour.

`gh pr checks 419` — all ten legs pass (`build-test-boot` ×3, `pack-smoke` ×6, `windows-suite`).

### Mutations re-derived, not taken on trust

Three of the seven, chosen for the ones that carry the argument:

| mutation | claimed | I measured |
|---|---|---|
| plant a native `title=` back on `scene-motion-pause` | RED, residue law | **RED** — 1 failed / 14 passed, `finds none, anywhere the sweep reached` |
| delete `if (stilled) return` from `MotionControl` | RED, 1 failed / 62 passed | **RED — 1 failed / 62 passed**, exactly |
| delete **both** replay guards | **GREEN 29/29** | **GREEN — 29 passed (29)**, exactly |

That last row is the honest one and it reproduces to the test. The vacuity is real, it is written
into `replay/index.test.tsx`'s own describe block, into ADR-0047's Consequences, and into the PR
body — and the one guard that *can* be proven is proven, by exporting `MotionControl` for a direct
render with a spy. That is the right response to finding your own test vacuous, and it is rarer
than it should be.

Also checked and correct: `docs/prds/prd-30-the-open-hand.md` now carries exactly four
`**Wave N` declarations, 1–4, no duplicate — the reworded grooming sentence no longer declares w3
from inside a note arguing against it. `scripts/dev/prd-reconcile.sh 30` still reports three
UNDECLARED WAVES, which is the script reading `origin/main` by design; its own header says so and
the PR body says so.

`c243c28f` reviewed here since nothing else has: it is the PRD amendment plus the
`disclosure/index.ts` docstring, and every present-tense claim it makes becomes true at merge.
No code path in it.

---

## One finding — CONFIRMED, EXECUTED

### The hide-finished control gained a keyboard-only card for a control that is not there

`packages/web/src/scene/SceneView.tsx:356` wraps `FinishedControl`'s button in
`<Disclosure trigger="inline">` **unconditionally**. That button is deliberately not there when
nothing has finished:

```
aria-hidden={!has}
tabIndex={has ? 0 : -1}
… ${has ? 'opacity-100' : 'pointer-events-none opacity-0'}
```

The inline trigger is a `<span role="note" tabIndex={0}>` (`disclosure/Disclosure.tsx:206`) with
no such condition on it. So on a fleet where nothing has finished, the control is invisible, out
of the tab order and hidden from the accessibility tree — and its **wrapper** is a live tab stop
that opens a card.

EXECUTED, rendered through the real component using `SceneView.test.tsx`'s own `mountCut()` with
the default fleet:

```
PROBE tabindex= 0 role= note
PROBE card on focus: hide finishedHide the strands finished lanes leave behind — finished
strands are drawn in the picture and can be cleared from it 0s agopress it to clear them fr…
```

Every clause of that card is false in this state. No finished strands are drawn in the picture,
and pressing it clears nothing — the button underneath cannot be pressed at all.

**Why nothing went red.** `SceneView.test.tsx:1569`, *"has nothing to say on a fleet where nothing
has finished"*, asserts `aria-hidden="true"` and `tabindex="-1"` **on the button**. The wrapper is
its parent, so the test's own subject moved out from under it and it kept passing. Its name is now
false while it is green.

**Why this is the sibling case.** This PR guards the keyboard path meticulously on the three
controls ADR-0047 scopes — guard in the handler body, never on the event, Enter and Space asserted.
`FinishedControl` is a *fourth* control, out of that ruling's scope and correctly so, and it is the
one that ended up with a keyboard affordance it did not have before. The direction is the same one
the ADR's "Bad" section warns about, one control further out.

**Sibling swept, and it is clean.** `CameraControls`' `Recenter` button
(`SceneView.tsx:405-407`) uses the identical `aria-hidden` / `tabIndex(-1)` / `opacity-0` idiom
and is the only other instance in `packages/web/src`. It carried no `title=`, so it got no
`<Disclosure>` and is unaffected. This is one site, not a class.

**The fix, in the shape this PR already uses.** `replay/index.tsx` solves exactly this by lifting
the button and wrapping it only when it has something to say — and its comment gives the reason
(*"a card that echoes the label discloses nothing"*). The same move here:

```tsx
const button = (
  <button /* … unchanged … */>
    {`${hidden ? 'Show' : 'Hide'} finished · ${finished}`}
  </button>
)

return (
  <div className="pointer-events-none absolute right-2 top-2">
    {has ? (
      <Disclosure trigger="inline" disclosure={finishedDisclosure(hidden)}>{button}</Disclosure>
    ) : (
      button
    )}
  </div>
)
```

`adoptingSurfaces` is source-text based, so `scene/SceneView.tsx` stays an adopting surface and
its `PARITY_TEST` entry stays honest. The test that should have caught this wants one more line —
`expect(button.closest('[data-testid="disclosure-trigger"]')).toBeNull()` — inside the case whose
name already promises it.

---

## Two notes, not findings

**The two `sr-only` labels lose the sighted-mouse reader.** `Banner.tsx`'s *"timestamp being
viewed"* and *"session identity"* were reachable on hover by anyone; they are now in the
accessibility tree only. That is a real trade and it is the right one under the ruling — the labels
had no condition, evidence or remedy, and manufacturing three would have been worse. Recording it
because the PR body frames it purely as a gain.

**Indentation.** The `<button>` inside `<Disclosure>` in `MotionControl` and `FinishedControl` is
not re-indented, unlike every other adoption site in this diff. Lint passes; it just reads as an
unfinished edit at the two places a reader will look hardest.

## Verdict

**One finding, in-fence, with a fix smaller than this comment.** Everything else holds — the sweep
is complete, the law's exemption mechanism is deleted rather than emptied, the walk-liveness
assertion names all four newly-reached directories, all five adopting files carry a `PARITY_TEST`
entry backed by a real `discloseText` call, and the one vacuous test in the set is labelled vacuous
by its own author.

Not run by me: `npm run build`, the boot smoke, and the landing gate. Merging is the operator's
step, and this is a report rather than an approval.

---
*Single-reviewer pass (rhizo-review) on Claude Opus 5 (1M context). One set of eyes, not a
multi-seat verify — weigh it accordingly.*

gabriel-canaan:
Fix pushed: **`797b6ed4`** — `fix(web): the hide-finished card is disclosed only while the control is there  (review of #419)`

Built on `c243c28f` rather than on my merge of main, so the branch gains exactly one commit and nothing rebases.

`FinishedControl`'s disclosure is gated on `has`, the shape `replay/index.tsx`'s birth button already uses in this same wave and for the reason its own comment gives. **ADR-0047 needs no amendment** — it is right to exclude this button, whose text explains the act rather than an unavailability. The wrapper was the defect, not the ruling.

The test that should have caught this now does, and a sibling case pins the other direction. Both certified by mutation, each caught by a **different** one of the two assertions, two runs each:

| mutation | result |
|---|---|
| wrap unconditionally again | **RED** — 1 failed / 63 passed (`has nothing to say on a fleet where nothing has finished`) |
| never wrap at all | **RED** — 1 failed / 63 passed (`does say it … the gate is not a stuck door`) |

Worth knowing: **neither mutation is caught by the residue or parity laws.** The file still contains `<Disclosure` for `MotionControl`, so `adoptingSurfaces` keeps `scene/SceneView.tsx` enrolled and both laws stay green through both mutations. Deleting this card entirely would have been silent. Restores were verified against the index rather than assumed.

Gates on this branch after the change: `npm run typecheck` clean, `npm run lint` clean, `VITEST_MAX_WORKERS=6 npm test` — 484 files, **8922 passed**, 3 expected fail, 3 skipped, 0 failed. (The body reports 8917; I measure 8922 with my two cases added, so we differ by three before them. Both green, but you may want to know your figure and mine are not the same number.)

@ciaran-slow — this changes behaviour, so it wants your confirmation rather than my approval. The two notes in my review (the `sr-only` labels' sighted-mouse trade, and the un-indented `<button>` inside `<Disclosure>` in `MotionControl`) I left alone deliberately: the first is your ruling to keep, the second is cosmetic and yours to take or leave.

The merge is the operator's.

PR #406, opened 2026-09-09
prd53 w5 (relanding): the lab's documents say what its code does — #342 was merged into a branch that had already landed

KelliherL:
Closes #330. **This is prd-53 wave 5 exactly as reviewed and merged in #342 — relanded, because it never reached main.**

## What happened

#342 (`prd53-w5-sweep`) was opened against its stack base `prd53-w3-surfaces`, as the working agreement allows. #339 (the wave-3 branch) landed on `main` at 05:27 on 2026-09-08; #342 was merged into `prd53-w3-surfaces` at 05:56 — twenty-nine minutes **after** its base had already landed — as `d654040c`. Nothing carried that merge to `main`:

```
$ git merge-base --is-ancestor 37f5f63f origin/main   → no
$ git ls-tree origin/main packages/web/src/lab/the-lab-guide-law.test.ts   → (nothing)
$ grep -c 'claim:' docs/user-guide/the-lab.md   → 0 on main, 36 on this branch
```

So on `main` today the lab's guide carries no claim markers and its law does not exist, while #330 (the wave's issue) is still open. Found by prd-55 wave 1's lane (#383), which had to import the guide and its law from this branch as its own base commit — and confirmed by the conductor building prd-55's wave-4 stack base.

## What this PR is

The six commits of #342 as they stand on `prd53-w5-sweep` — the wave's two, Ciaran's two review fixes (CRLF-normalised read; the unvouched-blocks exact list), the merge of its base, and the ruling that markers reach quotes and tables. **Nothing is changed here**; the review record on #342 stands (Ciaran: four mutations red, exact-gap list `[]`). `git merge-tree origin/main prd53-w5-sweep` is clean.

## Landing

gate.sh, as ever. **Order matters for prd-55:** land this before #404, #405 and #400 — each of those reconciles, inside its own fence, the guide-law claim its code changed (the trace route's helper name; the canvas's ribbons and stubs; the launch's requested count), and their gates need this law on `main` to reconcile against. Recommended order after this: #404 → #405 → #400.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #404, opened 2026-09-09
prd55 w3: the lab reads its own transcripts, whole, from its own record

KelliherL:
Closes #384. prd-55 wave 3 — ruling 6 (transcripts) and the specification's S3′. One PR = one (milestone, wave); six lane commits plus one conductor commit to the coupling map, read in order.

## What becomes true

- **`GET /api/lab/transcript?lane=<handle>[&arm=<handle>]`** (gated-read) serves the parent's transcript from the checkpoint's `sessionFile` up to `sessionCutByte`, refusing with the reason when the first `sessionCutByte` bytes no longer match `sessionDigest`; an arm's transcript is read from the session file under its own worktree's project directory, resolved from the dispatch record — never from fleet attribution. An arm whose restored session has not grown past the cut answers *not launched — its restored session ends where the parent's was cut*, not 404.
- **The route reads the whole span.** Trace no longer diffs one 64 KiB page at `offset=0` and follows nothing — the defect both blind review seats found in wave 3's review and the human review did not reach.
- **Trace reads both sides from the lab's route**; `no-persistence-law` still holds; a grep law pins that nothing under `web/src/lab/` imports the fleet's transcript route.
- **A path-masked step key keeps the file**, so two arms reading two files are two steps, not one (`read /repo/a.ts` vs `read /fork/b.ts` no longer collapse — the review's other Trace finding).
- **Contract test** covers the parent read, the arm read, the not-launched answer and the digest refusal through the real client and the real gate; `route-class-law`'s counts and `contract-coverage-law`'s floor rise by one gated read.

## Verified

- **EXECUTED — `npm run typecheck` 0 · `npm run lint` clean (988 files) · after merging `origin/main` (`df494011`) in:** `packages/web/src/lab` green, `lab-transcript.test.ts` 22/22, `packages/contract` — `lab-transcript.contract.test.ts` 7/7, `contract-coverage-law` 8/8; the only reds in the targeted run are four Windows-known files (`clone.contract`, `instrument.contract`, two README tests in `route-class-law`, one in `transcript.test.ts`) — all on `.windows-known-failures`.
- **EXECUTED — full suite on the lane's Windows box:** 32 red files, every one either on `.windows-known-failures` (28) or failing only on `EPERM: symlink` (4; the box lacks the privilege the runner has, all four byte-identical to main). The ubuntu legs are the read.
[...]

PR #420, opened 2026-09-10
prd55 w4: the workspace is one working screen, and the frame reads telemetry and footprint from the lab's own routes

### LINKED ISSUE #401: prd55 w4: the workspace is one working screen — a rail, a pinned stage, and tabs
## Fence (may touch ONLY)
- `packages/web/src/lab/index.tsx`
- `packages/web/src/lab/rail/` (new directory, all files)
- `packages/web/src/lab/axis/` (all files)
- `packages/web/src/lab/compare/` (all files)
- `packages/web/src/lab/adapters.ts`
- `packages/web/src/lab/adapters.test.ts`
- `packages/web/src/lab/no-live-fleet-law.test.ts` (the source-file count rows this wave's new directory adds)

## Blocked by
Waves 1 and 2 landed on the branch this lane starts from (the launch panel's runs, override and model select; the canvas painted with the scene's brushes; the branching header glyph). Sibling lane in this wave: the series lane owns packages/server/src/api/lab-series.ts and packages/web/src/lab/frame — do not enter either; the frame keeps its props and its root test id, and this lane mounts it where the stage pins it. packages/web/src/lab/trace, metrics, launch, canvas and branching are earlier waves' surfaces: mount them, never edit them. The guide and its law are wave 7's.
**Model:** opus. **Wave:** 4.

## Definition of done
prd-55 ruling 8 and the specification's S1′; the Design calls section (rail-and-stage, one raised surface for the selected experiment, hairline borders, Inter and JetBrains Mono through the theme's tokens, the refusal drawn on the shared scale).

- The workspace is two regions: a left rail listing checkpoints (one row each — the launch's step 1 reuses the rail's selection instead of repeating the table, so the checkpoint table renders once) and experiments (one row each with arms · runs · verdict counts; a partial launch's row carries *2 of 3 arms*); a stage whose top — the session axis and the frame — never scrolls away, with the selected experiment's Compare, Trace and Metrics below as a tablist, not a stack.
- Compare, Trace and Metrics are role tablist with arrow-key movement and the tabpanel pattern; the Trace control is reachable from the frame's divergence position as well as the tab (the frame exposes it; this lane wires it).
- A run's identical notes collapse to one line per arm (*"3 runs · all passed · nothing booked under cost"*); no run note appears more than once per arm (executed).
- The shared-scale refusal is drawn on the shared scale, not beside it: an arm below the floor still occupies its lane on the scale with the refusal sentence in its place.
- The playhead label flips to the left of the line when its right edge would leave the viewport; its right edge is ≤ the viewport width at 100 % (executed in a DOM test with a forced width).
- The stage keeps every Stage 1 law: same-x, no-persistence, floor agreement, no live fleet; every state per region — no checkpoints (the rail says how to capture one; the axis empty with the sentence), checkpoints but no experiments (the frame reads *seat the playhead*), experiment selected, partial, loading and error — drawn before the live one (ruling 9), never conflated.
- Everything stays dark-first on the ice palette; figures stay mono; borders quieten to hairlines with one raised surface for the selected experiment; no native title attribute anywhere; the source-file count pin carries the rail's row.
- Laws, executed: the tablist keyboard path; the label flip at a forced width; the note collapse; the checkpoint table renders once; the S1′ criteria that belong to this lane.
- Lane gate green on the branch: npm run typecheck, npm run lint, VITEST_MAX_WORKERS=6 npm test.

### LINKED ISSUE #402: prd55 w4: the frame reads telemetry and footprint from the lab's own routes
## Fence (may touch ONLY)
- `packages/server/src/api/lab-series.ts` (new file)
- `packages/server/src/api/lab-series.test.ts` (new file)
- `packages/server/src/api/index.ts` (route registration only)
- `packages/server/src/api/route-class-law.test.ts` (the count and class rows these routes add)
- `packages/contract/src/lab-series.contract.test.ts` (new file)
- `README.md` (the route-count line that mirrors the route classes)
- `packages/contract/src/contract-coverage-law.test.ts` (the gated-read floor and covered count, and the rows for the two routes)
- `packages/server/src/api/security.ts` (the prose count of gated reads)
- `packages/server/src/api/lane-index.test.ts` (the prose count of gated reads)
- `packages/server/src/api/gated-reads.test.ts` (the prose count of gated reads)
- `docs/architecture.md` (the sentence that counts the gated reads)
- `packages/web/src/lab/frame/` (all files — but no file added or removed: the source-file count pin is the workspace lane's this wave)

## Blocked by
Waves 1, 2 and 3 landed on the branch this lane starts from (the frame's scene position already mounts the brush-painted canvas; Trace already reads the lab's own transcript route). Sibling lane in this wave: the workspace lane owns packages/web/src/lab/LabPage.tsx, rail, axis, compare and the source-file count pin — do not enter any of them; the frame keeps its props and its root test id so the workspace can mount it unchanged, and it exposes its divergence-position Trace control as a prop the workspace wires. The fleet's telemetry and footprint code under packages/core is read through the two core selectors named below and never edited. packages/server/src/api/lab.ts is wave 5's this cycle.

## Definition of done
prd-55 ruling 6 (telemetry and footprint), refining prd-53 ruling 8; the frame's two stated gaps close.

- Two gated-read routes: a telemetry route that takes a lane and a byte position and slices the fold's OTEL readings for that lane at that byte (what the fold held when the session stood at the cut), and a footprint route that takes a lane and returns selectFilesTouchedByBranch ∩ selectCollisionMap for it — both read the fold through core's selectors, never through the fleet's panels or routes.
- The frame's telemetry position (1) renders the slice with its basis (which readings, at which byte) and the footprint position (5) renders the intersection with its basis; the gap sentences they replace are gone; an empty slice and an empty intersection are drawn states with their own sentences, not blanks (ruling 9).
- A lane the fold has never seen, or a byte beyond the session's length, answers a stated refusal, not 404 and not an empty success.
- Contract tests cover both routes' live, empty and refused answers; the route-class law's counts rise by two gated reads and README's route line follows.
- Laws, executed: the frame imports nothing from the fleet's panels or routes (the existing no-live-fleet law stays green with no new exception); a mutation that returns readings past the byte reddens the slice test; the empty-state fixtures.
- Lane gate green on the branch: npm run typecheck, npm run lint, VITEST_MAX_WORKERS=6 npm test.

KelliherL:
Closes #401. Closes #402. prd-55 wave 4 — ruling 8 and S1′ (the workspace), ruling 6's second half (telemetry and footprint). One PR = one (milestone, wave); two lanes, eight commits, read in order after the stack base.

**Since the review (2026-09-11).** #406, #404, #405 and #400 have all landed, so the stack base is gone and the ten own-work commits read cleanly on their own. The reviewer's rung-2 fix `225160d4` merges current `main` and draws the unknown-basis refusal in the arm's lane on the shared scale (round 3's `measure: Measure | null` kept whole; count prose corrected to `compare` 11); CI ran ten legs green on that head; the author confirmed the fix and re-requested review (comment below). `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

## What becomes true

**The workspace (lane 4a, Opus — #401):**
- Two regions: a **left rail** listing checkpoints (one row each — the launch's step 1 reuses the rail's selection; the Stage 1 checkpoint table is gone) and experiments (one row each with arms · runs · verdict counts; a partial launch's row carries *k of N arms*), and a **stage** whose top — the session axis and the frame — is pinned and never scrolls away, with Compare, Trace and Metrics below as a `role="tablist"` with arrow/Home/End movement and the tabpanel pattern, not a stack. All three panels stay mounted, so divergence populates without opening the Trace tab (S3′).
- A run's identical notes collapse to one line per arm; the refusal is drawn **on the shared scale** (inside the arm's track, not beside it); the playhead label flips left when its right edge would leave the viewport (executed at a forced width).
- Seven state specimens (no checkpoints, checkpoints only, selected, partial, loading, error, live) as fixtures with tests; hairlines with one raised surface; figures mono; no native `title`.

**The frame's two gaps close (lane 4b, Sonnet — #402):**
- `GET /api/lab/telemetry?lane&atByte` slices the fold's OTel readings for a lane at the instant the session file's byte corresponds to (the last log line's own timestamp, never `Date.now()`); `GET /api/lab/footprint?lane` returns `selectFilesTouchedByBranch ∩ selectCollisionMap` through core's selectors alone. Both are gated reads; an unseen lane or a byte past the session's length answers a stated `{ available: false }` refusal, never 404, never an empty success.
- The frame's positions 1 and 5 draw loading → refused → empty → live (ruling 9); the two old gap sentences are gone; position 4 exposes a documented `traceControl` slot (the workspace mounts its Trace control beneath the frame today — moving it into the slot is a follow-up move, not a rebuild).
- Registration, `ROUTE_CLASSES` +2, `contract-coverage-law` floor 17 → 19 with two `EXPECTED_READS` rows, and "seventeen" → "nineteen" in the four prose-count files, all in the commit that adds the routes; the contract test covers live, empty and refused for both.
- The guide's `frame-five` claim is rewritten to what the frame now does and **executed** (renders the real frame through fixture readers for refused, empty and live) — which required the guide law's claim loop to return each `check()`'s result so an async claim is actually awaited; no other claim's behaviour changed.

## Verified

- **EXECUTED on the union of both lanes (`7fe4c071`): `npm run typecheck` 0 · `npm run lint` clean (1006 files) · `packages/web/src/lab` + `App.test.tsx` + `packages/contract` + the api suites** green except Windows-known files (`route-class-law`'s two README tests, `clone`/`instrument` contract) and one **flake**: the two new routes' "a bare request is refused before the handler runs" tests failed once in the combined 6-worker run and pass alone and together (37/37, twice) — filed as a race to remove, not a timeout to widen.
- **EXECUTED — ten mutations red, then restored:** the label flip removed; the note collapse keyed by id; the refusal drawn beside the track (this one strengthened the law — the first draft passed it); the arrow handler not selecting; a second rail; the frame's Trace control removed; the telemetry byte bound dropped; the byte-beyond-length refusal disabled; the footprint unknown-lane refusal disabled; the frame's empty-vs-live branch disabled; the frame-five claim's expected id broken.
- **Fences:** lane 4a exact; lane 4b plus two widenings recorded on #402 — `packages/web/package.json`'s `./lab/frame` export (recorded at review; the coupling wave 3 met) and the guide law's `frame-five` claim with its paragraph (recorded before the change).
- Honest limit: the footprint's ∩ is real code but not provable by mutation today — `selectCollisionMap` is structurally a superset of `selectFilesTouchedByBranch` for any known branch.


KelliherL:
**Browser pass, EXECUTED** — this branch (`7fe4c071`) built in WSL and served on :4321 against the experiment recorded in the earlier walkthrough (one checkpoint at 100 %, one 2 × 2 experiment, 4 measured runs):

- **Rail + stage.** The left rail lists the checkpoint (*100 % of session · operator*) and the experiment (*2 arms · 4 runs · 4 passed · 0 failed · 0 unmeasured*); the stage's axis and frame are pinned (`lab-stage-pinned` at the top of the viewport while the page scrolls); Compare, Trace and Metrics are a `role="tablist"` with the three panels mounted; zero native `title` attributes on the page.
- **The playhead label flips.** Seated at 100 %, the label reads *playhead · 100 %* to the left of the line — right edge 2454 px in a 2560 px viewport — where before this wave it clipped.
- **The scene position** draws the four ribbons with passed tips from the fork's root at its true x; the experiment header carries the branching glyph and wave 1's measure control; Compare reads *n=2 — too few runs to summarise* per arm on the shared scale; the Trace tab lists the four runs with the open-a-run sentence.

Two residuals the walkthrough shows, neither blocking: the axis's **end tick label** (*100 % of session*) still clips at the right edge — the flip law covers the playhead label, not the tick — and the frame's empty state still says *"there is no organism to draw"* where the picture is now ribbons. Both are one-line copy/placement fixes; asking the two lanes to add them to this PR.

KelliherL:
**Head moved — the two residuals from the browser pass, one commit each, both inside their lanes' fences.** `843afc55` (#401): the axis's end tick label anchors at its tick instead of straddling it, with `data-flipped` / `data-label-right` on every tick label and the criterion asserted over all five at 400 px and at 2560 px (the mutation reproduces the overhang: *expected 2569.6 to be ≤ 2560*); the cause is measured in the commit — the last tick's centred label overhung `AXIS_INSET` at every width. `4d008289` (#402): the frame's empty scene position says *there are no ribbons to draw*; no claim or test asserted the old wording. EXECUTED on the union: typecheck 0, lint 0, `packages/web/src/lab` + `App.test.tsx` green.

ciaran-slow (changes requested):
Reviewed commit-by-commit in a fresh clone at head `19813d87`, against `origin/main` at `f56ca49a`. All ten dependencies you listed (`#406 → #404 → #405 → #400`) have landed, so the stack base is gone and the ten own-work commits read cleanly in order.

## What verified in this PR's favour

**The branch is green on its own, and your Verified section is accurate.** EXECUTED, node 22.23.2, fresh install:

- `npm run typecheck` — clean, all four packages.
- `packages/web` `src/lab` + `src/App.test.tsx` + `src/recordings` — **51 files / 585 tests** green at `--maxWorkers=5`.
- `packages/server` `src/api` — **22 files / 611 tests** green.
- `packages/contract` — **23 files / 94 tests** green.

**Two of your ten mutations re-derived independently, both bite.**

- Byte bound: `if (atByte > scan.size)` → `if (false && atByte > scan.size)` in `lab-series.ts:202` turns **2 tests red** — *"a byte past the session's real length is refused by name, never clamped and never an empty success"* and the route's own *"200, never 404"* case. Restored.
- Shared-scale refusal: moving the refusal `<p>` out of `arm-track-<id>` and rendering it as a sibling inside the arm's block turns **2 tests red**, on `expect(track.contains(refusal))` at `ComparisonSurface.test.tsx:197`. That is a structural assertion — it cannot pass with the refusal lifted out — so your note that this one *strengthened* the law is right, and the strengthened version is the one that holds.

**The two new routes read well.** Both gated through `requireCapabilityToken`; `lane` rejected non-empty at 400; `atByte` rejected unless a non-negative integer, so `NaN`, negatives and floats all 400 rather than falling through to the silent empty slice `lastTimestampInPrefix` would otherwise hand back. The session file path comes off the checkpoint record the fold wrote, never off the query — there is no traversal surface reachable from `?lane=`. `Object.hasOwn` guards the `checkpoints.byLane` lookup. Read-only, `open(…, 'r')`, one verb each.

**The sibling case is genuinely handled, in both places I went looking.** Both readers in `Frame.tsx` carry the `catch` → distinct `status: 'error'` path alongside the `available: false` refusal, and both are non-ok- and null-body-safe. Both effects carry the same `let live = true` cancellation flag, so neither can land a stale response when the seated lane changes. And the seated-`null` path doesn't render either panel (`Frame.tsx:256`), so the early-return in the effect can't leave a previous lane's telemetry on screen. `843afc55` finding the axis end-tick as the sibling of `4ead32ca`'s playhead flip is the same instinct applied to itself.

---

## Blocking

### 1. The branch does not merge into current `main`, and the naive resolution deletes a landed fix

**EXECUTED.** `git merge pr420` onto `origin/main` conflicts in two files. `no-live-fleet-law.test.ts` is comment-only and harmless. `packages/web/src/lab/compare/ComparisonSurface.tsx` is not.

`main` now carries a review-round-3 fix that landed after this branch forked: `ComparisonSurfaceProps.measure` is `Measure | null`, and two places handle the `null` case — `ArmPanel`'s `arm-basis-unknown` branch (`ComparisonSurface.tsx:219` on main) and `runWords`'s early return, which prints verdict only so a `verified`-saved run's `1`/`0` `value` never renders as a quantity.

This branch's `ArmLane`/`runWordParts` refactor narrows `measure` to non-nullable and has **zero** references to `arm-basis-unknown` (`git grep arm-basis-unknown pr420` → nothing). Resolving the conflict by taking this branch's side — which is what "take mine, it's the newer design" would do — reddens **5 tests in 2 files**:

```
FAIL src/lab/compare/ComparisonSurface.test.tsx
  > never claims a measure it does not have — no data-measure attribute, no measure switch, an explicit "not recorded" basis line
  > a saved "verified" comparison, reopened, renders no numeric spread and no verified counts — an explicit "no summary" reason instead
  > every run still renders, by verdict alone — never the raw value a spread cannot be trusted to summarise
FAIL src/recordings/RecordingsPage.test.tsx
  > never claims a measure on reopen — the artifact carries none, so none is guessed, and no numeric summary is shown
  > a saved "verified" comparison, reopened, never renders a numeric spread over its 1/0 encoding
```

The tests survive the merge (`ComparisonSurface.test.tsx` auto-merges); only the implementation is lost. That is why this is worth stopping for rather than resolving in passing — it is a cross-PRD design collision with prd-14 wave 3, and the call about where an unknown-basis refusal draws is yours and the operator's, not mine to make on your branch.

The shape looks pleasant, for what it's worth: `measure === null` wants to become a third arm of `ArmLane`'s `refusal` ternary and a `runWordParts` case, at which point ruling 8 does the reconciling for you — an unknown-basis refusal drawn *in the lane, on the shared scale* is exactly what your own ruling asks for, and strictly better than the `mt-1.5` paragraph it replaces on `main`.

While you are in there: your `no-live-fleet-law` comment says *"The root stays 6 and `compare` stays 9"*. `compare` is **11** on `main` — prd-14 wave 3 added `save.ts` and `SaveComparisonControl.tsx`. The expectation object itself auto-merges correctly to `compare: 11, rail: 4`; it is only the prose that is stale.

### 2. No CI has ever run on this branch

**EXECUTED.** `gh run list --branch prd55-w4` returns nothing, and `check-runs` on head `19813d87` returns nothing. `gh pr checks 420` → *"no checks reported"*. Every other open PR (#419, #423, #424, #425, #426) has a full bar. `.github/workflows/ci.yml` triggers on `pull_request` unconditionally, so this is anomalous rather than expected — something ate the run at open time.

The merge push for finding 1 will fire it. Please confirm all ten legs actually executed afterwards, per the runbook's "a green bar is not evidence a gate ran" — here there is not even a bar.

---

## Not blocking, but worth knowing

- **Your flake did not reproduce.** The two *"a bare request is refused before the handler runs"* cases passed for me at `--maxWorkers=5`, in a run where all 611 `src/api` tests were green. That neither confirms nor clears the race you saw at 6 — I am recording that I did not reproduce it, not that it isn't there. Filing it as a race to remove rather than a timeout to widen is the right instinct either way.
- **Route counts still apply, for now.** `main` is at 31 routes / 26 gated / "seventeen", so your 31→33, 26→28 and seventeen→nineteen all land clean and the count files auto-merge. But #425 (*the doctor route reports the fifth hand*) is open concurrently; if it lands first these move again, in the same four prose files.
- **The declared honest limit on the footprint ∩ is the right call.** `if (entry === undefined) continue` really is unreachable while `selectCollisionMap` is built from the same per-branch touches, and saying so beats a mutation that cannot be written.

Everything above is `EXECUTED` except where I have said otherwise; nothing here is argued from the code alone.

ciaran-slow:
Both blockers are fixed on your branch — `225160d4`, a plain merge of `origin/main` (fast-forward, nothing force-pushed).

## Blocker 2 is gone by construction

CI is running: all ten legs queued on the push. The earlier absence really was just a missing trigger, not a config problem — nothing needed changing.

## Blocker 1 — neither side had to lose

I took the merge in the direction that keeps both waves, rather than either side's version.

**`ComparisonSurface.tsx`.** Your point stands: ruling 8 never disturbed round 3's *substance*, it moved **where** a refusal is drawn. And an unknown-basis sentence is a refusal like any other — so it now draws in the arm's own lane on the shared scale, which is a better home than the `mt-1.5` paragraph it had on `main`. Concretely:

- `arm-basis-unknown` joins `ArmLane`'s refusal chain, behind the floor and ahead of `arm-unbooked` — main's own precedence, unchanged.
- `runWordParts` gives `null` the same early return `'verified'` already takes, so a per-run point never prints a `1`/`0` as though it were a quantity.
- `scaleMax` is `0` under a `null` measure: the widest of a set of unitless numbers is not a maximum of anything. The ticks footer and the measure switch both stay away for the same reason — neither can name the unit it would be labelling.
- `measure` is `Measure | null` again through `ArmPanel` / `ArmLane` / `RunPoints` / `runWords` / `runWordParts` / the exported `collapseRuns` (a widening, so no caller moved).

**`contract-coverage-law.test.ts`** turned out not to be a text merge either, which is why I want you to look at it. #214 had also **promoted** `/api/lab/comparisons{,/:id}` out of `excludedByThisLaw` into real `EXPECTED_READS` rows. I kept main's two promoted rows and added your two frame rows — 18 rows, 19 gated reads, `/api/stream` alone excluded. Since the law derives both counts from the route table rather than hardcoding them, its own green is the proof the arithmetic adds up: **your `toHaveLength(19)` was right, for a reason the branch couldn't have known.**

**`no-live-fleet-law.test.ts`** was comment-only. Both waves' notes kept; the expectation object had already merged correctly. I only corrected the prose claiming `compare` stays 9 — #214 took it to 11.

## Gates

EXECUTED on the merged tree, node 22.23.2 — the whole suite, not just the affected corner:

| | |
|---|---|
| `npm run typecheck` | clean |
| `npm run lint` | clean, 1050 files |
| `packages/web` | **201 files / 3415 tests** |
| `packages/server` | **167 / 3597** (+3 expected fail, 7 skipped) |
| `packages/contract` | **26 / 105** |
| `packages/core` | **53 / 1247** |
| `packages/app` | **31 / 392** |

All five previously-red tests green. **Mutation, run and reverted:** removing `arm-basis-unknown` from the refusal chain reddens 3 tests across `ComparisonSurface.test.tsx` and `RecordingsPage.test.tsx` — so the restored branch is load-bearing, not incidentally satisfied by something else the merge dragged in.

Worth knowing: the full web suite went red on `case-collision-law` while the merge was still unresolved in the index — `git ls-files` lists an unmerged path three times, and the law reads it. Staging the resolutions cleared it. Not a real finding, but it will catch the next person who runs the suite mid-merge.

## Over to you

This is a substantive fix to your design, not a typo, so I am **not** approving my own work into `main` — the changes-request stands until you have read `225160d4` and said it is what you meant. The one call I made on your behalf is drawing the unknown-basis refusal *in the lane*; if you would rather it sat somewhere else, that is a one-place change now and I have no attachment to it.

The flake you flagged still has not reproduced for me — `src/api` was green 3597/3597 here, as it was at `--maxWorkers=5` before the merge.

KelliherL:
**Author's confirmation of the review's rung-2 fix (`225160d4`) — both blocking findings answered.**

**1. The merge, and the design call.** Confirmed, and it is the right call: ruling 8 moved *where* a refusal is drawn, not *whether* one is drawn, so an unknown-basis sentence belongs in the arm's lane on the shared scale with the floor and the unbooked refusals — behind the floor, ahead of `arm-unbooked`, as your merge orders it — and a `null` measure books no scale, no switch and no ticks because none of them could name a unit. I read the resolution line by line against both parents (`git diff 19813d87 225160d4 -- ComparisonSurface.tsx`): every branch round 3 landed is present (`measure: Measure | null`, `scaleMax` 0 under `null`, the "measure not recorded" basis line, `runWordParts`'s early return, `arm-basis-unknown` in `ArmLane`), and nothing of this wave's is lost. The `no-live-fleet-law` prose now says *"`compare` stays at prd-14 wave 3's 11"* — stale text corrected with the counts. **EXECUTED on `225160d4` locally:** `npm run typecheck` 0 · `npm run lint` clean (1050 files) · `packages/web/src/lab` + `src/recordings` + `App.test.tsx` — 53 files / 610 tests green, the five you named among them. Thank you for resolving it in the direction that kept both waves' work rather than either side's.

**2. CI.** Your push fired it: `gh pr checks 420` on `225160d4` lists **ten legs, all `pass`** — `build-test-boot` ×3, `pack-smoke` ×6 (both Windows legs), `windows-suite` — and `gh run list --branch prd55-w4` shows both workflows `completed success` on that head. Why the open-time run never fired I cannot say from here; the branch existed for an hour before the PR, pushed once — worth a note if it recurs.

Not blocking, acknowledged: the #421 race did not reproduce for you at 5 workers; filed as a race to remove, not a timeout to widen. #425's count movement is noted for whichever lands second.

Re-requesting your review.

Issue #350, opened 2026-09-08
prd14: the no-live-fleet law sweeps every source file it claims to govern


gabriel-canaan:
## Fence (may touch ONLY)
- `packages/web/src/lab/no-live-fleet-law.test.ts`

## Blocked by

Nothing. The law this issue repairs is on main and no open issue claims the file.

**EXPLICITLY OUT, reported rather than fixed:** the sibling walker in
packages/web/src/drawer/readonly.test.ts carries the identical extension filter
and the identical hole — verified below. This issue's Definition of done asks
only that it be checked and said either way, and that is what happens here. A
second walker in a different law's territory is not this issue's to change.

**EXPLICITLY OUT:** the per-directory pin's VALUE is not edited by hand. If the
widened walk changes it, it is re-derived by running the law, per that comment's
own standing instruction.

**Model:** sonnet. **No wave** — this issue is not part of ruling 5's sequencing,
the only sequencing prd-14 declares, and minting a wave for it would be a PRD
amendment rather than a tracker edit. Milestoned to prd-14 because the law it
repairs is prd-14's own lab-tab law.

## Why

packages/web/src/lab/no-live-fleet-law.test.ts enforces prd-14's "the lab tab
renders no live-fleet surface". Its walker sees less than the law claims to
govern:

```
if (!/\.(ts|tsx)$/.test(entry)) continue
```

So a JavaScript module under the lab directory is invisible to both of the
file's load-bearing assertions at once: the per-directory source-file pin does
not count it, so it can be added silently; and the forbidden-identifier sweep
never reads it, so it may import fleet machinery freely.

That is the shape prd-43 ruling 2 names for prose counts, one level out: **a
claim wider than the instrument that checks it.** Two spellings of the governed
set, and the narrower one is the one that runs.

Found by the Codex seat during the verify round on #343 (a6bae212, 2026-09-08)
and re-executed independently. It predates #343 and is unaffected by #349
(f3107e19), which correctly moved only the count.

## The decision, taken 2026-09-10 — widen the walk

The issue as filed offered two honest resolutions and required a check before
choosing: **is there another rule that already keeps JavaScript out of the lab
directory?** If so, narrowing the claim in prose would be enough and cheaper.

The check was run. **There is no such rule — only a partial one, which a
two-file spelling walks straight through.** EXECUTED on a scratch worktree at
main, restored clean afterwards:

| probe | result |
|---|---|
[...]

PR #422, opened 2026-09-10
prd14: a saved comparison is listed and reopened, and the launch confirm re-checks every conjunct

### LINKED ISSUE #214: prd14 w3: the recordings library lists a comparison as its own kind and reopens it

### LINKED ISSUE #379: prd14: the launch confirm re-checks every conjunct its arm gate checks

gabriel-canaan:
Two prd-14 issues and a docs correction, bundled so the queue toll is paid once.

Closes #379
Closes #214

## Declared review gap — read this first

**#214's final fix has not been through independent seats.** Its earlier state was
reviewed twice over (two seats on the initial pass, two more on the repairs), and
both rounds' findings are fixed. The last repair — the one that stops a reopened
`verified` comparison rendering a numeric spread — was verified by the orchestrator
alone, by mutation, and that is a lighter class of evidence than the rest of this PR
carries. It is disclosed rather than recorded away.

That was a deliberate stopping decision, not an oversight. Five review rounds had by
then been spent on one test-only law elsewhere in this wave, and the operator asked
whether the return still justified the cost. It did not, so the line was drawn here.

The mutation, EXECUTED: reverting the null-measure branch reddens three tests by
name — "renders no numeric spread and no verified counts", "never claims a measure on
reopen", "never renders a numeric spread over its 1/0 encoding" — with the control
green. It is load-bearing, not decorative. What it is not is a second pair of eyes.

## The commits, against the review that read each

| commit | review record |
|---|---|
| `docs(prd-14)` | Both seats read the amendment and found no defect in it; one found a stale CHANGELOG forward reference contradicting it, corrected here. That correction is prose and is the only unreviewed content. |
| `#379` | Two seats on the initial pass, two on the repairs. **Sound in full** — each independently probed the `isWatched` unreachability claim and confirmed it. |
| `#214` | Two seats initial, two on the repairs, all findings fixed. **Final repair orchestrator-verified only** — see above. |

## What is NOT in this PR

**#350** — the no-live-fleet law's extension sweep — was cut deliberately. Its work is
real, strictly better than `main` at every step, and sitting on its branch. It is
held back because two findings still stand against it and the shape of the last two
review rounds said to stop: each closed one spelling of the CommonJS grammar and
missed or opened another. Its issue carries the full record, the executed evidence,
and the recommendation — narrow the law's claim rather than extend the regex a sixth
time. Nothing in this PR depends on it.

## What is the sibling case?

- **#379** generalises #352, which closed one of the arm gate's three conjuncts and
  left two. This closes the other two, and `confirmRetarget` — which already
  re-checked both of its own — is now genuinely symmetric with it rather than
  described as such.
- **#214**'s save is the app's seventh mutating call, and it joins the enumeration in
  `replay/mutating-calls-law.test.ts` rather than being an exception to it.
- The measure fix has a sibling *inside itself*, and it is the reason there was a
  third round: the first repair stopped a reopened artifact printing a false unit on
  a real number, and by doing so let a `verified` comparison's `1`/`0` verdict flags
  fall through to the numeric spread renderer — `min 0 · median 1 · max 1`, formally
  indistinguishable from a genuine spread. A false label was traded for a false
  statistic. Both are now closed: with no recorded measure, every run is still shown
  by its verdict and no summary is printed, which is ruling 3's own vocabulary for a
  summary the data does not support.

- **#379** — each conjunct was probed alone. Removing `live` reddens its own test;
  removing the harness-status conjunct reddens its own. Removing `isWatched` reddens
  nothing, and that is stated in the code rather than papered over: the guard is
  structurally unreachable, two independent reviewers confirmed it by probe, and the
  commit says no test covers it and why. A guard kept as defence in depth, with an
  honest note, beats a test that forces an impossible state.
- **#214** — the three-test mutation above, plus each earlier repair was certified
  against its own dedicated test with baseline controls.
- **docs** — none is claimed; prose, no test reads it. Every factual claim was
  executed instead: #23 and #220 closed, PRs #351 and #393 merged, and #214's own
  body carrying the renumber.

## Fence widenings — recorded here, and not on the tracker

#214 touches four paths outside its declared fence, all recorded in its commit body
before the change:

- `packages/web/package.json` — the exports map, without which the two contract tests
  its fence already named cannot resolve
- `packages/web/src/lab/no-live-fleet-law.test.ts` — the per-directory pin, which this
  wave's new files move
- `packages/server/src/api/route-class-law.test.ts` and `README.md` — the egress
  enumeration and the prose derived from it, which the new mutating call moves

The lane attempted to record these on the issue and was stopped: three of the four are
claimed by other open issues with open PRs, so writing a claim there would have
manufactured an overlap against work in flight. They are disclosed here instead, which
is where the runbook wants a widening recorded anyway. Each of the last three is a
DERIVED count, and each was re-derived by running its law rather than edited to pass.

## Assembly, and two defects it introduced

These commits were cherry-picked onto `main` after prd-55 wave 1 landed mid-flight and
moved the same counts. Three files conflicted; both sides were kept and every count
re-derived by running the laws.

That resolution was wrong twice, and the laws caught it both times — recorded because
the second one was invisible on inspection:

1. it took a stale `canvas: 3` over the current `canvas: 5`;
2. it fused two entries of the gated-reads table into a single object literal, so
   `/api/lab/transcript` was silently overwritten by duplicate keys and disappeared
   from the enumeration.

Both are fixed, and the two laws now pass. Neither would have been found by reading
the merged file; only running the thing found them. That is the same standard this PR
holds #214's counts to.

## Gates, EXECUTED on the assembled head

`VITEST_MAX_WORKERS=6 npm test` rc=0 — 489 test files, 8947 passed, 3 expected fail,
3 skipped.

---

## Declared review gap

> #214's final repair (the verified-spread fix) was verified by the orchestrator with a certified mutation, not by independent seats — a deliberate stopping decision after the operator questioned the return on further rounds. Its earlier states were seat-reviewed twice. See the first section of this body.

```
  UNVERIFIED  c9b4e154  feat(comparisons): the recordings library lists a saved comparison and reopens it (#214)
```


ciaran-slow (approved):
## Review — the declared review gap verified independently, one vacuous assertion fixed in `506b7bb3`

Read commit by commit in a fresh clone, Node 22.23.2. The body declares one commit as orchestrator-verified only, so that is where I spent the effort.

### Verified in your favour

**The declared gap's own mutation reproduces exactly, and reddens exactly the three tests you name.** `EXECUTED` — baseline 53 passed across the five touched test files; removing `ArmPanel`'s `measure === null` branch:

```
Tests  3 failed | 50 passed (53)
  × a saved "verified" comparison, reopened, renders no numeric spread and no verified counts…
  × never claims a measure on reopen — the artifact carries none, so none is guessed…
  × a saved "verified" comparison, reopened, never renders a numeric spread over its 1/0 encoding
```

Three by name, as claimed. Its siblings hold too: `runWords`' `measure === null` early return reddens 2, and the "not recorded" basis line reddens 2. **The lighter class of evidence you disclosed is, on re-derivation, load-bearing.**

**#379's guards are sound, and the disclosure about them is honest.** `EXECUTED`, baseline 61:

| mutation | result |
|---|---|
| drop harness-status from the button's `disabled` | **1 failed / 60** — matches your figure exactly |
| drop harness-status from `confirmLaunch` | **61/61 green** — as you disclosed |
| drop `isWatched` from **both** guards | **61/61 green** — as you disclosed |
| drop `!live` from the button's `disabled` | **1 failed / 60** |

I also checked the structural-unreachability claim rather than taking it: `ConductorStep` branches at `wizard.tsx:897` on `!isWatched ?`, and the confirm button is at `1035`, inside the true-half. There is no render with the button present and `isWatched` false. The claim holds, and declining to write a test that would have to tear down the button it tests is the right call.

I checked one thing the body does not mention: `confirmLaunch`'s `facts` is computed fresh at `wizard.tsx:389`, and `ConductorStep`'s at `792`. They are **byte-identical**, `?? HARNESSES[0]` fallback included — so the re-check cannot drift from the gate it mirrors. Good.

**The assembly's two self-reported defects are genuinely fixed.** `canvas: 5` (not the stale `3`), `/api/lab/transcript` still enumerated, and the four count laws green together — `route-class-law`, `no-live-fleet-law`, `mutating-calls-law`, `contract-coverage-law`: **130 tests**. `EXECUTED`

**`ClaimBanner` was my main suspicion and it is clean.** The surface claims *"no number this surface cannot name the unit of is ever printed"* while still rendering the claim banner unconditionally on the null path — but `claimDetail` returns only configuration prose (`arms differ in {dimension} only`), never a measure-derived figure. The claim holds as written.

**All ten CI legs executed and passed** on the pre-fix head. Read from `gh pr checks`.

### One finding, fixed rather than handed back — `506b7bb3`

The test named **"never claims a measure it does not have — no data-measure attribute, no measure switch"** renders `<ComparisonSurface measure={null} />` with **no `onMeasureChange`**. So `queryByTestId('measure-switch')` is null because no handler was given — nothing to do with `measure` being null.

`EXECUTED`: deleting `|| measure === null` from the switch's render condition left all 12 tests **green**. With `onMeasureChange` passed, the same deletion reddens exactly this test. One prop; the assertion now fails for the reason its name gives.

The `data-measure` half was **already sound** — I checked separately, since React omits null attributes and a careless mutation proves nothing there. A real `measure ?? 'cost'` fallback reddens it. Only the switch half was vacuous.

That guard is defence in depth today (the reopen path passes no handler at all), which is exactly why the assertion was the only thing holding it — and it was holding nothing.

The body records that removing `facts?.status !== 'implemented'` from `confirmLaunch` leaves the suite green, because `fireEvent.click` never reaches a disabled handler. That is honest and I reproduced it, but it means **both** of `confirmLaunch`'s new conjuncts are carried entirely by the button. If you want the inner guard pinned, calling the handler directly rather than through a click is the shape that would do it. Not a blocker — the refuse-before-the-wire posture is worth keeping either way. `REASONED`

### What I did NOT run

The full suite locally, and the load probe. My runs were the five touched web test files, the four count laws, and `typecheck` + `lint` across the tree (both rc=0). CI on `506b7bb3` is still finishing as I write. I did not exercise #350, which you correctly cut.

### Verdict

**Approve.** The commit carrying the declared gap is the one I probed hardest and it survived its own strongest mutation, by name. The one gap I found was in a test's assertion, not in behaviour, and it is fixed. The merge is the operator's.

---
*One careful reading by Claude Opus 5 via `rhizo-review` — a single reviewer, no second seat.*

PR #423, opened 2026-09-10
prd17: every live document's status claim matches the tree, and the PRD stops overclaiming

gabriel-canaan:
Closes #416

Three documents, one commit. prd-17's status claims now match the tree, and — more usefully — the PRD stops claiming something it could not establish.

**This is a docs change with no mutation to certify.** Every line of it is a claim about the tree, so the review question was "is each sentence true", and the proof is re-derivation rather than a red test. Treat every figure below as re-derivable, and the enumeration's commands are on the issue.

## What was wrong

`docs/roadmap.md` called prd17 "partially landed" and, eleven lines below, listed rulings 2, 4 and 5 as "ruled but not yet landed" — all three had landed. `docs/architecture.md` contradicted the tree in four clauses plus a heading. The PRD's own header, family table and six body passages described the pre-wave-6 world.

## The three documents

- **`docs/architecture.md`** — four false clauses and a heading. The sharpest: *"`eras.test.ts` lists all eight in its corpus gap list"* is wrong on **both** halves — nine families, and the list holds **four**. The defined/emitted split is now a per-family table rather than one clause attached to all nine, which is the shape that kept going wrong.
- **`docs/roadmap.md`** — rewritten as a **whole bullet**. A previous attempt (in wave 8, reverted) changed only the status line and left the list beneath it, producing a bullet that contradicted itself within eleven lines. Whole or nothing.
- **`docs/prds/prd-17-complete-record.md`** — header, table, six body passages, and the Outcome.

## What is NOT claimed, and this is the point

The Outcome asserts **the landings only** — shas present, issues closed, checkable in a minute.

An earlier version asserted that every live document describing this programme agreed with the tree. That is a claim over a set nobody bounded, and this issue's own history is the evidence it cannot be established by inspection: wave 8 made it and withdrew it, three review rounds each found one more contradicting document, and then this issue's method was **falsified twice** — once on the mechanism it named, once on a document it missed.

So the document lists what was checked and does not claim the list is complete. That is narrower, and it is the only version I can defend.

## What is the sibling case?

**One fix, several places that say it — four rounds running, and twice inside my own repairs.**

- The beacon-doorway sentence was corrected in `architecture.md` while the identical claim stood in the PRD's header, four lines above new text saying the opposite.
[...]

PR #424, opened 2026-09-10
prd14: the no-live-fleet law claims only what it checks, and says what it does not

### LINKED ISSUE #350: prd14: the no-live-fleet law sweeps every source file it claims to govern

### LINKED ISSUE #411: prd14: the no-live-fleet law sees a forbidden symbol renamed on its way in
## Fence (may touch ONLY)

**Co-fenced with #350 on purpose. These two are ONE lane, never a parallel wave.**
scripts/fence-lint.sh will hard-fail the pair, correctly — an overlap between two
issues dispatched side by side is a scheduled rebase conflict. It is not one here:
both issues are the same defect on the same file and are built sequentially in one
worktree, as two commits. If you are about to dispatch these to two agents, stop.

## Blocked by

Nothing. #350's branch carries the earlier work and this issue is built on top of it,
in the same lane.

**EXPLICITLY OUT:** the shared extractor that this law and two other laws read for
import specifiers. It is ESM-only, it is not in this fence, and the third consumer of
it — the concierge invocation law — has the same gap. Reported, not fixed here.

**EXPLICITLY OUT:** the sibling walkers in the drawer and recordings laws, and the
four other laws under the lab directory. Named in #350's commit body, unchanged.

**Model:** sonnet. **No wave** — not part of ruling 5's sequencing.

## Why

The law sweeps the lab tab's own source for forbidden identifiers as literal text. An
identifier renamed on its way in is invisible to it, and the rename need not happen
anywhere the law can see.

Found by both independent review seats on 2026-09-10 and re-executed. EXECUTED: a
module outside the governed directory re-exports the banned symbol under another name,
a lab file imports the alias, no literal banned token appears anywhere under the lab
directory — and with the count pin re-derived as a maintainer normally does, the law is
fully green.

The forbidden-import prefixes cover the fleet and panels directories only, so any other
directory is a clean channel through.

## The decision, taken 2026-09-10 — narrow the claim

Ruled by the operator. The two routes this issue offered were following the import
closure, or narrowing the claim and naming what is uncovered. **Narrowing is chosen**,
on the same proportionality ground that paused #350: a resolver that crosses package
boundaries is real work, and the gap is latent — nothing in the tree launders a fleet
symbol, and nothing ever has.

**The PRD is told, not only the code.** Narrowing is a real weakening of what prd-12
ruling 1 and prd-14's lab-tab law are read as promising, so the amendment lands in
`docs/prds/prd-14-experiment-console.md` as part of this work rather than being left
implicit in a test file. That file is out of this fence and is the orchestrator's
commit, not this lane's.

## Definition of done

- The law's doc comment states what it actually checks: direct-text coverage over one
  directory, forbidden identifiers by literal name, forbidden import paths by prefix.
- It states, in words, what that does NOT cover — a symbol renamed outside the
  directory — with this issue's re-export case named as the worked example.
- The claim and the instrument agree. No sentence survives that a reader could take as
  "the lab tab cannot reach live-fleet machinery"; the honest guarantee is that nothing
  in the lab tab NAMES it.
- **A mutation proves the boundary is where the words say.** Plant the re-export case
  and show the law green, in the test file, as a recorded fact rather than a claim —
  or if that is too strange to keep, certify it and record the result in the commit.
- **The mutation is certified with the count pin already re-derived.** With the pin
  stale the count fails first and proves nothing about the sweep. That is how this gap
  reads as caught when it is not, and it is the same trap #350 recorded twice.

gabriel-canaan:
Two prd-14 issues on one law, plus the PRD amendment they oblige.

Closes #350
Closes #411

## Read the UNCOVERED rows as the deliverable, not as a defect list

This law's axis table now carries six rows marked UNCOVERED. That is what this PR
is FOR. The law is a text sweep over one directory; it cannot parse CommonJS
reliably, and five review rounds of trying proved it — each closed one spelling
of the same grammar and revealed the next. The operator ruled: stop extending the
instrument, narrow the CLAIM until it is true of what the code checks.

So the guarantee is now stated as what it always was — **nothing in the lab tab
NAMES fleet/panels machinery, by the forms marked CAUGHT** — and every way past
it that anyone found is written down beside it with a mechanism and a verified
probe: a concatenated `require` argument that splits the forbidden prefix,
`createRequire(x)(...)`, `require` behind a renamed binding, `require.resolve`,
escaped ECMAScript spellings, and a symbol renamed outside the directory.

None is a live breach. No file in the tree uses any of them, and none ever has.

## Declared review gap

Two independent seats read this branch and their findings are all fixed. Since
they read it, two things changed and neither has independent eyes:

- **the squash** — six rounds of #350 became one commit. History only: the law
  file's diff against `main` has patch-id `132b75af…` both before and after, so
  what the seats verified is byte-identical to what is here;
- **the PRD sentence fix** — prose, in the commit that was already mine.

## The commits

| commit | |
|---|---|
| `#350` | the walk inverted to a deny-list, the import-path claim narrowed, one regex repair, four titles retracted |
| `#411` | the identifier-sweep claim narrowed |
| `docs(prd-14)` | the constraint corrected where a reader outside the test file will see it |

The two #350 commits both seats told me to squash are squashed. Their reasoning
was one I had missed: rounds 1-4 never landed on `main`, so the broken pattern
was *introduced* by the first and repaired by the second — same branch, same
issue, not two units of review.

## What is the sibling case?

Two, both left deliberately with reasons recorded rather than quietly not noticed.

The shared specifier extractor is ESM-only and has three consumers — this law,
the `lab/launch` invocation law and the concierge invocation law. Its CommonJS
gap is theirs too, and it is outside this issue's one-file fence.

The identical `/\.(ts|tsx)$/` filter survives in other walkers. The commit body
carries the command that enumerates them rather than a count, and says why:
review reported four under `lab/`, the orchestrator repeated four, and running
the command returned five — `lab/canvas/laws.test.ts` was missed by every
reader. A tally quoted from a review is not a measurement.

The question that mattered here, because the answer used to be "all of them".
The only code change on the branch — the `TEST_FILE` repair — had no committed
witness: reverting it left the whole suite green at 15/15. Both seats found that
independently. There is now a direct test on both patterns, certified by
reverting `TEST_FILE` toward the broken shape and watching that test go red on
its own.

Every UNCOVERED row was verified green with the count pin ALREADY re-derived, and
every CAUGHT row red, with controls. The ordering is not incidental: with a stale
pin the count assertion fails first and proves nothing about the sweep. That trap
misled three readers on this file, the orchestrator among them.

## The regression this PR repairs is its own

An earlier round widened `TEST_FILE` to exclude `.test.d.ts` and did not restrict
the `d.` infix to declaration extensions — so `*.test.d.cjs`, `*.test.d.mjs` and
`*.test.d.js`, ordinary executable JavaScript, were excluded from both the count
pin and the sweep, with the pin never needing to move. A total, silent miss, and
worse than the hole that round closed. Reverted, with the replacement measured
over a 22-extension corpus: it excludes a strict SUBSET of the old pattern, five
names moved from excluded to swept, zero newly excluded.

## The PRD amendment, and what it does not weaken

The constraint used to read *"a law test asserts the lab surface renders no live
fleet state"*. That is the stronger of two readings and was never what the law
checked. Corrected in place.

Its load-bearing sentence — that prd-12 ruling 1's read-only guarantee over the
watched repo is untouched — was verified by a seat rather than asserted: that
ruling is enforced by `packages/server/src/lab/namespace-law.test.ts`, a
different law in a different package, green on this head, and this branch touches
no file under `packages/server`.

The first draft of that amendment appended a caveat and left the over-claim
standing above it. A seat reported the contradiction it had just created. The
commit body records that, because "narrow the claim" was the instruction and the
first attempt did the other thing.

## Fence

One file for both issues — `packages/web/src/lab/no-live-fleet-law.test.ts` —
plus `docs/prds/prd-14-experiment-console.md` in the orchestrator's own commit.
Co-fencing two issues is deliberate and #411's body says so: `fence-lint` will
hard-fail the pair and is right to for a parallel wave, which this never was.
They were built sequentially in one worktree. No widenings.

## Gates, EXECUTED on this head

`VITEST_MAX_WORKERS=6 npm test` rc=0 — 489 test files, 8961 passed, 3 expected
fail, 3 skipped.

---

## Declared review gap

> the PRD commit's sentence fix is unreviewed — a seat reported that the first draft caveated the over-claim instead of replacing it, and the correction was made afterwards. Prose only. The other two commits' content was read by both seats; the #350 squash is history-only and patch-id identical to what they verified.

```
  UNVERIFIED  33b5ceb4  docs(prd-14): the lab-tab law guarantees nothing NAMES fleet state, not that it cannot reach it
```


ciaran-slow (approved):
**Approve.** Reviewed commit-by-commit in a clean clone at `33b5ceb4`, against `origin/main` `c4c38d8c`. This closes the declared review gap: the `UNVERIFIED 33b5ceb4` line is the PRD prose commit, and an independent pass has now read it.

## The mutation certification holds, and I re-ran it rather than trusting it

**EXECUTED.** Baseline on this head: 16/16 green. Reverting `TEST_FILE` to round 4's shape (`/\.test\.(?:d\.)?[^./]+$/`):

```
× TEST_DECLARATION_FILE stays scoped to real TS declaration extensions…
AssertionError: foo.test.d.js not matched by TEST_FILE: expected true to be false
Tests  1 failed | 15 passed (16)
```

Exactly one test reddens, and it is the new one. That simultaneously certifies the new test **and** confirms the claim that the repair had no witness before it — the other 15 stayed green under the mutation.

## The core defect is real, and the miss was total. EXECUTED, both directions

I built the probe rather than taking the description: `packages/web/src/lab/zz-probe.cjs`, containing `const fleet = require('../fleet/manifest.js')`.

- **Against the base law** (`122c0dc1`), pin untouched: **12/12 passed, fully green.** A forbidden fleet reach sitting in `lab/`, invisible to both load-bearing assertions at once.
- **Against this head**, pin bumped 6→7 first: **red, and named** — `zz-probe.cjs imports fleet/ or panels/ machinery: expected [ '../fleet/manifest.js' ] to deeply equal []`.

## The axis table is accurate in both directions — which is the part I most expected to find wrong

The dangerous error would be a row claiming CAUGHT that isn't. I probed the two novel ones:

| row | probe | result |
|---|---|---|
| `require()` with a second argument | `require('../fleet/manifest.js', 'ignored')` | **red**, extracts `'../fleet/manifest.js'` ✓ |
| split *after* the prefix | `require('../fleet/' + n)` | **red**, extracts `'../fleet/'` ✓ — the subtle distinction the table draws is exact |

Then all five UNCOVERED rows at once — five real `.cjs` modules under `lab/`, each reaching `../fleet/manifest.js` via a concatenated split, `createRequire(x)(…)`, a renamed binding, `require.resolve`, and an `\x2f`-escaped path. Pin bumped 6→11: **16/16 green.** Those gaps are real, and they are documented rather than hidden. That is the deliverable, and it is the right call — an allow-list found short in three consecutive rounds is evidence about the instrument, not about the next regex.

The `TEST_FILE` subset claim also holds: the new pair matches a strict subset of round 4's pattern (`*.test.d.{js,mjs,cjs}` move from excluded to swept; nothing newly excluded). REASONED from the two regexes, consistent with the executed test above.

## The PRD amendment's load-bearing sentence checks out

`prd12` ruling 1's read-only guarantee is enforced by `packages/server/src/lab/namespace-law.test.ts` — the file exists, it is green on this head, and **this branch touches no file under `packages/server`** (verified: the true diff from merge-base `122c0dc1` is exactly two files). Correcting the over-claim in place rather than caveating it is the right shape, and the commit is honest that the first draft did the other thing.

## Offered, not asserted — one residual the table doesn't quite say

**EXECUTED.** The UNCOVERED rows are all verified with *added* files, where the count pin does move and does force a look. I tried the other shape: appending `createRequire('x')('../fleet/manifest.js')` to an **existing** lab file (`adapters.ts`), no new file, pin untouched — **16/16 green.**

So the pin is a backstop for *added* files only, not for in-place edits. Nothing in the table is wrong about this — UNCOVERED is UNCOVERED — but several rows lean on "pin already correct" as reassurance, and for the edit case there is no pin movement to rely on at all. A sentence to that effect would make the honest guarantee harder to over-read later. **Your call entirely; not a condition of the approval.**

## Gates, EXECUTED on `33b5ceb4`

`npm run lint` rc=0 (1041 files) · `VITEST_MAX_WORKERS=6 npm test` rc=0 — 489 files, 8957 passed, 3 expected fail.

**Fence:** exactly the two files declared, no widenings. (The diff against current `origin/main` shows five files; three are stale-base artifacts — the branch's base is 16 commits behind. The true merge-base diff is two.)

## Said plainly, what I did not run

`npm run typecheck` did not complete in my clone: `packages/team` fails on a missing `postgres` module. That is my clone's stale `node_modules` — the dependency is declared at `packages/team/package.json:15`, and this branch touches no file in that package. **`@rhizomorph/web`, the package this PR does touch, typechecks rc=0.** CI's ten legs are green and I confirmed each of the three expected jobs actually executed.

## Landing together with #423 — EXECUTED

CI tests each branch alone, so it cannot see a collision between two PRs queued at once. I merged **both** onto current `origin/main` and ran the suite against the merged tree:

- `origin/main` + #423 → clean · `origin/main` + #424 → clean · #423 + #424 → clean
- merged tree, full suite: **rc=0, 489 files, 8957 passed, 3 expected fail**

No textual conflict and no semantic collision between the two. They can land in either order.

**LANDABLE.** Not merging — merges here are the operator's manual act on GitHub.

PR #425, opened 2026-09-10
prd51 w4: the hand and the fold stop wedging in silence, and the doctor route reports the fifth hand

### LINKED ISSUE #387: prd51 w4: the doctor route reports the fifth hand, not only the CLI
## Fence (may touch ONLY)

## Blocked by
#386 (prd-51 wave 3) merging. The shipper and its CLI `doctor` check land there; this issue is the
half that did not.

## Why this exists
Found building #372 and confirmed by its verify pass, both of which flagged it rather than swapping
in a weaker test.

ADR-0034 clause 2 requires `doctor` to report the ingest credential's **presence, never its value**,
and prd-51 ruling 12 says a person must be able to see that the hand is on. #372 delivered that on
the CLI — `cli/doctor.ts` carries the `shipper` check — but **not** on the HTTP surface.
`api/doctor.ts` composes its own list of checks, so the route returns a doctor report with no
`shipper` id at all.

The two files above were outside #372's declared boundary, so the lane recorded the gap instead of
widening for it. That was the right call and it is why this issue exists rather than a silent
coverage drop.

The consequence is narrow but real: anything reading the instrument's doctor over HTTP — including
a future `/connect` row — cannot see whether the fifth hand is enabled, while the CLI can.

## Definition of done
- `GET /api/doctor` includes a `shipper` check whose verdict matches what `cli/doctor.ts` reports
  for the same data root: enabled and reachable, enabled and failing with the exact remedy, or not
  enabled. "Connected" never means preconditions passed (ruling 12).
- **The check reports presence and never the value.** A test asserts no part of a planted key
  appears anywhere in the serialised response body, not merely that the field is absent — the CLI
  side has such a law already and this is its sibling.
- The route reaches the hand's facts the way the CLI does, **through `cli/connect-team.ts`**, and
  does not import the shipper directory. `hand-law.test.ts` holds `DECLARED_IMPORTERS` to exactly
  one file, and a second entry would widen ADR-0034's clause-3 seam — which is the one thing that
  law exists to keep narrow. If that proves impossible, STOP: it is a ruling, not a workaround.
- `api/doctor.test.ts`'s exact `toEqual([...])` id list gains `shipper` in the same commit. That pin
  is an enumeration one directory from the feature — the same shape that has now forced three fence
  widenings across prd-51 — so re-derive it rather than appending blindly, and check whether the
  replay-mode exclusions (`target-path`, `web-build`, `port`) should include it too.
- Falsifier stated and its verdict recorded: if the route cannot reach the hand without a second
  declared importer, then ADR-0034 clause 2's "doctor reports it" and clause 3's one-importer seam
  are in tension, and that is an amendment to the ADR rather than a widened law.

### LINKED ISSUE #398: prd51 w4: a torn journal tail is repaired at open, not escalated by the next append
## Fence (may touch ONLY)
- `packages/team/src/journal/journal.ts`
- `packages/team/src/journal/read.ts`
- `packages/team/src/journal/journal.test.ts`
- `docs/adr/0046-the-ingest-journal-is-a-crc-framed-append-log.md`

## Blocked by
#386 (prd-51 wave 3) merging. The journal lands there; this is a defect in what it lands.

## Why this exists
Finding 1 of the review on #386, and reproduced independently before filing rather than inherited.

`readJournal` gets torn-versus-corrupt exactly right — ruling 4's distinction is implemented, and
`read.ts`'s own table is correct. The gap is that the **writer's open path never acts on the verdict
it is handed**. `openJournal` checks `existing.ok` and never `existing.verdict`, then opens `'a'`.
The torn bytes stay in place and the next frame is appended after them.

One crash mid-write plus one accepted batch is enough to turn a legal state into a permanent one.

**EXECUTED** — two clean records (292 bytes), truncate 3 bytes, then one more accepted batch:

```
after tear:            ok=true  verdict=torn  lastSeq=1
reopen after tear:     ok=true  lastSeq=1
one more accepted batch, seq=2, 435 bytes
after that append:     ok=false
  journal … is corrupt at byte offset 146, and bytes follow it, so this is not a
  torn tail: crc32 is 89b78be5 but the header claims ab252241
reopen (what every later boot does): ok=false
```

Two consequences, and the second is the serious one:

- **The records ahead of the tear become unreadable through the normal path.** They are still on
  disk and still valid, but every `readJournal` now aborts at offset 146.
- **The server cannot boot again.** `startTeamServer` in `packages/team/src/api/main.ts:75` returns
  `{ok:false}` on that result, so every subsequent boot fails on a state ruling 4 explicitly calls
  legal.

Note the shape of the escalation: the append after a torn tail is handed `seq = lastSeq + 1 = 2` —
the **same seq the torn record already claims** — so the new frame is written at the torn record's
own offset boundary and the chain rule sees a crc that cannot be reconciled. This is not a
near-miss; the torn tail guarantees the collision.

## The test that should have caught it, and could not
`journal.test.ts:221`, *"a torn tail is NOT a refusal — it is the state a crash legitimately
leaves"*. It asserts `openJournal` ok and `lastSeq() === 1`, then closes. Its own comment claims
*"the next append continues from the last COMPLETE record, so the torn bytes are overwritten by
nothing and the chain stays gapless"* — **the test never performs that append**, and the claim it
makes turns out to be false. A test whose comment states the property it does not exercise is worse
than no test, because it reads as coverage.

## Definition of done
- A torn tail is **repaired at open**, not merely tolerated: `openJournal` truncates to the end of
  the last complete record before opening for append. `readJournal` already knows that point — its
  `offset` at the moment it returns `torn()` is exactly it — so the shape is to carry that on the
  torn result rather than to re-derive it.
- The existing case at `journal.test.ts:221` **performs the append its comment promises**, and
  asserts the file reads back `clean` afterwards with a gapless chain. Extending that case is
  preferred to adding a sibling beside it: the false comment is the defect's other half.
- A case for the tail the repair must not eat: a torn tail on a journal with **zero** complete
  records truncates to 0 and still opens, and one on a journal of many truncates to the last
  complete record and no earlier.
- **A mutation is recorded**: removing the truncation reddens the extended case. If it does not, the
  case is still not exercising the append.
- ADR-0046 states what open does to a torn tail — repair is part of the durability contract, not an
  implementation detail, and the ADR currently describes only the reader's verdicts.
- The `seq` reuse above is stated wherever the repair is documented, since it is the reason a torn
  tail cannot simply be appended past.

## What this is NOT
Journal rotation, which ruling 4 leaves out of v1 and ADR-0046 records as known debt. Truncating a
torn tail at open is not rotation and must not grow into it.

## Falsifier
If truncating at open turns out to conflict with ruling 4's *"a torn tail … reads as never-acked"*
— because some reader is entitled to see the torn bytes — then the repair belongs at the reader and
this is a ruling on ADR-0046 rather than a patch. Record the verdict either way.

### LINKED ISSUE #399: prd51 w4: a ledger line larger than the read window does not wedge the hand in silence

ciaran-slow:
Closes #387
Closes #398
Closes #399

Three defects found by the review of #386 (prd-51 wave 3) and by building #372, each one a case
where the wave-3 code was silently wrong rather than loudly wrong. Two implement rulings written
for them in the 2026-09-10 *two wedges, two rulings* amendment; the third completes a clause
ADR-0034 has required since it was accepted.

Base is `origin/main` at `f56ca49a`. Three commits, one per issue, cherry-picked from their lane
branches with **no conflict** — which is the wave's own fence audit: nothing here touched a file
anything else here touched.

---

`AGENTS.md` requires that every commit in the range has been read by an independent review pass and
that the record is somewhere a later reader can ask. Each verify pass ran on **Sonnet 5** against a
builder that ran on **Opus 5** (the `Co-Authored-By` trailers carry it), in a separate worktree, and
ran all four gates plus its own mutations rather than reading the build report's.

| commit | issue | plan | build report | verify (Sonnet 5) |
|---|---|---|---|---|
| `40018b22` | #387 | [plan](https://github.com/launchpad-26/rhizomorph/issues/387#issuecomment-5626049844) | [report](https://github.com/launchpad-26/rhizomorph/issues/387#issuecomment-5626229668) | [**ready to merge**](https://github.com/launchpad-26/rhizomorph/issues/387#issuecomment-5626460827) |
| `a9998e8b` | #398 | [plan](https://github.com/launchpad-26/rhizomorph/issues/398#issuecomment-5626014041) | [report](https://github.com/launchpad-26/rhizomorph/issues/398#issuecomment-5626211069) | [**ready to merge**](https://github.com/launchpad-26/rhizomorph/issues/398#issuecomment-5626507567) |
| `46efda7a` | #399 | [plan](https://github.com/launchpad-26/rhizomorph/issues/399#issuecomment-5625986779) | [report](https://github.com/launchpad-26/rhizomorph/issues/399#issuecomment-5626296127) | [**ready to merge**](https://github.com/launchpad-26/rhizomorph/issues/399#issuecomment-5626463026) |

**Two limits on that record, stated because a reviewer should weigh them rather than discover them.**

1. **A unanimous verdict is the tell, and this one is 3/3 approve.** `AGENTS.md` names that shape for
   refute-first seats manufacturing findings; the inverse deserves the same suspicion. What is
   offered against it is that no verdict rests on reading alone — each rests on a mutation the
   verifier executed itself, and one returned a finding nobody pointed it at (§ *deferred*, below).
2. **The three verify passes were dispatched by the same session that wrote the three plans**, and
   that session supplied each verifier with four targeted questions. Different model, different
   context, public record — but not an independent *choice of where to look*. The human review on
   this PR is the first pass that is independent in that sense.

---

## `40018b22` — #387: the doctor route reports the fifth hand, not only the CLI

ADR-0034 clause 2 requires `doctor` to report the ingest credential's presence, never its value, and
prd-51 ruling 12 says a person must be able to see that the hand is on. #372 delivered that on the
CLI and not on the HTTP surface, so anything reading the instrument over HTTP — a future `/connect`
row included — could not see whether the fifth hand was enabled while the terminal could.

`runServerDoctor` now calls the CLI's own exported `checkShipper`, in the CLI's own position
(after `lane-manifest`, before `cli-version-drift`).

**What is the sibling case?** `session-boundary` and `lane-manifest`, which are already replaced with
an honest *not applicable* during a replay because a replay's `repoPath` is a synthetic
`record:<slug>` string. `shipperDoctorFacts` opens with `sessionDirFor(repoPath, …)`, so an
unguarded `shipper` check would have reported a confident **"shipper: off"** about a session
directory belonging to no record anyone is replaying — the identical defect, one line down. It is
guarded the same way. The issue did not ask for this; it was found by reading what the fact-fetcher
actually resolves.

**What mutation would this test survive?** Four were run, each reverted:

- delete the `shipper` entry from `baseChecks` → **8 red**
- drop the replay arm → **exactly 1 red**, and the failure message shows a real leaked *on* state,
  so the case plants a genuine enabled hand rather than passing over a no-op
- `checkShipper(repoPath, undefined)` → **3 red** (the plan predicted 1; the pin is stronger than
  claimed, and nothing was changed to chase the prediction)
- make `checkShipper` leak the fixture key → **7 red**, including both never-value assertions

The never-value law searches the **serialised** body for the key and for fragments of it, and also
asserts the check ran at all — without that, every negative passes on a build where it never ran.

**Fence:** `api/doctor.ts`, `api/doctor.test.ts`. No widening. `shipper/hand-law.test.ts` is
untouched and `DECLARED_IMPORTERS` is still a one-entry set.

**The issue's falsifier resolved, and it resolved the easy way.** It asked what to do if the route
could not reach the hand without a second declared importer, since that would put ADR-0034 clause 2
against clause 3. It does not arise: `checkShipper` is exported from `cli/doctor.ts`, which
`api/doctor.ts` already imported, so this adds **no import edge at all**.

---

## `a9998e8b` — #398: a torn journal tail is repaired at open, not escalated by the next append

`readJournal` got torn-versus-corrupt exactly right; `openJournal` checked `existing.ok` and never
`existing.verdict`, then opened `'a'`. One crash mid-write plus one accepted batch turned a state
ruling 4 calls **legal** into a permanent refusal: the torn record was assigned `lastSeq + 1` before
it was cut off, so the next append is assigned the same `seq` at the same offset boundary, the
reader sees a CRC it cannot reconcile with bytes following it, and `startTeamServer` returns
`{ok:false}` on **every subsequent boot**.

`readJournal` now carries `completeBytes` out on its ok arm — the reader is standing on the
truncation point when it gives its verdict — and `openJournal` truncates and `fsync`s to it before
the append descriptor opens.

**What is the sibling case?** A **corrupt** journal. The whole defect was keying on `ok` instead of
the verdict, so a repair that made the same mistake in the other direction would truncate a journal
whose interior is corrupt — destroying evidence rather than escalating it. The repair keys on
`verdict === 'torn'`, which TypeScript's discriminated union makes structurally unreachable for the
`ok:false` arm (`completeBytes` is not on it), and a test asserts a corrupt interior is still
refused at open with the file byte-identical afterwards.

**What mutation would this test survive?** The extended case at `journal.test.ts:221` is the point of
this issue — it previously carried a comment claiming *"the next append continues from the last
COMPLETE record"* and **never performed that append**. It performs it now.

- remove the truncation → **4 red**, including that case (file 727 bytes vs expected 584 — it really
  is appending)
- key the repair on `ok`, ahead of the refusal guard → **2 red**, including the corrupt-interior
  sibling

**A correction to the plan, kept because it matters.** The plan prescribed a second mutation —
swapping `verdict === 'torn'` for `!existing.ok` — which is **unreachable**: the `if (!existing.ok)
return` guard one line above already sends every corrupt journal home, so the swap degenerates into
a no-op. The builder said so rather than reporting a mutation it had not really run; the verifier
then reproduced the degeneracy itself, **built a reachable variant of the same defect class** (repair
keyed on `!existing.ok`, running ahead of the refusal guard), and confirmed the corrupt-interior case
catches that. So the case is non-vacuous against the defect, not merely against the plan's phrasing.
The clause it guards is the **ordering** of guard-then-repair.

**Fence:** `journal/journal.ts`, `journal/read.ts`, `journal/journal.test.ts`,
`docs/adr/0046-…md`. No widening. ADR-0046 is `proposed`, so editing it in place is within
`docs/adr/README.md`'s append-only rule, which binds once *accepted*; its status line and number are
untouched.

**Not rotation.** Ruling 4 leaves rotation out of v1 and ADR-0046 records it as known debt.
Truncating a tail nothing ever acked discards no acknowledged record.

---

## `46efda7a` — #399: a ledger line larger than the read window advances or names a failure

`shipActor` reads at most `MAX_READ_BYTES` per actor per tick. A single line longer than that window
contains no newline, so `splitLines` yielded nothing and the pass returned `unchanged(null, null)` —
no failure, no skip, no advance, `ok: true` with an empty `failures` array, **forever**. Worse than
the case ruling A already handles, because ruling A at least tells the operator something is wrong.

Implements **ruling 15** as written, both arms.

**What is the sibling case?** A window with no newline has *two* causes and conflating them is the
defect. The discriminator is `size - before.offset`, never the window's contents: at or below
`MAX_READ_BYTES` the whole remainder was read and its tail is simply unterminated — a partial write,
and waiting is correct. Above it, the line is already a megabyte with no terminator inside it and no
amount of waiting shortens it. The second sibling is the **unterminated** oversized line — a writer
that died mid-line hits EOF on every forward scan forever, so that arm, and only that arm, names a
`ShipActorFailure` instead of advancing.

**What mutation would this test survive?** Four, each reverted:

- oversized arm → `unchanged(null, null)` → **6 red / 17 green**
- `remaining <=` → `remaining <` → **exactly 1 red**, the fits-the-window case. The test pair is one
  byte apart, which is what pins `>` against `>=`; the verifier re-ran this one itself.
- `findLineEnd`: `at + index + 1` → `at + index` → **3 red**, every case reading behind the fat line
- cursor schema `kind` back to a closed union → **exactly 1 red**, the leniency case only

**Fence:** `shipper/ship.ts`, `shipper/ship.test.ts`, `shipper/cursor.ts`. No widening.

**Plan deviations, all three minor.** Case 5 written as two `it`s (the plan allowed either); the
`remaining > MAX_READ_BYTES` comparison appears negated as an early-return guard (same comparison,
and the mutation was run against the form in the tree); and `findLineEnd` is wrapped in
`try/catch → 'unreadable'`, which the plan did not cover and which is what keeps the
no-leaked-descriptor law satisfied on a throw mid-scan. The verifier judged the third a correct
judgement call consistent with the file's existing convention, not a swallow.

**Forward compatibility, per ruling 15's amendment.** `ActorSkip.kind` gains `'oversized'`, and the
read side is made lenient in the same commit so a cursor written by a newer build stays readable by
an older one. A `CURSOR_VERSION` bump is rejected there and is not taken. The negative half still
holds: a genuinely broken entry (`kind: ''`, `kind: 7`) still resets that actor alone through
`readCursor`'s per-actor reject path, and a test says so.

**The issue's falsifier resolved as ruling 15 predicted:** `findLineEnd` allocates one 64 KiB buffer
regardless of line length, so *advance past the whole line* and *no unbounded allocation* are not in
tension and ruling 15 needs no amendment.

---

## Gates — run on the assembly, not only on the lanes

This is the first run in which the three commits coexist. On `prd51-w4-wedges-and-the-doctor`:

- `npm run typecheck` — green, all workspaces
- `npm run lint` — green, 1041 files
- `VITEST_MAX_WORKERS=6 npm test` — **489 files, 8975 passed**, 3 expected fail, 7 skipped
- `npm run build` — green
- boot smoke — `rhizomorph running at …`, `GET /` → 200, `GET /api/meta` → 401 (capability-gated,
  ADR-0012, unchanged by this wave)

`scripts/gate.sh` was **not** run: running it is the landing, and the landing is the operator's step.

---

## Deferred, non-blocking — named here rather than left in a comment thread

Both were found during this wave and belong to prd-51 wave 5's grooming, which has not happened yet;
neither has an issue, deliberately, because filing one requires a milestone and wave 5 is not groomed.

- **#399 — `findLineEnd` holds the handle across many reads** without re-checking the file has not
  changed underneath it, unlike the first read, which `bytesRead !== wanted` guards. A rotate, prune
  or archive landing mid-scan could produce a spurious `'oversized-unterminated'` report for one
  tick. Self-healing (a failure never writes the cursor) and strictly better than the permanent
  silent wedge being fixed. **Labelled REASONED, not executed** — proving it needs a concurrency
  harness this test file does not have.
- **#387 — `registerDoctorRoute` passes no `dataRoot`**, so through `buildApp` this check reads the
  process-default data root. That is correct on a live server and is pre-existing behaviour shared
  with `checkSessionBoundary` one line above. The `buildApp` case therefore asserts only presence and
  the never-value law, never a machine-dependent status. Threading `dataRoot` through
  `server/context.ts` is a change to that file and to `session-boundary` both.
- **#398 — `docs/architecture.md` has never cited ADR-0046**, including before this change. A wave-3
  gap, outside this fence.

---

## One deviation from the wave-PR process, recorded rather than glossed

**This wave is three issues, not four.** #410 (*a line the fold cannot read does not discard the
records around it*) is prd-51 wave 4 and is **not** in this PR: it is blocked by #407 (prd-55 wave 5,
In progress, no PR), which holds `packages/core/src/events/index.ts` — a collision the PRD's own
*Consequence for the wave-4 fences* section records, and which this repo has already paid for once
when #405 and #400 conflicted in a guide.

The `open-wave-pr` skill says a wave with an unbuilt issue is not ready and to stop. That was put to
the operator as a choice and **the operator ruled to assemble the three now**, accepting a second
queue toll for wave 4 rather than holding three verified issues behind an unrelated PRD's in-progress
work. #410 follows in its own PR once #407 lands.

CHANGELOG: no entry. All three are defects in `[Unreleased]` features (the shipper bullet, and
`packages/team`, which has never been in a tagged release). No new ADR: #387 and #399 implement
clauses already accepted, and #398's decision is recorded in ADR-0046 itself.


gabriel-canaan:
## Review — read commit-by-commit, `origin/main` merged in

Read as three commits, not one diff. The branch was 9 behind `main` at review time, so everything below was exercised on a detached worktree with `origin/main` merged in — a clean merge, no conflict with any file this PR touches.

### What verified in your favour — all EXECUTED

| check | result |
|---|---|
| `git merge origin/main` into the branch | clean, no conflict with any PR file |
| `npm run typecheck` | green |
| `npm run lint` | green |
| `VITEST_MAX_WORKERS=6 npm test` (post-merge) | **489 files, 8983 passed**, 3 expected fail, 3 skipped |
| `gh pr checks 425` | 10/10 legs — `build-test-boot` ×3, `pack-smoke` ×6, `windows-suite` ×1 |
| fences vs. the three issue bodies | exact match, no widening, no overlap |

**The PR body's suite count is stale in your favour, not against it.** It records 8975 passed / 7 skipped; the merged tree reads 8983 / 3. The difference is `main`'s nine commits — `no-live-fleet-law.test.ts` grew 238 lines. You measured a tree that no longer exists and the newer number is the better one.

**All three headline mutations re-derived here rather than read off the reports**, and all three reproduce exactly as claimed:

- `#387` — delete the `shipper` entry from `baseChecks` → **8 red**
- `#398` — `ftruncateSync(fd, completeBytes)` → a no-op → **4 red**
- `#399` — `remaining <=` → `remaining <` → **exactly 1 red**, the fits-the-window case

Each was applied by exact-string substitution, proven to have landed with `git diff HEAD`, run, then restored with `git checkout HEAD --` and the restore proven clean before the next one.

Two claims nobody pointed a verifier at, checked because the repair-opens-the-next-hole shape lives in what a changed file now *participates* in:

- **`#387` — adding `shipper` to `baseChecks` feeds `checkEnrichmentLadder`.** It reads only five named ids (`target-path`, `session-logs`, `tmux`, `workmux`, `telemetry`), so the new entry changes no ladder output. The `hand-law.test.ts` claim holds too: `checkShipper` is exported from `cli/doctor.ts`, which `api/doctor.ts` already imported, so there genuinely is no new edge.
- **`#399` — the discriminator's soundness rests on `splitLines`.** It pushes a line for *every* newline byte, blank segments included, so `scan.lines.length === 0` is exactly equivalent to "no newline in the window". Had it filtered empty segments, a window of blank lines would have taken the oversized arm and skipped past real positions. It does not.

`docs/adr/0046-…` is still `proposed`, so editing it in place is inside `docs/adr/README.md`'s append-only rule, which binds on `accepted`.

### One finding, fixed rather than handed back — EXECUTED

**`packages/server/src/shipper/ship.ts:353` — the `oversized-unterminated` detail promised the line "will terminate and ship". It terminates and is *skipped*.**

Once `remaining > MAX_READ_BYTES` the remainder only ever grows, so the tick that sees the terminator arrive takes the oversized-**skip** arm: the cursor advances, an `'oversized'` skip is recorded, nothing is posted. That is what this file's own `'the terminator arriving turns the failure into a skip'` case already asserts — `skipped: 1`, `sent: []`. The message contradicted the test one screen above it, in the single sentence an operator acts on: it told them to expect data that is discarded by construction.

Fixed in `c0cdad53`, and the assertion is the other half of that fix — the sentence was wrong precisely by being prose no suite could contradict. Certified: with the new assertion in place, restoring the old wording turns that case red (1 failed / 22 passed in `ship.test.ts`).

### Your call, offered not asserted — EXECUTED, mutation survives

**`ship.ts:249` — `findLineEnd`'s `size` cap has no test that can fail for it.** The docblock states `size` is the stat'd size "deliberately, not the live end of file". Mutating `Math.min(SCAN_CHUNK_BYTES, size - at)` → `SCAN_CHUNK_BYTES` leaves the whole shipper suite green (10 files, 124 passed).

I am **not** calling this a defect: `while (at < size)` still bounds the loop, and a newline found past `size` is still a real line boundary, so I could not construct a wrong output from it. It is a documented clause carrying no load-bearing test. Left alone — it is yours to weigh.

### What I did not run, and one thing to weigh

- **`scripts/gate.sh` was not run.** Running it is the landing and the landing is the operator's step, as your own PR body says.
- **CI has not yet re-run on `c0cdad53`.** The green bar in the table above is against `46efda7a`; the full bar on the new head is another 13–15 min, `Windows suite` being the long pole.
- **Commit 4 is mine and has had exactly one pair of eyes — my own.** That is the position a review exists to abolish, arriving one step later, so it is stated rather than recorded away. It is a string literal and a test assertion, which is why it was fixed here instead of handed back, but a second reader before merge costs nothing.
- Nit, not actionable: `a9998e8b` and `46efda7a` carry `Co-Authored-By: Claude <claude-opus-5>`, which is not an address, so GitHub will not attribute it. `40018b22` has the well-formed trailer. Not worth rewriting your branch over.

**No blocking findings remain.** The formal approval and the merge are the operator's, not mine.

---
*Single-reviewer pass (rhizo-review), Claude Opus 5. One reviewer is weaker than the multi-seat `verify` pass — findings above are labelled EXECUTED where a command was run and its exit status read, and nothing here was posted unverified.*

PR #426, opened 2026-09-10
prd-56 — the shelf agrees with the ledger (blessed PRD)

gabriel-canaan:
One file: a blessed PRD, landing the session it was blessed.

There is no `Closes` line because a PRD is not an issue — it is the document its issues will be groomed from. Its milestone `prd56` exists and holds nothing yet; grooming is the next act, and its two issues are drafted but deliberately unfiled until this is on `main`.

## Why it exists

**Where a PRD file sits is read as a claim about whether its rulings are live, and nothing keeps that claim in step with the tracker.** Five PRDs are on the wrong shelf right now — `prd-25`, `prd-27`, `prd-43`, `prd-50`, `prd-52`, all with closed milestones and unarchived files.

The cost is not untidiness. A stale location makes a **finished** ruling look open, so work gets planned against it and prose gets edited that should have stayed a record. That is not hypothetical: #416's enumeration classified `prd-43` as a live document because of its path, and a fence was widened to edit a complete PRD's prose before anyone checked its milestone. Both were withdrawn — the finding is what produced this PRD.

And the drift recurs. `docs/prds/reconciliation-2026-08-22.md` records **nine** PRDs moved to `done/` in one batch; three weeks later there are five. A third tidy is not the fix.

## Three rulings

1. **Location is derived from milestone state, and a law says so.** Path, filename and header conventions are rejected as the source of truth because they are what drifted. Milestone state reaches the law through a **committed manifest**, not a live API call — the suite must run offline, this repo's sibling law already derives tracker facts from `git log` rather than `gh`, CI runs ten legs, and a review seat this week could not run the suite at all because its sandbox denied subprocess and socket access. A law that needs a token is a law that gets skipped. **The cost is named in the ruling rather than hidden**: the manifest can itself go stale, which is the open question wave 1 must close instead of hand-waving.
2. **A declared reason is a lawful exemption, and it lives in the file.** Without it the law would force prd-17's move against a live argument — `docs/follow-up-292.md` reasons from that file's location to call its ruling 3 live rather than archival, and holds an open item on that basis. A law that convicts a correct state is one people learn to override.
3. **A move and its citers are one commit.** Never a follow-up: the citation law is red in between, and a red gate that is someone else's follow-up is how a repo learns to ignore it.

A PRD with **no** milestone is out of scope rather than a violation. Exactly one is in that state and it is not obviously wrong; convicting it would make the law's first act a false positive.

## The keystone check, which changed the sequencing

The PRD standard asks: *what does ruling 1 make false on the day it lands, and is that thing inside this wave or a later one?*

Ruling 1's law **convicts all five files this PRD cites as its own evidence.** So the law and the fix cannot be separate waves — the law could not be green the day it landed. Wave 1 is therefore a **stack in one issue**, landed in commit order: five moves with their citers updated in the same commit, then the law.

That check is the standard's own, and skipping it is what once forced a re-sequence of an already-blessed PRD. Recording it here because the sequencing looks odd until you know why.

## What is deliberately not in it

- **Not a rule about when a PRD ships** — closing a milestone stays the operator's act; this only makes the file agree afterwards.
- **No change to append-only discipline** — a moved PRD's text is not rewritten, and its ruling citations must keep resolving.
- **Not a verdict on prd-17's own move**, which has a stated reason to stay. That is ruling 2's first subject, not an exception to it.
- **No third shelf.** `parked/` exists and is out of scope.

Rejected alternatives are named in the document with why each loses: a periodic tidy (this would be the third), deriving liveness at read time and leaving files anywhere (loses the property `docs/follow-up-292.md` depends on), and moving automatically on milestone close (a move is a content change that must update citers, so it cannot be a side effect).

## Evidence, re-derived rather than carried

Every bullet was re-derived against the trunk rather than from the drafting session: the five closed milestones and their closure dates, the five citing files (`AGENTS.md`, `docs/architecture.md`, `docs/roadmap.md`, `docs/prds/prd-30-the-open-hand.md`, `docs/prds/prd-51-the-split.md`), and zero `milestone` hits across the three laws that already read `docs/prds`.

One evidence claim was **wrong in the first draft and caught by my own validation**: it said three PRDs carry no milestone. Exactly one does. A false evidence bullet in a PRD about claims matching reality would have been a poor start.

## Verification

`doc-citation-law` was run **first and alone**, because a new document's entire risk is what it points at: **49/49**. Then the full set — typecheck clean, lint clean (1041 files), `VITEST_MAX_WORKERS=6 npm test` → **489 files, 8957 passed**, 3 expected-fail, 3 skipped. No `package-lock.json` change.

Rebased onto current `main` and the gates re-run there rather than carried over.

## Review status, stated plainly

**This document has had no independent review pass.** That is deliberate rather than an omission: it is an operator-blessed decision document, and the two choices inside it — the committed manifest over a live API call, and silence on PRDs with no milestone — were put to the operator and decided before blessing, which is recorded in the Status line and stamped on both rulings.

What a reviewer would usefully check is format compliance and whether the evidence re-derives; both are stated above with their commands.


ciaran-slow (approved):
Reviewed commit-by-commit (one commit, `866fe65`). **LANDABLE** — every gate green against the real post-merge tree. Two findings, neither blocking, both about claims rather than about the file that lands.

Every evidence bullet re-derives against current `origin/main` (`040c9ab`), not against the drafting session. EXECUTED:

- **The five adrift are exactly the five named.** Cross-referencing `gh api repos/:owner/:repo/milestones?state=all` against `git ls-tree -r origin/main -- docs/prds`, the unarchived PRDs with closed milestones are precisely `prd-25`, `prd-27`, `prd-43`, `prd-50`, `prd-52`. Closed 2026-09-07 → 2026-09-08, as stated.
- **"Exactly one has no milestone" holds** — `prd-34`, and it is the only unarchived PRD in that state.
- **The `docs/follow-up-292.md` quote is verbatim**, at line 136.
- **`reconciliation-2026-08-22.md:16` is the "Nine PRDs" line** — the `:16` pin is correct, not approximate.
- **0 `milestone` hits in all three named laws** — `packages/server/src/doc-citation-law.test.ts`, `packages/app/src/host/manifest-law.test.ts`, `packages/server/src/api/route-class-law.test.ts`.
- **The five-citer list is complete.** `packages/core/src/eras/era-2/recording.jsonl` and `session-state.snapshot.json` also contain those path strings, but both are `.jsonl`/`.json` and outside every sweep — so they are correctly excluded, and wave 1's fence is not short.
- **The sequencing precondition is satisfied**: `#416` closed 2026-09-11T01:05Z.
- **`#416`'s forward reference to prd-56 does not redden `main`** — `docs/prds/prd-17-complete-record.md:37` writes it bare, not backticked, so `CITATION_RE` cannot see it.
- **Heading convention is correct** — `## Rulings` / `## Ruling N` at the same level matches prd-51, prd-53, prd-54 and prd-55. I checked before "fixing" it.

Gates, in a disposable worktree at `origin/prd56-bless` + `origin/main` merged (`44340ca5`), node v22.23.2:

| | |
|---|---|
| merge | clean; post-merge tree differs from `origin/main` by **exactly one file** |
| `npm install` | clean, **no `package-lock.json` change** |
| `npm run lint` | 1041 files, no fixes |
| `VITEST_MAX_WORKERS=6 npm test` | **489 files, 8957 passed**, 3 expected-fail, 7 skipped |
| `doc-citation-law` alone | 49/49 |

I did **not** push the merge commit to your branch, so CI has not run on `44340ca5` — the merge was verified locally only. Say the word if you want it pushed.

## Finding 1 — ruling 3's red-in-between guarantee does not cover two of the five citers (EXECUTED)

Ruling 3 justifies "a move and its citers are one commit" with *"the citation law is red in between"*. That is only partly true. `doc-citation-law.test.ts:493`:

```
const EXCLUDED_DIRS = ['docs/research/', 'docs/review/', 'docs/prds/']
```

`docs/prds/` is excluded **as a citing source**, and the sweep is `docs/*.md` + `packages/*.ts` — so `AGENTS.md` is not swept by this law at all. Mutation, run against the post-merge tree:

- **positive control** — broke `docs/architecture.md`'s cite of prd-25 → `every in-scope citation exists` **FAILS**, reporting `docs/architecture.md -> docs/prds/prd-25-MOVED.md`. Harness proven.
- **`docs/prds/prd-30-the-open-hand.md:44`** — rewrote its cite of prd-27 to a path that does not exist → **49/49, green**.
- same for `docs/prds/prd-51-the-split.md`'s cite of prd-52, and `AGENTS.md`'s cite of prd-43.

Per-citer scope for the five wave 1 must move:

| PRD | citers | seen by the law? |
|---|---|---|
| prd-25 | `docs/architecture.md`, `docs/roadmap.md` | both |
| prd-27 | `docs/prds/prd-30-…`, `docs/roadmap.md` | roadmap only |
| prd-43 | `AGENTS.md`, `docs/roadmap.md` | roadmap only |
| prd-50 | `docs/roadmap.md` | yes |
| prd-52 | `docs/prds/prd-51-…`, `docs/roadmap.md` | roadmap only |

All five are cited from `docs/roadmap.md`, so a **total** omission still reddens. The gap is a **partial** one: update `roadmap.md` and `architecture.md`, forget `prd-30`, `prd-51` or `AGENTS.md`, and wave 1 lands a dead citation with a green suite. That is the sibling-case shape AGENTS.md names.

Success criterion 2 is unaffected — its second clause ("or a citer still names the old path") is independent of the law, so the criterion is sound as written. It is ruling 3's *rationale* that overstates.

**Your call, offered not asserted:** I have not touched ruling 3 — amending a blessed ruling is yours, not a reviewer's mechanical fix. The cheap version is a clause noting the law covers `docs/*.md` outside `EXCLUDED_DIRS` only, so the wave-1 commit must check the two `docs/prds/` citers and `AGENTS.md` by hand. Worth recording on #427 either way, alongside the manifest-staleness question the PRD already books for that wave.

## Finding 2 — the PR's own verification claim is true but vacuous (EXECUTED)

The body and the commit message both say the citation law was run first and alone *"because a new document's entire risk is what it points at: 49/49."* By Finding 1, `docs/prds/` is excluded as a citing source — so that run never read prd-56's citations. Mutation: I broke `` `scripts/dev/prd-reconcile.sh` `` inside prd-56 and the law still returned **49/49**.

This is in PR prose, not in the file that lands, so it changes nothing on `main`. Flagging it because the claim reads as coverage it does not have.

**So I re-derived it by hand instead**, replicating the law's extractor (`CITATION_RE`, fenced blocks stripped, `:NNN`/`#anchor` suffixes stripped) against the merged tree: **16 distinct citations, 0 broken.** Self-controlled — injecting one bad path makes the checker report it. Your citations are in fact all good; only the evidence for that was hollow.

## One nit

The body says prd-56's two issues are "drafted but deliberately unfiled until this is on `main`" — #427 and #428 are already filed under the `prd56` milestone. Stale PR prose only.

---

Approving. The merge itself is the operator's.

PR #437, opened 2026-09-11
prd55 w5: the R&D hand is the operator's own claude — schema, events and folds in core; engine, CLI, route and settings on the server; the lab's listing folds the log

KelliherL:
Closes #407. Closes #412. Closes #409. prd-55 wave 5 — rulings 1, 2, 3 and 4, and the ADR ruling 1 asked for (ADR-0048). One PR = one (milestone, wave); two lanes, fifteen own-work commits plus two conductor commits to the coupling map, read in order after `main`.

## What becomes true

**Core (lane 5a, Sonnet — #407):** a fixed R&D result schema in the events module's idiom — `patterns[]` (shape, source items, count, `heldBack`), `proposals[]` (one pattern, one `varies` dimension of `model | brief | checkpoint | gate`, 2–3 arms differing in that dimension only, a `checkpointPick` with every rejected checkpoint and its reason) — and two pure laws beside `laws.ts` (`isHeldBack` = count < 2; `rdRefusalReason` with two verbatim refusal sentences). Four events with provenance — `rd.patterns`, `rd.proposal`, `rd.refused`, `rd.override` — carrying the CLI's own `total_cost_usd` and `duration_ms`, the prompt and corpus digests, `claude --version` and the corpus name; `state.rd` per lane, by pattern, by proposal; the reducer folds them; every enumeration a new event type reaches is reconciled (two hand-enumerated lists, the era snapshots re-blessed with only the empty `rd` slice, web/fleet's slice ratchet — ruled: `rd` is the lab's, never fleet input — and the recorder's frozen-fold list).

**Server (lane 5b, Opus — #412):**
- **The engine** spawns the operator's own `claude -p --output-format json --model <m> --tools "" --strict-mcp-config [--max-turns n] -- <prompt>` (the no-tools flag read off `claude --help` 2.1.266 and recorded), reads the JSON result, validates against core's schema — a two-dimension or held-back proposal is recorded as `rd.refused` with core's sentence, never patched — and records the events through the live session's recorder. The binary is whatever `lab.agentCommand` names, resolved on PATH; absent, the exact sentence *"no claude on this machine's PATH — the R&D hand is your CLI, installed by you"* and nothing spawned.
[...]

Issue #436, opened 2026-09-11
prd51 w7: the architecture document cites the journal's decision record


ciaran-slow:
## Fence (may touch ONLY)
- `docs/architecture.md`

## Blocked by
**Nothing.** #427 landed in PR #458 (`a893af22`), releasing `docs/architecture.md`. Re-linted
2026-09-14 against every live In-progress issue: `docs/architecture.md` reported `ok`, one owner.

**Re-homed to wave 7 (operator, 2026-09-14).** This issue was groomed into wave 5 and left it
before dispatch behind #427; wave 5 has since merged as PR #454, so there is no wave-5 bundle
left to land in. Rather than pay the ~21 h per-PR queue toll for one 16-line docs commit, it
rides wave 7's bundle PR. This issue's single path is claimed by no other
wave-7 lane, so the re-homing costs the wave nothing. Nothing about the work changed: same one
path, same Definition of done.

**Already built and verified** at `c64a4cb7` on `436-architecture-cites-adr-0046` — two review
findings raised and fixed (the entry cited a PR number where every sibling cites an issue), full
gates green, both required mutations reproduced by an independent pass on a different model.

Sibling lanes hold other surfaces: the router lane owns the team package's api and auth
directories; the /connect lane owns the web connect directory and the doctor files in the server
package. Neither goes near a document.

**Prohibition.** The record itself is not edited here — its status line is `proposed` and moving
it to `accepted` is the operator's act. `docs/adr/README.md` already carries its index row, and
no lane this wave claims that index.

**Model:** sonnet. **Wave:** 7.

## Definition of done
Recorded as a deferral in #425's body: `docs/architecture.md` has **never** cited ADR-0046,
including before that change. Wave-3 residue, not a regression.

- `docs/architecture.md` cites ADR-0046 where it describes the team server's ingest, in the same
  style the document already uses for its sibling records.
- The text says **what the journal is** — a CRC-framed append log whose ack follows a durable
  write — not merely that a record exists. A pointer with no claim is the failure this repo
  already records for citations.
- Every backticked repo-rooted path in the new text resolves on that commit
  (`doc-citation-law.test.ts` green), and any backticked symbol it names still exists in source
  — that law checks paths, not symbols.
- **The prose count this file carries for gated reads is untouched.** `docs/architecture.md` is a
  coupling point because `api/route-class-law.test.ts` reads that sentence; this lane adds no
  route and must not move the number.
[...]

PR #438, opened 2026-09-11
The CI leg moves into the landing gate

ciaran-slow:
GitHub Actions is being retired on this repo for cost. This moves the CI leg
onto the landing gate, and rescues two checks that would otherwise have been
deleted along with the workflows.

**Measured 2026-09-11, this repo alone:** 45,209 billed minutes over Sept 1–11
(~4,490/day). macOS was **17% of raw machine time and 60% of the bill** at the
10x multiplier; Windows 23% at 2x; ubuntu 17%. Derived from job-level durations
via the API — the `/timing` endpoint returns zeros on this plan.

## Closes

**Nothing — this PR has no issue and no milestone.** It was worked directly
from an operator instruction. Both are yours to assign if you want it tracked;
I did not invent a milestone, and `charter-laws` did not obviously fit since
no charter section states this rule.

## What changed

1. **`refactor(ci)`** — `packaging-guard.mjs` and `boot-smoke.sh` extracted from
   inline YAML; `ci.yml` now calls them; `ci-local.sh` composes the leg.
2. **`feat(gate)`** — `gate.sh` gains lint, build, packaging guard, boot smoke,
   pack smoke, before the merge.

New sequence: `… → suite → typecheck → **lint → build → packaging guard → boot
smoke → pack smoke**`. `lint` had never been in this gate at all.

## What is the sibling case?

Three, swept deliberately:

- **Other laws reading `ci.yml`.** `boot-line.test.ts` asserted ci.yml contains
  the `rhizomorph running at` marker — moving the grep to a script would have
  left it passing **vacuously** against a file no longer containing the marker.
  Fixed here. I checked the rest: `node-floor-law` (reads `node-version:`),
  `route-class-law` (job shape and matrix), `runbook-delivery-law` (gated step
  names, which is why step names and `if:` conditions are byte-identical) — all
  unaffected, none vacuous.
- **Other workflows with inline logic — NOT fixed here.** `windows-suite.yml` is
  already clean (it calls a script). But `desktop.yml` (2 blocks) and
  `release.yml` (2 blocks, including "Version guard — tag, package.json and
  `--version` all agree") still carry logic that exists nowhere else and dies
  the same way. Out of scope for this PR — and **not filed**, contrary to what
  this line said until the review caught it: there is no such issue, and there
  is no obvious milestone for one (no PRD claims workflow hygiene, and
  `charter-laws` wants a charter section that states the rule). Naming it here
  is the honest holding position until that decision is made.
- **Pre-merge vs post-merge build.** `gate.sh` already ran `npm run build`
  after the merge. Adding one before it means the same command has two failure
  meanings, so `build-red` (holds the merge, lane keeps its work) is a distinct
  verdict from `build-broken` (holds only the push, fix forward on main).

Each check was broken deliberately and observed to redden:

| mutation | result |
|---|---|
| drop `packages/server/dist/` from the pack allowlist | guard flags the files, **exit 1** |
| drop the `x-rhizomorph-capability` header from `/api/meta` | boot smoke **exit 1** — the endpoint is genuinely gated, the check is not vacuous |
| change `rhizomorph running at` in `boot-smoke.sh` | `boot-line.test.ts` **1 failed**; restored → 11 passed |
| insert 23 lines into `gate.sh` | `gate-honesty-law` producer citations **red**; re-derived → 228 passed |

The citation one was not planned — the insert reddened it, which is the law
working. Worth knowing for the next person: the law reads **every** `:<digits>`
token between its markers as a citation, so writing the mapping as
`:659 -> :682` silently re-cites the stale numbers and keeps it red. That
mistake is recorded in the block itself.

`ci-local.sh`'s own gating was harness-tested across four scenarios: build-red
skips the three vacuous-green steps, test-red still yields typecheck and lint,
exit 0 green / 1 red, `--no-pack-smoke` honoured.

## Verification

Run in the lane worktree at node 22.22.2 (`.nvmrc`), on this exact base:

- `npm run typecheck` **exit 0** · `npm run lint` **exit 0**
- `VITEST_MAX_WORKERS=6 npm test` — **493 files, 9,081 passed, 0 failed**
- Each new gate step run by hand: lint 0 · packaging guard 0 · boot smoke 0 ·
  **pack-smoke 0, no leaked processes**. Pack-smoke specifically, because adding
  an unverified check to the landing gate breaks every landing.

`gate.sh` itself was **not** run — running it *is* the landing.

**Re-run on the merged tree** (`318db245`, which carries the review's two fixes,
`613a7422` and `origin/main` through #439), macOS, node 22.22.2:
`npm run typecheck` **0** · `npm run lint` **0** (1058 files) ·
`VITEST_MAX_WORKERS=6 npm test` — **496 files, 9,216 passed, 3 expected fail,
7 skipped, exit 0**.

**CI on `e7988d5f` was green on all ten legs**, `windows-suite` and both macOS
legs included — so the extracted scripts, and the `set -m` boot-smoke fix, are
witnessed on ubuntu, macOS and Windows while Actions still runs. That is the
"worth doing before Actions is switched off" item below, done.

## Fence widening

No fence — no issue.

## Review round — 2026-09-11

[KelliherL's review](https://github.com/launchpad-26/rhizomorph/pull/438#pullrequestreview-5174683107)
read both commits, re-derived both of the mutations above, and proved by diff
that the extracted step names, `id:`s and `if:` conditions are byte-identical —
so `runbook-delivery-law` and `node-floor-law` cannot have gone vacuous by
construction, not merely by passing. It pushed two fixes and merged `origin/main`.

**`e74e56c4` — boot smoke printed "server shut down cleanly" with the server
still running.** `kill -TERM "$SERVER_PID"` signalled the `npm start` wrapper,
not the node server under it. Measured on Linux (npm 10.9.8): one re-parented
survivor per run, still holding its listening socket, twice. A CI runner is
disposable, so the inline version never paid for it; the landing gate and
`ci-local.sh` run on the operator's machine. Fixed with the shape
`pack-smoke.sh` already uses (#225) — `set -m`, TERM to the process group.

Corroborated on macOS here, EXECUTED, on a proxy rather than the real server (a
scratch package whose `start` is a bare `node -e`, driven by copies of the
before/after shapes, with a no-kill control that finds 1 survivor so the probe
is not vacuous): under node 22.22.2 / npm 10.9.7 and node 24.12.0 / npm 11.6.2,
**both** shapes leave zero survivors, including when the script is interrupted
mid-run — which is the case `set -m` changes, since the job is no longer in the
script's process group and cleanup rides entirely on the EXIT trap. So the leak
looks Linux-specific and the fix is inert-but-harmless on darwin, under bash 3.2.

**`613a7422` — the honesty law now holds the CI leg's guards.** The review's
open item: `gate-honesty-law` does not police the five new lines. Re-derived
rather than taken on report — replacing the lint guard's `fail "lint red"
lint-red` with a bare `echo "GATE FAILED: lint red"` left the law **green, 228
passed**, while the control (dropping `lint-red` from `GATE_VERDICT_VOCAB`)
reddened it. So `4494b277`'s body was wrong to say the predicate "already reads
[them] as checked": it does not read them at all. Closed as a class rather than
a list — every `( cd "$W" && … )` line in the tracked file must honour its fault
through `fail()` — because an enumeration of today's five is blind to the sixth.
Three mutations red with it in place: the lint guard printing without failing,
the brace-less typecheck twin doing the same (which an enumeration would have
missed), and a guard deleted outright.

**A fourth sibling, REASONED from the source and not fixed here:**
`scripts/dev/soak.sh --spawn` has the identical wrapper-kill shape at its last
line — and worse, it sets `PID=$!` to the npm wrapper and then samples *that*
process's RSS, CPU and fd count for the whole soak. If the wrapper is not the
server, a `--spawn` soak measures the wrong process. Out of this PR's scope,
owned by #213's author, and not filed.

## Condition 1, per commit

| commit | read by an independent pass? |
|---|---|
| `2e86c42c` refactor(ci) | yes — the review linked above |
| `4494b277` feat(gate) | yes — the review linked above |
| `e74e56c4` fix(scripts) | **no** — the reviewer's own fix, which this repo says they may not review into `main`. Corroborated on macOS above, which is not a review |
| `89bdd06c` docs(contributing) | **no** — the reviewer's own, prose only |
| `613a7422` test(gate-honesty) | **no** — written here in answer to the review's open item |
| `e7988d5f`, `318db245` | merges of `origin/main`, no content of their own |
| `b2d3d23c` feat(ci) | **no** — added after the review below, and not read by anyone else |

Four shas still want a reader who did not write them. Stated here rather than
recorded away, and more than usually worth it: this changes the landing gate
every collaborator depends on.

## Added 2026-09-12 — `b2d3d23c`, the verdict reaches the PR

Actions is now actually off: all four workflows are `disabled_manually` on the
repo (`ci`, `desktop`, `release`, `windows-suite`). The repo-level switch
returns 404 for a non-admin token, so this is per-workflow — a *new* workflow
file added later would still run, and `release.yml`'s `v*`-tag publish and
`desktop.yml`'s installers are off until re-enabled individually.

That closes the loop this PR opened and leaves a gap it did not: with no check
runs at all, a PR says nothing about whether the code works. `ci-local.sh` ran
the leg and the result died on the operator's terminal.

`scripts/pr-verdict.sh` publishes it, and `ci-local.sh` gains `--pr <n>`. **Two
marks, deliberately not interchangeable:**

- the **`Passed local CI` label** — repo-level, pinned to nothing. One push
  after a green run and it claims a result for a tree nobody tested.
- a **`local-ci (<uname -s>)` commit status on the head sha** — per-sha by
  construction, so it lands in the check box and goes quiet the moment the
  label starts lying. The context names the platform because one operator
  machine is one of the three legs `ci.yml` ran; a `local-ci (Darwin)` row with
  nothing beside it states that silence rather than hiding it.

Preflight refuses a PR that is not open, a `HEAD` that is not the PR's head sha,
and a dirty tree — **before** the leg, so a refusal costs a second and not a
suite. `--pr` refuses `--no-install` and `--no-pack-smoke`: a label saying the
full leg passed with two steps skipped is a claim wider than its evidence.

### What is the sibling case?

A status posted to `statuses/HEAD` rather than to the sha. GitHub accepts it,
it renders identically, and it silently re-points itself on every push — the
same staleness the label has, in the half that exists to be immune to it. The
law pins the sha into the URL and mutating it to `HEAD` reddens.

The other one, found while writing the harness: `report` re-runs preflight
rather than trusting the sha preflight pinned at the start. A tree dirtied
*during* the leg is a real case — it is what a mid-run "just one fix" produces
— and without the recheck the verdict names a tree that was never tested.

`pr-verdict-law.test.ts` drives the script for real with `spawnSync`, against a
scratch git repo and a stubbed `gh` on PATH, the way `windows-suite-law` drives
`windows-triage.sh`. Six mutations, each red, control green at 17/17:

| mutation | result |
|---|---|
| drop preflight's dirty-tree guard | **3 red** |
| `report` skips its own preflight | **1 red** |
| post to `statuses/HEAD`, not the sha | **1 red** |
| a red run leaves the green label in place | **1 red** |
| never cut the description to GitHub's 140 | **1 red** |
| `--pr` accepts `--no-install` / `--no-pack-smoke` | **2 red** |

Every refusal is asserted on an **empty gh log**, not on the exit code: an
exit 2 *after* the status was already posted is exactly what an exit-code-only
assertion cannot see.

One harness detail worth the next reader's time: the stub's `bin/` and its log
must live **outside** the scratch repo. Inside, they read as untracked files,
every case refuses on a dirty tree, and five assertions fail for a reason that
has nothing to do with the code under test.

### Verification

macOS, node 22.22.2, on `b2d3d23c`: `npm run typecheck` **0** · `npm run lint`
**0** (1059 files) · `VITEST_MAX_WORKERS=6 npm test` — **497 files, 9,233
passed, 3 expected fail, 7 skipped**. Both refusal paths also exercised against
the **real** `gh` on live PRs #438 and #446.

**The posting path is proven against the stub only.** It cannot be run against
a real PR yet: posting requires a clean tree at that PR's head sha, and
`pr-verdict.sh` does not exist on any current PR head. The first live green
will be its first real exercise — worth watching rather than assuming.

`CONTRIBUTING.md` and `AGENTS.md` both gained the standing rule: read the
status, not the label, and never write "CI passed".

## Platform caveat, which is permanent

CI ran ubuntu + macOS at two node legs plus a native Windows suite. The landing
gate runs on whichever machine the operator is at. **A green gate is now
evidence about one platform and silence about the other two** — `case-collision-law`
exists because a macOS leg caught 16 failures Linux cannot see, and
`windows-suite` is prd-25's whole deliverable. AGENTS.md now says "do not write
CI passed". prd-25 still needs amending or superseding; that ruling is not in
this PR.

**Worth doing before Actions is switched off:** these scripts are proven on one
Mac. While the workflows still run, they will witness themselves on ubuntu,
macOS and Windows for free. Git Bash is where I would expect trouble —
`mktemp -d`, the `grep -o` on the meta tag, `curl -sI` line endings.


KelliherL (commented):
## Review of #438 — two fixes pushed (one substantive, one mechanical), `origin/main` merged in, the whole leg run on the union

Read commit-by-commit (`2e86c42c` the extraction, `4494b277` the gate), in a clean clone at node 22.23.2 / npm 10.9.8 on Linux. **Most of what I checked verified in the PR's favour**, and the specifics come first because they are the evidence.

### What verified

**EXECUTED, on your head `4494b277`:**

- `npm run typecheck` **0** · `npm run lint` **0** (1052 files) · `npm run build` **0**.
- The seven laws that read `ci.yml` or `gate.sh` — `boot-line`, `node-floor-law`, `route-class-law`, `doc-citation-law`, `runbook-delivery-law`, `static.test`, `gate-honesty-law` — **7 files, 436 passed, 2 skipped**.
- **Step names, `id:`s and `if:` conditions in ci.yml are byte-identical** before and after (diffed the extracted lines of `origin/main` against this branch: no output). So `runbook-delivery-law`'s gated-step-name reads and `node-floor-law`'s `node-version:` read cannot have gone vacuous — the claim in the body holds by construction, not only by the tests passing.
- `node scripts/packaging-guard.mjs` **0** — "65 files, all allowlisted". `bash scripts/boot-smoke.sh` **0**.
- **Your mutations, re-derived:** allowlist entry `packages/server/dist/` dropped → guard **exit 1**, naming `packages/server/dist/cli/index.js`. Marker changed in `boot-smoke.sh` → `boot-line.test.ts` **1 failed / 10 passed**, restored green. And one you did not list: `lint-red` removed from `GATE_VERDICT_VOCAB` while the new `fail … lint-red` call stands → `gate-honesty-law` **red** on *"fail() call site(s) naming a category outside GATE_VERDICT_VOCAB"*. The vocabulary addition is load-bearing.

### One substantive finding — fixed, and pushed as `e74e56c4`

**`boot-smoke.sh` printed "server shut down cleanly" with the server still running.**

`kill -TERM "$SERVER_PID"` signals the `npm start` wrapper. npm spawns `sh -c rhizomorph --port 0`, which spawns the node server, and the TERM never reached node. EXECUTED, twice, before any change: run the script, exit 0, last line "server shut down cleanly" — and `pgrep -f 'rhizomorph --port 0'` finds **one survivor**, re-parented (ppid no longer the script), still holding its listening socket. Killing a bare `npm start` wrapper by hand leaves its node alive the same way, so this is npm's signal handling, not the script's timing. (When I came back ~3 minutes later the orphans had exited on their own; I did not chase why. The gate runs pack smoke immediately after this step, so the overlap is real either way.)

The inline version in ci.yml had the identical shape and never paid for it, because a runner is disposable. This PR moves the check onto the operator's machine, once per landing, and onto every developer who runs `ci-local.sh` — which is exactly where the leak lands.

**The fix** is the shape `pack-smoke.sh` already uses for the same reason (its own header cites #225, "six servers were leaked live"): `set -m` so the backgrounded npm is its own process group, and `kill -TERM -- "-$SERVER_PID"` from both the trap and the clean path. Verified **EXECUTED**: two runs, exit 0 each, **zero survivors** after each; the log is unchanged line for line (no job-control noise under `set -m` in a non-interactive bash).

**Not approving, on purpose:** it changes process-control behaviour in a script you authored, and this repo's rule is that a reviewer does not approve their own substantive fix. **Please confirm `e74e56c4` reads right to you.** If you'd rather the leak stay documented than fixed here, revert the commit — it is one commit, yours to drop.

### One mechanical fix — pushed as `89bdd06c`

`CONTRIBUTING.md` cites `ci.yml:60`, `:69`, `:72`, `:97` and `:166` for Build-before-Test, Lint, the packaging guard, the boot smoke and the pack-smoke job. **All five had already drifted on `origin/main`** (`:60` is `node-version-file: package.json`, `:97` is `set -e`), and this PR's extraction takes 96 lines out of the file, leaving `:166` past the end of the job it names. No law reads them, so nothing reddened — which is why they rotted. Replaced with step and script names, the rule AGENTS.md applies to `gate.sh`. Prose only; the sibling of your `boot-line.test.ts` move, for readers instead of laws.

### The merge and the union

`origin/main` moved under you while this was open (#437 merged at 03:05Z as `74d4c984`). Merged it **into** the branch (no rebase, no force-push) as `e7988d5f` — clean, no conflicts. Then the whole leg, on the union, with your own `ci-local.sh --no-install`:

```
  SKIP  Install (npm ci)        (--no-install; the clone was npm ci'd fresh 40 min earlier)
  PASS  Build
  PASS  Test                    496 files · 9219 passed · 3 expected fail · 3 skipped
  PASS  Typecheck
  PASS  Lint
  PASS  Packaging guard
  PASS  Boot smoke
  PASS  Pack smoke
  ci-local: green — exit 0 in 154 s, Linux / node v22.23.2
```

`pgrep -f 'rhizomorph --port 0'` after the leg: **0** — with `e74e56c4` in place, boot smoke and pack smoke together leave nothing behind.

**One platform (Linux), as your own header says.** I did not run this on macOS or Windows, and the `windows-suite` leg is CI's to report.

- **`gate-honesty-law` does not police the five new lines.** Mutation, EXECUTED: replace the lint step's `|| { …; fail "lint red" lint-red; }` with `|| { …; echo "GATE FAILED: lint red"; }` — the fault string printed, `fail` never called — and the law stays **green (227 passed)**. Its structural predicate walks `VAR=$(...)` producers (the `sameLine + nextLine + unchecked = allProducers` invariant), and a `( cd "$W" && npm run lint … ) || …` subshell is not one, the same as the pre-existing typecheck line at :603. So "a shape the predicate already reads as checked" in the commit body is slightly generous: it does not read them at all. Not this PR's defect and not a blocker — the five lines *are* honest — but the law's reach is narrower than the AGENTS.md sentence about it suggests, and that is worth a line somewhere if you agree.
- **`ci-local.sh`'s four harness scenarios** (build-red skips, test-red still typechecks/lints, exit codes, `--no-pack-smoke`) — I did **not** re-derive them; the one scenario I ran is the all-green one above. Taking the body's word, labelled as such.
- **"Filed separately"** for `desktop.yml` and `release.yml`'s inline blocks: I could not find the issue (`gh issue list --search release.yml`, and your open issues since 02:00Z are #432–#436 only). If it exists under another wording, ignore this; if not, it is one issue away from being true.
- No issue, no milestone: as the body says, yours to assign. It reads as `charter-laws`-adjacent to me too, and I have no better home for it.

### What I did not run

macOS and Windows. The Actions billing numbers in the body (I have no access to that view). The `--version` guard in `release.yml`, which is out of this PR's scope by its own statement.

KelliherL:
**One measured piece of evidence for the design, from an hour ago — offered, not asserted, and it is about what the ubuntu legs were doing for us.**

Gating prd-55 wave 6 on a native-Windows box, the lane's gate and mine both reported green-with-known-reds. **CI's three `build-test-boot` legs went red**, on a real defect: the wave's new web client is the app's seventeenth outbound call site, and README's outbound-fetch recipe names every module and site by hand while `api/route-class-law.test.ts` sweeps the same roots against it. Both count pins were stale.

**Why neither local gate saw it.** `route-class-law.test.ts` is on `.windows-known-failures` for a *line-endings* failure (its README-paragraph reader does `indexOf('\n\n')` on a CRLF checkout). So on Windows the whole **file** reads as expected-red, and a **new** failing test inside an already-listed file is invisible to a per-file attribution. It is prd-25's own *"a count masks a swap"* one level down: a known-red **file** masking a new red **test**.

**What that implies for this PR**, given Actions is going away: `windows-triage.sh` is a per-FILE verdict by design (ruling 2, and rightly — a count masks a swap), and `ci-local.sh` currently runs `step "Test" npm test || true`. On a Windows box after the retirement, the signal that caught this defect does not exist anywhere: the ubuntu legs are gone, and the local verdict cannot tell "this listed file failed for its listed reason" from "this listed file also failed for a new one".

The cheap form, if you want it: have the suite step emit per-test results (`vitest --reporter=json`), and have the triage compare the failing **test names** inside each listed file against the evidence the list already states for it — `.windows-known-failures` carries that evidence per row today (`# evidence: expected '  build-test-boot:\r\n…' not to match`), so the data is there. A listed file failing with a test name the row does not describe is the red the list was never meant to absorb. That keeps ruling 2's per-file verdict and closes the hole underneath it.

Not a review of this PR — it is yours and it is draft. Just the one thing I learned today that your design can still act on, with the numbers: one real defect, three ubuntu legs, zero local signal.

(Unrelated, for whoever assigns it: the `desktop.yml` / `release.yml` blocks you name as the honest holding position are the same class, and the version guard in particular is the one whose silent death would be worst.)

ciaran-slow (commented):
Reviewed ahead of landing, at the operator's request. Recording it because this PR changes the landing gate itself and is about to merge with no CI available to re-check it.

## Verified — EXECUTED

- **CI green is on the CURRENT head.** Run `34560232621` sha `318db245` == PR head `318db245`. All **10** legs pass — `build-test-boot` ×3, `pack-smoke` ×6, `windows-suite` — counted against the three-jobs/ten-legs list. This ran *before* the Actions billing cap.
- **Still merges clean against today's main**, 55 commits of drift later: `git merge origin/main` → rc=0, no conflicts. The tree GitHub will produce on merge is the tree tested below.
- **The four laws this PR can reach all pass on that merged result** — `gate-honesty-law`, `boot-line`, `runbook-delivery-law`, `doc-citation-law`: **300 tests, 4 files, rc=0**.
- **Every new script parses under this box's actual bash** — `/bin/bash` here is **3.2.57**, not 4+, and `bash -n` is clean on `boot-smoke.sh`, `ci-local.sh` and `gate.sh`; `node --check` clean on `packaging-guard.mjs`. Checked deliberately: this repo has been bitten before by a bash-4 construct in a script the gate runs under 3.2.

## The review record, stated rather than averaged

| commits | record |
|---|---|
| `2e86c42c`, `4494b277` | KelliherL's review, 2026-09-11T03:23:40Z (COMMENTED) |
| `e74e56c4`, `89bdd06c` | authored in response to that review |
| `613a7422`, and the two merge commits | **no independent review record** — `613a7422` post-dates the review by 28 minutes |
| the merged-with-today's-main result | this pass |

The review on record is **COMMENTED, not APPROVED**. I am not upgrading it to an approval on someone else's behalf; what I can attest is the list above.

## Why it is landing anyway, said plainly

The Actions billing cap hit this repo between 22:51 and 23:19 UTC on 2026-09-11. Every run since — including `main`'s own — dies in 2–7 seconds with *"The job was not started because recent account payments have failed or your spending limit needs to be increased."* Nothing has executed since. So the repo currently has **no verification at all**, and this PR is the thing that restores some. Holding it for a CI run that cannot happen would leave the gate emptier, not safer.

That is a judgement about sequencing, not evidence that the change is correct. The evidence is the section above, and it is a local, single-platform pass — **macOS only**. The pre-cap CI legs are the only cross-platform witness this change will get.

## For whoever reads this later

The primary worktree still holds **uncommitted copies of eight of these files** — an *earlier* draft than this branch. It lacks `e74e56c4`'s process-group fix (`set -m` and `kill -TERM -- "-$PID"`, without which the boot smoke leaves a live server behind) and the 62 lines `613a7422` added to `gate-honesty-law.test.ts`. Those local edits are superseded by this merge and should be discarded, not committed.

PR #368, opened 2026-09-08
prd50 w2: the fixed ceiling says so, in the code and in the note

gabriel-canaan:
Closes #237
Closes #238

prd-50 wave 2 — the two issues that make the fixed ceiling **legible**, now that wave 1 (#236, merged `fda4af65`) makes it enforced. Same milestone, same wave, disjoint fences: one PR, two commits, one file each.

The PRD's problem was never that the ceiling could be raised — it is that nothing said whether it *may* be. prd-35's whole lifetime passed with this question sitting as a residual, because a reader who wanted the value configurable found a well-argued number and no answer to their actual question.

**`ed34dedc` (#237)** — the docblock above `LAB_CLI_LOCK_CEILING_MS` gains the who-may-change-it answer: fixed by prd-50 ruling 1, not operator-settable, because raising it rebuilds the queue prd-41 ruling 2 refused — spend hidden behind latency, which prd-12 ruling 3 forbids. The existing rationale for *why 30 seconds* is untouched; this adds an answer rather than rewriting one.

**`372757d5` (#238)** — `lab-launch-ceilings.md` stops inviting an action the instrument cannot perform. It told the reader of `MAX_ARMS` to *"raise it if a real workflow needs more"*; there is no config store, no env var, no flag.

## What is the sibling case?

**#237's, named by the issue and done:** `withLabCliLock`'s `ceilingMs` parameter carries a comment calling itself *"a test seam; production takes the default."* A reader told the constant is fixed who then finds a parameter that overrides it sees a contradiction. That comment now also says production has **no path** to the seam, so the pair agrees.

**#238's, which required a grep rather than a fix to the quoted sentence:** any entry telling the reader to change, raise, lower or tune a number makes the same promise. Swept — no others remain.

**The one that is deliberately NOT swept:** the rest of the ceiling family. prd-50 open question 1 leaves `MAX_ARMS`, `RESTORE_EXEC_TIMEOUT_MS`, `COMPARE_*`, `FORK_EXEC_TIMEOUT_MS`, `ROUTE_EXEC_TIMEOUT_MS` and `COLLECTOR_EXEC_TIMEOUT_MS` unruled. #238's difficulty is that `MAX_ARMS` — the constant carrying the invitation being corrected — is *in* that unruled family, so the correction must not resolve by declaring the family fixed. It separates two facts instead: **no number here is operator-settable today**, for any of them, because the mechanism does not exist; and **the lock ceiling is fixed by ruling as well as by absence**, which is true of one constant only.

[...]

PR #440, opened 2026-09-11
test(server): a symbol cited from a document or a comment still exists  (#369)

### LINKED ISSUE #369: doc law: a symbol cited from a document still exists in source
## Fence

Ungroomed, deliberately. The instrument's **grammar** is a decision that has to be made before a fence can be drawn — see "The hard part" below — and the two honest resolutions touch different files.

## Why

`packages/server/src/doc-citation-law.test.ts` enforces prd-43 ruling 1: *a path cited from a document or a comment must exist*. It works, and it is thorough.

**It cannot see a cited SYMBOL, and that is structural rather than accidental.** Its extractor, at `packages/server/src/doc-citation-law.test.ts:683`:

```
const CITATION_RE = /`((?:packages|scripts|docs)\/[A-Za-z0-9_./*:#-]+)`/g
```

A match requires a `packages/`, `scripts/` or `docs/` prefix **and** a slash. A backticked constant, function or type name — `` `MAX_ARMS` ``, `` `withLabCliLock` ``, `` `LabRunDTO` `` — cannot match by construction. So a document may name a symbol that was renamed or deleted years ago and every law in the repo stays green.

## What it cost, measured

`8f7a60ec` (#324, prd-53 wave 2, merged 2026-09-08 10:11 +1200) renamed `MAX_ARMS` to `LAUNCH_CEILING_LANES` and recounted it in **spending lanes (arms × runs)** rather than arms. EXECUTED on `main` today:

```
grep -rn 'MAX_ARMS' --include='*.ts' packages/ | grep -v '\.test\.'
  (no output)
```

`docs/design-notes/lab-launch-ceilings.md` went on saying `MAX_ARMS` in two places. The citation law swept that file and passed, correctly — every *path* in it resolved.

The rot then propagated the way stale prose does:

1. **#238's Definition of done was written from the stale note** and instructed, verbatim, that the entry say *"no mechanism exists to change it at runtime … whether one should is unruled"*.
2. **A lane executed that brief faithfully** and committed prose that was false in three clauses: the constant did not exist, a `--ceiling-override <n>` flag did (`packages/server/src/cli/lab-fork.ts:65`), and prd-53 ruling 6 had ruled it **configurable** on purpose.
3. **Two design notes in one directory then gave opposite answers about the same ceiling** — inside the commit whose stated purpose was prd-50 ruling 3, *"the design notes stop inviting what cannot be done."*
4. It was caught by a **human reviewer on PR #368**, as a blocking finding, after the review round had been paid for.

A doc-symbol law would have reddened `8f7a60ec` at the moment of the rename, which is the cheapest possible place to catch it.

## The hard part — read this before writing a regex

The check is trivial. **The grammar is not**, and getting it wrong produces one of the two failure shapes this repo names most often.

Backticks in this corpus carry at least: file paths (already covered), shell commands, CLI flags (`` `--ceiling-override` ``), env vars, HTTP routes (`` `POST /api/lab/launch` ``), npm scripts, git refs and shas, branch names, external API and library symbols this repo does not define, and **deliberately historical references** — `doc-citation-law`'s own doc cites `` `docs/prd2.md` `` *because* it is the old name.

So:

- a matcher too broad fires constantly on prose and gets suppressed or allowlisted into uselessness;
- a matcher too narrow catches nothing and becomes a law that cannot fail for the reason it claims — the exact defect prd-43 exists to prevent, recreated inside prd-43's own enforcement.

**Enumerate the productions as a table before writing the instrument**, with a verdict per row, including the forms the table's author does not intend to catch and why. A form not listed is a form not reviewed. #236 spent six review rounds learning this on a smaller grammar; the table was available at round three.

Worth considering as the level-above move: rather than classifying every backticked token, invert it — sweep only tokens matching an identifier shape that **also appear as a declaration somewhere in the corpus's history**, so the law asks "was this ever a symbol here, and is it still?" rather than "is this a symbol?".

## Definition of done

- A symbol cited from an in-scope document or comment either resolves in source, or is named in an explicit, reasoned exception — not silently skipped.
- The law reddens on `8f7a60ec`'s rename. That commit is the regression fixture: check out its tree, run the law, watch it fail. A law that passes there has not closed this gap.
- The productions the matcher can receive are enumerated **in the file**, with a verdict per row, including the ones deliberately out of scope.
- The exclusions `doc-citation-law` already reasons about are inherited or explicitly re-decided: `docs/research/`, `docs/review/` and `docs/prds/**` are excluded as *citing sources*, and a `**Tree:** <ref> at <sha>` pin exempts a dated run record when the sha landed. A symbol law that re-litigates those will fire on every historical artefact in the corpus.
- The law's own assertions are load-bearing: deleting the real-source check must redden something. (#236's `expect.assertions(n)` is the worked example — its absence was a blocking finding there.)

## Sibling case

**`doc-citation-law.test.ts` itself.** If the new law lives beside it rather than inside it, the two extractors are two spellings of "what a document cites", and they will drift. Decide deliberately whether this is a new file or a second axis in the existing one, and say why in the doc comment.

The second sibling is **line/anchor suffixes**, which the path law handles (`` `foo.ts:111` ``) and which a symbol law will meet as `` `withLabCliLock()` ``, `` `Foo.bar` `` and `` `Type<T>` ``.

The named one: **check out `8f7a60ec` and run the law.** If it passes, the gap is open and the law is decorative.

The weaker probe worth having: rename any constant currently cited by a tracked document and confirm the law names the document, the symbol and the file it expected to find it in — a failure that says only "a symbol is missing" costs the next reader the search this law exists to save.

## Milestone note

Filed to `charter-laws` on the operator's instruction. Being precise about the fit, since `AGENTS.md` asks for it: the rule this enforces is **prd-43 ruling 1**, and prd-43 has **shipped with zero open issues** — it is not a design-charter §-section rule, so this is the milestone's broader "a law's coverage is narrower than its claim" sense, as #311, #353 and #358 already are. This would be its sixth issue. `AGENTS.md` says that if the milestone ever holds a *wave*, that is the signal it has become a PRD and should be written as one — worth watching rather than acting on today.

ciaran-slow:
Closes #369

`prd-43 ruling 1`'s path law cannot see a cited **symbol**, structurally: its extractor
requires a `packages/`/`scripts/`/`docs/` prefix and a slash, so a backticked constant or
type cannot match it. `8f7a60ec` (#324) renamed `MAX_ARMS` to `LAUNCH_CEILING_LANES`; the
design note went on saying `MAX_ARMS`, the law swept it and passed *correctly*, the stale
prose was read into an issue's Definition of done, executed faithfully by a lane, and
caught by a human on PR #368 after the review round had been paid for.

One commit, two files: the law as a third `describe` in
`packages/server/src/doc-citation-law.test.ts`, and `.symbol-citation-baseline` at the
repo root.

## Condition 1 — what was read, by whom, and the part that is NOT met

Per AGENTS.md this belongs in the body where a reviewer sees it, not recorded away.

| commit / delta | read by an independent pass? |
|---|---|
| `5850b2b8` — the law as first built | **YES.** Verify skill, Sonnet 5, against an Opus 5 build. Record: [issue comment](https://github.com/launchpad-26/rhizomorph/issues/369#issuecomment-5628540580) → verdict and findings in [this comment](https://github.com/launchpad-26/rhizomorph/issues/369#issuecomment-5628774239), full report in `VERIFY-369.md` (left untracked in the lane) |
| `5850b2b8` → `dc1554d5` — the two review fixes | **NO.** Authored by me in response to the findings |
| `dc1554d5` → `32ebf823` — rebase onto `74d4c984` | **NO.** Clean rebase, no conflicts |
| `32ebf823` — as it stood at review round 2 | **YES.** Independent pass by @gabriel-canaan, Claude Opus 5, [review of `32ebf823`](https://github.com/launchpad-26/rhizomorph/pull/440#pullrequestreview-5188238611) — commit-by-commit, in a detached worktree with `origin/main` merged in. Verdict **changes requested**, four findings plus one offered call. That review says of itself what this table says: single-model, not the multi-seat verify, and it read `32ebf823` merged with `aaf9a50c` rather than the branch tree alone |
| `32ebf823` → `62a1f561` — rebase onto `aaf9a50c`, then the five review answers | **NO.** Authored by me in response to the findings. Clean rebase, no conflicts |
| `62a1f561` — read end to end, twice | **YES.** Round 3: two independent passes, Claude Opus 5 and Claude Sonnet 5, neither given the author's reasoning. **They disagreed** — Sonnet approved, Opus requested changes with three findings (a trailing `#` in any shell script silenced a real violation; the `# measured:` stamp was unverified prose; `[not-ours]`'s `citedBy` was unchecked). All three reproduced before being believed, all three fixed. [Full record](https://github.com/launchpad-26/rhizomorph/pull/440#pullrequestreview-5188940858) |
| `62a1f561` → `373365bc` — the three round-3 fixes | **YES.** Round 4: two delta-only passes, Opus and Sonnet, **both approved**, all three fixes certified by mutation. The Opus seat returned two prose corrections — my over-strip safety claim was true of the sweep and false of the three honesty tests, and "no live instance" needed a measurement. Both reproduced and taken. [Full record](https://github.com/launchpad-26/rhizomorph/pull/440#pullrequestreview-5189024359) |
| `373365bc` → `de2e618b` — those two prose corrections | **NO.** Authored by me. **Prose only**: two comment blocks and one input-table row, no behaviour change, so round 4's mutation evidence is unaffected by it |

**The gap, stated plainly rather than recorded away.** `de2e618b` has been read end to end
by nobody but its author. Five independent passes across four rounds read the two shas
before it, and every one of the eight findings they produced is closed and
mutation-backed — but a force-pushed fix re-opens condition 1 rather than inheriting it,
and the last push was a fix.

What is unread is **two comment blocks and one table row**. No behaviour changed, so the
mutation evidence round 4 certified still describes this tree. That is the smallest the
delta has been in four rounds, and it is the basis on which the operator undrafted this on
2026-09-13 — a decision this section exists to make visible, not to pretend away. A
reviewer who wants to close it properly reads `373365bc..de2e618b`, which is a five-minute
job.

**Where the process stopped paying.** Rounds 2 and 3 produced seven real defects, several
of them silencing. Round 4 produced two prose corrections. A fifth pass over prose is where
the regress stops earning its queue time.

Condition 2 is met: the operator asked for this PR in session on 2026-09-11, asked for the
review answers on 2026-09-13, and asked for it to be undrafted on 2026-09-13.

## Review round 2 — every finding answered

| finding | answer |
|---|---|
| **1. BLOCKING** — the law is red on the merged tree, and the baseline has no category for a citation that names a removed symbol *on purpose* | **FIXED, and ruled first.** Reproduced: one violation, `docs/roadmap.md` citing `NOT_YET_SWEPT`. The reviewer is right that `[renamed-away]` is the wrong home — its own header calls its entries debt — so a third **`[historical]`** section carries it: same mechanical checks, opposite meaning, pinned at **1** so every addition is a decision rather than a chore. The ruling is written into the law's docblock and the baseline header, both citing the path law's "cited from history, on purpose … that mechanism does not exist yet" as the gap this closes. A new input-table row declares the form, and a pinned-counts test asserts the two sections are disjoint on `(file, symbol)` |
| **2.** arm 2's history scope narrower than arm 1's source scope — 145 names | **FIXED, the widening direction.** Arm 2's pathspec is now `SOURCE_SWEEP_PATTERNS` **itself**, not a second list that happens to agree — the same "one list, so the two can never disagree" argument `DECLARATION_RES` already makes. Re-measured independently: **zero of the 72 `[not-ours]` entries flip**, reproducing the reviewer's number. The reviewer's shell-noise caveat is real and is why the fix is to share the constant rather than to retype a wider one: the noise (`die`, `try`, `print`) is not identifier-shaped under `SYMBOL_SHAPE_RES`, and if that ever changes it changes for both arms at once. The violation message no longer says "never declared" flatly. A `CONTROL` test on `setClip` (a `.mjs` const) and `summaryAndExit` (a shell function) reddens if the pathspec narrows again |
| **3.** the memoisation test passes only because of test ordering | **FIXED.** Primed first, then measured. Verified the way the reviewer found it: `-t 'both token sets are built exactly once'` alone **reddened before, passes now** — and breaking the memo with the prime in place still catches it, so the fix removed the fragility without removing the test's teeth |
| **4.** the `# measured:` stamp names a tree it was not verified against | **FIXED, re-derived rather than retyped.** `2026-09-13 tree=aaf9a50c candidates=2187 renamed-away=40 historical=1 not-ours=72`, with a comment recording what the old stamp said and why that mattered. The arithmetic closes: **34** distinct renamed-away symbols + **1** historical + **72** not-ours = **107** absent, which is the probe's own count |
| *offered, not asserted* — `stripCommentsFromSource` has no string-awareness | **TAKEN, as a declared limitation rather than a fix** — the verdict the path law next door already reached for the mirror case. The harm here is the worse of the two directions: a symbol declared *after* a `//`-bearing string literal drops out of arm 1's token set, so a citation of it reads absent and reddens as a violation that is not one. Closing it needs a string-literal-aware tokenizer, categorically bigger than the regex extractor this file deliberately is. The docblock claims to list every form, so the form is now listed, with a `CONTROL` test pinning the behaviour. No live instance — the corpus is green |

Two of the reviewer's observations are recorded here because they were **right and cost
nothing to confirm**: the "lands green" claim was true when it was made (`715f0a21` is not
an ancestor of `74d4c984`, so the citation arrived afterwards), and the body's old
496/9232 suite figures described a tree that no longer exists. Both are now restated
against `62a1f561` below.

## The ruling — why the obvious matcher is the wrong one

Measured before anything was written. Of **2,092** identifier-shaped candidates in scope,
**106** are absent from source — and **72 of those were never this repo's symbols**:
`shadowBlur`, `URIError`, `OTLP_ENDPOINT`, `isolatedModules`, and the `userId`/`orgUuid`
fields belonging to **Claude Code's own session files**. A 72-entry allowlist of things
that were never ours is a law allowlisted into uselessness.

So the question is inverted, as the issue itself suggested: not *"is this a symbol?"* but
**"was this ever a symbol *here*, and is it still?"** Three arms — present in non-comment
source; else a declaration of that name was once added under `packages/`, which is a
**violation**; else not ours.

**Arm 2 matches a DECLARATION, never a raw pickaxe**, and that is the finding that decided
the design. A `git log -S` pickaxe reports hits for `canonicalizeUnderRoot` (2),
`pathIsContained` (1), `humanizeTime` (1) and `RAW_COLLECTORS` (1) — and *every one* of
those hits is inside a comment, a hypothetical spelling this repo was warning itself
about. A pickaxe arm would turn four of this repo's own cautionary comments into
violations.

Lands **green** with 40 expected violations over 34 distinct symbols, each naming the
commit that removed it, **1** cited-on-purpose entry, and 72 pinned not-ours entries so
arm 3 cannot widen in silence.
`docs/architecture.md` carries 14 of the 40 — **correcting the documents is follow-up;
this is the instrument.**

## What is the sibling case?

**`doc-citation-law.test.ts` itself** — the issue's own question. Answered by living *in*
that file: the symbol law calls `sweepFiles`, `isExcludedCitingFile`, `isPinnedArtefact`,
`stripFencedCodeBlocks`, `extractComments`, `readSweptFile`, `OWN_FILE` and
`repoIsShallow` literally. Scope — which files a citation can come from — is where two
separate files would have drifted; the resolver is genuinely different, which is why it is
a separate law. The citation-ceiling law's docblock already makes that argument for itself.

**The second sibling is the law's own data, and it nearly shipped broken.** Arm 1 reads
string literals deliberately (an env var such as `RHIZOMORPH_DATA_DIR` appears in source
nowhere else), so the list of missing symbols — held as a TS const, as the plan
prescribed — would have put every one of those names into arm 1's own token set and
resolved each against itself. Both honesty tests would have passed vacuously. Proven with
a scratch before changing anything. Hence the data file, copying `.citation-prior-tracker`
one law over.

Excluding `OWN_FILE` wholesale was the obvious fix and is **wrong**: measured, it invents
three fresh violations against `CITATION_RE`, `EXCLUDED_DIRS` and `stripCitationSuffix` —
real symbols declared there and cited from `remedy-reachability-law.test.ts` and
`manifest-law.test.ts`. So `OWN_FILE` contributes only what it *declares*.

**Nineteen** run across the two rounds, each reverted, **baseline green first** — an
unquoted-glob harness under zsh reads every mutation as reddening, so the green baseline
is the control. Round 1's nine:

| mutation | result |
|---|---|
| arm 2 returns an empty set | **5 red** |
| arm 1's `OWN_FILE` carve-out removed | **6 red** |
| arm 1 stops stripping comments | **6 red** |
| shallow-clone refusal removed | **1 red** |
| trailing `()` no longer stripped | **2 red** |
| drop the `MAX_ARMS` `[renamed-away]` entry | **2 red** |
| drop a `[not-ours]` entry (`shadowBlur`) | **2 red** |
| add a fabricated `[renamed-away]` entry that still resolves | **4 red** |
| revert the `export default` widening | **1 red** |

Round 2's ten, over the four fixes — every one against a re-confirmed green control:

| mutation | result |
|---|---|
| arm 2's pathspec reverted to `packages/*.ts`, `*.tsx` | **1 red** |
| the `[historical]` row dropped | **2 red** |
| that row pointed at a symbol that still resolves | **2 red** |
| the same pair filed in **both** `[renamed-away]` and `[historical]` | **1 red** |
| the violation loop stops honouring `[historical]` | **1 red** |
| the memo prime removed, test run **alone** | **1 red** |
| the memo itself broken, prime left in place | **1 red** |
| the comment strip made string-aware | **4 red** |

The last two are the ones worth reading together. The memo fix could have been made
vacuous — priming a test into passing is exactly the shape `AGENTS.md` warns about — so it
is mutated in **both** directions: remove the prime and it reddens alone, break the memo
and it still reddens primed. The string-awareness mutation reddens the `CONTROL` that
declares the limitation, which is what makes that row a pin rather than a paragraph.

The named one from the issue — *the law reddens on `8f7a60ec`'s rename* — is a test, not a
tree checkout: it asserts the sha is a landed ancestor, that
`git grep -l MAX_ARMS 8f7a60ec~1` names `packages/server/src/api/lab.ts`, that no source
carries the name today, that history *declared* it, and that the verdict is
`renamed-away`. Deleting either arm breaks it.

## Fence widening

The issue was filed deliberately un-fenced, and the plan fenced exactly one file:
`packages/server/src/doc-citation-law.test.ts`. **This PR touches two** — the second is
the new root file `.symbol-citation-baseline`, forced by the self-certification problem
above.

Recorded here rather than before the change, because there was no PR to record it on at
the time. Flagging the ordering rather than quietly meeting the letter of the rule.

Round 2 widened nothing further: the four fixes land in the same two files.

Footprint checked rather than assumed: five other files mention `doc-citation-law`, and
none asserts a count, a line number or a literal from it — `windows-suite-law.test.ts`
uses it as a "real tracked file" fixture that need only keep existing.

## Gates — all EXECUTED on `de2e618b`

| gate | result |
|---|---|
| `npm run build` | OK |
| `VITEST_MAX_WORKERS=6 npm test` | 501 files, **9436 passed**, 3 expected fail, 7 skipped, **0 failed** |
| `npm run typecheck` | OK |
| `npm run lint` | 1067 files, no findings |
| `node scripts/packaging-guard.mjs` | passed — 65 files, all allowlisted |
| `bash scripts/boot-smoke.sh` | passed — `/api/meta` 200, `/` 200 `text/html`, clean shutdown |

The last two are CI's own scripts rather than a hand-rolled imitation of them, which is
also the answer to "did the root file leak into the tarball": neither baseline file is
among the 65.

**`scripts/ci-local.sh --pr 440` is green on `de2e618b`** — all eight legs including the
lockfile-honest `npm ci`, published as a `local-ci (Darwin)` **commit status pinned to the
head sha**, which is the mark to read; the `Passed local CI` label is repo-level and
survives a push that invalidates it.

Two things that leg taught, kept because both are reusable: it **refuses to run on a dirty
tree**, since a verdict pinned to a sha that is not what was tested is worse than none. And
it went **red on `npm ci` under node v24.12.0** — `jsdom@30` supports `^22.22.2 ||
^24.15.0 || >=26.0.0`, and the default node here sits in that gap. That was the machine,
not the change; re-run on **v22.23.2** it is green. `npm install`, which the lane had been
using throughout, tolerates the mismatch silently.

**Under load**, because this file is test-heavy and the landing gate runs suites ~4× at
once: four concurrent runs of this file at `--maxWorkers=5`, **all green, ~5 s wall**. No
`// @gate-timing` marker — this law makes no wall-clock assertion.

**One platform.** These are macOS only. GitHub Actions is disabled on this repo, so
`62a1f561` carries **no check runs at all** — the ten green legs a reader may remember
belong to `32ebf823`, a tree that no longer exists. A reviewer wanting the leg run before
merge runs `scripts/ci-local.sh --pr 440` and reads the **commit status** pinned to the
head sha, not the repo-level label.

**What the rebase cost, honestly.** `main` moved **72 commits** (`74d4c984` → `aaf9a50c`)
and this is the round where that mattered: one genuinely new citation arrived in
`docs/roadmap.md` and the law caught it, which is the instrument working rather than the
instrument being brittle. Both original lists — 40 renamed-away, 72 not-ours — came
through **unchanged**; the only addition is the one `[historical]` row that new citation
earned.

## Three things for the reviewer

- **The plan and the build came from the same session**, at the operator's direction. The
  pipeline assumes they do not. Four deviations from the plan were caught and are written
  up on the issue; a cold builder would likely have surfaced more, and that is the class
  of defect most worth hunting here.
- **`TurnExtraction` / `TranscriptDialect`** were flagged as possibly a real doc defect the
  law cannot see. The verify pass read `docs/adr/0017-...` directly and found both are
  documented *rejected* names, never built — so the `[not-ours]` pin is correct. Recorded
  because the question was raised in the open, not because it is outstanding.
- **The `[historical]` section is the part to attack next.** It is a new place for a
  violation to go and be quiet, which is the shape every allowlist starts as. Three things
  hold it: the count is pinned at 1, so growth is a deliberate edit; each entry is
  re-checked both ways (the file must still cite it, the symbol must still be absent); and
  it cannot overlap `[renamed-away]`. What it does **not** have is any test that the
  *reason* is true — a row's justification is prose, and only a human reads it. That is
  the same tolerance `[renamed-away]`'s reason column already carries, but it is worth
  saying out loud rather than discovering later.

No ADR and no `CHANGELOG.md` entry: nothing user-visible changes, and the path law shipped
the same way in `174afc05`.


gabriel-canaan (changes requested):
## Review of `32ebf823` — the independent pass the body asks for

Read commit-by-commit (one commit), against #369, in a detached worktree with `origin/main`
merged in. Everything below is **EXECUTED** unless a line says REASONED.

**Base moved under this PR: the branch is 72 commits behind `main`** (`74d4c984` → `aaf9a50c`).
That is the whole shape of this review — see finding 1.

### What verified, in the author's favour

- **The design holds up where I tried hardest to break it.** The arm-2-is-a-declaration-not-a-pickaxe
  ruling is right, and the `CONTROL` test for it is real: `canonicalizeUnderRoot` has pickaxe hits
  and still resolves `not-ours`.
- **The self-certification analysis is correct and the fix is the right one.** I re-derived the
  `OWN_FILE` claim: excluding the file wholesale does invent violations against its own real
  exports, and the declare-only carve-out is the narrower correct instrument.
- **`expect.assertions(n)` is on both honesty tests**, which is the thing #236 was blocked for.
- **The "lands green" claim was TRUE when it was made.** The violation in finding 1 is not the
  author's arithmetic going wrong — `715f0a21` (2026-09-11, `docs/prd30`) is not an ancestor of
  `74d4c984`. It arrived with main's 72 commits, *after* the measurement.
- Gates on the **post-merge** tree: `typecheck` OK; `lint` 1067 files, no findings; suite
  **501 files, 9434 passed, 3 expected fail, 3 skipped — 1 failed**. The body's 496/9232 was a
  tree that no longer exists; both numbers are honest, the second is just later.
- `gh pr checks 440`: all **10 legs** green across `build-test-boot`, `pack-smoke` and
  `windows-suite` — the full three-job checklist, not one job.

---

### 1. BLOCKING — the law is RED on the merged tree, and the baseline has no category for what it caught

EXECUTED, post-merge:

```
FAIL  doc symbol law > every symbol cited from an in-scope document or comment still exists
docs/roadmap.md cites `NOT_YET_SWEPT`, which this repo declared once and no longer has
```

`b33da639` (#389) deleted `NOT_YET_SWEPT`; that deletion was already in this branch's base.
What arrived with main's 72 commits is the *citation*: `715f0a21` added to `docs/roadmap.md:224`

> the `title=` sweep is finished in both halves — #220's fifty and #389's last twelve, with
> `NOT_YET_SWEPT` deleted rather than emptied.

**That sentence is correct, and it is correct *because* it names a removed symbol.** So the
interesting part is not the red bar, it is that the only home the baseline offers is
`[renamed-away]`, whose own header says *"EXPECTED VIOLATIONS … Removing entries is the
destination — correcting the documents is follow-up work."* Filing a true sentence there
records it as debt against a document that has nothing to correct.

This is not a new category for this repo — it is the one the path law already ruled on, in this
same file's docblock, as its reason for **not** widening to `.mjs`:

> Before widening, the law needs a way to say "cited from history, on purpose" so a comment like
> that one can opt out — **that mechanism does not exist yet**, so the scope stays exactly
> ruling 1's.

The symbol law widened into that gap, and the first deliberately-historical citation showed up
within 72 commits. Worth a ruling on the issue before this lands, not a silent `[renamed-away]`
row. Two defensible answers — a third section (`[historical]`, cited-on-purpose, never expected
to be removed), or `[renamed-away]` plus a reason string that says so explicitly — but the choice
should be recorded, because the next reader of that row will otherwise go and "fix" a true sentence.

Whichever you pick, the mechanical part is: entry added, `renamedAway.length` 40 → 41, and the
`# measured:` stamp re-derived.

### 2. Arm 2's history scope is narrower than arm 1's source scope — 145 names (latent, same shape as the `export default` gap)

EXECUTED. Arm 1 sweeps `packages/*.ts`, `*.tsx`, `*.mjs` **and** `scripts/*`. Arm 2's history pass
is `['--', 'packages/*.ts', 'packages/*.tsx']` only.

```
declared-in-history under packages/*.ts|tsx (arm 2 scope): 13752
declared-in-history under packages/*.mjs|scripts/*      :   360
names arm 2 can NEVER see                               :   145
```

Among them `EVIDENCE_RE`, `trackedViolations`, `measuredCount`, `summaryAndExit`
(`scripts/windows-triage.sh`), `interactPath`, `setClip`, `capturePath`
(`scripts/dev/visit.mjs`). Arm 1 counts all of these as *ours and present* today; arm 2 cannot
know they were ever ours.

**Consequence is narrower than it first looks, and I want to be exact about it** — a new
`not-ours` verdict is *not* silent, it still reddens. So the defect is a **wrong diagnosis with a
harmful instructed remedy**: remove one of those declarations and the law says *"which this repo
never declared — add it to `[not-ours]`"*, which is false, and following it pins a genuinely
renamed-away symbol into the never-ours list permanently. That is the silencing, one step later.

This is structurally the same finding as the `export default` gap already fixed in this commit —
a name this repo really declared reading as not-ours — one scope dimension over.

**Blast radius measured, the same way you measured the `export default` widening:** adding
`packages/*.mjs` and `scripts/*` to arm 2's pathspec flips **0 of the 72** `[not-ours]` pins.
So the widening is safe on today's tree. Caveat worth your judgement, not mine: `scripts/*` is
mostly shell, and `DECLARATION_RES`'s last row (`name(`) reads shell function definitions — the
360 includes `die`, `try`, `finally`, `print`. None is identifier-shaped under `SYMBOL_SHAPE_RES`,
which is why the flip count is 0, but it is noise entering a set that only ever grows.

Your call between widening arm 2 and narrowing arm 1 — but the violation *message* should stop
saying "never declared" flatly either way, since the data file's own header is already more
careful ("never declared in `packages/` history").

### 3. The memoisation test passes only because of test ordering

EXECUTED:

```
$ npx vitest run packages/server/src/doc-citation-law.test.ts -t 'both token sets are built exactly once'
AssertionError: arm 1 rebuilt — the memo is broken and this law costs 4x under the load probe:
  expected 1 to be +0
 Tests  1 failed | 65 skipped (66)
```

It captures `sourceTokensBuilds` *before* touching `symbolVerdict`, so `delta === 0` only holds if
an earlier test in the file already warmed both caches. Run alone — `-t`, `.only`, or any
reordering — it measures the first build and reddens for a reason that has nothing to do with the
memo.

It does still catch the real mutation (break the memo and the delta is 6, not 1), so this is
fragility rather than vacuity. One-line fix: prime first, then measure.

```js
symbolVerdict('MAX_ARMS')                       // prime both caches
const sourceBuildsBefore = sourceTokensBuilds
const historyBuildsBefore = everDeclaredBuilds
```

### 4. The `# measured:` stamp names a tree the baseline was not last verified against

`.symbol-citation-baseline:40` reads `measured: 2026-09-11 tree=a3c5b305`. The commit was rebased
onto `74d4c984` and the body states the lists were re-verified there — so the stamp is one tree
stale already, and finding 1 makes it two. Cosmetic, except that this is a stale provenance claim
inside the artefact built to catch stale claims, and you already fixed the sibling of it in this
same commit (the "30 distinct symbols" → 34).

---

`stripCommentsFromSource` strips `//` to end-of-line with no string awareness, so a URL in a string
literal (`'https://…/FooBar'`) truncates the rest of that line out of arm 1's token set. REASONED,
no live instance found. The path law's input table already declares this exact limitation for
`extractComments` ("review of #186 item 2") and pins it with a CONTROL test rather than fixing it —
the same declaration here would close the gap in the table, given the docblock claims to list every
form.

---

**Verdict: changes requested.** Finding 1 blocks — not for the red bar, which is a one-row fix, but
because the row needs a ruling first. 2 and 3 are worth taking in the same push. The body's own
statement that condition 1 is unmet is accurate and this review does not discharge it: this is a
single-model pass (Claude Opus 5), not the multi-seat verify, and it read `32ebf823` merged with
`aaf9a50c` rather than the branch tree alone.

gabriel-canaan (dismissed):
## Review of `32ebf823` — the independent pass the body asks for

Read commit-by-commit (one commit), against #369, in a detached worktree with `origin/main`
merged in. Everything below is **EXECUTED** unless a line says REASONED.

**Base moved under this PR: the branch is 72 commits behind `main`** (`74d4c984` → `aaf9a50c`).
That is the whole shape of this review — see finding 1.

### What verified, in the author's favour

- **The design holds up where I tried hardest to break it.** The arm-2-is-a-declaration-not-a-pickaxe
  ruling is right, and the `CONTROL` test for it is real: `canonicalizeUnderRoot` has pickaxe hits
  and still resolves `not-ours`.
- **The self-certification analysis is correct and the fix is the right one.** I re-derived the
  `OWN_FILE` claim: excluding the file wholesale does invent violations against its own real
  exports, and the declare-only carve-out is the narrower correct instrument.
- **`expect.assertions(n)` is on both honesty tests**, which is the thing #236 was blocked for.
- **The "lands green" claim was TRUE when it was made.** The violation in finding 1 is not the
  author's arithmetic going wrong — `715f0a21` (2026-09-11, `docs/prd30`) is not an ancestor of
  `74d4c984`. It arrived with main's 72 commits, *after* the measurement.
- Gates on the **post-merge** tree: `typecheck` OK; `lint` 1067 files, no findings; suite
  **501 files, 9434 passed, 3 expected fail, 3 skipped — 1 failed**. The body's 496/9232 was a
  tree that no longer exists; both numbers are honest, the second is just later.
- `gh pr checks 440`: all **10 legs** green across `build-test-boot`, `pack-smoke` and
  `windows-suite` — the full three-job checklist, not one job.

---

### 1. BLOCKING — the law is RED on the merged tree, and the baseline has no category for what it caught

EXECUTED, post-merge:

```
FAIL  doc symbol law > every symbol cited from an in-scope document or comment still exists
docs/roadmap.md cites `NOT_YET_SWEPT`, which this repo declared once and no longer has
```

`b33da639` (#389) deleted `NOT_YET_SWEPT`; that deletion was already in this branch's base.
What arrived with main's 72 commits is the *citation*: `715f0a21` added to `docs/roadmap.md:224`

> the `title=` sweep is finished in both halves — #220's fifty and #389's last twelve, with
> `NOT_YET_SWEPT` deleted rather than emptied.

**That sentence is correct, and it is correct *because* it names a removed symbol.** So the
interesting part is not the red bar, it is that the only home the baseline offers is
`[renamed-away]`, whose own header says *"EXPECTED VIOLATIONS … Removing entries is the
destination — correcting the documents is follow-up work."* Filing a true sentence there
records it as debt against a document that has nothing to correct.

This is not a new category for this repo — it is the one the path law already ruled on, in this
same file's docblock, as its reason for **not** widening to `.mjs`:

> Before widening, the law needs a way to say "cited from history, on purpose" so a comment like
> that one can opt out — **that mechanism does not exist yet**, so the scope stays exactly
> ruling 1's.

The symbol law widened into that gap, and the first deliberately-historical citation showed up
within 72 commits. Worth a ruling on the issue before this lands, not a silent `[renamed-away]`
row. Two defensible answers — a third section (`[historical]`, cited-on-purpose, never expected
to be removed), or `[renamed-away]` plus a reason string that says so explicitly — but the choice
should be recorded, because the next reader of that row will otherwise go and "fix" a true sentence.

Whichever you pick, the mechanical part is: entry added, `renamedAway.length` 40 → 41, and the
`# measured:` stamp re-derived.

### 2. Arm 2's history scope is narrower than arm 1's source scope — 145 names (latent, same shape as the `export default` gap)

EXECUTED. Arm 1 sweeps `packages/*.ts`, `*.tsx`, `*.mjs` **and** `scripts/*`. Arm 2's history pass
is `['--', 'packages/*.ts', 'packages/*.tsx']` only.

```
declared-in-history under packages/*.ts|tsx (arm 2 scope): 13752
declared-in-history under packages/*.mjs|scripts/*      :   360
names arm 2 can NEVER see                               :   145
```

Among them `EVIDENCE_RE`, `trackedViolations`, `measuredCount`, `summaryAndExit`
(`scripts/windows-triage.sh`), `interactPath`, `setClip`, `capturePath`
(`scripts/dev/visit.mjs`). Arm 1 counts all of these as *ours and present* today; arm 2 cannot
know they were ever ours.

**Consequence is narrower than it first looks, and I want to be exact about it** — a new
`not-ours` verdict is *not* silent, it still reddens. So the defect is a **wrong diagnosis with a
harmful instructed remedy**: remove one of those declarations and the law says *"which this repo
never declared — add it to `[not-ours]`"*, which is false, and following it pins a genuinely
renamed-away symbol into the never-ours list permanently. That is the silencing, one step later.

This is structurally the same finding as the `export default` gap already fixed in this commit —
a name this repo really declared reading as not-ours — one scope dimension over.

**Blast radius measured, the same way you measured the `export default` widening:** adding
`packages/*.mjs` and `scripts/*` to arm 2's pathspec flips **0 of the 72** `[not-ours]` pins.
So the widening is safe on today's tree. Caveat worth your judgement, not mine: `scripts/*` is
mostly shell, and `DECLARATION_RES`'s last row (`name(`) reads shell function definitions — the
360 includes `die`, `try`, `finally`, `print`. None is identifier-shaped under `SYMBOL_SHAPE_RES`,
which is why the flip count is 0, but it is noise entering a set that only ever grows.

Your call between widening arm 2 and narrowing arm 1 — but the violation *message* should stop
saying "never declared" flatly either way, since the data file's own header is already more
careful ("never declared in `packages/` history").

### 3. The memoisation test passes only because of test ordering

EXECUTED:

```
$ npx vitest run packages/server/src/doc-citation-law.test.ts -t 'both token sets are built exactly once'
AssertionError: arm 1 rebuilt — the memo is broken and this law costs 4x under the load probe:
  expected 1 to be +0
 Tests  1 failed | 65 skipped (66)
```

It captures `sourceTokensBuilds` *before* touching `symbolVerdict`, so `delta === 0` only holds if
an earlier test in the file already warmed both caches. Run alone — `-t`, `.only`, or any
reordering — it measures the first build and reddens for a reason that has nothing to do with the
memo.

It does still catch the real mutation (break the memo and the delta is 6, not 1), so this is
fragility rather than vacuity. One-line fix: prime first, then measure.

```js
symbolVerdict('MAX_ARMS')                       // prime both caches
const sourceBuildsBefore = sourceTokensBuilds
const historyBuildsBefore = everDeclaredBuilds
```

### 4. The `# measured:` stamp names a tree the baseline was not last verified against

`.symbol-citation-baseline:40` reads `measured: 2026-09-11 tree=a3c5b305`. The commit was rebased
onto `74d4c984` and the body states the lists were re-verified there — so the stamp is one tree
stale already, and finding 1 makes it two. Cosmetic, except that this is a stale provenance claim
inside the artefact built to catch stale claims, and you already fixed the sibling of it in this
same commit (the "30 distinct symbols" → 34).

---

`stripCommentsFromSource` strips `//` to end-of-line with no string awareness, so a URL in a string
literal (`'https://…/FooBar'`) truncates the rest of that line out of arm 1's token set. REASONED,
no live instance found. The path law's input table already declares this exact limitation for
`extractComments` ("review of #186 item 2") and pins it with a CONTROL test rather than fixing it —
the same declaration here would close the gap in the table, given the docblock claims to list every
form.

---

**Verdict: changes requested.** Finding 1 blocks — not for the red bar, which is a one-row fix, but
because the row needs a ruling first. 2 and 3 are worth taking in the same push. The body's own
statement that condition 1 is unmet is accurate and this review does not discharge it: this is a
single-model pass (Claude Opus 5), not the multi-seat verify, and it read `32ebf823` merged with
`aaf9a50c` rather than the branch tree alone.

ciaran-slow (commented):
## Answering the round-2 review — all five taken, on `62a1f561`

Thank you for this one in particular: three of the four would have shipped, and finding 1
is a ruling I would not have reached from inside the change. Everything below is
**EXECUTED** unless it says REASONED. Each answer names what changed; the commit message
carries the long form.

### 1 — FIXED, and ruled before the row was written

Reproduced exactly as reported, on the rebased tree:

```
FAIL  every symbol cited from an in-scope document or comment still exists
docs/roadmap.md cites `NOT_YET_SWEPT`, which this repo declared once and no longer has
```

And I confirmed the part that makes it a ruling rather than a chore: `715f0a21` is **not**
an ancestor of `74d4c984`, so the citation arrived with main's 72 commits — the "lands
green" claim was true when made, and false by the time you read it. `NOT_YET_SWEPT` is
gone from source; the sentence naming it is correct.

Your reading of `[renamed-away]` is right and it is decisive: its own header says removing
entries is the destination, so a true sentence filed there is debt against a document with
nothing to correct. Taken your **first** option — a third **`[historical]`** section:
cited-on-purpose, never expected to be removed, mechanically checked the same way (the
file must still cite it; the symbol must still be absent). Pinned at **1**, so every future
addition is a deliberate edit rather than a chore, and a new test asserts the two sections
are **disjoint** on `(file, symbol)` — a pair in both would make "is this fixed yet?"
unanswerable.

The ruling is recorded in three places a later reader can reach without this thread: the
law's docblock (as its own section), the baseline's header, and a new input-table row so
the form is declared. All three cite the path law's *"the law needs a way to say 'cited
from history, on purpose' … that mechanism does not exist yet"* as the gap being closed —
your observation that this law widened into that gap is the clearest framing of it, and I
have used your wording rather than inventing my own.

Mechanical part as you specified it: entry added; the stamp re-derived; `renamedAway.length`
deliberately **stays 40** rather than going to 41, since the row went to the new section.

### 2 — FIXED, the widening direction

Arm 2's pathspec is now `SOURCE_SWEEP_PATTERNS` **itself**, not a wider list retyped beside
it. That is the point of the fix: the same "one list, so the two can never disagree"
argument `DECLARATION_RES` already makes for the declaration shapes. Retyping a wider
pathspec would have closed today's gap and left the failure mode open.

I re-measured your blast radius independently rather than taking it: adding those paths
flips **zero of the 72** `[not-ours]` entries — the honesty test over that list stays green
with all 65 other tests passing. Your number reproduces.

Your shell-noise caveat is the reason I did not simply widen: `die`, `try`, `print` and
friends do enter the ever-declared set, and none is identifier-shaped under
`SYMBOL_SHAPE_RES` today. Sharing the constant means that if `SYMBOL_SHAPE_RES` ever
changes, it changes for both arms at once rather than one of them silently. Noted as a
tolerance rather than claimed as a non-issue.

A `CONTROL` test now pins the scope with one name per path shape the narrow list dropped —
`setClip` (a `const` in `scripts/dev/visit.mjs`) and `summaryAndExit` (a shell function in
`scripts/windows-triage.sh`), both declared there and never under `packages/*.ts|tsx`.
Reverting the pathspec reddens it. And the message no longer says "never declared" flatly,
as you asked — it now says "never declared in the paths arm 2 reads".

### 3 — FIXED, and mutated in both directions

Primed first, then measured, as you wrote it. Verified the way you found it:

```
$ npx vitest run packages/server/src/doc-citation-law.test.ts -t 'both token sets are built exactly once'
  Tests  1 passed | 66 skipped (67)     # was: 1 failed
```

Because "prime it until it passes" is the obvious wrong fix, it is mutated both ways:
remove the prime and run alone → **red**; break the memo with the prime in place → **red**.
Your point that it was fragility rather than vacuity is right, and the second mutation is
what keeps it that way.

### 4 — FIXED, re-derived rather than retyped

`# measured: 2026-09-13 tree=aaf9a50c candidates=2187 renamed-away=40 historical=1 not-ours=72`,
with a comment recording what the old stamp said and why a stale stamp inside this
particular artefact is worse than cosmetic. Re-derived from the tree, not adjusted by hand:
**2,187** candidates, **107** absent — and 34 distinct renamed-away symbols + 1 historical
+ 72 not-ours = 107, so the arithmetic closes.

### The offered call — TAKEN, as a declared limitation

You were right that the docblock claims to list every form, which makes an undeclared one a
gap in the table rather than a tolerable simplification. Declared, with the path law's
verdict for the mirror case and a `CONTROL` test pinning it — mutating
`stripCommentsFromSource` to be string-aware reddens four tests.

One correction to the direction of harm, in your favour: here it is worse than the path
law's version. `stripCommentsFromSource` **removes** comments, so a `//` inside a literal
drops the rest of that line from arm 1's token set — a symbol declared after such a literal
reads as *absent*, and a citation of it elsewhere reddens as a violation that is not one.
The path law's row loses coverage; this one would manufacture a false positive. Same fix
(a real tokenizer), same verdict (out of scope), but the row says which way it fails.

---

### Gates on `62a1f561`, all EXECUTED

| gate | result |
|---|---|
| `VITEST_MAX_WORKERS=6 npm test` | 501 files, **9434 passed**, 3 expected fail, 7 skipped, **0 failed** |
| `npm run typecheck` / `npm run lint` | OK / 1067 files, no findings |
| `npm run build` | OK |
| `node scripts/packaging-guard.mjs` | passed — 65 files, neither baseline file among them |
| `bash scripts/boot-smoke.sh` | passed — `/api/meta` 200, `/` 200 `text/html`, clean shutdown |
| 4× concurrent, `--maxWorkers=5` | all green, ~5 s wall |

**macOS only, and `62a1f561` carries no check runs** — Actions is disabled on this repo, so
the ten green legs you read belong to `32ebf823`, a tree that no longer exists. Nothing
here should be reported as "CI passed".

### Condition 1 is open again, and the PR body says so

This push re-opens it: `62a1f561` has been read end to end by nobody but its author. The
body's table now carries your review as the row it earns for `32ebf823`, and a row saying
the five answers are mine. Still a draft. If you would rather not re-read the whole thing,
the four fixes are separable — 2, 3 and 4 are small and mechanical; **finding 1 is the one
worth a second opinion**, because `[historical]` is a new place for a violation to go and
be quiet, and the thing it lacks is any test that a row's *reason* is true.

ciaran-slow (commented):
## Round 3 — two independent passes over `62a1f561`, and the three findings they produced

Condition 1 requires the record to live where a later reader can ask, so both passes are
recorded here rather than summarised away. **Neither seat was given the author's
reasoning**; both read the code, the commit, #369 and this PR. They were run because the
author of the round-2 answers cannot be the pass that checks them.

| seat | model | verdict |
|---|---|---|
| A | Claude Opus 5 | **changes requested** — F1 blocking, F2 and F3 non-blocking |
| B | Claude Sonnet 5 (cross-model: the code was authored by Opus) | **approve**, no blocking findings |

**They disagree, which is the useful outcome.** `AGENTS.md` warns that a unanimous verdict
from model seats is a tell; a split one, where the disagreement is legible, is worth more
than either verdict alone. Seat B attacked `[historical]` directly — relocating a real
`[renamed-away]` violation into it — and showed that hiding something there costs a row
move plus two pinned numbers plus a prose reason, *exactly* what hiding one in
`[renamed-away]` costs. That is the specific reassurance the section needed, and it is why
its approval is not just "found nothing".

**Nothing below was taken on a seat's word.** Each finding was reproduced here before it
was believed, and each fix was then mutated back. `AGENTS.md`: nothing from a seat may be
posted unverified.

### F1 — BLOCKING, and fixed. A trailing `#` comment in any shell script silenced a real violation

`buildSourceTokens` stripped `^\s*#` — **whole-line** comments only. A symbol named in a
*trailing* shell comment therefore entered arm 1's token set, and a document citing that
dead name resolved `present`.

Reproduced independently as an A/B, same doc line in both legs:

| tree | before the fix |
|---|---|
| `docs/roadmap.md` cites `` `SomeNeverDeclaredNameXyz` `` | **RED** |
| same, **plus** `scripts/zz-review-probe.sh` containing `true # SomeNeverDeclaredNameXyz` | **GREEN, 69 passed (69)** |

A real violation, silenced by a comment. It is the file's own named defect — *"arm 1 reads
string literals but NOT comments — the stale-pair defect, where a dead doc and a dead
comment certify each other"* — holding for the JS family and not for shell, and the input
table (which states its own rule, *a form not listed is a form not reviewed*) had **no row
for a shell source at all**.

**Fixed, not declared**, and the seat's reasoning for that is the part worth keeping: this
is the *third* instance of the class the `export default` gap and the arm-2 pathspec gap
belong to, and both of those were closed in-commit while still latent, with the docblock
arguing "closed here before it stopped being latent". Declaring the third would have been
the inconsistent choice. Zero live instances either way.

The over-strip direction is deliberate and now declared in the table: `#` inside a shell
string costs the rest of that line, which makes a symbol read **absent** and **redden** —
where under-stripping goes **silent**. For a regex extractor, noisy beats quiet.

After the fix, both legs of the A/B are red, and reverting to whole-line stripping reddens
the new `CONTROL`.

### F2 — the stamp had a grammar one law over, and not here

Round 2 answered the stale-stamp finding by **re-deriving the value**. The seat's point is
that the *mechanism* was not carried across: `.citation-prior-tracker`, 700 lines earlier
in this same file, has had `BASELINE_MEASURED_RE` and an assertion against its own content
since it was written. Reproduced — rewriting the stamp to

```
# measured: banana tree=deadbeef candidates=1 renamed-away=999 historical=999 not-ours=999
```

left the suite **green, 69 passed (69)**. Every number falsified, the date turned into
`banana`.

That is review finding 4's *class* rather than its instance, and the honest reading is that
round 2 fixed the instance and left the class open. Now: the three list counts are checked
against the file they describe, and the `tree=` sha must **resolve and be a landed
ancestor** — so a lane sha, which vanishes under squash-and-merge, cannot sit there.

`candidates` is deliberately left as prose, and the regex says why: it counts a corpus that
grows whenever anyone writes a document, so pinning it would redden this law on edits that
have nothing to do with it — that is a ceiling, and the ceiling law already owns that job.

### F3 — two of three sections checked their file field; the third took it on trust

`[not-ours]`'s `citedBy` was unverified. Reproduced, both mutations **green** before the
fix: pointing a row at `docs/this-file-does-not-exist.md`, and orphaning a row by deleting
the only citation. The section header calls that column *"a file that cites it"* — a claim
nothing held.

A sibling case sitting inside the same parser, which is this repo's most-cited defect
shape. Now asserted exactly as `[renamed-away]` and `[historical]` already did. **All 72
rows pass unchanged** — the list was honest; nothing was keeping it that way.

### Mutations — five that were GREEN before this commit are RED after it

Each re-run against a re-confirmed green control (the zsh word-splitting trap makes an
unverified control worthless):

| mutation | before | after |
|---|---|---|
| the garbage stamp above | green | **red** |
| counts right, `tree=` sha not landed | green | **red** |
| a single count off by one | green | **red** |
| `[not-ours]` row → missing file | green | **red** |
| `[not-ours]` row → real file that does not cite it | green | **red** |
| whole-line-only shell stripping (F1 reverted) | — | **red** |

### Gates on `373365bc`, all EXECUTED

`VITEST_MAX_WORKERS=6 npm test` → 501 files, **9436 passed**, 3 expected fail, 7 skipped,
**0 failed**. `typecheck` OK. `lint` 1067 files, no findings. Re-run on **node v22.23.2**,
which matters — see below.

### Two things the seats said about their own limits, kept rather than dropped

- **Seat A ran the packaging guard on an unbuilt tree** and got *"passed — 4 files"* against
  the real 65. That is the vacuous-pass case `AGENTS.md` names, and the seat flagged it as
  **no signal** rather than banking it. Correct call, and worth repeating: a guard that
  passes over an empty `dist/` has verified nothing.
- **Both seats were single-platform** (macOS, node v22.23.2), and seat A noted that
  `doc-citation-law.test.ts` is already in `.windows-known-failures`, so the new law's
  behaviour on native Windows is **witnessed by nothing**. The CRLF handling in the parser
  is correct and unexercised.

### One live corpus defect, offered and NOT taken here

Seat A found `packages/contract/src/instrument.contract.test.ts:115` naming
`` `MigrationResult` `` in a comment directly above the code that reads the real
`MigrationOutcome`. It is genuinely the rot #369 was filed about, currently misfiled in
`[not-ours]` as external. Arm 3 is mechanically right — a misspelled internal name is
indistinguishable from an external one by shape — so this is **not** a defect in the
instrument.

Deliberately not fixed here: it is a third file, outside this PR's two, and fixing it also
moves a pinned count. It wants its own issue rather than a third fence widening on a PR
already carrying one.

### Condition 1, stated plainly

These two passes read `62a1f561`. This push makes `373365bc`, and **a force-pushed review
fix re-opens condition 1 rather than inheriting it** — so the head is again author-read
only. The delta is three fixes in one file, each with a mutation in this comment. The
regress is real and I am not pretending it away: at some point a human decides the delta is
small enough, and that decision is the operator's, not mine.

ciaran-slow (commented):
## Round 4 — two delta-only passes over `62a1f561..373365bc`

Scoped deliberately: `62a1f561` was read end to end by two passes in round 3, so these two
read only the three fixes that answered them. Same split-seat arrangement.

| seat | model | verdict on the delta |
|---|---|---|
| A | Claude Opus 5 | **approve**, two non-blocking prose corrections |
| B | Claude Sonnet 5 (cross-model) | **approve**, no findings |

Both certified all three fixes **by mutation** rather than by reading the commit message —
each broke the thing and watched the named test redden. Seat A ran eight mutations of the
stamp alone, including the one that matters most: `tree=` pointed at an **unmerged lane
sha** reddens, which is the property the comment claims and the reason the check exists.
Seat A also confirmed `expect.assertions(notOurs.length * 3)` is exact — `* 3` passes and
`* 4` fails, so 216 assertions really run across 72 rows with none skipped.

### The seats disagreed, and the Opus seat was right

Seat B traced my over-strip safety argument and called it *structurally true*. Seat A
**demonstrated that it is only half true**, and that is a correction to a claim this
commit's own comments were making.

> "a dropped token can only make a symbol read ABSENT, which REDDENS"

True of the **sweep**, which asserts presence: a smaller token set there can only add
violations. **False of the three baseline honesty tests**, which assert a NON-present
verdict — absence is their passing condition, so over-stripping is the *quiet* direction.

Reproduced here before believing it, EXECUTED:

| `scripts/zz-probe.sh` contains | result |
|---|---|
| `echo "plain foldReplay"` | **RED** — the `[renamed-away]` honesty test fires, because that symbol has genuinely come back |
| `echo "a#b foldReplay"` | **GREEN, 71 passed (71)** — the same re-introduction, silenced |

Seat B reached the opposite conclusion by tracing only the sweep path. That is exactly why
the seats are split across models, and it is the second round running in which the
disagreement carried the signal rather than the agreement.

**The fix is kept**, and the trade is now stated in the code instead of glossed: what this
costs is a stale baseline **row** surviving — a symbol re-introduced only after a `#`
inside a shell string and nowhere in `packages/*.ts|tsx|mjs` — and what it buys is closing
a gap that silenced a real **citation**. The JS half has carried the identical property,
declared, since it was written.

### "No live instance" was imprecise, so it is now a measurement

The sibling row two lines above glosses that phrase as "the corpus is green"; the shell row
dropped the gloss, so read literally it claimed no line is affected. MEASURED instead of
reworded: **16 tokens** drop from the sweep today, **4 symbol-shaped** — `Fence`, `Ff`,
`Field`, `Option` — and **none is cited** in `docs/` or `packages/`. The row carries the
number now.

One correction to my own working, recorded because the number nearly went in wrong: my
first replication scoped the measurement to shell files alone and got **48**. That
overstates it — a token dropped from a script that still lives in a `.ts` file is not lost
from the sweep at all. **16 is the sweep-wide figure**, and it matches seat A's
independently.

### Gates on `de2e618b`, all EXECUTED

`VITEST_MAX_WORKERS=6 npm test` → 501 files, **9436 passed**, 3 expected fail, 7 skipped,
**0 failed**. `typecheck` OK. `lint` 1067 files, no findings. Node v22.23.2, macOS.

### What these passes did not cover, in their own words

- **One platform.** No Linux, no Windows. Seat A noted the delta adds no new literal
  comparison against a checked-out file, so the CRLF-sensitive class this file already
  worries about is not extended — but that is a judgement, not a run.
- Seat A probed `pinVerdict`'s two git calls in a real `--depth 1` clone and confirmed the
  inversion is real (`git cat-file -e` exits 128 for a non-tip sha), then confirmed the
  pre-existing shallow-refusal test reds alongside it with the correct diagnosis, so an
  operator would not be misled. It did **not** run the full suite shallow.
- Neither seat re-audited the baseline lists or the other two laws — out of scope by
  instruction.
- Seat A made **no `gh` call at all** and did not read this PR's checks; its verdict is
  about the tree, not about CI state.
- Seat A's closing note, kept because it is the honest boundary: the commit message's claim
  that round 3's two passes happened is *not something a reviewer of the tree can verify*.
  It is durable and per-sha, which is the form condition 1 asks for; whether the passes
  occurred is the operator's call.

### Condition 1

This push makes `de2e618b`. The delta since the reviewed `373365bc` is **prose only** — two
comment blocks and one input-table row, no behaviour change, and the mutation evidence
above is unaffected by it. That is the smallest the delta has been in four rounds, and it
is still, strictly, author-read only.

gabriel-canaan (commented):
## Review of `de2e618b` — commit-by-commit, in a detached worktree with `origin/main` merged in

Read end to end: one commit, two files, 1,020 lines. This closes the condition-1 gap the
body declares — `de2e618b` has now been read in full by someone other than its author, not
just the `373365bc..de2e618b` delta.

### Verified in the author's favour (all EXECUTED)

- **Suite green on the merged tree.** `origin/main` is `aaf9a50c` and the branch is already
  rebased onto it, so the merge was a no-op. `VITEST_MAX_WORKERS=6 npm test` → **501 files,
  9440 passed, 3 expected fail, 3 skipped, 0 failed**, node v22.22.2. The body says
  9436 passed / 7 skipped; both sum to 9446, so four tests that skip on the author's macOS
  run here. The difference is in the author's favour.
- **The baseline's arithmetic holds.** 40 `[renamed-away]` rows over **34 distinct** symbols,
  1 `[historical]`, 72 `[not-ours]` — matches the stamp and the pinned counts exactly.
- **Every `[renamed-away]` reason sha resolves and is an ancestor of `origin/main`.** Nothing
  in the law asserts this (the reason column is prose, as the body says), but all 34 check out.
- **The `[historical]` ruling is correct on its one row.** `docs/roadmap.md:224` says the
  `title=` sweep finished "with `NOT_YET_SWEPT` deleted rather than emptied" — `b33da639`
  deleted it and `715f0a21` wrote the sentence. The sentence is true *because* it names the
  removed symbol. `[renamed-away]` would have sent the next reader to "fix" it.
- **`local-ci (Darwin)` is genuinely pinned to the head sha.** `de2e618b` committed
  02:03:16Z, status created 02:06:10Z. It describes this tree, and it is the only check —
  Actions is disabled, as the body states.
- **The `OWN_FILE` self-certification trap is real and the fix works.** Confirmed the file's
  own control-test literals (`'MAX_ARMS'`, `'shadowBlur'`) do not enter arm 1's token set.

---

### Finding 1 — EXECUTED — arm 2 reads JSDoc continuation lines as declarations, so a prose word gets diagnosed as one of ours

The docblock's central precision argument is this:

> The regexes are applied to raw diff lines, comments included — they are precise not because
> comments are stripped … but because **a comment almost never takes a declaration's shape**.

That is false, and the cause is one character. `DECLARATION_RES`' last row admits `\*` as a
modifier prefix (for `*gen()`), with `\s*` after it:

```
/^\s*(?:public|private|protected|static|async|\*)?\s*([A-Za-z_$][\w$]*)\s*\(/
```

So every JSDoc continuation line of the form `` * Word (…) `` is a declaration. Exhibited
from the actual history pass:

```
"Measured"      <= "+ * Measured (theme/oklch.ts): H ≈ 335–341 in both worlds; ≥ 33° from broken,"
"Completeness"  <= "+ * Completeness (every tracked screenshot has exactly one entry, and no"
"Opens"         <= "+ * Opens (or creates) a journal for appending."
"States"        <= "+ * States (ruling 7): live (markers on the track) · empty (no checkpoints —"
```

**Measured over the real corpus:** 389 names enter `everDeclaredTokens()` through this row
and no other. **20** of those are symbol-shaped *and* absent from source today — meaning any
document that cites one gets the wrong verdict:

```
Completeness Exposed Input Lenient Measured Only Opens States
conductorSubagentAt manualChunks observatoryEvent setActiveTab setActivityHighlight
setClock setCollapsedState setConverging setPanning setReducedMotion setShowCard useFrame
```

**A/B through the law itself**, one run, two words that are equally never-ours:

```
docs/zz-review-probe.md cites `ZzQqNeverDeclaredHere`, which this repo never declared in the paths arm 2 reads — add it to [not-ours] with its citing file
docs/zz-review-probe.md cites `Measured`, which this repo declared once and no longer has — fix the citation, record it in [renamed-away], or in [historical] if the sentence names the removal on purpose
```

This is the same defect shape as the two latent gaps this PR already closes, one dimension
over — and it is the **harmful** direction the arm-2-pathspec CONTROL test is written for: a
wrong diagnosis whose instructed remedy files a never-ours word into `[renamed-away]`, the
section whose own header calls its entries debt to be removed.

`setPanning` is worth reading as the near-miss: its `[renamed-away]` verdict rests entirely
on `+        setPanning(event.sourceEvent.type !== 'wheel')` — a **call**, not a declaration.
The classification is right (it was a destructured state setter, genuinely ours) but the
mechanism is luck, not the argument the docblock makes.

By contrast `+  conductorSubagentAt(): number | null {` is a genuine method declaration that
only this row catches, so the row is load-bearing and must stay.

#### Tested fix — require the `*` to be adjacent to the name

Generator methods are written `*gen()`; JSDoc continuations always have the space.

```diff
-  /^\s*(?:public|private|protected|static|async|\*)?\s*([A-Za-z_$][\w$]*)\s*\(/,
+  /^\s*(?:(?:public|private|protected|static|async)\s+)*\*?([A-Za-z_$][\w$]*)\s*\(/,
```

EXECUTED on this branch: `doc-citation-law.test.ts` **71/71 green**, all 40 `[renamed-away]`,
1 `[historical]` and 72 `[not-ours]` rows still hold (`conductorSubagentAt` and `setPanning`
included), and the A/B above now returns the correct verdict for both words:

```
docs/zz-review-probe.md cites `Measured`, which this repo never declared in the paths arm 2 reads — add it to [not-ours] with its citing file
```

It does not close the *call-expression* half (`foo(` at line start is ambiguous on a diff
line by construction). If you take the fix, the docblock sentence quoted at the top wants
narrowing to match — and the call-expression residue is worth a row in the input table
rather than silence, on the same "a form not listed is a form not reviewed" rule the table
is built on.

---

### Offered, not asserted

- **The DoD's regression fixture is substituted, and the substitution is disclosed.** #369
  asks to "check out its tree, run the law, watch it fail"; the test instead asserts the
  chain (sha landed → `8f7a60ec~1` declared it in `lab.ts` → absent today → history declared
  it → verdict `renamed-away`). I think the substitute is the better test — it runs every
  time for ~0 cost — but it does not demonstrate the law would have reddened *at that commit*,
  and the body is right to say so out loud rather than claim the DoD line verbatim.
- **`.windows-known-failures` already lists this file** under `path-separator`. This commit
  adds a second `file === OWN_FILE` comparison against slash-separated git output, so the
  entry's cause class stays accurate — no action, just noting the entry now covers two
  mechanisms, not one.
- **Unrelated to this PR:** local `main` here carries merges that are not on `origin/main`.
  Worth untangling before anything lands, since the gate pushes local `main`.

---

*Single-reviewer pass (rhizo-review), Claude Opus 5 — not the multi-seat verify. Every
finding above was reproduced by running it; nothing is carried from a model seat unverified.*

gabriel-canaan:
Finding 1 is fixed on the branch as `93fb0622` — pushed as a fast-forward, no force-push.

**What it changes** (one file, the one the issue fenced; +66 −3):

- `DECLARATION_RES`' method row — the generator star must now be adjacent to the name, so
  `*gen()` is a method generator and `* Word (…)` is a sentence.
- The docblock sentence the finding falsified ("a comment almost never takes a declaration's
  shape") is corrected rather than left standing beside the fix, with the measurement in it.
- Two input-table rows: the JSDoc shape now closed, and the half this does **not** close —
  a call expression at the start of a line, where `foo(` is a call and a method declaration
  written identically and one diff line carries nothing that separates them. `setPanning` is
  named there as the live instance.
- A CONTROL test pinning it, in the file's own idiom.

**Evidence, all EXECUTED on the merged tree:**

| gate | result |
|---|---|
| `doc-citation-law.test.ts` | **72/72** (was 71) |
| `VITEST_MAX_WORKERS=6 npm test` | **501 files, 9441 passed**, 3 expected fail, 3 skipped, **0 failed** |
| `npm run typecheck` | OK |
| `npm run lint` | 1067 files, no findings |
| mutation — revert the regex alone | **1 failed, 71 passed** — only the new CONTROL reddens |
| the A/B from the finding, re-run | `Measured` now reads *"never declared in the paths arm 2 reads"* |

All 40 `[renamed-away]`, 1 `[historical]` and 72 `[not-ours]` rows still hold — including
`conductorSubagentAt`, which is a genuine method declaration no other row in the list can
see, and `setPanning`, which survives on the call-expression residue the new table row
declares.

@ciaran-slow — this is a substantive change to your classifier, so it wants your
confirmation rather than my approval, and I have not approved or merged. Linux only; the
`local-ci (Darwin)` status still points at `de2e618b`, so that leg wants re-running on the
new head before this lands.

*Single-reviewer pass (rhizo-review), Claude Opus 5 — not the multi-seat verify. The fix
above has had exactly one perspective on it: mine.*

gabriel-canaan:
**Correction to my review above.** The last "offered, not asserted" bullet — that local `main`
carried merges not on `origin/main` — is **false**, and nothing on this PR depends on it.

`git rev-list --left-right --count main...origin/main` → `0  64`. Local `main` is a clean
ancestor of `origin/main`, 0 ahead and 64 behind. Ordinary, fast-forwardable, no unpushed
landing.

How I got it wrong is worth naming, since it is this repo's own lesson one shape over: I ran
`git rev-parse … && git merge-base --is-ancestor … || echo "NOT ancestor"`. The `rev-parse`
died with `fatal: Needed a single revision`, the `&&` short-circuited so `merge-base` never
ran, and the `||` printed a definitive-sounding verdict that no check had produced. I read a
command that never executed as a result — and then labelled it `EXECUTED`.

The rest of the review stands; every other finding was reproduced by running the law itself.

Issue #412, opened 2026-09-10
prd55 w5: the R&D hand is the operator's own claude, spawned as an explicit act — engine, CLI, route, and the lab's live listing folds the log


KelliherL:
## Fence (may touch ONLY)
- `packages/server/src/lab/rd.ts` (new file)
- `packages/server/src/lab/rd.test.ts` (new file)
- `packages/server/src/lab/__fixtures__/` (fixtures this lane adds)
- `packages/server/src/cli/lab-rd.ts` (new file)
- `packages/server/src/cli/lab-rd.test.ts` (new file)
- `packages/server/src/cli/lab.ts` (the subcommand's registration)
- `packages/server/src/cli/lab.test.ts`
- `packages/server/src/cli/cli-surface-law.test.ts` (the row the subcommand adds)
- `packages/server/src/api/lab.ts` (the R&D route)
- `packages/server/src/api/lab.test.ts`
- `packages/server/src/api/index.ts` (route registration only)
- `packages/server/src/api/route-class-law.test.ts` (the count and class rows this route adds)
- `packages/contract/src/lab-rd.contract.test.ts` (new file)
- `README.md` (the route-count line and the CLI surface line)
- `packages/contract/src/contract-coverage-law.test.ts` (the route's row and any count it moves)
- `packages/server/src/api/security.ts` (the prose route counts, if the mutation count is spelled there)
- `docs/architecture.md` (the sentence that counts the routes)
- `packages/web/src/settings/registry.ts` (two entries: the agent command and the corpus switch)
- `packages/web/src/settings/registry.test.ts`
- `packages/web/src/settings/coverage-law.test.tsx`
- `docs/adr/0046-the-instrument-spawns-the-operators-own-tools-as-an-explicit-act.md` (new file)
- `docs/adr/README.md` (the index row)

## Blocked by
[...]

PR #441, opened 2026-09-11
prd55 w5 (follow-up): the tracker read has its own ceiling — a network call is not a plumbing probe

KelliherL:
Follow-up to #437, which landed two minutes before this commit was pushed to its branch; cherry-picked onto `main` as its own PR so the record stays one PR per wave and nothing rides in unreviewed. Refs #412.

**What becomes true.** `readTrackerItems`'s `gh issue list` ran under `RD_PROBE_TIMEOUT_MS` (5 s) — the ceiling this repo gives `git` plumbing and a `--version` string, where five seconds means wedged. A network call resolves a host, opens TLS, authenticates and pages an API, where five seconds is ordinary latency on a slow link; the degrade was loud (a recorded `trackerRefusal`), but the cost was a corpus the operator declared and kept not getting, for a reason that read like a broken `gh`. `RD_TRACKER_TIMEOUT_MS = 30_000` sits beside the probe ceiling and is used for the `gh` call only. (The reviewer's own "your call, offered not asserted" on #437.)

**Verified.** EXECUTED on this head: typecheck 0, lint 0, `rd.test.ts` 34/34. The new law reads the ceiling each subprocess actually ran under off the recorded call (`withTimeout` sets `options.timeoutMs`), pins the three ceilings in the module as distinct numbers with the tracker's above the probe's, and a mutation swapping the two call sites goes red (1 of 34), then restored.


gabriel-canaan:
Reviewed at `d937f70`, in a detached worktree with `origin/main` merged in (the merge brought one file, `docs/prds/prd-51-the-split.md`; nothing in this PR's fence moved).

Every claim in the PR body reproduces, and two of them are stronger than stated.

**EXECUTED** — post-merge tree, node 22.22.2:

| | |
|---|---|
| `npm run lint` | exit 0, 1057 files |
| `npx vitest run packages/server/src/lab/rd.test.ts` | **34/34 passed** |
| `VITEST_MAX_WORKERS=6 npm test` | **496 files, 9220 passed, 3 expected-fail, 3 skipped, 0 failed** |

**EXECUTED — the mutation, certified in both directions.** The PR body claims the swap goes red 1 of 34. It does, and so does the other half of the swap, which the body did not claim:

1. Reverting the tracker call site to `RD_PROBE_TIMEOUT_MS` — taken from `origin/main` rather than retyped, run twice, restore verified against `HEAD`:
   ```
   CERTIFIED: the mutation was caught 2/2 run(s).
   ```
2. The reverse — `probeExec` wrapped with `RD_TRACKER_TIMEOUT_MS` instead:
   ```
   Tests  1 failed | 33 passed (34)
   ```

So the law catches a single-token slip in either direction between the two ceilings that now live eight lines apart. Reading `options.timeoutMs` off the recorded call rather than off which constant the source mentions is what buys that, and the `new Set([...]).size === 3` assertion is what stops "some timeout was set" passing on the very swap the test exists to catch. That is the right shape.

**The sibling case, which the body does not answer — I checked it, and there is none.** `gh` has exactly one call site in product code across the whole tree:

```
$ grep -rn "exec('gh'\|'gh'," packages --include "*.ts" | grep -v node_modules | grep -v "\.test\.ts"
packages/server/src/lab/rd.ts:222:    'gh',
```

Every other `5000` in the exec-ceiling family — `ROUTE_EXEC_TIMEOUT_MS`, `RETARGET_EXEC_TIMEOUT_MS`, `COMPARE_EXEC_TIMEOUT_MS`, `CHECKPOINT_EXEC_TIMEOUT_MS` — bounds local `git` plumbing, and each says so in its own doc comment. No second network call is sitting under a plumbing number.

**The composition trap is clear too.** `docs/design-notes/lab-launch-ceilings.md` records the `dispatchFork` failure where an outer 5s wrap beat an inner 120s one, because `withTimeout` applies its override on the way *down* and the wrap closest to the raw `exec` wins. This change adds a second internal wrap point in `rd.ts`, so it is the shape that bug had. It does not reproduce: `runRdHand` hands `readRdCorpus` `rawExec` (`packages/server/src/lab/rd.ts:560`), not `probeExec`, so the tracker's 30s is the wrap nearest the raw exec. The only other door in is `packages/server/src/cli/index.ts:365` → `runCli` from `packages/server/src/api/lab.ts:868`, and neither wraps. Traced, not assumed.

**CI: all three jobs, ten legs, green** — `build-test-boot` ×3, `pack-smoke` ×6, `windows-suite`. Not read off the bar: on `build-test-boot (ubuntu-latest, current)` the step list shows `Build`, `Test`, `Typecheck`, `Lint`, the packaging guard and the boot smoke each with conclusion `success`, so no gate passed by being skipped.

**Fence.** `packages/server/src/lab/rd.ts` and `rd.test.ts` are both named in #412's declared fence. No widening, and none needed. `Refs #412` rather than `Closes` is right — #412 closed COMPLETED at 03:05, four minutes before this branch's commit was pushed.

## Two facts, offered not asserted

**1. Your 13 failures do not reproduce here.** The commit body records `packages/server/src/lab` + `packages/core/src/lab` as 176 passed / 13 failed, "identical to the pre-change baseline of 175/13", naming the five `.windows-known-failures` files. On this machine the **full** suite is 9220 passed / 0 failed, those five files included. Both numbers are honest measurements of different environments — yours carried a baseline of failures that is not a property of the tree. Worth knowing that the wider run is available to you, because "identical to baseline" is a weaker statement than "green", and you had the stronger one.

**2. What the law survives.** It pins the wiring, not the value. Mutating `RD_TRACKER_TIMEOUT_MS = 30_000` to `5_001` leaves it 34/34 — `toBeGreaterThan(RD_PROBE_TIMEOUT_MS)` still holds. **EXECUTED.** I think that is correct and would not change it: pinning `30_000` in the test makes the law a restatement of the constant and turns every deliberate retune red for no reason, and the argument for 30s belongs where you put it — in the doc comment, where a reader deciding whether to move it will actually be standing. Recording it so the next reader knows the boundary rather than discovering it.

## Verdict

**No findings.** Nothing to fix, so nothing for me to push. The change is one constant, one call site and one law that is certified to catch the defect it was written for.

Approval and landing are the operator's act via `gate.sh` — I have not approved or merged.

---
*Reviewed by a single reviewer (`rhizo-review`), Claude Opus 5. One pass, not a multi-seat verification: the executed results above are re-runnable, the judgement calls in "offered not asserted" have had one perspective on them.*

gabriel-canaan (approved):
Approved at `d937f70` — the head this review read, re-checked immediately before approving.

No findings. Verdict and evidence are in the review comment above: typecheck 0, lint 0, `rd.test.ts` 34/34, full suite 496 files / 9220 passed / 0 failed on the post-merge tree, and the mutation certified in **both** directions (reverting the call site, 2/2 caught; the reverse swap, 1 failed of 34). Sibling case checked — `gh` has one call site in product code. Composition trap checked — `readRdCorpus` gets the raw exec, so the 30s wrap is the one nearest the subprocess. CI green on all three jobs and ten legs, with the steps confirmed executed rather than read off the bar.

Recorded in the verify ledger as `rhizo-review`, a single-reviewer record — a lighter class than seats, and labelled as such so an auditor sees exactly what backed this.

Landing stays the operator's act via `gate.sh`.

---
*Single-reviewer pass (`rhizo-review`), Claude Opus 5. One perspective, not a multi-seat verification.*

PR #439, opened 2026-09-11
prd-51: the wedge wave took number 4, and wave 5 is groomed

ciaran-slow:
Documentation only. One commit, one file, 73 added lines, no code.

Base is `origin/main` at `a3c5b305` — the merge of #425, prd-51 wave 4.

Closes no issue, deliberately: this is the grooming record for wave 5, and the wave's issues
(#432–#436) are the output it describes rather than work it performs. It follows the shape of
#417, which carried rulings 15 and 16 the same way.

---

## Condition 1 — stated as a gap, not recorded away

`AGENTS.md` requires every commit in the range to have been read by an independent review pass,
with the record somewhere a later reader can ask. **This commit has not had one.**

What it has had, exactly and no more: the operator was sent the drafted amendment text in
session, read it, and said to land it. That is a human read by the person the amendment binds,
and it is why this PR exists — but it is the same session that wrote the text, so it is not
independent in the sense the condition means. No agent verify pass ran.

That is written here rather than dressed up, because a docs-only PR is precisely where the
temptation to skip the pass is strongest and the cost of skipping it is a map the next groomer
inherits and trusts. **A second reader before merge costs nothing.**

## What this records

**1. The wedge wave took number 4.** The 2026-09-09 *wave numbering, as built* amendment put
*seal → archive → verify → tombstone → prune* and *membership is the boundary* at wave 4. The
2026-09-10 *two wedges* amendment then filed #387, #398, #399 and #410 as wave 4 and did not
renumber the map. From that day the archive/membership pair had no wave number at all, and the
map read as describing work that was already in flight. The pair moves to 5+; everything the old
map listed as "5+" moves down one place. The map table is superseded, not edited in place — the
same treatment the 2026-09-09 amendment gave the Sequencing section.

**2. Wave 5 is not that pair, and #410 is the reason.** Membership needs a keys port, and a keys
port is `storage/contract.ts`, `fake.ts` and `postgres.ts`. #410 holds all three — this document
widened it onto them itself, for ruling 16's `unfoldable` marker — and #410 is open, still blocked
by #407. Grooming membership into a wave that dispatches before #410 lands would bundle across a
live fence, which the *Consequence for the wave-4 fences* section already refused once. So wave 5
was inverted to touch **no path under `packages/team/src/`**, and membership goes to wave 6.

**3. Wave 5's five issues**, filed and boarded before this PR: #432 (ruling 11's `rhizomorph
[...]

PR #442, opened 2026-09-11
prd14 w4: a saved comparison carries the measure and the facts that judged it, and a v1 artifact still reads

### LINKED ISSUE #347: prd14 w4: a saved comparison carries the measure and the facts that judged it, and a v1 artifact still reads
## Fence (may touch ONLY)
- `packages/web/src/lab/compare/artifact.ts`
- `packages/web/src/lab/compare/artifact.test.ts`
- `packages/web/src/lab/compare/types.ts`
- `packages/web/src/lab/compare/fromExperiment.ts`
- `packages/web/src/lab/compare/fromExperiment.test.ts`
- `packages/web/src/lab/compare/ComparisonSurface.tsx`
- `packages/web/src/lab/compare/ComparisonSurface.test.tsx`
- `packages/web/src/lab/compare/save.ts`
- `packages/web/src/lab/compare/SaveComparisonControl.tsx`
- `packages/web/src/lab/compare/parser-agreement-law.test.ts`
- `packages/server/src/comparisons/artifact.ts`
- `packages/server/src/comparisons/artifact.test.ts`
- `packages/server/src/comparisons/store.ts`
- `packages/server/src/comparisons/store.test.ts`
- `packages/server/src/comparisons/parser-agreement-law.test.ts`
- `packages/contract/src/fixtures/comparison-artifact/` (all files)
- `packages/contract/src/lab-comparison.contract.test.ts`
- `packages/contract/src/lab-comparisons.contract.test.ts`
- `packages/contract/src/lab-comparison-save.contract.test.ts`
- `packages/web/src/recordings/comparisons.ts`
- `packages/web/src/recordings/comparisons.test.ts`
- `packages/web/src/recordings/RecordingsPage.tsx`
- `packages/web/src/recordings/RecordingsPage.test.tsx`
- `CHANGELOG.md`

This fence is drawn from the real footprint, not from the two artifact modules
the issue used to name. A version bump lands in BOTH parser copies by
construction — that is ADR-0047's whole condition — so a fence naming one
schedules the drift this PRD has already paid for once, in #376.

## Blocked by

**Nothing. Cleared 2026-09-11.** #401 fenced the entire compare directory and
landed in PR #420 alongside #402. No open issue claims any path above —
verified by scanning every open issue's fence on 2026-09-11.

**EXPLICITLY OUT:** docs/prds/prd-14-experiment-console.md and docs/adr/ — the
ruling and ADR-0047 are already written and land in the same PR as the
orchestrator's own commit, not this lane's.

**EXPLICITLY OUT:** packages/core. ADR-0047 rejects it as the shape's home and
says why; do not move the shape there, and do not import across the package
boundary in either direction.

## Why

A finished comparison saves and reopens, but the artifact is a projection of one
measure and nothing else. Per run it stores `{ verdict, detail?, value }` where
`value` is whichever measure happened to be selected at save time — so
`durationMs`, `commits` and the gate's `provenance` are folded away, and the
measure itself is not recorded at all.

The measure is the half a reader can see. A reopened comparison currently says
*"measure not recorded — this artifact does not carry which measure produced
these values, so no summary is shown for any arm"*, and shows no summary,
because #214 correctly refused to guess a basis rather than print a false one.

## The ruling this implements

**prd-14 ruling 6**, ruled 2026-09-11, with **ADR-0047** superseding ADR-0042
beside it. Read both before starting — every decision below is already made and
none is yours to reopen.

## Definition of done

- **v2 stores facts in the compare surface's own vocabulary**, never
  `LabRunOutcomeDTO`: per run `{ verdict, detail?, cost, duration, commits }`;
  per artifact the `measure` and the gate's provenance (`verifyCommand`,
  `source`, `measuredAt`). Artifact-level, not repeated on every run.
- **The reopened surface re-derives through `runForMeasure`** — the same
  function the live surface uses, not a second path that happens to agree. The
  measure switch works on a saved comparison, which v1 cannot offer at all.
- **A v1 artifact is READ, not migrated and not refused.** The parser takes 1
  and 2 as a discriminated union and rewrites nothing on disk. No upcast: v1 →
  v2 is not total, because the measure is unrecoverable, and prd-17's chokepoint
  is for translations that are complete.
- **3 and above still refuse by name.** The shared fixture row pinned to
  `unsupported comparison artifact version: 2` moves to `3`.
- **The `measure === null` rendering path is kept and retargeted at v1
  artifacts** — every run shown by its verdict, no summary, an explicit reason.
  Its wording is already exactly right for a v1 file. Do not delete it.
- **Both parser copies move together**, and both agreement laws prove it on the
  shared fixture. A v2 case that only one side accepts is the defect #376
  existed to close.
- The three contract tests cover the v2 round trip, and the coverage law's rows
  still match the routes.

State it for each half, and certify at least these:

- change the server parser's v2 run shape alone, leaving the web copy — an
  agreement law must redden, not just one side's unit test;
- store a v2 artifact and read it back with the measure dropped — the surface
  must fall to the v1 path and say so, never guess a basis;
- feed a v1 artifact to the v2 reader — it must READ, and a mutation making it
  refuse must redden a test that says so by name.

## Sibling case

The two parser copies are the standing sibling and always will be — ADR-0047
keeps them deliberately, and its falsifier is that the agreement law's
throw-site coverage stays derived from source. If this work hand-lists those
sites for any reason, say so loudly on the PR: it voids the ADR's argument.

gabriel-canaan:
prd-14 wave 4 — the last build in this PRD. A saved comparison gains a
`version: 2`, and the ruling that shapes it lands beside it.

Closes #347

## Declared review gap

Two independent seats read this branch at `656d735d` and found one real gap
(below). **Everything since has had no independent eyes**: the fixture
expansion that closes it, the ADR citation sweep, and the orchestrator's own
ADR renumber. What exists instead is the orchestrator re-running the exact
mutations the seats used, which is verification of the specific holes they
found and not a fresh pass over the repair.

That was a deliberate stopping decision by the operator, taken after an earlier
issue in this PRD absorbed five review rounds for findings that stopped
changing what shipped.

## What the seats found, and what closed it

Both defeated the two-parser drift guard, from different angles, and the root
was the same: **axis B covers throw SITES, not accepted VALUES.** Narrowing a
union removes no site and adds no message, so nothing reads as uncovered.

Measured before the fix: narrowing ONE copy's accepted measures from
`cost|duration|commits|verified` to `cost|verified` left both agreement laws at
63/63 and the whole suite green, while the parsers had genuinely diverged. The
same held for `provenance.source`, and for a finite-number guard applied to
`cost` alone — that one passed both agreement suites at 126/126 while the
server accepted a non-finite `duration` the web copy threw on.

The cause was fixture coverage: the v2 accept cases only ever carried the
values the first build happened to use. `duration`, `commits` and
`measure-route` never appeared as an accepted value anywhere in the repo.

Fixed by spanning the value space — every member of `MeasureV2`, every
`provenance.source`, and both finite and non-finite forms of all three numeric
fields. **Verified by the orchestrator, EXECUTED**: both mutations that
previously passed now redden at 2 failed / 74.

The fix is coverage, not more throw sites. The one-site-for-three-fields
choice was deliberate and is unchanged.

## This changed the ADR, not just the build

ADR-0049's falsifier as first drafted said the two-copy decision holds while
throw-site coverage stays derived from source. The seats proved that
insufficient. It now names **two** kinds of coverage — sites and accepted
values — carries the measurement that defeated the first version, and says the
first draft was wrong. Corrected before it shipped rather than after, which is
the only reason it is a footnote instead of an incident.

## The ADR was renumbered twice, and the sweep is the risky part

The record was numbered 0047, then 0048, and is now **0049**. Both collisions
are the orchestrator's: the number was taken from the highest on `main` at the
time, and `main` gained an ADR from another wave before this could land, twice.

**Eight files in this repo carry `ADR-0048` meaning the real 0048** — prd-55's
record that the instrument spawns the operator's own tools. Sweeping by grep
would have corrupted every one of them. The sweep was scoped **by diff**: only
the fifteen citations this commit added, across four files, each verified to
carry zero pre-existing `ADR-0048` occurrences before being touched. The real
0048's citations are confirmed unchanged.

## What is the sibling case?

The two parser copies, permanently — ADR-0049 keeps them deliberately and this
is the first version bump to test that. Both move here by construction, and the
agreement laws on the shared fixture are what prove it.

Within the change: a v1 artifact is READ, not migrated and not refused. The
upcast is ruled out because v1 to v2 is not total — the measure is
unrecoverable — so the existing `measure === null` rendering path is kept and
retargeted at v1 files, where its wording was already exactly right.

Seven certified by the lane, and the load-bearing one re-certified independently
by the orchestrator **before** the seats ran: forcing the SERVER copy alone to
parse v2 `cost` as `null` reddens the server agreement law 4/63 while the web
agreement law stays 63/63 green. That is a version bump landing in a single
copy, caught — the property ADR-0049's whole argument rests on.

Plus the two post-repair mutations above, and the three the issue demanded: a
dropped measure must fall to the v1 no-summary path and never guess a basis; a
v1 artifact must READ under the v2-aware parser, with a mutation making it
refuse reddening a test that says so by name.

## Fence

26 declared, 22 touched, nothing outside. No widening on a change spanning
three packages and a stored format version.

## One thing found in the tooling, and it reaches past this PR

The lane hit a fence-audit failure it could have rationalised away — the repo's
own `scripts/fence-lint.sh` passed clean on the same diff — and instead stopped
and reported it. Two defects in the operator's pre-commit gate, now fixed
outside this repo:

- its fence matcher was an exact whole-line match, so a directory-style fence
  entry never matched a file inside it;
- more seriously, in normal mode it audited `BASE..HEAD`, which is **empty
  before the first commit** — the normal case, and the point of a pre-commit
  gate. The audit was guarded by a non-empty check, so it silently did nothing
  and printed neither pass nor fail.

**An earlier lane commit in this repo reporting a clean fence audit may be
reporting a check that never executed.** Recorded here because a later reader
needs to know which era of that phrase means something. `scripts/fence-lint.sh`
in this repo was correct throughout and is unchanged.

## Unrelated, and flagged so the next person does not re-derive it

`packages/web/src/lab/trace/TraceDiff.test.tsx` is load-sensitive and not in the
timing set. It failed once in a full-suite run at loadavg 15.5, and passes
isolated twice, on `origin/main` alone, and in a full suite at loadavg 10.5.
This branch does not touch it. It will redden the next load-probed landing.

## Gates, EXECUTED on the assembled head

`VITEST_MAX_WORKERS=6 npm test` rc=0 — 496 test files, 9305 passed, 3 expected
fail, 3 skipped.

---

## Declared review gap

> the repairs made after the seats read this branch have no independent eyes — the fixture value-space expansion, the ADR-0049 citation sweep, and the orchestrator's ADR renumber. The orchestrator re-ran the seats' own defeating mutations and both now redden (2 failed / 74 each), but that verifies the named holes rather than the repair as a whole. A deliberate stopping decision by the operator.

```
  UNVERIFIED  88b86ed1  feat(lab): a saved comparison carries the measure and the facts that judged it, and a v1 artifact still reads (#347)
  UNVERIFIED  193b802a  docs(prd-14): ruling 6 — the saved artifact carries facts, and ADR-0049 supersedes ADR-0042
```


KelliherL (approved):
## Review of #442 — approve, with three mechanical fixes pushed and `origin/main` merged in

This is the independent pass the body's *Declared review gap* asked for: the post-seat repairs — the fixture value-space expansion, the ADR-0049 citation sweep, and the renumber — read commit-by-commit in a clean clone at node 22.23.2 on Linux, against #347's own Definition-of-done bullets. **Most of it verified in the PR's favour**, and that comes first because it is the evidence.

### What verified — EXECUTED on your head `88b86ed1`

- `npm run typecheck` **0** · `npm run lint` **0** (1057 files) · every test file this PR touches plus `contract-coverage-law` — **13 files, 338 passed**.
- `scripts/fence-lint.sh 347` **PASSED**; and by hand, all 22 touched paths sit inside the 26 #347 declares, with the four docs files being the orchestrator's commit the issue explicitly carves out.
- **The ADR-0048 sweep really was scoped by diff.** `git grep ADR-0048` on `origin/main` and on this branch are **byte-identical** (9 files) — every citation of the real 0048 is untouched. No added line carries `ADR-0048`.
- **Fixture value space, parsed from each accept case's raw bytes, not grepped:** 48 cases, 14 accept (4 v1, 10 v2). v2 `measure` values present: cost ×2, duration, commits, verified (+5 absent). `provenance.source`: compare-cli, measure-route (+ null, + absent). Every complete v2 run field — cost, duration, commits — appears both as a number ×5 and as an explicit null. That is the coverage ADR-0049's falsifier asks for, on the three fields it names.
- **Your four certified mutations, re-derived:**
  1. server copy alone parses v2 `cost` as `null` → **server agreement law red** (the v2 accept, the MAX_VALUE boundary, the not-a-number refuse), **web law green**. The property the ADR rests on.
  2. web `parseMeasureV2` narrowed to `cost|verified` → **both laws red**, on exactly the two new "every member of MeasureV2 must appear as an ACCEPTED value" cases. The seats' finding, now caught.
  3. web parser's v1 branch disabled → the v1 round-trip and refusal cases red, and *"a v1 artifact still READS under the v2-aware parser — no upcast, no refusal"* **red by name**.
  4. `RecordingsPage` defaulting a dropped measure to `'cost'` → *"a v2 comparison with its measure dropped falls back to the v1 no-summary path, and offers no switch"* **red by name**.
- **REASONED, on the code:** the v2 route branch keys on `'measure' in input`, so an older client with no `measure` still lands a v1 file and the wire body stays `{ input }` — the `mutating-calls-law` literal is untouched by construction. `parseRunV2` ignores the `value`/`note` keys the live `Run` still carries, and the reopened surface regenerates the note through `runForMeasure`, so nothing a reader sees is lost by not storing it. `toMeasurableRun` ↔ `runForMeasure` round-trips `detail`/`verifiedDetail` and the three facts. `latestProvenance` is the newest judged run's, as the ruling says — one per artifact.

### Three fixes, all mechanical, pushed

**1. `95657264` — the fixture did not reach `provenance.measuredAt`'s finite boundary.** A fifth mutation, mine: delete the server copy's `Number.isFinite(measuredAt)` check, keeping the type check and the message → **both agreement laws green** (EXECUTED). The only `measuredAt` refuse case carries a string (`"soon"`), which reaches the `typeof` half of the guard; nothing reaches the finite half, so a one-sided drop reads as fully covered — the exact shape the seats found on `cost`, on the one v2 numeric field the widened fixture left out. ADR-0049's own condition is "both the finite and non-finite form of every numeric field", and `measuredAt` is one. Two cases added, mirroring your cost pair: `1e400` refuses with the guard's own sentence; `Number.MAX_VALUE` accepts with provenance passthrough. Re-run of the same mutation with them in place: **server law red** (EXECUTED). No parser changed.

**2. `1ccff7fb` — the renumber sweep never reached the docs commit.** `docs/adr/0042-…md:3` and its `README.md` row both said **"superseded by ADR-0047"** — which on this tree is the aria-disabled record. The body's sweep was scoped to the feat commit's four files, correctly; these two live in `193b802a`. Now `ADR-0049`. No law reads an ADR's Status against the README, which is why nothing reddened.

**3. Same commit — prd-14's ruling 6 states the first-draft falsifier.** The amendment says the decision holds *"only while the agreement law's throw-site coverage stays derived from source"* — one kind of coverage — while ADR-0049 in the same commit was corrected to two kinds after the seats defeated that sentence. The ruling now names both and points at the ADR's Consequences for the measurement, per AGENTS.md's "link the ADR rather than restate it".

### The merge and the union

`origin/main` moved under you twice (#437, #439). Merged **into** the branch as `27939cdb` — clean, no overlapping files. The lane gate on the union:

```
npm run lint                     exit 0 — 1057 files
VITEST_MAX_WORKERS=6 npm test    exit 0 in 158 s
  Test Files  496 passed (496)
       Tests  9309 passed | 3 expected fail | 3 skipped (9315)
```

9309 is your 9305 plus the four the two new fixture cases add (two cases × two agreement laws). `TraceDiff.test.tsx` stayed green at this load.

One platform (Linux), one node. CI's ten legs are green on your head; they will re-run on this one.

- **#347's body says ADR-0047 throughout** (title of the ruling, the *EXPLICITLY OUT* lines, the sibling-case paragraph). It is the issue's text, not the tree's, and it closes with this PR; a one-line comment there saying "0047 in this body = 0049 as landed" would stop the next reader from opening the wrong record. I did not edit your issue.
- **`TraceDiff.test.tsx`** did not redden here, at loadavg well under the 15.5 you saw. Nothing to add; noting that this run is one more green data point for "load, not this branch".
- The `--unverified` / verify-ledger section: per the operator's 2026-09-07 ruling that AGENTS.md section describes one contributor's local kit, not a repo gate. Harmless here.

### What I did not run

macOS and Windows. The four-scenario harness for the seats' original mutations against the *pre*-fix fixture (I re-derived them against the fixture as shipped, where they redden — which is the claim that matters). A live save from the surface in a browser; the round trip is exercised through the real route by `lab-comparison-save.contract.test.ts`'s v2 case, which I read and ran.

**Approving.** Three fixes are a fixture pair, two citation corrections and one sentence aligned to the ADR it cites — rung 1 on the ladder, so I am not blocking on confirmation; drop any of the three if you disagree, they are separate commits. Landing is the operator's act.

Issue #430, opened 2026-09-11
prd55: the R&D hand's cost is carried on its events' provenance but not booked as llm.cost — booking needs a third telemetry source


KelliherL:
## What happened

prd-55 ruling 1 says the R&D hand's cost "is booked as spend with its basis, like a fork's." Wave 5's server half (#412) carries `total_cost_usd` and `duration_ms` from the CLI's own JSON result on every `rd.*` event's provenance — auditable, beside the model, the prompt and corpus digests and `claude --version` — but does **not** emit an `llm.cost` event for it, and says so in its report rather than hiding it.

## Why not

`llm.cost`'s `source` is the closed set `TELEMETRY_SOURCES = ['sessionlog', 'otel']` in core, and neither is true of a spawned CLI: no OTLP receiver saw the call and no transcript was tailed for it. Signing a cost event `sessionlog` or `otel` would be a false provenance in the record — exactly what ruling 1's "the instrument holds no credential and forwards nothing" is careful about on the other side.

## What would make it true

A third source (`agent-cli`, or `lab-rd`) in core's telemetry schema, folded by the reducer like the other two, with the spend surfaces' basis line saying *from the CLI's own result*. A new enum value reaches every exhaustive map and round-trip test the coupling map now names (#407 found four) — so it is its own fenced issue, not a line in wave 5. Until then the hand's cost is readable on the R&D tab (wave 6 renders the provenance line) and absent from the ledger's total, and the guide says so in the honest-gap voice.

---

## Fence (may touch ONLY)
- `packages/core/src/events/telemetry.ts`
- `packages/core/src/events/telemetry.test.ts`
- `packages/core/src/reduce.telemetry.test.ts`
- `packages/core/src/reduce.ts`
- `packages/core/src/fixtures.ts`
- `packages/core/src/fixtures.test.ts`
- `packages/server/src/lab/rd.ts`
- `packages/server/src/lab/rd.test.ts`

## Blocked by
Nothing. Sibling lanes of this wave hold other surfaces: the fork's dispatch ceiling (#408), the two route tests (#421) and core's id factory (#429) — this issue enters none of them. Two files this wave does not claim and this issue must not edit: core's event-source map, which keeps `llm.cost`'s primary source exactly as it is, and the web spend surfaces, which render whatever the fold books and need no change to do it. If either turns out to be needed, STOP and report it as a widening.
**Model:** opus. **Wave:** 8.

## Definition of done
[...]

Issue #421, opened 2026-09-10
lab routes: 'a bare request is refused before the handler runs' flakes under a loaded 6-worker run for lab-transcript and lab-series — a race, not a timeout


KelliherL:
## What happened

Gating prd-55 wave 4's union (`prd55-w4` at `7fe4c071`) on the conductor's Windows box with `npx vitest run packages/web/src/lab packages/web/src/App.test.tsx packages/contract packages/server/src/api/{lab-series,lab-transcript,route-class-law,gated-reads,lane-index}.test.ts` (6 workers), two tests failed together:

```
FAIL src/api/lab-series.test.ts > GET /api/lab/telemetry … > the route > is a gated read — a bare request is refused before the handler runs
FAIL src/api/lab-transcript.test.ts > GET /api/lab/transcript … > the route > is a gated read — a bare request is refused before the handler runs
```

Run alone: 15/15 and 22/22. Run together, twice: 37/37. Run with the whole `packages/server/src/api` directory: green. The same `lab-transcript` test also failed once earlier today in the union of waves 1–3 and passed alone. So the assertion is right and the test races something only a loaded run exposes — most likely the app's readiness or the capability-token setup relative to the first bare request.

## RULED OUT (2026-09-11): the readiness race

The hypothesis below — that the bare request outran the app's setup — is **disproven**. Against the real installed Fastify 5.10.0, an async plugin's registration delayed 0–100 ms and raced against a timer, `inject()` already serialises on `ready()` internally for a single call: the old shape and an `await app.ready()` shape answer identically at every delay. Do not spend time there.

The three cases now assert the refusal's **body**, from the exported header constant, not just its status — so a recurrence reports what answered. That is diagnostics, not a repair, and this issue stays open until one recurrence tells us what it was.

## What to do

The repo's rule: never fix a flake by widening a timeout — remove the race. Both tests share a shape (wave 4b copied wave 3's), so one fix fits both: await the app's `ready()` (or the same setup the other tests in those files use) before the bare request, and assert the 401 body, not just the status. Owner: whichever lane next touches `packages/server/src/api/lab-transcript.test.ts` / `lab-series.test.ts`; both are in open PRs (#404, wave 4's PR).

EXECUTED as above; neither route's behaviour is in question — the gate is `requireCapabilityToken` on both, the same preHandler the other lab routes use.

---

## Fence (may touch ONLY)
- `packages/server/src/api/lab-transcript.test.ts`
[...]

Issue #429, opened 2026-09-11
recorder: createIdFactory restarts at <prefix>-000001 in every process, so one session file carries two events named lab-000001 — 'unique within a session' is no longer true


KelliherL:
## What happened

Found while fixing #409 in prd-55 wave 5 (the lab listing now folds the live session's log, not only the server's buffer). Merging the buffer's events with the file's needs a key, and `event.id` is not one: `createIdFactory` restarts its counter at `<prefix>-000001` in every process, and the lab's fork (the lab fork module (unbackticked: it is #408's fence, not this issue's), its own `SessionRecorder`) and the server's measure route (`ctx.recorder`) both mint `lab-…` ids into the same session file. EXECUTED: one session log on this machine carries two different events named `lab-000001`. The #409 merge keys on id + ts + type because of it, and a mutation that dedupes by id alone drops one of the pair — recorded red, then restored.

`packages/core/src/events/index.ts`'s comment says an id is "unique within a session". Today it is unique within a process.

## What would make it true

Either the id carries the process (a per-process nonce or the recorder's own instance in the prefix), or every writer to a session file goes through one recorder (which #409 shows they do not, by design — the CLI must work standalone). The first is contained in `createIdFactory` and the tests that pin id shapes; the second is an ADR. Until one lands, any consumer that dedupes lab events by id is wrong, and the #409 merge's three-part key is the workaround, not the fix.

---

## Fence (may touch ONLY)
- `packages/core/src/events/index.ts`
- `packages/core/src/events/events.test.ts`
- `packages/server/src/api/lab.ts` (the live-merge key's doc comment only — the reasoning it states changes)
- `packages/server/src/api/lab.test.ts`

## Blocked by
Nothing. Sibling lanes of this wave hold other surfaces: the fork's dispatch ceiling (#408), the two route tests (#421) and the telemetry source vocabulary (#430). The live-merge key added by the wave-5 fix stays as it is — this issue makes it belt-and-braces rather than load-bearing, and the change here is to the comment that explains why, not to the key.
**Model:** sonnet. **Wave:** 8.

## Definition of done
- An event id is unique within a session, which is what its own doc comment already claims: a second process writing to the same session file cannot mint an id a first process already used. The shape is the lane's to choose and to justify in the doc comment — a per-process discriminator inside the id, or a factory that reads the log's high-water mark before it starts — but it must keep ids human-readable and diffable by eye, which is the other half of what that comment promises.
[...]

Issue #409, opened 2026-09-10
lab: a launch or a CLI checkpoint is invisible until the server restarts — the live session's listing reads the server's recorder buffer, and the fork writes through its own


KelliherL:
## What happened

Walking the prd-55 stack base (`451693ab`) in Chrome against a live server on this machine:

1. `rhizomorph lab checkpoint <lane>` from a shell — the CLI printed the capture and the session log gained a `fork.checkpoint`; the lab page showed *there are no checkpoints yet* on every reload. After restarting the server (which resumes the session and re-reads the log) the checkpoint appeared.
2. A launch from the panel — *2 of 2 requested arm(s) dispatched … launched — fork-6eee90d4…*, both `fork.dispatched` events in the log; the Experiments section showed *there are no experiments yet* on every reload, and the frame *no experiment was forked from this checkpoint*. Wave 1's reload-after-launch fires; there is nothing new to read.

## Why

`packages/server/src/api/lab.ts`'s event merge (around lines 133–154): for every session but the live one it reads the session file; for the live one it takes `ctx.recorder.eventsSoFar()` — the server's in-memory buffer — on the reasoning that the buffer is fresher than the disk. But `packages/server/src/lab/fork.ts:368` constructs its own `new SessionRecorder(resumed.sessionId, logFilePath, { resumeFrom })` and appends `fork.dispatched` through it (the CLI must work standalone), and `rhizomorph lab checkpoint` does the same from another process. Those events reach the file and never the server's buffer, so the listing folds a session that is missing its own lab events until the next boot re-reads the file. `POST /api/lab/measure` records through `ctx.recorder`, so a measurement would be visible — of an experiment that is not.

Every unit test of these routes passes because the tests append to the server's recorder.

## What would fix it

In `api/lab.ts`'s merge, fold the live session from the union of the buffer and the file's events beyond the buffer's last offset (dedupe by event id — the recorder already carries a resume-from shape), or have the launch route hand `ctx.recorder` into `dispatchFork` and the checkpoint CLI signal the running server. The first is contained in `api/lab.ts`, which prd-55 wave 5's server half already fences; it takes the fix.

EXECUTED: the two runs above; `grep -c '"type":"fork\.'` on the live session log (checkpoint 1, dispatched 2) against the page's empty listings; the appearance of the checkpoint only after a restart.

PR #445, opened 2026-09-11
prd55 w6: the R&D tab reads the operator's own hand — patterns by shape, proposals that vary one thing, every empty state drawn first

### LINKED ISSUE #413: prd55 w6: the R&D tab reads the operator's own hand — patterns by shape, proposals that vary one thing, and every empty state drawn first
## Fence (may touch ONLY)
- `packages/web/src/lab/rd/` (new directory, all files)
- `packages/web/src/lab/LabPage.tsx` (the R&D tab and its mount)
- `packages/web/src/lab/api.ts` (the R&D route's client and the events' read)
- `packages/web/src/lab/api.test.ts`
- `packages/web/src/lab/types.ts` (the R&D DTOs)
- `packages/web/src/lab/no-live-fleet-law.test.ts` (the source-file count rows this wave's new directory adds)
- `packages/web/package.json` (the `./lab/rd` exports entry the contract test imports through)
- `packages/web/src/replay/mutating-calls-law.test.ts` (the `MUTATING_MODULES` row for the R&D client)
- `packages/contract/src/lab-rd.contract.test.ts` (new file — moved here from #412)
- `packages/contract/src/contract-coverage-law.test.ts` (the `EXPECTED` row for it)

## Blocked by
Waves 4 and 5 landed on the branch this lane starts from (the rail-and-stage workspace with its tablist; the R&D route, events, schema and the two settings entries). Sibling lanes hold other surfaces: nothing under packages/server or packages/core is entered — a shape the route does not serve is a gap drawn in the honest-gap voice, not a client-side invention; packages/web/src/lab/launch, compare, trace, metrics, canvas, frame, rail and axis are earlier waves' surfaces: the R&D tab reuses the launch review (ruling 4) and Compare's run drawing (the counterfactual) by mounting them, never by editing them. The guide and its law are wave 7's.

## Definition of done
prd-55 rulings 1, 2, 3, 4, 5 and 9 (the surface's half) and the specification's S5.

- The workspace gains an R&D tab beside Compare, Trace and Metrics. Top: the R&D control — a native model select reading lab.models (ruling 5), a corpus switch reading lab.rdCorpus (ruling 2), a *read and propose* button that posts to the R&D route, and the provenance line of the last run (model · cost · turns · corpus digest · claude --version). Left: patterns by shape, count and corpus, the held-back ones dimmer with their sentence. Right, for the selected pattern: its proposals (one dimension varied, arms listed, the checkpoint pick with what it rejected), the estimate, *restore n arms* and *restore and run*, and *against what actually happened* once an experiment exists.
- Every state is drawn before the live one (ruling 9), each a fixture and a test: no CLI (control disabled with the PATH sentence verbatim from the server); no corpus (*"nothing to read yet — a retro, or a measured experiment, is where a pattern comes from."*); all held back (every row dim and *"no pattern recurs — nothing is proposed."*, with the copy *"1 issue · not yet a pattern — testing a shape that may not recur spends real money."*); refused (the schema reason verbatim and the raw result in a details element); live; launched (the proposal row links to its experiment); partial launch (prd-53 ruling 7's copy); replay (the tab disabled as the lab is). No demo fixture invents a pattern.
- *Restore n arms* and *restore and run* open the launch review prefilled from the proposal (its arms and checkpoint) and dispatch only through the launch the lab already has, with its estimate and one confirmation unchanged; changing the checkpoint pick before launching sends the override so the server records rd.override, and the surface says *"operator override — the choice is never re-attributed to the agent"*.
- The counterfactual is drawn as Compare's runs against the source item's own observation, labelled *what actually happened*, as one observation, never a spread; a retro with no measurable figure yields *"no measured baseline — the retro's own words"*, quoted as text, never converted to a number.
- Keyboard path: ↑/↓ moves through patterns, Enter selects, Tab reaches the proposals, Enter on a proposal opens the launch review prefilled, Esc returns; the model dropdown is a native select; no native title attribute anywhere; type through the theme's tokens.
- Laws, executed: a proposal's arms differ in exactly one dimension as rendered (a fixture adding a second renders the server's refusal, never a patched proposal); a held-back pattern carries no proposal; the provenance line's cost equals the event's; the hand never runs without a click (no effect posts to the route on mount — a mutation that does goes red); nothing under rd imports the fleet's panels or routes; the source-file count pin carries the new directory's row.
- The R&D route's contract test, moved here from #412: through the web client this wave writes (a `.ts` module under rd/, exported as `./lab/rd`), it covers the run, the schema refusal, the no-CLI sentence and the override record, and is enumerated where every contract test is (the coverage law's row, the mutating-calls law's `MUTATING_MODULES` row).
- Lane gate green on the branch: npm run typecheck, npm run lint, VITEST_MAX_WORKERS=6 npm test.

KelliherL:
Closes #413. prd-55 wave 6 — rulings 1–5 and 9 (the surface's half) and the specification's S5. One PR = one (milestone, wave); ten commits, read in order.

## What becomes true

- **The workspace gains an R&D tab**, the tablist's fourth reading. The control carries a native model select over `lab.models`, the corpus switch over `lab.rdCorpus`, *read and propose*, and the provenance line of the last run (model · cost · turns · corpus digest · `claude --version`). Left: patterns by shape, count and corpus, the held-back ones dim with their sentence. Right, for the selected pattern: its proposals (the dimension varied, the arms, the checkpoint pick with every rejection and its reason), the estimate, *restore n arms* and *restore and run*, and *against what actually happened* once an experiment exists.
- **Every ruling-9 state is a fixture and a test, drawn before the live one:** no CLI (the control disabled with the server's PATH sentence verbatim), no corpus, all held back (with *"1 issue · not yet a pattern — testing a shape that may not recur spends real money."*), refused — both refusal shapes, including the declared-dimension one the wave-5 review added — live, launched, partial launch, replay. No fixture invents a pattern.
- **The web client is the app's eighth mutating call**, `requestRd`, contract-proven end to end: the gate, the schema refusal, and the run and no-CLI sentence for real (a deliberately absent `agentCommand`, no fake CLI), enumerated in `contract-coverage-law`'s `EXPECTED` and `mutating-calls-law`'s `MUTATING_MODULES`.
- **Ruling 4 is whole now** (the recorded widening below): *restore n arms* and *restore and run* prefill the proposal's **arms** and checkpoint and send its `proposalId`, so the experiment carries the proposal durably; and the launch route, when a launch names a proposal whose checkpoint differs from the pick, looks it up in the fold and records `rd.override` through `recordRdOverride` before dispatching — naming both checkpoints, never re-attributing the choice. An unknown `proposalId` is refused by name.
- **A refusal carries the hand's raw result text** (bounded), so ruling 9's refused state shows it in its `<details>` rather than an honest gap.
- **The tab is reachable whenever a lane is known** — a seated checkpoint — while Compare, Trace and Metrics keep their experiment gate: the hand reads retros and reviews, and S5's *no corpus* state exists for a repo with no experiment at all.

## Verified

- **EXECUTED after merging `origin/main` (`74d4c984`, waves 1–5): `npm run typecheck` 0 · `npm run lint` clean (1064 files) ·** `packages/web/src/lab` + `App.test.tsx` + `mutating-calls-law` + `packages/contract` + `api/lab.test.ts` + `lab/rd.test.ts` + `lab-ceiling-law` — the only reds under load are the two Windows-known contract files (`clone`, `instrument`), confirmed by listing the failing files across four loaded runs.
- **EXECUTED — ten mutations red, then restored** (the lane's): header smuggling in the client; the `rd: 4` count pin; the coverage-law cross-check; the click-only law (no effect posts on mount); the one-dimension law as rendered; the override sentence; the counterfactual baseline; the post-merge `varies` field; the corpus label; the estimate. The widening commits carry their own: a launch body dropping `proposalId`; the override not recorded; the raw text dropped.
- **One flake, not this branch's:** `TraceDiff.test.tsx`'s not-launched case failed once in a loaded run and did not reproduce (3/3 alone, green on `main` under the same load, green in four further loaded runs; this branch does not touch `trace/`). Recorded on #421 beside its two siblings rather than filed again.

## Widening — recorded on #413 before the change

`launch/LaunchPanel.tsx` + `launch/launch.ts` and their tests (prefilled arms and `proposalId`), `server/src/api/lab.ts` + its test and `server/src/lab/rd.ts` (the override recorded at the launch, the raw text on a refusal), and `api/lab-ceiling-law.test.ts` (the launch entry point now hands `sessionDir` and `recorder` through — the same two the measure entry already carries, for the same reason: recording an event needs a place to write it).

## Known gaps, stated

The lane's own report is at `reports/w6-413.md` through commit `0921e6b1`; the five commits after it are the widening above and are described here. The hand's cost is on every `rd.*` event's provenance and rendered in the provenance line, but is still not booked as `llm.cost` — that needs a third telemetry source in core (#430).


KelliherL:
**CI caught one thing the lane's gate could not see — fixed in `68835ea1`.** The three `build-test-boot` legs went red on `api/route-class-law.test.ts`: the R&D client is the app's **seventeenth** outbound call site, and README's outbound-fetch recipe names every module and site by hand while the law sweeps the same roots against it. Reconciled together, so no number is typed twice: README's two number words, its path list, its own back-reference further down (*"every one of the seventeen targets this instrument's own loopback origin"*), the law's `EXPECTED_CALL_SITES` row for `lab/rd/rd.ts`, `EXPECTED_TOTAL`, and the two test titles that quote the numbers. Widening recorded on #413 before the change.

**Why the lane's gate missed it, which is the part worth keeping:** `route-class-law.test.ts` is on `.windows-known-failures` (line-endings — its README-paragraph reader does `indexOf('\n\n')` on a CRLF checkout), so on a Windows box the whole **file** reads as expected-red, and a **new** red test inside an already-red file is invisible to a per-file attribution. That is prd-25's own *"a count masks a swap"* hazard one level down: a known-red file masking a new red test. The ubuntu legs are what see it, which is what they are for — and it is an argument for the CI rework's local verify scripts reporting per test rather than per file.

EXECUTED after: typecheck 0, lint 0, `route-class-law.test.ts` 93/95 — the two remaining reds are that file's own known line-endings pair (the support-matrix check and the README path-list reader), both red on `main` on this box too.

KelliherL:
**CI green on `68835ea1`** — ten legs, all pass, including both Windows legs and `windows-suite`. The recipe reconciliation was the only thing between this branch and a full bar.

ciaran-slow (changes requested):
Read commit-by-commit (14 commits). Most of this is well-built and verified below. **One finding blocks**, and it is the kind this repo's own PR template asks for: a correct fix whose structurally identical sibling was missed.

- **All 10 CI legs pass** — `build-test-boot` ×3, `pack-smoke` ×6, `windows-suite` — counted against AGENTS.md's three-jobs/ten-legs checklist rather than read as a green bar.
- **`packages/server/src/api/lab.test.ts` → 86 passed**; `packages/server/src/lab/rd.test.ts` → 39 passed.
- **Both fence widenings were recorded before the change**, which I checked by timestamp rather than by the PR's say-so: widening 1 posted on #413 at 15:22:29 NZ, its first commit `9bcd523f` at 16:08:12; widening 2 at 16:37:12, its commit `68835ea1` at 16:40:17. ✓
- **The ordering claim holds.** `await recordOverrideIfNeeded(request, options)` runs *before* dispatch, so a failed override write refuses the launch rather than dispatching an unrecorded one. That is the safe direction and it is the one AGENTS.md's "seal released on a failed write" note would have caught if it were reversed.
- **The raw-result bounding is applied at every push site**, not just the obvious one — `boundedRawResult()` wraps `answer.result` at both the refusal paths, with `RD_REFUSAL_RAW_RESULT_CHARS = 8_000`. I went looking for the unbounded sibling and there isn't one.
- **An unknown `proposalId` is refused by name before dispatch**, as claimed.

## Finding — BLOCKING · EXECUTED · every `rd.override` event in a session mints the same id

`packages/server/src/api/lab.ts`, in `recordOverrideIfNeeded`:

```js
const nextId = createIdFactory('lab')   // ← constructed per call
…
{ id: nextId(), ts: now() },            // ← exactly one id drawn, so always -000001
await options.recorder.record(event)
```

The factory is constructed **inside** the per-invocation function and exactly one id is drawn from it, so its counter restarts at 1 on every call. Two operator overrides in one session are both written as `lab-000001`.

I proved it rather than reasoned it — a probe driving the real route twice through the existing harness:

```
PROBE override ids: ["lab-000001","lab-000001"]
AssertionError: two overrides minted ids ["lab-000001","lab-000001"]: expected 1 to be 2
```

(Probe inserted into `api/lab.test.ts`, run, then reverted; baseline restored to 86/86.)

**What mutation would the current tests survive?** This one. Every override assertion in the suite is `toHaveLength(0)` or `toHaveLength(1)` — no test records two overrides in one session, so the id being a constant is invisible to it.

**Why it matters here specifically.** This is the exact collision class #409 surfaced and #429 (PR #450, landing in the same window) exists to close. `api/lab.ts`'s own doc comment on `liveEventKey` says a bare-id dedupe would "silently DROP one of the two writes it exists to surface". The composite key means both events still survive a read today, so this is not data loss — but it re-opens a class that the tree is simultaneously declaring closed.

**The one-line fix does NOT work — I tried it first and it failed.** Passing #429's writer tag looks like the answer and is not:

```
PROBE2 override ids: ["lab-override-000001","lab-override-000001"]
```

The tag fixes collisions *between different writers*; it does nothing about a factory whose counter restarts, which is this bug. The repair has to make the id actually advance — hoisting the factory to a scope that outlives the call, or deriving the id from something already unique (the proposal, the ts). **Which of those is right is your call, not mine**, which is why I am not pushing a fix: I would be guessing at the id scheme you want.

## Offered, not asserted — the same shape lives in `lab/rd.ts`

`recordRdOverride` (`packages/server/src/lab/rd.ts`) constructs its factory the same way, per call, and mints one event. I did not execute a probe against that path, so treat this as **REASONED** — but if you are choosing an id scheme for the route, it is probably the same decision.

## Not run, stated as fact

I did not run the full suite against this branch alone. I did run the full suite against the merged `#449 + #450` tree, which contains all of these commits: **499 files, 9386 passed, rc=0**.

## Landing order — please read before merging

`#450` was built without this branch and the two genuinely conflict. Details are on #450; the short version is that the conflicts are additive and I have verified the resolution, but **#445 → #449 → #450** is the order that works.

ciaran-slow (commented):
**Fixed in `19e1ef29`, pushed to this branch.** Rung 2 — substantive, so this is not an approval: it needs your confirmation before it lands.

## What the fix does

`recordOverrideIfNeeded` now seeds its factory from the overrides already on the record instead of restarting the counter:

```js
const priorOverrides = events.filter((candidate) => candidate.type === 'rd.override').length
const nextId = createIdFactory('lab-override', priorOverrides)
```

The function already reads and folds the whole session two statements earlier, so the count is free.

**Why seeding and not the writer tag.** I tried the tag first and it fails — `["lab-override-000001","lab-override-000001"]`. The tag fixes collisions *between* writers; this bug is a counter that restarts, which is a different axis. Seeding also survives a server restart, where a module-scoped factory would not.

**Why that prefix.** `'lab-override'` produces a string byte-identical to what `createIdFactory('lab', n, 'override')` will produce once #429 lands, so adopting the tag later does not change the shape of ids already written.

## The test, and what it survives

`api/lab.test.ts` gains one test: two launches, **one recorder** — the same session file, which is the scope `createIdFactory` promises uniqueness within.

It asserts the exact pair `['lab-override-000001', 'lab-override-000002']` rather than `new Set(ids).size`, deliberately: a set assertion passes for *any* two distinct ids, including a random one, and the claim here is about a readable counter that advances by one.

**Proven to bite, both ways — EXECUTED:**

| mutation | result |
|---|---|
| the original bare per-call factory | `expected [ 'lab-000001', 'lab-000001' ]` → **red** |
| the tag-only repair (counter still restarts) | **red** |
| restored | **87 passed (87)** |

The second one is the one I care about: the test discriminates the real fix from the plausible wrong one, not merely from no fix.

Every pre-existing override assertion counts events (`toHaveLength(0|1)`) and none reads an id — so all of them survive the defect, which is why it was invisible.

## Gates — EXECUTED

- `npm run typecheck` rc=0
- `npm run lint` clean (1064 files)
- **full suite: 499 files, 9256 passed, 3 expected fail, rc=0**
- I swept for anything keying on the `lab-` id prefix before changing it. Nothing does — the fold indexes overrides by lane and `proposalId`, not by event id, and no test pinned an override id.

## Not fixed here, on purpose

`recordRdOverride` in `lab/rd.ts` and the four writers #429 tags share this per-invocation shape — `captureCheckpoint` most cleanly, one fresh factory and one id drawn, so every capture in a session is `lab-checkpoint-000001`. That is #450's fence and the finding is recorded there rather than widened into from here. **REASONED** for those four; only this path is EXECUTED.

## Still true: land in order

**#445 → #449 → #450.** #449 contains this branch in full; #450 conflicts additively with both, resolution verified and written up on #450.

Over to you — confirm the approach and I'll treat it as settled.

ciaran-slow (approved):
**APPROVE** at `19e1ef29`. All 10 CI legs pass at this sha, including `windows-suite` (15m 4s).

## The review record, split by who wrote what

Condition 1 asks that every commit in the range have an independent review pass on record. It is not uniform here, so it is stated rather than averaged:

| commits | reviewed by |
|---|---|
| `f8c56785` … `68835ea1` (14 commits, the author's) | this review pass — read commit-by-commit, findings above |
| `19e1ef29` (the override-id fix) | **authored by me.** Not independently reviewed. The approach was put to the operator and confirmed in session, 2026-09-12 |

I am approving a range that contains a commit I wrote. That is a weaker guarantee than the table's first row and is recorded here so a later reader can see which it was, rather than inferring one review covered all fifteen. If you want the fix to carry a genuinely independent read before it lands, that is a reasonable thing to ask for and I would not argue against it.

## What the approval rests on — EXECUTED

- `npm run typecheck` rc=0 · `npm run lint` clean (1064 files)
- full suite **499 files, 9256 passed, 3 expected fail, rc=0**
- the new test proven red against the original defect **and** against the tag-only repair, then restored to 87/87
- no consumer keys on the `lab-` id prefix — swept before changing it; the fold indexes overrides by lane and `proposalId`
- both fence widenings verified recorded *before* their commits, by timestamp
- all 10 CI legs pass, counted against the three-jobs/ten-legs list

## Caveats that survive the approval

**The local runs are macOS only.** CI covers ubuntu and Windows across ten legs, but every suite I ran myself witnessed one platform. This is not "CI passed" in the cross-platform sense.

**Landing order still binds: #445 → #449 → #450.** #449 contains this branch in full. #450 conflicts additively with both; the verified resolution is written up on #450 and is still unapplied.

**My merged-tree evidence predates #451 and #453**, which landed mid-review. The 9386-passing integration run was against an older `main`. Worth re-running at landing rather than trusting it as current.

Merging stays the operator's act.

PR #446, opened 2026-09-11
prd51 w4: a line the fold cannot read lands unparsed, and one actor's skew stops only that actor

### LINKED ISSUE #410: prd51 w4: a line the fold cannot read does not discard the records around it

ciaran-slow:
Closes #410

prd-51 wave 4's fourth issue, implementing **ruling 16** — both answers. One commit,
`0413d872`, base `main`.

**Why this is its own PR and not part of #425.** #410 was blocked by #407 (prd-55 wave 5), which
held `packages/core/src/events/index.ts`, so it could not ride the wave-4 bundle. The operator
ruled at the time to assemble the other three rather than hold verified work behind an unrelated
PRD, accepting a second queue toll for this wave. #425's body records that deviation and this PR
is its other half. #407 merged as `74d4c984`, confirmed an ancestor of `origin/main`.

---

| commit | plan | build report | verify (Sonnet 5) |
|---|---|---|---|
| `0413d872` | [plan](https://github.com/launchpad-26/rhizomorph/issues/410#issuecomment-5629109855) | [report](https://github.com/launchpad-26/rhizomorph/issues/410#issuecomment-5629462500) · [correction](https://github.com/launchpad-26/rhizomorph/issues/410#issuecomment-5629576507) | [**ready to merge**](https://github.com/launchpad-26/rhizomorph/issues/410#issuecomment-5629569568) |

**Two weaknesses in that chain, stated because a reviewer should weigh them rather than find them.**

1. **The plan and the build came from one session, on the same model.** The operator asked for it
   explicitly, and it is not the pipeline's shape: the builder is meant to execute a plan cold and
   the verifier to read it as a contract rather than as the author's memory. The verify pass is
   therefore the *first* independent read of this work at any stage, where it would normally be
   the second. It was told so, and told to weigh the plan's provenance as a weakness rather than
   as reassurance.
2. **One verifier, not a panel.** It ran on Sonnet 5 against an Opus 5 build and confirmed the
   `Co-Authored-By` trailer itself rather than taking the claim. It ran every gate in its own
   clean worktree rather than reading the build report's numbers — and the numbers matched.

**The verify pass corrected the build report, which is the evidence that it read rather than
skimmed.** The report's mutation table claimed the first mutation reddens 5; it reddens **6**. The
figure was measured accurately and then went stale *inside the same session* — taken against a
141-test scope before two more test files were added, and written down after. True when executed,
false when recorded. The corrected table is in the correction comment and is the one to trust; the
build report's is wrong as written.

## What is wrong, and what lands instead

`runOnce` assembled every row from every record past the cursor **before** opening the
transaction, and returned on the first refusal. One line a newer sender shipped therefore
discarded the records *before* it as well as after: nothing inserted, the cursor file never
created, and **every actor in the journal stopped**, not only the skewed one. Reproduced three
times on the issue — four good lines landing nowhere, cursor 0 on every pass, and the only
operator escape was editing the journal by hand.

The ingest cannot catch it, deliberately: ruling 4's ordering 1 journals and answers 202 before
anything parses, so the sender has already been told the batch is durable and has already advanced.

**Answer (2) — the line lands unparsed.** `toEventRow` lands an `unknown` with its `type` recorded,
its envelope intact and its `line` verbatim. There is no gap, so ruling 3's *no gaps in `n`* is
untouched and ruling 5's identical-`chainDigest` claim needs no union with a second relation —
that is why this beat a quarantine table. `malformed` still refuses, loudly: a corrupt line in an
fsynced journal is a fault in the journal, not skew between builds.

**Answer (3) — the fold is per actor.** Records are grouped by `(project, actorInstance)`; a
refusal stops that group only, keeping every row built before it. The cursor grows a per-actor
mark and its `seq` becomes the **minimum** over them, so a stuck actor holds the journal back and
is re-read while healthy actors' spans dedup through `ON CONFLICT DO NOTHING`.

**No projection derives a value from an unvalidated payload.** `EventRow` carries an `unfoldable`
marker the adapter does not persist. `spend.events` still counts the row — it is an event and it
reached storage — while `costUsdOf`, `agentStateOf` and `dirtyFiles` are gated.

## What is the sibling case?

**The one the ruling itself missed, and the one this PR turns on.** Ruling 16 first said the marker
closes the leak and `projectionsFor` reads it. True for the three values that module derives
itself; **false for `row.lane` and `row.worktree`**, which are derived *upstream* in `toEventRow`
and arrive as ordinary columns. A marker read inside a projection cannot see a derivation that
already happened — a projection reads a column, it does not derive a value. So they are nulled at
source instead.

**`unknown-shape` walks through every `row.type` guard.** The issue's stated falsifier was
`unknown` vs `malformed`; the sibling one level down is `unknown-type` vs `unknown-shape`, where
the type *is* known. `dirtyFiles` gates on `row.type === 'worktree.dirty'`, which protects the
unknown-*type* arm and nothing else. Covered by the marker, not by the type gate.

**The widening is three fields, not two, and the third rests on a different argument.** Ruling 16
names `id` and `source` and justifies them as *"additive, in the one place that already proved
them"* — the envelope probe validates and discards both. That argument does **not** extend to
`payload`: the probe is `{ id, ts, source, type }` and never touches it, so this puts an
unvalidated field into a core type. Defensible, because `line` beside it is already the whole
unvalidated line and the marker exists so nothing derives from either — but it is a different
argument, and it is written into the commit message so a later reader cannot infer the probe
validated it.

**A sibling deliberately left alone, and it is a live trap.** `packages/core/src/wire/reserialize.ts`
projects the unknown arm as four explicit fields and drops the rest. That is the *only* reason
`reserialize.test.ts`'s veil assertions stay green under a new `payload` field. The verifier
mutated that projection into a spread to prove the point: **2 tests redden with leaked marker and
line bytes.** Tidying it into a spread would breach the veil. Outside this fence, flagged here.

Five, each reverted, on the 143-test `fold` + `lenient` scope — **and every one re-run
independently by the verifier**:

| mutation | red |
|---|---|
| `projectionsFor`'s `unfoldable` check removed | **6** |
| `row.ts` restores `laneOf`/`worktreeOf` on the unfoldable arm | **3** |
| `row.ts` refuses an `unknown` again (the pre-ruling-16 wedge) | **8** |
| `lowWaterMark` returns `Math.max` instead of `Math.min` | **2** |
| `runOnce` returns `ok: false` on the first bad row | **3** |

**One of these exposed a hollow test during the build, and that is the most useful thing in this
PR.** The second mutation was required to redden a `lane_state` assertion. Run as first written it
reddened **2**, both in `row.test.ts`, while `projections.test.ts` stayed green — because that
suite hand-builds rows with `lane: null` and so **bypasses `row.ts` entirely**. It could not fail
for the reason it claimed. That is exactly the half the review of #417 exists to correct, guarded
by a test that only read as though it guarded it. Fixed with an end-to-end case that folds a real
journal whose unfoldable payload carries a live `lane`, `worktreePath` and `costUsd`; the mutation
now reddens 3 including it, confirmed independently.

The verifier also chose **two assertions no mutation was claimed for** — `payloadOf`'s
null-vs-undefined coalesce and `nonNegativeInt`'s negative guard — and reddened both.

## Gates

Run in the lane worktree on node 22.22.2, and **re-run in full by the verifier** in a separate
clean worktree:

- `npm run typecheck` — green, 6 workspaces
- `npm run lint` — green, 1058 files
- `VITEST_MAX_WORKERS=6 npm test` — **497 files, 9253 passed**, 3 expected fail, 7 skipped
- `npm run build` — green
- boot smoke — booted from built `dist`, `GET /` → 200, `GET /api/meta` → 401 (ADR-0012, unchanged)

**The merged result was tested, not just the branch.** `origin/main` moved 6 commits after the
verify pass (prd-14 wave 4, `#442`). Merging it into this branch is a clean automatic merge with
**zero file overlap** — that PR touched `contract/`, `server/comparisons/`, `web/lab/`; this one
touches `core/events/` and `team/fold/`. On the merged tree: **0 type errors, 9343 passed**. The
trial merge was then discarded so the sha in the table above is byte-for-byte what the verifier
read. This is done by hand because Actions was retired on 2026-09-11 and nothing tests a merge
result automatically any more.

**One platform.** All of the above is macOS. It says nothing about ubuntu or the Windows suite.
Not "CI passed".

`scripts/gate.sh` was not run: running it is the landing, and the landing is the operator's step.

## Fence

Eleven files changed, all inside the issue's declared 17-path boundary. **No widening.**
`storage/fake.ts` and `storage/postgres.ts` were in the fence and needed no edit — `EVENT_COLUMNS`
did not grow, which is the no-migration claim holding rather than being asserted.
`scripts/fence-lint.sh` was re-run against the live board before dispatch and again by the
verifier against what is in progress now; no overlap either time.

## Not in this PR

**No migration** — `events.type` is `text NOT NULL` with no CHECK, no enum and no FK, so prd-51
wave 5's single migration-adding lane is still unspent. That is load-bearing for the wave, not
incidental. **No ADR** — the decision is ruling 16, already accepted in
`docs/prds/prd-51-the-split.md` with both rejected alternatives (quarantine table, refuse at
ingest) recorded there by name. **No CHANGELOG entry** — `packages/team` has never been in a
tagged release.

## One contract change to weigh

`FoldResult`'s `ok: false` now means **only** a journal-level failure; a fold where every group
refused returns `ok: true` with a populated `refused` array. A single-actor journal returning
`ok: false` while a two-actor one returns `ok: true` for the identical actor-level fact would be
the coupling ruling 16 exists to remove. `runOnce` has **zero callers outside its own tests**
anywhere in the tree — grepped by the verifier — so nothing depends on the old meaning today.
Whoever wires the fold into a scheduler needs to read `refused`, not just `ok`.

## Not covered

No live Postgres is exercised anywhere in this work — the suite uses the `fake.ts` double by
design, and the real adapter is asserted through recorded SQL. The per-actor cursor's behaviour
against a real partitioned table is unwitnessed until prd-51's integration drill.


gabriel-canaan (commented):
Reviewed commit-by-commit in a clean worktree with `origin/main` merged in. One finding, fixed and pushed as `14394917`; everything else verified in your favour.

## What verified

**The gates hold on today's `main`, not just the tree you measured.** Merged `origin/main` into the branch and re-ran: typecheck 0, lint clean (1058 files), **502 files / 9456 passed**. Your body reports 497/9253 on the branch and 9343 on a merge with `main` six commits back; `main` has moved again since, so all three numbers are right for their own tree.

**CI genuinely ran, and I checked the steps rather than the bar.** All ten legs SUCCESS on `0413d872`. On both `build-test-boot` legs, `Test`, `Typecheck`, `Lint`, the packaging guard and the boot smoke each have a real start and end time — no skipped step riding on a green job. The "one platform … not CI passed" caveat in the body was more conservative than the evidence required.

**The mutation table's corrected figure is right.** EXECUTED: `const unfoldable = row.unfoldable !== undefined` → `false` reddens **exactly 6**, restored byte-identical (sha256). The build report's 5 is the stale one and your correction comment is the one to trust, as you say.

## The finding — a *retired* actor is not a stuck one

`lowWaterMark` is a minimum, and `runOnce` seeded it with the seq each actor last *happened to appear at*, carrying forward every actor the cursor had ever seen. `actorInstance` is the session id — `packages/server/src/shipper/cursor.ts` says exactly that of its own `actors` map — so a deployment accumulates one per session forever, and a session that simply **ended** kept its final seq in `actors` where the minimum then sat for good.

EXECUTED against unmodified `0413d872`: session-1 folds cleanly at seq 1, twelve later sessions fold cleanly, `refused` is empty on every pass and every row lands — and `readCursor().seq` is still **1**, `actors` holds twelve entries, and the next `runOnce` reports `records: 13`. The whole journal is re-read on every tick, on a journal `journal/journal.ts` records as growing without bound (ADR-0046), with nothing stuck and nothing for an operator to fix. Before this wave the cursor always reached the head.

The two per-actor tests exercise an actor that *refuses*, so neither can see this: the mechanism is identical and only the diagnosis differs.

There is a second, smaller cost from the same wrong claim — even with retired actors dropped, a minimum over last-appearances sits on the earliest of them, so one clean read over interleaved actors leaves the cursor mid-span and re-reads the rest next tick.

## The fix, and why it is one sentence

A group that folded everything it was shown is committed through `head`, whether or not it ever appears again; only a group that **stopped** may sit below it. `readJournal` returns a contiguous span, so an actor absent from the span is absent from it and its rows are all at or below `cursor.seq`. Entries at or below the resulting `seq` are then dropped before the write — lossless, because `worker.ts` reads them back as `actors[key] ?? cursor.seq` — which is what bounds the file: a clean pass leaves `actors` empty, and entries exist only while someone is behind.

The stuck-actor behaviour is untouched and is asserted beside the new cases in the same shape, so "the cursor reached the head" cannot quietly become "the fold stopped holding anything back".

**My first set of tests was hollow, and saying so is the useful part.** Reverting only the mark rule left 119/119 green — dropping retired actors alone already stops the permanent pin, so four of the new assertions could not fail for the reason they claimed. That is the shape your own report names on `projections.test.ts`, arriving one step later, in the fix for it. An interleaved-clean-pass case is the only one that distinguishes the two halves. Certified separately, restored byte-identical (sha256) after each:

| mutation | red |
|---|---|
| mark reverts to `furthest`, pruning kept | **1** — the interleaved pass |
| pruning removed, `head` rule kept | **5** |
| the shipped rule in full, all three reverted | **5** |

Two of your assertions move in that commit, both pinning the old `actors` contents: the repaired-cursor case now reads `actors: {}`, and the stuck-actor case carries `lane-b` alone with `lane-a`'s omission asserted lossless through the reader the fold actually uses. Neither check is removed.

Gates on the fix, in its own worktree: typecheck 0, lint clean (1058 files), **497 files / 9260 passed**, build clean.

## What my push cost you, and it is not nothing

Your ten-leg green belonged to `0413d872`. The head is now `14394917` and **zero check runs exist against it** — Actions was retired on 2026-09-11, so nothing will replace them. Everything above is Linux, one machine. Nothing here is a Windows or macOS witness for the fold's new cursor shape, and the `path`-free nature of this change is an argument, not a measurement.

## Not blocking

`packages/team/src/storage/contract.ts`'s note that `unfoldable` is absent on a row read back through `readEvents` is the right call and already written down — worth knowing that `readEvents` has no production caller today, so the round-trip that would make an unfoldable row read as foldable is latent rather than live. `projectionsFor` is only ever fed freshly built rows.

## Where this leaves the PR

Rung 2 under this repo's review ladder: the fix changes behaviour, so it does not go in on my say-so. **Confirm `14394917` reads right to you, and let a second pair of eyes land it** — I will not approve my own substantive fix, and no merge-time review record has been written, since nothing is being merged here.

---
*Single reviewer (Claude Opus 5, 1M context) — one pass, not a panel, and not a substitute for the independent read this repo's condition 1 asks for.*

PR #449, opened 2026-09-11
prd55 w7: the lab's documents say what stage two's code does — guide, law, architecture, roadmap, and the paper answers itself

KelliherL:
Closes #414. prd-55 wave 7 — ruling 10, and prd-53 ruling 9's guide-law pattern. The last wave of prd-55. One PR = one (milestone, wave); four commits.

**Stacked on #445** (wave 6), which is CI-green and open; its commits appear here until it lands. Land #445 first.

## What becomes true

- **The guide gains the R&D section and the rail-and-stage workspace**, and every new sentence is a claim the law tests against code: **61 claims green, up from 42**. The launch, measure, Trace, canvas and frame claims that waves 1–4 rewrote stay as they were; what is added is stage two's own vocabulary — the R&D hand as the operator's CLI, the corpus and its two declared acts, held-back patterns and one-dimension proposals with their verbatim refusal sentences, the tab's states, the rail, the pinned stage and the tablist.
- **The architecture section says what stage two shipped** — the lab's routes, its events, its record boundary (the lab reads its own files), the R&D hand's spawn boundary, and the six scene brushes it borrows. Two facts wave 2 left stale are corrected while here: it still said *"n organisms"* and *"two named scene files"* where ruling 11 made them *n ribbons* and *six named brush modules*.
- **The roadmap's prd-55 line** says what shipped and what stays unfiled: atomic launch, recursive forking, a scoring measure.
- **The paper answers itself.** Two of four open questions are answered in place where the tree answers them; the other two stay open, honestly, because neither is built (viewer gating on a shared host; a dollar ceiling for the R&D hand). Two dated notes go under the rulings the tree read differently than written — **ruling 3**: zero varying dimensions passes, as a replication (the review of wave 5, #437); **ruling 1**: the hand's cost is carried on every `rd.*` event's provenance and not yet booked as `llm.cost` (#430). No ruling renumbered, no question deleted.

## A correction this wave made to its own brief

The brief cited **ADR-0046** for the R&D hand's spawn boundary. The tree names it **ADR-0048** — 0046 is prd-51's ingest-journal ADR, and 0047 landed from prd-30 while wave 5 was in flight. Verified against `docs/adr/`'s own listing and corrected everywhere it is cited. This is the third time that number moved under a branch; the coupling map now says an ADR number is claimed at landing, not at drafting.

## Verified

[...]

PR #450, opened 2026-09-11
prd55 w8: the findings — the hand's cost booked, ids that cannot collide, a dispatch ceiling that fits, and a disproven race

### LINKED ISSUE #408: lab: the fork's 5 s dispatch ceiling is shorter than the worktree's own setup — workmux add is killed mid npm-install and every launch reports 'exit null'
## What happened

Walking the prd-55 stack base (waves 1–3, `451693ab`) in Chrome against a live server on this machine, every launch from the panel failed at arm 1:

```
arm 1 failed and dispatch stopped there — workmux add fork-091e974b-…-arm-1 -b -P /tmp/rhizomorph-lab-brief-….md failed: exit null.
```

`exit null` is a killed child, not a refusal. Run by hand in the same WSL with the same PATH, `workmux add wsl-probe -b -P …` succeeds — after `.workmux.yaml`'s setup commands run `npm install` in the new worktree, which takes ~6 s on this box (`added 238 packages … in 5s`, then audit). `packages/server/src/lab/fork.ts` wraps that spawn in `FORK_EXEC_TIMEOUT_MS = 5000` ("a hung `workmux add` must not hang `dispatchFork` itself"), so the dispatch is killed while the worktree is still installing.

The surfaces behaved exactly as ruled: the panel read *0 of 2 requested arm(s) dispatched* with the failure verbatim (prd-53 ruling 7, prd-55 wave 1). The defect is the ceiling, not the reporting.

## Why it matters

A 5 s ceiling on a command whose configured setup is an `npm install` cannot succeed on a cold cache, and whether it succeeds on a warm one is a property of the box. Stage 1's seed on 2026-09-08 dispatched six lanes through the same path, so the ceiling is sometimes enough — which is worse than never, because it reads as a flake.

## What would fix it

One of: a dispatch ceiling sized to the setup it wraps (a minute, or `RESTORE_EXEC_TIMEOUT_MS`'s 120 s, with the hung-process protection kept by a liveness check rather than a wall clock); or `workmux add --no-setup` for the lab's arms with the install owed elsewhere — both are `server/src/lab/fork.ts`, which prd-55's waves do not enter except for the R&D engine. Not in a wave yet; filed from the walkthrough so the ledger says it.

EXECUTED: two launches (3×2 and 2×2 arms × runs) from the panel, both `exit null` at arm 1; one hand-run `workmux add` in the same environment, exit 0 in ~7 s; `FORK_EXEC_TIMEOUT_MS = 5000` at `packages/server/src/lab/fork.ts:132`.

---

## Fence (may touch ONLY)
- `packages/server/src/lab/fork.ts`
- `packages/server/src/lab/fork.test.ts`

## Blocked by
Nothing. Sibling lanes of this wave hold other surfaces: the two lab route tests (#421), core's id factory (#429) and the telemetry source vocabulary (#430) are each another issue's, and this one enters none of them.

## Definition of done
- The dispatch of an arm is bounded by a ceiling sized to what it wraps, not by the plumbing probe's. `workmux add` runs the worktree's configured setup — `npm ci` in this repo — so a five-second ceiling cannot succeed on a cold cache and succeeds or fails by the box on a warm one. Follow the precedent prd-55 wave 5 set for the tracker read: a separate exported constant beside the existing one, a doc comment saying why a command that runs a setup is not a plumbing probe, and the wider ceiling used for that spawn only.
- The git plumbing calls this module makes keep the five-second ceiling they have; nothing else moves.
- The hung-process protection the ceiling exists for survives: the dispatch still returns, and still reports, when the spawn never completes.
- Laws, executed: a test reads the ceiling each subprocess actually ran under off the recorded call (the injected exec sees its own `timeoutMs`), pins the two ceilings as distinct numbers with the dispatch's above the plumbing's, and a mutation swapping the two call sites goes red.
- Lane gate green on the branch; every red attributed by failing test name, not by path.

### LINKED ISSUE #429: recorder: createIdFactory restarts at <prefix>-000001 in every process, so one session file carries two events named lab-000001 — 'unique within a session' is no longer true

### LINKED ISSUE #430: prd55: the R&D hand's cost is carried on its events' provenance but not booked as llm.cost — booking needs a third telemetry source

KelliherL:
Closes #430. Closes #429. Closes #408. Refs #421. prd-55 wave 8 — the four findings the earlier waves surfaced and did not do. One PR = one (milestone, wave); four lanes, fifteen lane commits and two conductor commits to the coupling map.

**Why a wave and not four loose fixes:** their fences are disjoint, the queue toll is per PR, and one of them could only be finished once the others were in the same tree — see #429 below.

## What becomes true

**#430 — the R&D hand's cost is booked as spend** (Opus, 7 commits, 10 mutations). Ruling 1 asks for it and wave 5 could not keep it honestly. `llm.cost` may now name the lab as its source, hand-built on the precedent of the lab's own events: a checkpoint the lab *captured* and a cost the lab *booked* are the same kind of fact, neither a reading a collector polled, so `'lab'` stays out of `eventSourceSchema` and out of `TELEMETRY_SOURCES` — the lab books spend and still cannot sign a reading. The engine books one cost per run from the CLI's own `total_cost_usd`, never re-derived, as `auxiliary` (the instrument's research money is not a lane's agent doing that lane's work), with no `sessionId` (an id no collector will corroborate reads downstream as an uninstrumented session — a setup gap that does not exist), and books **nothing** on a run that reports no cost. The fold needed no new arm, and the absence of one is now the law. One consumer must *not* treat it like any other: `selectConnection` gets a guard, not a sixth flow, keeping the promise `CONNECTION_SOURCES`' own comment already makes — a cost the lab booked proves nothing about an external link being wired.

**#429 — an id that cannot collide** (Sonnet, 4 commits, 6 mutations). `createIdFactory` gained an optional writer tag, byte-for-byte unchanged when omitted; then the four writers that share one session file each named themselves — `lab-fork-…`, `lab-checkpoint-…`, `lab-rd-…`, `lab-measure-…`. **EXECUTED:** the exact pair the record contains, the measure route's and the fork's first ids, was `lab-000001` twice and is now `lab-measure-000001` and `lab-fork-000001`. `liveEventKey`'s composite key is unchanged and its comment moves from *still load-bearing* to *belt-and-braces* — insurance against a future writer that omits its tag, never a fix for a collision today's four can produce.

**#408 — a dispatch ceiling sized to what it wraps** (Sonnet, 1 commit, 1 mutation). `workmux add` runs the worktree's configured `npm ci`, so the five-second plumbing ceiling could not succeed on a cold cache and succeeded or failed by the box on a warm one — every launch died `exit null` at arm 1 in the walkthrough that found this. It now has its own two-minute ceiling; `workmux path` keeps five seconds, the restore path keeps its own, and the test reads the ceiling each subprocess *actually ran under* rather than which constant the source mentions.

**#421 — the hypothesis is disproven, and the issue stays open** (Sonnet, 1 commit). The lane could not reproduce the flake, so it went after the mechanism: against the real installed Fastify 5.10.0, with a plugin's registration delayed 0–100 ms, `inject()` already serialises on `ready()` — the readiness race named in that issue **cannot** be the cause, and its body now says so at the top so nobody re-chases it. What landed is a diagnostic upgrade, not a repair, and is not claimed as one: the three cases assert the refusal's *body* from the exported header constant, so a recurrence tells us **what answered**. #421 stays open with its cause unknown, which is why this PR only refs it.

## Verified

- **EXECUTED on the integration branch: `npm run typecheck` 0 · `npm run lint` clean (1057 files).**
- **Attribution by failing TEST NAME against a same-worktree baseline**, which this wave adopted after a sibling PR showed a known-red *file* hiding a new red *test*: revert the touched files to `origin/main`, run, restore, run, diff the names both ways. Every lane did it; the integration run did it again over `server/src/lab` — 13 failing names on both sides, **empty in both directions**.
- **Twenty-two mutations across the four lanes**, each shown red then restored. The ones worth naming: folding the lab record into a connection flow instead of skipping it (only the two new connection laws go red); `source: 'lab'` → `'otel'` read from the *route's* end, which the count pin alone would have missed; dropping any one writer's tag, which reddens naming that writer.
- **Fences:** each lane's exact, plus four recorded widenings — two measured by #430's lane (the connection guard, the route's count pin), one deferred and then unblocked by this PR's own integration (#429's wiring), one deferred out (see below).

## Two coupling entries, from what the lanes measured

`97d048ec` records that **a widened source enum is a different fan-out from a new event type** — the three hand-enumerated event-*type* lists are untouched by it, and what it reaches instead is every exhaustive `Record` keyed by that enum (today `selectors/connection.ts`, which had no entry) and the count pins on a producer whose event list grows.

## Deferred, with reasons

- The guide's `fork-exec-ceiling` claim now reads stale, **and its law still passes** because the check greps for the unchanged 5 s constant rather than the sentence's content — the guide law's own blind spot, recorded on #408. The sentence and its check are the conductor's follow-up once #449 lands, which holds both files.
- #430's dated line under ruling 1 waits on #449 for the same reason.
- One nit for the reviewer, not fixed here: `lab/rd.ts`'s module doc still describes its factory as `createIdFactory('lab')`, which is now inexact by one argument.


KelliherL:
**CI green on `bc242db2`** — ten legs, all pass, both Windows legs and `windows-suite` among them.

ciaran-slow (commented):
Read commit-by-commit (17 commits, four lanes). The engineering is careful and most of it verified clean. **One doc claim is now too strong** and wants a sentence, plus the conflict resolution below. I pushed the nit you left for the reviewer.

- **The sibling hunt on the widened source enum came back clean.** The coupling entry claims the fan-out is "every exhaustive `Record` keyed by that enum — today exactly one, `selectors/connection.ts`". I swept every consumer of `TelemetryOrigin` / `telemetryOriginSchema` across `packages/`: `state.ts`, `fleet/plumbing.ts`, `spend.ts`, `spend-cursor.ts`, `connection.ts`. All but `connection.ts` use `Set<TelemetryOrigin>`, arrays, or a membership predicate — none is an exhaustive `Record`. **The claim holds.**
- **`fromOrigin` was the one I expected to be wrong, and it is right.** `filter.origins === undefined || filter.origins.includes(origin)` is a plain allowlist, so a lab-booked cost counts when nothing is filtered and is excluded when a caller names the collectors explicitly. No silent misclassification into a collector bucket.
- **The `'lab'`-stays-out-of-`EventSource` design is consistent where it matters**: `telemetryOriginSchema` widens, `LLM_COST_SOURCES` is its own tuple, `TELEMETRY_SOURCES` keeps the collector pair, and `llmCostEventSchema` is hand-built with the reason stated. The lab books spend and still cannot sign a reading, as advertised.
- **`createIdFactory`'s backward compatibility is real** — omitting `writer` is byte-for-byte unchanged, and the test that pins it asserts the *bare* shape rather than a defaulted tag.
- **Full suite on the merged `#449 + #450` tree: 499 files, 9386 passed, rc=0**, `npm run typecheck` rc=0, `npm run lint` clean (1057 files).
- **9 of 10 CI legs pass**; `windows-suite` was still pending when I looked, re-triggered by my push.

## Pushed — the nit you flagged

`7e46c824` — `rd.ts`'s module doc now names the writer tag its factory actually passes. Verified: `rd.test.ts` 39/39, lint clean.

While there I checked the obvious sibling, `api/lab.ts:164`, which also says `createIdFactory('lab')`. **It is correct as written** and I left it alone: that passage is narrating the historical collision and closes itself sixteen lines later with "can no longer produce the pair this paragraph describes". Recording the non-finding so nobody else re-opens it.

## Finding — non-blocking · EXECUTED · the "no longer collides" claim is stronger than the fix

`api/lab.ts`'s comment says the four tagged writers "**can no longer produce the pair this paragraph describes**" and "nothing among these four writers can hand it a collision to resolve any more."

The writer tag removes collisions **between different writers**. It does not remove a writer colliding **with itself**, because all four construct their factory inside a per-invocation function:

| writer | site | events per call |
|---|---|---|
| `dispatchFork` | `lab/fork.ts:421` | factory built per call |
| `captureCheckpoint` | `lab/checkpoint.ts:79` | factory per call, **exactly one** `nextId()` |
| `recordRdOverride` | `lab/rd.ts` | factory per call |
| `measureExperiment` | `api/lab.ts:1348` | factory per call |

`captureCheckpoint` is the cleanest case: one fresh factory, one id drawn, so **every checkpoint capture in a session is `lab-checkpoint-000001`**.

**EXECUTED**, on the merged tree, against the structurally identical override path:

```
PROBE2 override ids: ["lab-override-000001","lab-override-000001"]
```

That run had the tag applied and still collided — which is the whole point: the tag is not what fixes this.

**This is not a regression and I am not blocking on it.** Before #429 all four minted bare `lab-000001` and collided with each other *and* with themselves; you have strictly removed one of those two axes, and the composite `liveEventKey` still covers the remainder. What needs changing is one sentence — the guarantee is "no two *different writers* collide", not "no collisions remain". Given the comment explicitly frames the composite key as "belt-and-braces now, not load-bearing", the overclaim is load-bearing prose and worth a minute.

Related: **#445 introduces a fifth, untagged call site** — `createIdFactory('lab')` in `recordOverrideIfNeeded` — which post-merge sits at `api/lab.ts:1065` beside your tagged `measure` one. It is the "future call site that omits its own `writer` tag" your own comment anticipates, arriving in the same landing window. I have blocked #445 on it.

## Blocking for landing, not for approval — the conflicts

`#450` was built without `#445`/`#449` and conflicts with both. **EXECUTED:**

```
CONFLICT (content): Merge conflict in .swarm/coupling.txt
CONFLICT (content): Merge conflict in packages/server/src/lab/rd.test.ts
```

Both additive. Resolution, verified:

```js
expect(result.refusals).toEqual([{ patternId: 'pattern-one-off', reason: RD_HELD_BACK_REFUSAL, rawResult: document }])
expect((await recordedEvents()).map((event) => event.type)).toEqual(['rd.patterns', 'rd.refused', 'llm.cost'])
```

…and keep both coupling entries. Result: 128/128 on the two files, then the full suite green as above. **Order: #445 → #449 → #450.**

## Checked and accurate — the deferred guide gap

Your "the `fork-exec-ceiling` claim reads stale and its law still passes" is **true**, and the law is 61/61 on the merged tree, so it is not a landing blocker. One correction to the mechanism: the claim's `says` regex *does* hold the guide sentence (`/bounded at 5 s\*\*[\s\S]*FORK_EXEC_TIMEOUT_MS/`), so it is not only greping the constant — it passes because both halves remain literally true now that `FORK_EXEC_TIMEOUT_MS = 5000` and `FORK_LAUNCH_TIMEOUT_MS = 120_000` coexist. The sentence is incomplete rather than false, which is why nothing goes red. Same conclusion, slightly different reason.

## Not run, stated as fact

I did not execute a probe against `captureCheckpoint` or `dispatchFork` directly — my attempt nested an `it` inside an `it` and skipped, and I did not rewrite it. The per-call factory shape there is **REASONED** from four lines of source; only the override path is EXECUTED.

PR #453, opened 2026-09-11
prd-49 ruling 3: the tripwire is a count, not a clock — clause 2 retired, clause 1 made a law

### LINKED ISSUE #452: prd49: the tripwire is a law — no selectable stream source composes more than 90 threads
prd-49 ruling 3 (2026-09-12) retired clause 2 — the wall-clock proxy nobody could observe — and made clause 1, the scale the instrument actually renders, the whole trigger. This issue is the law that makes it fire by itself.

## Why a law and not a number in a document

The condition prd-47's NO-GO is conditional on is *the scale the instrument renders*. That is a fact about the product and a pure count, so it can assert. Everything else about this tripwire has been a percentage read from a wall clock on an unnamed box, printed to stdout, recorded nowhere — see ruling 3's audit.

The repo already proves the shape one fixture at a time: `packages/web/src/scene/geometry.test.ts` lays out `fleet20Spec()` and asserts twenty threads. This draws the same assertion once around **every source the operator can select**, not one of them.

## Fence (may touch ONLY)
- `packages/web/src/scene/tripwire-law.test.ts` (new file)

## Blocked by
Nothing. No other lane is in `packages/web/src/scene`. The wall-clock cells in that directory's perf suite are ruling 2's re-measurement recipe and are **not** touched, asserted against, or moved by this issue.
**Model:** sonnet. **Wave:** — (a single issue; prd-49 has no waves by design).

## Definition of done

- A law asserts that **no selectable stream source composes more than 90 threads**, with the threshold a named exported constant and the number appearing exactly once.
- Every selectable source is covered, derived rather than retyped: the sources a reader can actually choose are enumerated in the app's own key map (`STREAM_SOURCE_KEYS` in `app/StreamContext.tsx`, `live`/`fleet20`/`pathology` today), and the law must fail if a fourth is added and not covered — a law that silently skips the new source is the failure mode this one exists to avoid. `live` has no fixture to lay out; say so explicitly in the law rather than omitting it silently, and assert what *is* true of it (the frame loop composes exactly one colony, so its thread count is one repo's lanes and is not knowable here).
- The failure message names prd-49, ruling 3, and this tripwire's issue — the reader who trips it must land on the re-measurement order in ruling 2, not on a bare number.
- The law is shown to bite: a fixture grown past the threshold turns it red (executed by constructing an over-sized spec in the test, never by editing a shipped fixture).
- It is a **count**, with no wall clock anywhere in it: a run of this law on a loaded box and a quiet one gives the same answer.
- Lane gate green; every red attributed by failing test name against a same-worktree baseline of `origin/main`, not by path.

KelliherL:
Closes #452. Refs #190. Two commits: the ruling, and the law that makes it fire by itself.

## The audit that produced the call

prd-49 exists so that prd-47's conditional NO-GO is *watched rather than remembered*. Its clause 2 — *the model floor's composed 30x3 cell crossing a percentage of budget on the reference box* — could never have been watched, and this was checked against the tree rather than reasoned about:

- **Box-dependent, with our two readings moving two variables at once.** The paper's provisional baseline is **76.6 %** (one box, under `--maxWorkers`); a quiet serial reading on another box is **69.2 %** (30x1 32.3 %, 60x3 144.2 %, 180x1 116.3 %). Box *and* load differ, so neither supersedes the other and no honest baseline comes from the pair. That is not a third-run problem: `scene/perf.test.ts`'s own discipline already says a wall clock measures the box, which is why every law in that file is a count.
- **Nothing records the reading.** The cells are reported, never asserted, into stdout. The only timing state the landing gate keeps, `.swarm/timing-count`, is **gitignored** (`.swarm/*`), counts *files* rather than readings, and has never been written in this clone. There is no archive and no ratchet on any number — nobody could have seen a crossing.
- **And after the CI retirement there is no recurring reading at all** — the clock that remains is whoever lands a PR, on their box, writing the number nowhere.

The clause had been re-derived twice (2026-09-07 suspending it, 2026-09-10 re-arming it at ~90 %) without ever being observable. "The reference box" is named in no document, and after #438 there will be no candidate.

## The call — Ruling 3, nothing renumbered

**Clause 2 is retired as a trigger.** The cells stay exactly where they are as **ruling 2's re-measurement recipe**, which is where a clock belongs: to answer the question once something fires, not to notice it.

**Clause 1 — the scale the instrument actually renders — is the whole trigger, and becomes mechanical.** It was always the clause that expressed the condition: the NO-GO is conditional on the scale the instrument renders, which is a fact about the product and a pure count. `TRIPWIRE_THREADS = 90` carries over unchanged, because it was never a reading of a box — it is the halfway cell where the summed floor already read 49.8 % and the curve was steepening.

The live half stays a human observation and says so: the frame loop composes exactly one colony (`layoutWorld([{ id: LOCAL_COLONY, … }])`), so a session's threads are one repo's lanes, visible on the fleet table, and nothing about a stranger's repo is assertable here.

## The law

`packages/web/src/scene/tripwire-law.test.ts` — one new file, no widening.

- Sources are **derived, never retyped**: `FIXTURE_SPEC_FOR` is `Record<StreamSource, FixtureSpec | null>` over the operator's own `STREAM_SOURCE_KEYS`, so a fourth source with no line here fails `tsc` (TS2741, proven by mutation), and a runtime coverage check catches the direction the structural check cannot see — the silent-skip trap.
- `live` is not skipped. It gets its own block saying it has no shipped fixture, and asserts the fact that makes that true: the frame loop still composes one colony.
- The failure message names prd-49, ruling 3 and **#190** — the tripwire's own ticket, not this PR's issue — and states ruling 2's order, so whoever trips it lands on *re-measure first, against a production build*, never on a bare number.

**Verified by the conductor, not only reported:** lowering the threshold to 5 reddens exactly the two shipped sources with their real counts and the full sentence a reader needs —

```
AssertionError: fleet20 composes 20 threads, over prd-49 ruling 3's tripwire of 5
  (the whole trigger; https://github.com/launchpad-26/rhizomorph/issues/190). …
```

— then restores to 8/8. The lane ran four more (dropping a source from the record; filtering the known-source list, which silently dropped the per-source loop from 9 tests to 7 and is exactly the trap; fabricating a two-colony frame loop).

`npm run typecheck` clean, `npm run lint` clean (1058 files), the law 8/8 standalone. Full-suite failing-test-**name** diff against a same-worktree baseline: **empty in both directions**, 239 reds on each side, none of them in `scene`, `app` or `fleet`.

## What this does to #190

Its body has been rewritten to state the live trigger — it had been carrying the original pre-composition table and the superseded *"~60 % on the reference box"* while two comments beneath it held the corrections. The superseded numbers are kept below as history, which is where the number 90 comes from. The issue stays **open**: parked, and now watched by a law rather than by a person.


KelliherL:
**CI green on `6e051159`** — ten legs, all pass, both Windows legs and `windows-suite` among them.

ciaran-slow (approved):
**APPROVE.** The body makes three mechanical guarantees. I reproduced all three by mutation rather than taking the report's word, and each behaved exactly as claimed.

**Baseline.** `npx vitest run packages/web/src/scene/tripwire-law.test.ts` → **8 passed (8)**. `TRIPWIRE_THREADS` appears as a literal exactly once (`grep -n '\b90\b'` → one hit, line 73), so "every other reference reads it back off this constant" is true.

**Mutation 1 — the threshold.** `TRIPWIRE_THREADS` 90 → 5 reddens exactly the two shipped sources, with the full sentence:

```
AssertionError: fleet20 composes 20 threads, over prd-49 ruling 3's tripwire of 5
  (the whole trigger; https://github.com/launchpad-26/rhizomorph/issues/190).
  Ruling 2 binds: re-measure first, against a production build — this law is not
  evidence that the free-list pooling should be built.
AssertionError: pathology composes 9 threads, over prd-49 ruling 3's tripwire of 5 …
```

`2 failed | 6 passed`, restored to 8/8. The counts the PR body quotes (fleet20 20, pathology 9) are the counts I got.

**Mutation 2 — the structural check.** Widening `FixtureId` with a fourth member gives, on `tsc --noEmit` in `packages/web`:

```
src/scene/tripwire-law.test.ts(110,7): error TS2741: Property 'chaos' is missing in type
  '{ live: null; fleet20: FixtureSpec; pathology: FixtureSpec; }' but required in type
  'Readonly<Record<FixtureId, FixtureSpec | null>>'.
```

TS2741 at the claimed line. Restored → `tsc` rc=0.

*Worth recording how this nearly went wrong:* my first attempt ran in a worktree whose `node_modules` was symlinked to another clone, so `@rhizomorph/core` resolved to a **different tree** and the mutation silently did nothing — it read as "TS2741 does not reproduce". It reproduces; the first run was measuring the wrong tree. Flagging it because a mutation that quietly tests nothing is exactly the failure this law exists to prevent, and I walked into it.

**Mutation 3 — the `live` block.** Adding a second colony to `useFrameLoop.ts`'s `layoutWorld` call reddens both `live` tests with the intended message ("…no longer composes exactly one colony… this law's reason for skipping it needs re-deriving"). Restored → 8/8.

**Wider suites.** `packages/web/src/scene`, `packages/web/src/app` and `doc-citation-law` → **49 files, 1171 tests, rc=0**.

**CI.** All **10** legs pass — `build-test-boot` ×3, `pack-smoke` ×6, `windows-suite` — checked as a count against AGENTS.md's three-jobs/ten-legs list, not as a green bar.

**The doc half.** Rulings 1 and 2 are untouched, ruling 3 is appended — nothing renumbered, as claimed. #190 is rewritten to state the live trigger and stays OPEN. #452 is open and closed by this PR.

## The design call, which is the part I'd have argued if it were wrong

Retiring clause 2 is right and the audit supports it: a wall-clock percentage that was never recorded anywhere, whose only state (`.swarm/timing-count`) is gitignored and counts files, could not have been observed by anyone. Keeping the cells as ruling 2's re-measurement recipe rather than deleting them keeps the answer available without pretending it was a trigger. The residual — that fixtures are fixed, so this law can only fire when someone grows a shipped spec — is not a weakness: that *is* "the scale the instrument renders", and the live half is correctly left as a human observation that says so.

## Not run, stated as fact

I did not run the full suite against this branch in isolation; scene/app/doc-citation were run directly and CI covers the rest across ten legs.

## Nothing to fix

No findings.

Issue #427, opened 2026-09-10
prd56 w1: a finished PRD sits on the archival shelf, and a law says so


gabriel-canaan:
## Fence (may touch ONLY)
- `docs/prds/prd-25-the-third-platform.md`
- `docs/prds/done/prd-25-the-third-platform.md`
- `docs/prds/prd-27-the-declared-voice.md`
- `docs/prds/done/prd-27-the-declared-voice.md`
- `docs/prds/prd-43-the-claim-is-a-test.md`
- `docs/prds/done/prd-43-the-claim-is-a-test.md`
- `docs/prds/prd-50-the-fixed-ceiling.md`
- `docs/prds/done/prd-50-the-fixed-ceiling.md`
- `docs/prds/prd-52-the-world-composes.md`
- `docs/prds/done/prd-52-the-world-composes.md`
- `AGENTS.md`
- `docs/architecture.md`
- `docs/roadmap.md`
- `docs/prds/prd-30-the-open-hand.md`
- `docs/prds/prd-51-the-split.md`
- `docs/prds/retained-prds-review-2026-08-22.md`
- `.swarm/prd-milestones.txt`
- `scripts/dev/prd-milestones.sh`
- `scripts/dev/prd-milestones.test.sh`
- `packages/server/src/prd-location-law.test.ts`

Both halves of each move are claimed because a rename audits as a change to
each path.

**Fence widened 2026-09-11, before any change was made.**
`docs/prds/retained-prds-review-2026-08-22.md` is a SIXTH citer, found by
re-deriving the list at commit cea5a361. It links all five moved PRDs by bare
relative filename (`[25](prd-25-the-third-platform.md)`), so all five links
break on the move, and no law catches it: the citation law excludes
the PRD directory as a citing source and skips markdown-link hrefs. The operator
ruled that the links are re-pointed at the archival directory rather than left dead. Those
five links are the ONLY thing in that file this issue may touch; its table
rows, verdicts and prose are EXPLICITLY OUT.

The last four paths are new files. Their locations were fixed at grooming
rather than left to the lane, because wave 2 must name the law's file in its
own fence and "wherever wave 1 put it" is the vague boundary the wave lint
refuses. They follow the conventions already in the tree: a law beside its
siblings under the server package, a plain-text manifest beside coupling.txt
and lanes.json, and a generator with a shell test sibling under the dev
scripts directory.

- `.gitattributes`

## Blocked by

[...]

PR #454, opened 2026-09-12
prd51 w5: a recording seals itself into an archive, and the team server comes up from one image

### LINKED ISSUE #432: prd51 w5: seal, archive, verify, tombstone and prune are one local command
## Fence (may touch ONLY)
- `packages/server/src/log/archive.ts` (new)
- `packages/server/src/log/archive.test.ts` (new)
- `packages/server/src/cli/archive.ts` (new)
- `packages/server/src/cli/archive.test.ts` (new)
- `packages/server/src/cli/index.ts`
- `README.md`
- `packages/server/src/log/transcript-capture.ts`
- `packages/server/src/log/transcript-capture.test.ts`
- `packages/server/src/log/lane-index.ts`
- `packages/server/src/log/lane-index.test.ts`
- `CHANGELOG.md`

## Blocked by
Nothing — #425 landed prd-51 wave 4 and this wave's base is `main`.

**Prohibition, from ruling 11 itself.** The command *composes* the steps against the retention
grammar and does not edit it. `packages/server/src/log/retention.ts` is read and called, never
modified — if it needs a change, stop and surface it rather than absorbing it.

Sibling lanes hold other surfaces this wave: the deploy-image lane owns `packages/team/`; the
shipper-scan lane owns `packages/server/src/shipper/`; the doctor lane owns
`packages/server/src/server/context.ts` and `packages/server/src/api/doctor.ts`; the
ADR-citation lane owns `docs/architecture.md`.

This lane is the wave's **single `README.md` claimant**. `.swarm/coupling.txt` allows one per
wave, and `cli-surface-law.test.ts` holds `cli/index.ts`'s dispatch table and README's CLI
reference to the same subcommand set **in both directions**, so a new top-level command forces
both files.

**Model:** opus. **Wave:** 5.

## Definition of done
Ruling 11, whole. Each line below is a test that could fail.

- `rhizomorph archive` builds the portable record for a candidate session, gzips it,
  `verifyRecord`s the archive, writes the tombstone manifest at the transcript-capture path
  **while the log still exists**, and only then calls `applyRetentionPlan` for that candidate.
  A case observes the tombstone on disk *before* the log is gone — the ordering is the ruling.
- **Prune is not reachable from this command by any other route.** Removing the archive step
  must not leave a path that still prunes.
- The manifest's lane list is cross-checked against the log before the log goes, so a tombstone
  can never conjure a lane that never ran.
- Lanes only git knows about — a worktree with commits and no instrumented agent — are either
  named by widening the writer's lane source to `laneHandlesOf` (`log/lane-index.ts:354`), or
  the command **says out loud** that it cannot name them. Either answer is acceptable; which one
  was taken is stated in the commit and pinned by a test.
- `attributedFrom` gains `'tombstone'` as a declared value. It is declared once
  (`log/transcript-capture.ts:61`) and no exhaustive switch reads it, so this is additive —
  confirm that before assuming it.
- On Windows the command answers `EBUSY` by refusing the candidate **with the holder named**,
  never by pretending. Witnessed by injecting the fault, so the case passes on every platform.
- `README.md`'s CLI reference gains the `rhizomorph archive` row and `cli-surface-law.test.ts`
  is green in both directions.
- **What mutation would each new test survive?** Answer per case in the PR body. The ordering
  cases are the ones that matter: a test that passes whether or not the tombstone precedes the
  prune has not tested ruling 11.
- Falsifier: if the tombstone cannot be written before the prune without editing
  `log/retention.ts`, ruling 11's "composes without editing it" is wrong and needs an
  amendment — stop and say so rather than widening into that file.

---

## Two grooming corrections, recorded before dispatch (2026-09-11)

Both found by the plan pass on this issue, and both are defects in this issue as first written.

**1. CHANGELOG.md is added to the boundary above — a widening, recorded before the change.** A new
top-level subcommand is a user-visible change and AGENTS.md requires an Unreleased entry for one.
The CHANGELOG's own semver policy makes the CLI's subcommand set the breaking-change contract,
which the cli-surface law's docblock cites as the reason that law exists. The original list named
the README and the CLI index and missed the third file the same change forces.

Checked rather than assumed: **no test enforces a CHANGELOG entry.** The word appears exactly once
under packages/server/src — inside the cli-surface law's docblock prose, never in an assertion;
that law reads the CLI index and the README only. So the gate would have stayed green with the
entry missing, which is precisely why it needed declaring rather than leaving to be noticed.

**2. The Windows known-failures clause was unsatisfiable and is struck.** It asked for an entry in
that file if the suite could not witness EBUSY on the runner. Two independent reasons that cannot
be done, and they point the same way:

- that file is not in this issue's declared list, and no lane this wave claims it;
- it is an expected-FAIL list. AGENTS.md: *"a failure outside the list is red, a listed file that
  passes is a removal candidate."* The EBUSY case injects its fault, so it passes on every
  platform — an entry would be reported as a removal candidate on the next Windows run, and each
  entry needs a measured-on-windows-latest provenance line no lane can produce.

Replaced with what the case actually is: the fault is injected, so it is witnessed everywhere and
needs no platform carve-out.

### LINKED ISSUE #433: prd51 w5: one image, Postgres and Caddy come up from init.sh with the key printed once
## Fence (may touch ONLY)
- `packages/team/deploy/` (all files, new)
- `packages/team/Dockerfile` (new)
- `docs/team-server-runbook.md` (new)
- `.gitignore`
- `packages/team/tsconfig.json`

## Blocked by
Nothing in code — #355 landed `packages/team` and its migration runner, #373 landed the ingest.
Wave 0's operator acts are done: the VPS is `rhizomorph.devacademy.life` and the org is
`rhizomorph-team` (the 2026-09-08 amendment records both).

**Prohibition, and it is the reason this issue is in wave 5 at all.** This lane adds and edits
**no path under `packages/team/src/`**. #410 (prd-51 wave 4) is live in that subtree and holds
the storage port and the fold. If the image cannot come up without a source change, stop and
surface it — see the falsifier.

**Prohibition.** `docs/adr/0035-the-watcher-is-never-a-container.md` is accepted and is read,
not edited; ruling 13's decision is already recorded there. No new architecture record: the ADR
log is append-only and numbered, so only one issue per wave may claim the next number and its
index row, and no issue this wave has claimed it.

Sibling lanes hold other surfaces: the archive lane owns `README.md` and
`packages/server/src/cli/`; the ADR-citation lane owns `docs/architecture.md`.

## Definition of done
Ruling 13, minus what wave 0 parked.

- `packages/team/deploy/compose.yml` brings up one app image plus Postgres plus Caddy, and
  nothing above that rung.
- Migrations run on boot from the SQL files tracked in `_migrations`. A second boot applies
  nothing **and says so** — a silent no-op and a successful re-apply are indistinguishable to
  an operator, which is the failure this line exists to prevent.
- `init.sh` generates secrets and prints the first ingest key **exactly once**. A scripted run
  shows the key absent from a second invocation's output, and shows where it is written and
  with what mode.
- **Restore runs into the empty database *before* the app's first boot.** The reversed order
  produces `duplicate key` errors — the install spike found this — and the ordering is what
  wave 6's drill will time. The runbook states it and the drill proves it.
- **Backups are parked and the runbook says so in those words.** No off-VPS destination was
  provisioned, so a dump written to the same machine as the database is not a backup and this
  repo must not record a green backup drill against one. The runbook states that the deployment
  has no recovery story until a destination exists.
- The runbook records that **the Linode account belongs to the academy, not the cohort** — the
  team holds root on the machine but not the account, so the instance can be resized, rebuilt or
  removed without notice. Recorded so it is not rediscovered during the drill.
- The server has a `doctor` in the local one's discipline is **NOT** this issue — that is
  ruling 12's second half and belongs to a later wave with `/connect`'s row.
- Falsifier, and its verdict must be recorded either way: if the image cannot come up without a
  change under `packages/team/src/`, this issue is mis-scoped and belongs in wave 6 behind #410.
  Say so and stop; do not widen into a live lane's subtree.

---

## Boundary widening — recorded before the change (2026-09-11)

Two paths added above, both found by the plan pass, and both outside the team package's src
directory — so the prohibition that put this issue in wave 5 is untouched: #410 is still the only
lane in that subtree.

**The team package's tsconfig.json.** Verified rather than taken on trust: it reads
include ["src"] with rootDir "src". So anything added under deploy/ is **invisible to
npm run typecheck** — this lane could ship a type error in its own new code and every gate would
stay green. That is the same shape as the CHANGELOG gap on #432: a file the change forces that
nothing would have complained about. Note for the builder, since the two settings move together:
rootDir is src, so widening include alone is not enough or tsc refuses.

**.gitignore.** init.sh generates a dotenv file holding a generated secret, and nothing in the
repo ignores one today (checked). An unignored generated secret in a public repo is one
`git add -A` away from being committed, and this repo has already recorded that `git add -A` is
the wrong habit here.

Neither path appears in the coupling registry, and the pre-dispatch lint passes over this wave and
every live lane with both additions in place.

## The scoping question this issue was filed with is answered: correctly scoped

Its own falsifier asked what to do if the image could not come up without a change under the team
package's src directory. It can. That package's public barrel already exports startTeamServer,
bootstrapTeamStorage and resolveTeamConfig — confirmed by reading the barrel on origin/main, not
inferred — which is everything a deploy entrypoint needs. No widening into a live lane, and the
issue stays in wave 5.

## One honesty clause the plan adds, and it should survive review

The team server does **not verify ingest key values yet** — that is membership work, unbuilt. So
init.sh generates a bearer secret for future shippers rather than a working credential, and the
runbook must say so. The fold worker and the doctor are likewise deliberately not wired up. Nobody
should be able to read *"the image comes up"* as *"ingest is end-to-end"*, and the deliberate
exclusions are named for that reason.

### LINKED ISSUE #434: prd51 w5: the oversized scan does not report a stale file as unterminated
## Fence (may touch ONLY)

## Blocked by
Nothing — #425 landed the oversized-line arm this refines, on `main` as `a3c5b305`.

**Prohibition.** `packages/server/src/shipper/hand-law.test.ts` and its `DECLARED_IMPORTERS` set
are read, never widened. `packages/server/src/shipper/cursor.ts` belongs to nobody this wave and
needs no change — if it does, stop and surface it.

Sibling lanes hold other surfaces: the archive lane owns `packages/server/src/log/` and
`packages/server/src/cli/`; the doctor lane owns `packages/server/src/api/doctor.ts`.

## Definition of done
Recorded as a deferral in #425's body, **labelled REASONED and never executed** — that is the
first thing this issue has to fix.

- The claim: `findLineEnd` holds the handle open across many reads without re-checking that the
  file has not changed underneath it, unlike the first read, which `bytesRead !== wanted`
  guards. A rotate, prune or archive landing mid-scan could produce a spurious
  `'oversized-unterminated'` report for one tick.
- **Decide whether it is real, before changing anything.** Proving it needs a concurrency
  harness `ship.test.ts` does not have. Build one and reproduce the spurious report, or state
  with evidence why the interleaving is unreachable and **close this issue saying so**. A close
  with a recorded reproduction attempt is a result; a fix with no reproduction is not.
- If it is real: the scan re-checks, and removing the re-check reddens a case that observes the
  spurious report. If it is not: nothing is edited and the finding is recorded as refuted.
- Self-healing is not a reason to leave it unanswered. A failure never writes the cursor, so
  the tick recovers — but it reports a wedge that is not one, and an operator acts on that.
- **What mutation would the test survive?** If the answer is "none, the harness cannot fail",
  the harness is not evidence and the honest outcome is the refutation above.

### LINKED ISSUE #435: prd51 w5: the doctor route reads the data root it was given
## Fence (may touch ONLY)
- `packages/server/src/server/context.ts`

## Blocked by
Nothing — #425 landed the shipper check that this threads a root through, on `main` as
`a3c5b305`.

**Prohibition.** No new route and no change to any route's class. `packages/server/src/api/index.ts`
is not touched, so `api/route-class-law.test.ts` and `packages/contract/src/contract-coverage-law.test.ts`
stay out of this lane — both pin exact counts in both directions and neither is claimed this wave.

Sibling lanes hold other surfaces: the archive lane owns `packages/server/src/cli/index.ts` and
`packages/server/src/cli/cli-surface-law.test.ts`, so this lane changes no subcommand surface and
needs neither; the shipper-scan lane owns `packages/server/src/shipper/`.

## Definition of done
Recorded as a deferral in #425's body.

- `registerDoctorRoute` passes no `dataRoot`, so through `buildApp` the shipper check reads the
  process-default data root rather than the one the server was given. It is given one.
- **The sibling case is `checkSessionBoundary`, one line above, and it shares the defect.** Both
  are fixed together or neither is — a fix that threads a root to the new check and leaves the
  older one reading the process default is the same bug with one witness removed.
- `RHIZOMORPH_DATA_DIR`'s override (`log/paths.ts:6`) reaches the HTTP doctor exactly the way it
  reaches the CLI one.
- The `buildApp` case asserts a **real status against a planted root**, not merely presence.
  Today it can only assert presence and the never-value law, because the status is
  machine-dependent — that is the thing this issue removes.
- The never-value law still holds: no ingest key value, and no fragment of one, appears in the
  serialised body, and the case still asserts the check ran at all. Without that last assertion
  every negative passes on a build where the check never ran.
- **What mutation would the test survive?** At minimum, reverting the thread-through must redden
  a case that reads a planted root. If it does not, the case is asserting presence again.

ciaran-slow:
Closes #432
Closes #433
Closes #434
Closes #435

prd-51 wave 5, assembled from four lanes. Base `main` at `fada7de1`. Four commits, one per issue,
cherry-picked with **no conflict** — which is the wave's own fence audit: nothing here touched a
file anything else here touched.

Two pieces of ruling-shaped work — ruling 11's local archive command and ruling 13's container
image — plus the three deferrals PR #425's body parked *for this grooming*, because filing them
needed a milestone that did not exist yet. One of those three (#436, the ADR citation) left the
wave before dispatch and sits in Backlog behind #427, which holds `docs/architecture.md`; that is
recorded on the issue.

---

Every lane was verified on a **different model from its builder**, each verifier confirming the
`Co-Authored-By` trailer itself rather than accepting the claim, and each running all gates in its
own clean worktree rather than reading the build report's numbers.

|---|---|---|---|---|---|
| `995d5176` | #432 | Opus 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/432#issuecomment-5629845586) | [report](https://github.com/launchpad-26/rhizomorph/issues/432#issuecomment-5641408485) · [count correction](https://github.com/launchpad-26/rhizomorph/issues/432#issuecomment-5641524484) | [**ready** — Sonnet](https://github.com/launchpad-26/rhizomorph/issues/432#issuecomment-5641517308) |
| `8ea87ad7` | #433 | Sonnet 4.5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/433#issuecomment-5629876453) | [report](https://github.com/launchpad-26/rhizomorph/issues/433#issuecomment-5641844461) · [fixes](https://github.com/launchpad-26/rhizomorph/issues/433#issuecomment-5642031404) | [changes](https://github.com/launchpad-26/rhizomorph/issues/433#issuecomment-5641960084) → [**ready** — Opus](https://github.com/launchpad-26/rhizomorph/issues/433#issuecomment-5642063573) |
| `eb752702` | #434 | Sonnet 4.5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/434#issuecomment-5629850126) | [report](https://github.com/launchpad-26/rhizomorph/issues/434#issuecomment-5641175695) · [correction](https://github.com/launchpad-26/rhizomorph/issues/434#issuecomment-5641387777) | [ready](https://github.com/launchpad-26/rhizomorph/issues/434#issuecomment-5641304793) → [**re-read** — Opus](https://github.com/launchpad-26/rhizomorph/issues/434#issuecomment-5641500699) |
| `5da19bf8` | #435 | Sonnet 4.5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/435#issuecomment-5629848996) | [report](https://github.com/launchpad-26/rhizomorph/issues/435#issuecomment-5641148396) · [fix](https://github.com/launchpad-26/rhizomorph/issues/435#issuecomment-5641582908) | [changes](https://github.com/launchpad-26/rhizomorph/issues/435#issuecomment-5641495086) → [**ready** — Opus](https://github.com/launchpad-26/rhizomorph/issues/435#issuecomment-5641753394) |

**The verify records name the LANE shas, not the shas in this PR.** Cherry-picking rewrites them,
and this repo's condition 1 is per-sha, so that gap is closed by content rather than by assertion:
`git patch-id --stable` on each pair is **identical** for all four (`1581d4fe`→`995d5176`,
`72c65e01`→`8ea87ad7`, `4aefd15e`→`eb752702`, `52417594`→`5da19bf8`). What was verified is what is
here, byte for byte.

**Three lanes were sent back after their first verify** and re-read at the new sha afterwards, so
every amend has an independent read against it. The reasons are in the *what went wrong* section
below rather than buried, because they are the most useful thing in this PR.

---

## `995d5176` — #432: seal, archive, verify, tombstone and prune are one local command

Ruling 11, whole. `rhizomorph archive [path] --older-than <dur> [--out-dir] [--handle] [--dry-run]`.
Per candidate: `buildRecord` → gzip → write the `.gz` → **verify it read back off disk** →
cross-check and merge the tombstone manifest **while the log still exists** → `applyRetentionPlan`
for that one candidate.

**Ruling 11's own falsifier did not fire.** `log/retention.ts` is read and called, never modified —
the command composes against its published grammar exactly as the ruling requires.

**The ordering law cannot be tested by end state, and this is the interesting part.** The tree is
byte-identical whether the tombstone is written before or after the prune, so no post-hoc assertion
can test ruling 11. Three witnesses instead, and the decisive mutation *moves* the manifest write to
just after the prune — both writes still happen:

| witness | reddens under the moved write? |
|---|---|
| T1 — `{logExists, tombstoneExists}` observed **inside** the prune seam | **yes** |
| T2 — prune throws EBUSY, the manifest must still stand | **yes** |
| T7 / T8 / T9 / T11 — end-state assertions | **no** — and that is the proof |

Verified independently: the reviewer moved the write itself and got exactly T1 and T2 red, 101/103
green across the five touched test files, then restored and sha256-checked byte-identical.

**Prune is reachable by no other route.** `log/retention.ts` had **zero importers** on `main` before
this commit, so the shipped-source importer set is now exactly `{log/archive.ts}`. Two laws pin it —
T4a over shipped source, T4b including tests — and both redden when a direct import is planted in
`cli/archive.ts`. The reviewer re-grepped the whole repo and confirmed one route.

**What is the sibling case?** A **corrupt archive**. Verifying the in-memory record rather than the
bytes on disk would pass while shipping an unreadable `.gz`; the verify step re-reads the file, and a
`compress` seam producing a tampered gzip is what makes the word "archive" in ruling 11 mean anything.

**What mutation would this test survive?** Every mutation in the table on the issue reddened; none
survived. **Read the count off that table, not off the report's summary line** — the summary says 26
where the table has 28 rows and 29 distinct ids, an arithmetic slip corrected on the issue.

**Deviations, five, none structural:** `busyRefusal` gained an `archivePath` param; T13 uses a stale
lock because a **live** lock excludes the session from the plan entirely, so the busy branch was
unreachable as the plan wrote it; the T4 split above (one half **stronger** than the plan asked);
numeric session ids; `voiceOutcome`/`voiceArchiveRun` exported for assertability. All checked
against source by the reviewer and confirmed not to be quiet narrowings.

**Boundary:** 12 files. `CHANGELOG.md` was added before dispatch (grooming correction on the issue)
because a new subcommand is user-visible — and note **no law enforces that entry**, so the gate
would have stayed green without it.

## `8ea87ad7` — #433: one image, Postgres and Caddy come up from `init.sh`

Ruling 13, minus what wave 0 parked. `packages/team/deploy/` + `Dockerfile` + a runbook.

**The scoping falsifier resolved the good way, and it was settled by doing.** The issue asked what to
do if the image could not come up without touching `packages/team/src/` — where **#410 is a live
lane** (PR #446). It can: the package's public barrel already exports `startTeamServer`,
`bootstrapTeamStorage` and `resolveTeamConfig`. Confirmed by building and booting the image, not by
reading exports. **Nothing under `packages/team/src/` moved.**

**Ruling 13's ordering claim is executed, not argued.** Reversed restore order → `duplicate key`
plus 98 `already exists` (116 total); correct order → **zero** of either, and the next boot reports
`0 applied, 4 already applied … nothing to do` against the restored schema. The reviewer re-ran the
whole drill under an isolated compose project with `down -v` between phases so no cached volume could
fake a pass, plus a `--no-cache` build to prove `npm ci` and `tsx` hoisting work from cold.

**The drill found a bug the plan had wrong**: the plan's `postgres_data:/var/lib/postgresql/data`
mount crash-loops on `postgres:18.4-alpine`, which now wants `/var/lib/postgresql`. The reviewer ran
the plan's literal version to confirm the crash rather than accepting the upstream citation.

**The honesty clauses are the deliverable as much as the code**, and were checked as such. The
runbook states there is no recovery story until an off-VPS destination exists, records that the
Linode account belongs to the academy rather than the cohort, and names three deliberate exclusions —
ingest keys are **not verified yet**, the fold worker and the team doctor are **not wired up**. Nobody
should read "the image comes up" as "ingest is end-to-end". No green backup drill is recorded anywhere.

**What is the sibling case?** The one this lane first got wrong, and it is a secret path. `init.sh`
writes both secrets to `$ENV_FILE.tmp.$$` before the atomic `mv`, and that path was **not** ignored —
so the atomicity mechanism that protects `.env`'s contents created a second secret-bearing file, in a
public repo. Fixed. The reviewer proved the fix by shimming `mv` to fail so the **real script** left a
real 348-byte orphan holding all three secrets, then confirmed `git status` clean and
`git check-ignore` naming the rule.

**What mutation would this test survive?** A silent key rotation on the idempotent path is caught
**only** by the dotenv byte-identity assertion — a naive "old key absent from stdout" check passes
against it. Confirmed by both builder and reviewer. `chmod 600` and the applied-names list also
redden when broken.

**Boundary widening, recorded before the change:** `.gitignore` and `packages/team/tsconfig.json`.
The tsconfig one was necessary, not cosmetic — `include: ["src"]` meant everything under `deploy/`
was **invisible to typecheck**; the reviewer planted a type error and confirmed it is caught with the
shipped tsconfig and silent with `main`'s.

**A known gap, documented rather than hidden:** a plain `pg_dump` restored before roles exist yields
16 errors across three roles (`rz_viewer` ×8, `rz_readonly` ×4, `rz_ingest` ×4) with all 12 `GRANT`
statements among them, so a restored database has no privileges for any app role. The RLS policies
reference roles created by an out-of-fence migration. `pg_dumpall --globals-only` is named as the
available, unbuilt fix path; building it is backup work and backups are parked.

## `eb752702` — #434: the forward scan re-checks size before naming a wedge

**This issue was licensed to close as REFUTED and did not need to.** The claim came from #425's body
labelled **REASONED and never executed**, so the bar was: reproduce the spurious
`'oversized-unterminated'` against **unmodified** `ship.ts` first, or apply no fix.

It reproduced. Both race tests failed with exactly that report; the no-truncation control passed. The
reviewer re-established it independently by reverting `ship.ts` byte-for-byte to base while keeping
the tests. **The fix is licensed by evidence, not by argument.**

`findLineEnd` now returns `found` / `not-found` / `stale`, and any short read is `stale` rather than
falling toward `not-found`, taking the existing silent-wait path.

**What is the sibling case?** A **genuine** unterminated oversized line, which must still fail loudly
— a fix that converted a real permanent wedge into an endless silent wait would be worse than the
bug. Silencing `'not-found'` reddens four tests, three of them pre-existing.

**What mutation would this test survive?** Originally answered *"none"*, and **that was false** —
narrowing the trigger from `bytesRead < wanted` to `bytesRead === 0` survived all 26 tests, because
neither race test exercised a second read against different data. A fourth test closes it: a
shrink-then-**regrow** mid-scan, where a too-permissive scan reads across the seam and reports a skip
stitched from two files. On the final 27-test tree — fixed 27/27; narrowed → **exactly 1 red**, the
new test; full pre-fix revert → 3 red. The correction is on the issue.

**Deviation, gate-forced:** the plan typed the mock via `Parameters<typeof handle.read>[3]`, which
fails `tsc` (TS2493) — `FileHandle.read`'s overloads resolve that to the options-object form. Retyped
explicitly against the positional overload; the reviewer reproduced the error verbatim before
accepting it. Typing only.

**Boundary:** two files. No widening.

## `5da19bf8` — #435: the doctor route reads the data root it was given

`registerDoctorRoute` set no `dataRoot` in either options literal, so **both** `checkSessionBoundary`
and `checkShipper` fell through to the process default rather than the root the server was booted
against. `dataRootFor(ctx)` derives it from `path.dirname(ctx.sessionDir)` — no new stored field,
which is what keeps the change inside a fence that excludes `cli/run.ts`.

**What is the sibling case?** `checkSessionBoundary`, one line above the check the issue was filed
for. Fixing one and leaving the other is the same bug with one witness removed, so a single test
plants distinguishable state for both and asserts both in one request.

**The derivation is indirect and was checked against every caller**, not just the common one. The
reviewer enumerated all three `buildApp` callers and found a counterexample the build missed:
`packages/contract/src/harness.ts` builds a bare `mkdtemp` sessionDir with `readOnly` unset, so the
checks run and derive `os.tmpdir()`. Benign — that path has never existed, so both checks report
no-state defaults, which is strictly more hermetic than reading a real home root — and the doc now
names it.

**What mutation would this test survive?** Deleting the first `dataRoot` literal reddens 2; deleting
**only** the retarget literal reddens exactly 1; making `dataRootFor` return `sessionDir` instead of
`dirname` reddens both, so the derivation itself is asserted. Leaking a 12-character key fragment
reddens 4 — including the retarget test, which is the fix to the second finding below.

**Boundary:** three files touched of five fenced. `cli/doctor.ts` and `cli/doctor.test.ts` were in
the fence and deliberately left alone.

**Recorded, not fixed:** `api/retarget.ts` already has a private `dataRootOf(ctx)` with a
byte-identical body, so `dataRootFor` is a second copy. Out of fence; someone's follow-up.

---

## What went wrong, and it is one thing four times

Three lanes were sent back after their first verify. Every finding was the same shape: **a written
claim its own evidence did not support.** Not code defects — the code was substantially right each
time. Defects in reports, comments and docblocks: the artefacts a reviewer reads *as* evidence.

- **#435a** — the retarget call site had no test that could fail for it, and the report offered *"no
  existing test retargets `ctx.repoPath`"* — an accurate description of the gap, restated as a
  justification for it.
- **#434** — *"no mutation survives"*, a claim about a space that was never enumerated. One survived.
- **#435b** — a comment asserting a never-value guard, in a test that discarded the response and
  could assert nothing about it.
- **#433** — the sibling-case answer covered `.env`'s contents and missed the temp file the
  atomicity mechanism creates.

Two counts were also wrong and are corrected on their issues: #432's summary (26 vs a 28-row table)
and, on the wave-4 residual, #410's (5 vs 6, measured before two test files were added and written
down after).

**Not one was caught by a gate.** Every suite was green, every fence audit clean, every typecheck and
lint passing, at the moment each claim was written. Five of six were caught by verify; one was caught
only because a human read a build report closely.

The common cause is sequencing rather than discipline: the mutation discipline **was** followed on
all six, but it governs the act, and the report is written after the last edit while the evidence is
gathered before it. Worth weighing when reading the answers above — they have been re-derived, but
the process that produced them is the one described here.

## Gates — on the assembly, not only the lanes

First run in which the four commits coexist. On `prd51-w5-archive-image-and-deferrals`:

- `npm run typecheck` — **0 errors**, all workspaces
- `npm run lint` — clean, 1074 files
- `VITEST_MAX_WORKERS=6 npm test` — **504 files, 9451 passed**, 3 expected fail, 7 skipped
- `npm run build` — clean
- boot smoke — `GET /` → 200, `GET /api/meta` → 401 (capability-gated, ADR-0012, unchanged)
- `rhizomorph archive --help` exercised through `bin/rhizomorph.mjs` against the built `dist`

Run on a quiet machine (load 8.4). That matters: earlier runs today, under load average 175, produced
timeouts in two `packages/web` files unrelated to this wave, and both verifiers correctly refused to
report a green they had not seen.

**One platform.** GitHub Actions was retired on 2026-09-11, so every figure above is macOS and says
nothing about ubuntu or the Windows suite. **This is not "CI passed."**

`scripts/gate.sh` was **not** run: running it is the landing, and the landing is the operator's step.

## Not in this wave

**No migration** — `events.type` is unconstrained and prd-51's single migration-adding lane is still
unspent, which is load-bearing for the next wave rather than incidental. **No new ADR** — #432 and
#433 implement rulings already accepted, and the ADR log allows one claimant per wave with nobody
holding it. **#436** left the wave before dispatch, blocked by #427 on `docs/architecture.md`.

**Also unwitnessed:** no live Postgres in the suite (the `fake.ts` double by design); Caddy is
config-validated but never started, deliberately, since starting it fires live ACME at the org's real
hostname; a real non-injected `EBUSY`; two concurrent `archive` processes; and a restore carrying
actual rows — the ordering result covers schema collisions only. All of these belong to prd-51's
integration drill, not here.


gabriel-canaan (commented):
Read all four commits separately in a clean worktree with `origin/main` merged in. Two findings, both on `8ea87ad7`, both fixed and pushed (`cc72a84b`, `4f1a934c`). The other three commits I have nothing to hand back on.

## What verified

**Gates on today's `main`, merged in:** typecheck 0, lint clean (1076 files), **505 files / 9474 passed**, build clean.

**The ordering law is load-bearing, re-derived rather than taken from the table.** Moved the tombstone write below the prune — both writes still happen, so the tree is byte-identical either way — and got **exactly T1 and T2 red**, 36 others green. Restored and sha256-checked byte-identical. Your table's claim that the end-state assertions (T7/T8/T9/T11) stay green under that mutation is the interesting half and it holds: ruling 11 is testable only through the two witnesses that observe the seam.

**The `.gitignore` fix from verify pass 1 works.** Ran the real `init.sh` in a checkout: `git status` clean, file at mode 600.

**`eb752702` is licensed by evidence, as it says.** The bar was "reproduce against unmodified `ship.ts` or apply no fix", and the three-way `found`/`not-found`/`stale` outcome keeps a genuine unterminated oversized line loud. `bytesRead === 0` can no longer spin: at `at < size` the width is at least 1, so a zero read reaches `stale` rather than the old early `return null`.

**`5da19bf8`'s derivation is sound at every call site.** `path.dirname` inverts `sessionDirFor`'s `path.join(dataRoot, repoSlug(repoPath))` exactly, and the counterexample your reviewer found in `packages/contract/src/harness.ts` is named in the docblock rather than left for a later reader.

## Finding 1 — `init.sh`'s secrets are baked into the image · fixed in `cc72a84b`

`packages/team/Dockerfile`'s runtime stage does `COPY packages/team/deploy packages/team/deploy`. The runbook's order is `./init.sh` then `docker compose build`, so `.env` always exists when the context is assembled — and there was **no `.dockerignore` anywhere in the repo**.

EXECUTED: ran `init.sh`, built an image from the Dockerfile's own two runtime instructions against the same context, exported the filesystem, and read `POSTGRES_PASSWORD`, `RZ_TEAM_DATABASE_URL` and `RZ_TEAM_INGEST_KEY` back out of the layer — `-rw------- 348 packages/team/deploy/.env`. The same 348 bytes as the temp-file orphan your reviewer found. (Two instructions rather than the whole Dockerfile: the full build needs a network `npm ci` this box cannot reach.)

This is that same finding one boundary out. The atomic write's `$ENV_FILE.tmp.$$` got an ignore rule and the build context did not — and the runbook's own rotation recipe compounds it, since "Rotating the ingest key" re-runs `init.sh` and **restarts** the app, and a restart does not rebuild, so the image keeps the old secrets after a rotation that reads as having replaced them.

The fix is a repo-root `.dockerignore`. After it, the same build against the same on-disk `.env` (mode 600, still present) yields an image whose `packages/team/deploy/` holds the eight non-secret files and nothing else. The law added to `deploy/init.test.ts` runs over **both** ignore files and derives the artefacts from `init.sh`'s own `ENV_FILE`/`tmp_file` assignments rather than listing them, so a third secret-bearing path has to be ignored in both places before it goes green. Proven to bite: with `.dockerignore` removed — the state that shipped — it reddens the docker case and only that one.

`node_modules` and `.git` are excluded in the same file for context size. No stage of any Dockerfile here copies either, so that half changes no image.

## Finding 2 — the minted key is refused by the only thing that consumes one · fixed in `4f1a934c`

`init.sh` minted a bare `openssl rand -hex 32`, and the runbook tells the operator to save that value *now* because it is never printed again. `packages/server/src/shipper/key.ts` refuses anything without the `rzk_` prefix.

EXECUTED: `new IngestKey(<64 hex chars>)` throws *"that is not an ingest key: a team ingest key begins \"rzk_\""*. So `rhizomorph connect team` would reject the one value the operator was told to keep.

Your runbook is right that key **values** are not verified yet — that is exactly why nobody hit it. The **shape** is checked on the shipper's side, long before the value would be, by the one command that stores it. `shipper/key-mint-law.test.ts` holds the two sides together: it reads `init.sh`'s own `ingest_key=` assignment and `key.ts`'s own exported constants, retyping neither, and constructs a real `IngestKey` from a value of exactly the shape the script mints. Restoring the shipped form reddens two of its four cases.

The file is declared in `hand-law.test.ts`'s directory list in the same commit, as that law's own note instructs. It is a test file only — `SHIPPER_SOURCES_TODAY` is unchanged, so ADR-0034's importer clause is untouched, and it lives under `shipper/` precisely because clause 3 makes `cli/connect-team.ts` the sole importer of that directory from outside it.

Gates on both fixes: typecheck 0, lint clean (1075 files), **505 files / 9463 passed**, build clean.

- `findLineEnd`'s `stale` returns `unchanged(null, null)`, recording no reason. It mirrors the bounded read's own guard exactly, so it is consistent — but an actor wedged against a repeatedly shrinking file now ships nothing with nothing in the cursor to say why, where `'unreadable'` beside it names its cause.
- `compose.yml` never passes `RZ_TEAM_INGEST_KEY` to `app`, and nothing in `serve.ts` or `startTeamServer` reads it. Consistent with the runbook's honest exclusion list; noting it so a later reader does not assume the variable is wired because it is generated.
- `Caddyfile` puts a real production hostname in a public repo. Defensible, and the runbook is honest that ingest-key values are unverified — flagging it as a decision rather than an oversight.
- `--dry-run` prints `0 archived, 0 refused, 0 bytes freed` under the plan voice, which reads oddly for a run that was never going to archive anything.
- `packages/team/tsconfig.json`'s `rootDir: "."` is inert — the package has no build script, `main` is `src/index.ts` and typecheck is `--noEmit` — so the widened `include` buys the `deploy/` coverage with no emit-layout change. Confirming rather than querying.

## The CI line, said plainly

All ten legs on this PR are red and **not one of them ran**: every job has zero steps and ended in 2–5 seconds. That is the Actions retirement, not your code. It also means this PR has no CI witness at all, and my two commits do not change that — everything above is one Linux machine. Nothing here speaks for macOS or Windows, and `8ea87ad7` in particular is a shell script and a container that only ever executed here under Linux.

## Where this leaves the PR

Rung 2: both fixes change behaviour a reader would want to check — one alters the format of a minted secret. **Confirm `cc72a84b` and `4f1a934c` read right to you, and let a second pair of eyes land it.** I will not approve my own substantive fixes, and no merge-time review record has been written, since nothing is being merged here.

---
*Single reviewer (Claude Opus 5, 1M context) — one pass, not a panel, and not a substitute for the independent read this repo's condition 1 asks for.*

PR #456, opened 2026-09-12
docs(lab): the ceiling claim says what the code does, and prd-55 ruling 1 records the cost as booked

KelliherL:
Refs #408. Refs #430. The two follow-ups this fleet deferred on the record while prd-55 waves 7 and 8 were in flight — both were blocked on #449 holding those files, #449 has landed, so they are owed and paid. Docs and one law; no production code.

## 1. A claim that held while its prose lied

The guide said *"Every subprocess the fork module spawns is bounded at 5 s … except the install"*. Wave 8's #408 gave `workmux add` its own 120 s ceiling, because it runs the new worktree's configured `npm ci` and five seconds killed every launch mid-install. That made the sentence false — **and its claim stayed green**, because the check grepped for `FORK_EXEC_TIMEOUT_MS = 5000`, the one constant that had not moved.

That is the single failure the guide law exists to prevent, so both halves are fixed:

- the sentence now says what the code does — 5 s for the git plumbing and `workmux path`, because five seconds of a plumbing call means wedged; 120 s for `workmux add`, because it waits on an install; the restore keeps its own — and prd-41's two unowned questions are carried forward unchanged;
- the check greps **both ceilings and both numbers**, which is what the sentence states.

**EXECUTED, and this is the point:** renaming `FORK_LAUNCH_TIMEOUT_MS` now reddens the claim. Before this change it did not — the rot that got past the law is the same rot the law now catches.

The law also caught me while I was writing it: my first `says:` regex assumed the sentence sat on one line and it wraps, so the marked-block check rejected my own edit. The regex was made wrap-tolerant; the sentence was not bent to fit the regex.

## 2. prd-55 ruling 1's second dated note

Wave 7 recorded, honestly, that *"booked as spend with its basis"* was true only on the `rd.*` event's own provenance, because `TELEMETRY_SOURCES` named only `sessionlog` and `otel`. Wave 8 changed that. The note is superseded in place, dated, with nothing renumbered and the original left standing:

`llm.cost` admits `lab` — hand-built beside the lab's own events, on the reasoning those already carry: a cost the lab **booked** is the same kind of fact as a checkpoint it **captured**, and neither is a reading a collector polled, so `'lab'` stays out of `eventSourceSchema` and out of `TELEMETRY_SOURCES`. The lab books spend and still cannot sign a reading. One consumer refuses it deliberately, `selectConnection`, because a cost the lab booked proves nothing about an external link being wired.

## Verified

`npm run typecheck` 0 · `npm run lint` 0 · the guide law **61/61** · `corpus-eol-law` and `no-personal-paths-law` green.

One process note, because it nearly became a false bug report: a document law failed once immediately after a scripted write and passed on every rerun, twice across two days. The cause was my own tooling — a python heredoc writing without closing the handle, leaving the flush to GC, so the next command could read a partial file. Not the repo's. Recorded so nobody chases it.


KelliherL:
**No CI ran on this PR, and none will — Actions was retired between its opening and now (#438 landed 00:59). Here is the evidence in the form the repo now asks for, and why the verdict is not published.**

I ran the new leg, `scripts/ci-local.sh --pr 456`, on a native-Windows box:

```
  PASS  Install (npm ci — clean, lockfile-honest)
  PASS  Build
  FAIL  Test
  PASS  Typecheck · Lint · Packaging guard · Boot smoke · Pack smoke
  ci-local: RED     platform witnessed: MINGW64_NT-10.0-26200 / node v22.23.1
```

**The Test leg's red is the platform, not this change**, and I measured that rather than asserting it: unmodified `origin/main`, same worktree, same command — **247 failing tests, 38 files**; this branch — 250 / 40. Then the failing-test-**name** diff, which is the number that means something:

```
=== NEW on this branch ===
(empty)
=== GONE on this branch ===
× L11: a grow-in that finishes under a pause is not frozen a frame short
```

Nothing new fails; one timing-sensitive scene test that fails on main happened to pass here. The per-file counts wander run to run on this box because several of those reds are load-sensitive — which is exactly why the name diff, not the count, is the evidence.

**So no verdict is published**, and deliberately: `ci-local` is RED, and a label claiming otherwise would be the thing `pr-verdict.sh`'s own design warns about. The gap that makes a green verdict unreachable from any Windows machine — the Test leg runs the suite raw and never consults `.windows-known-failures` or `windows-triage.sh`, which the retired `windows-suite.yml` always did — is filed as **#457**, with the dirty-tree false positive on `packages/server/bin/rhizomorph.mjs` (zero-byte content diff, CRLF only) noted there too.

**What this PR itself rests on**, all EXECUTED here: `npm run typecheck` 0 · `npm run lint` 0 · the guide law **61/61** · `corpus-eol-law` and `no-personal-paths-law` green · the mutation that matters — renaming `FORK_LAUNCH_TIMEOUT_MS` reddens the `fork-exec-ceiling` claim, where before this change it did not. Per `AGENTS.md`: this witnesses one platform and is silent about the other two.

The branch is also now merged up to current `main` (`6ed92710`), so the head a reviewer sees is the tree I measured.

ciaran-slow (approved):
**Approve** — read at `6ed92710`. The headline claim is the strongest kind this repo asks for, and it reproduces with a clean control.

## The mutation, with the control that makes it mean something

Your body says renaming `FORK_LAUNCH_TIMEOUT_MS` reddens the claim now and did not before. **EXECUTED, both halves, same mutation applied to both trees:**

| tree | mutation | result |
|---|---|---|
| parent `aaf9a50c` | none | `61 passed (61)` |
| parent `aaf9a50c` | `FORK_LAUNCH_TIMEOUT_MS` → `FORK_LAUNCH_TIMEOUT_RENAMED` | **`61 passed (61)` — the law did not see it** |
| this PR | none | `61 passed (61)` |
| this PR | same rename | **`1 failed | 60 passed`** — `claim "fork-exec-ceiling" holds against the code` |

That is the rot getting past the old law and being caught by the new one, demonstrated rather than argued. The control is what makes it evidence, and you ran it.

## The prose now says what the code does — checked line by line

The sentence makes three finer claims than the law greps for, so I went to `fork.ts` for each:

- *"the git plumbing and `workmux path` get 5 s"* — `fork.ts:151` `FORK_EXEC_TIMEOUT_MS = 5000`, and `:468` says `for 'workmux path' only. NOT for 'workmux add'`. ✅
- *"`workmux add` gets 120 s"* — `fork.ts:176` `FORK_LAUNCH_TIMEOUT_MS = 120_000`, `:470` scopes it to *"the `workmux add` spawn alone"*. ✅
- *"the restore keeps its own longer one"* — `:466` hands restore its exec **unwrapped** so `RESTORE_EXEC_TIMEOUT_MS` governs. ✅

The guide claim is accurate at the level the sentence actually asserts, not just at the level the grep checks.

## prd-55 ruling 1's second note — every factual claim verified against source

Dated, superseded in place, nothing renumbered, wave 7's note left standing. The claims inside it:

| claim | where | ✓ |
|---|---|---|
| `'lab'` stays out of `TELEMETRY_SOURCES` | `telemetry.ts:66` — `['sessionlog', 'otel']` | ✅ |
| `llm.cost` admits `lab` | `telemetry.ts:76` — `LLM_COST_SOURCES = ['sessionlog', 'otel', 'lab']` | ✅ |
| `'lab'` stays out of `eventSourceSchema` | `common.ts:12` | ✅ |
| booked as `auxiliary` | `rd.ts:731` `RD_COST_ROLE = 'auxiliary'` | ✅ |
| `total_cost_usd` carried **unmodified** | `rd.ts:789` `costUsd: provenance.total_cost_usd` | ✅ |
| *"books nothing at all when the result reports no cost"* | `rd.ts:782` `if (!(provenance.total_cost_usd > 0)) return null` | ✅ |
| `selectConnection` refuses it | `connection.ts:85` — `CONNECTION_SOURCES` excludes `lab`, and `:82` gives your exact stated reasoning | ✅ |

The separate `LLM_COST_SOURCES` tuple rather than a third entry in `TELEMETRY_SOURCES` is the right call and its comment says why — widening the shared one would have claimed the lab can report token usage and tool activity. *"The lab books spend and still cannot sign a reading"* is enforced by the type, not just by the prose.

I have no findings on this PR.

## Gates — macOS only, and I am not calling this CI

Post-merge against current `origin/main`:

```
merge 0 · install 0 · typecheck 0 · lint 0 · test 0
Test Files 507 passed · Tests 9542 passed | 3 expected fail | 7 skipped
the-lab-guide-law.test.ts  61/61
```

Also merged with #458 and with #440's just-landed symbol-citation law, since all three are in flight together: **508 files, 9553 passed, 0 failed**, and `doc-citation-law.test.ts` 72/72 — your new prose in `prd-55` and `the-lab.md` cites backticked symbols, and #440's new arm is clean on all of them. No file overlap with either PR.

**What I did NOT run:** anything off `darwin`. No Windows leg, no Linux leg, no `pack-smoke`, no load probe. Actions are disabled repo-wide so there were no check runs to confirm — `gh pr checks 456` reports none. One platform witnessed, two not.

## On the process note in your body

Worth having written down. A document law that fails once after a scripted write and passes on rerun is going to get filed as a flake by the next person who hits it; naming the unclosed-handle cause saves them the hunt. That is the second time this month a law has been accused of a defect that belonged to the tooling around it — #458's finding 15 is the same shape from the other direction.

Reviewed on macOS, node v22.23.2, in a clean clone.

PR #466, opened 2026-09-13
prd-51: wave 5 is merged, and wave 6 is groomed into three lanes

ciaran-slow:
A PRD amendment, **two commits**, no code. Base `main` at `a893af22`.

The map has said "wave 6 — membership is the boundary and a key is a hash — not
groomed" since 2026-09-11. Wave 5 has since merged and wave 6 is groomed,
dispatched and building, so the document was contradicting the tracker. This is
the amendment that closes that, in the same form as the four before it.

This PR **closes no issue**. Amendments to a blessed PRD are operator paperwork;
PR #439 (`14c39468`) is the precedent.

---

## Condition 1 is NOT met, and this paragraph is the disclosure

**No independent review pass has read `3fa379c6`.** I wrote it, and a green suite
is explicitly not condition 1 — AGENTS.md says so in the section this paragraph
exists to obey: *"Evidence you authored yourself is not evidence."* The honest
move is to write the gap where a reviewer sees it rather than to record it away,
so: this needs a reader before it merges, and nothing in this PR substitutes for
one.

## What is the sibling case?

The structural sibling of "the map went stale and nobody noticed" is **every
other place a wave number is written down**. Checked rather than assumed:

- This amendment **renumbers nothing** — wave 6 was already called wave 6 in the
  superseded map, so there is no cascade of the kind the 2026-09-09 amendment had
  to record when three wave citations elsewhere were invalidated.
- The stale `wave 4` citations still sitting in `packages/team/src/ingest/handle.ts`
  (three of them, including the 401 body a client actually receives) and in
  `packages/team/src/api/http.ts` are real and are **not** fixed here. They belong
  to #462, whose fence holds them and whose Definition of done names all five
  sites. Fixing them here would be a second lane touching a file #462 owns.
- The section this amendment half-invalidates — *"#169 and #171 still carry their
  prd-48 bodies"* — is **not** edited in place. It is left standing and corrected
  by the new text, which is how this document has handled every superseded claim
  (the wave-0 open question, the 2026-09-11 map). Editing it would destroy the
  record of what was true when it was written.

## What mutation would this survive?

**Every one of them, and that is a finding rather than a rhetorical answer.**

I ran the doc laws, got 94 green, and then checked whether they could fail: I
planted `` `packages/team/src/auth/does-not-exist.ts` `` in this file and ran
`doc-citation-law` — 72 passed. Suspecting the law's own documented `git ls-files`
index trap, I staged the file and ran it again — 72 passed.

[...]

PR #474, opened 2026-09-13
prd30 w5: the strip counts what it hides, and a card states its age once

### LINKED ISSUE #464: prd30 w5: the attention strip's overflow marker is on screen at every supported width
`prd-30`'s first-glance acceptance is an operator act with a real lay viewer, and Check 2 asks
that viewer "which of these need you?" over the attention strip. Measured on `main` at
`a893af22` in a real browser at 1440x900, the strip answers with **one and a half of five
pathologies, and the `+1` marker that says "there is more" is itself off the right edge of the
viewport.**

## Evidence

**EXECUTED** — live DOM, Chromium, `innerWidth: 1440`, fixture `3` (one lane per pathology):

| chip | x | width | on screen? |
|---|---|---|---|
| `42-otel-receiver` | 848 | 282 | yes |
| `41-retry-parser` | 1136 | 334 | cut mid-string |
| `43-drawer-attach` | 1476 | 462 | no |
| `45-ledger-subrows` | 1944 | 431 | no |
| `+1` | 2381 | 13 | no |

The chip row is allotted **563 px** starting at x=848. Its content needs **1546 px**.

`MAX_CHIPS = 4` still slices correctly — this is not a count bug. The loss is CSS clipping in
the `overflow-hidden` flex row downstream of the slice
(`packages/web/src/panels/attention/AttentionStripView.tsx:129`), and the file's own comment at
`:180` names the failure exactly: *"its share, ran out, and the overflow left the page
entirely."*

## This is already known, and the guarantee was booked to a pass nobody ran

`AttentionStripView.test.tsx:128` is explicit that it bounds a count and not a width, that four
chips bound the strip at *"roughly 1,900px — a real ceiling, just one above every usable
viewport"*, that jsdom has no layout so no assertion in that file can see a width, and that the
width guarantee is therefore **`shell-bounds-law.test.ts` plus a browser pass**.

The measurement above is that browser pass, and the ceiling is exceeded. So the defect is not
that nobody thought about it — it is that the one check that could observe it was deferred to a
manual step, and the manual step had not been taken until now.

## What becomes true

The strip never renders a pathology set in which the reader cannot tell that pathologies are
hidden. Whatever the fix — a narrower chip at width, dropping the evidence span before the lane
name, or fewer chips with a wider `+N` — the `+N` marker is reachable on screen at every width
the shell supports, and `+N` counts every pathology the reader cannot see, not just the ones
past `MAX_CHIPS`. Today it says `+1` while three are invisible.

## Fence

- `packages/web/src/panels/attention/AttentionStripView.tsx`
- `packages/web/src/panels/attention/AttentionStripView.test.tsx`
- `packages/web/src/panels/attention/attention.css`
- `packages/web/src/panels/attention/index.test.tsx`
- `packages/web/src/app/shell-bounds-law.test.ts`

`shell-bounds-law.test.ts` is in the fence deliberately: `AttentionStripView.test.tsx:128` names
it as the file that owns the width guarantee, so a fix that does not move it has almost
certainly not closed the hole.

**Coupling, declared:** `packages/web/src/scene/palette.test.ts` imports across `panels/attention/`
(`.swarm/coupling.txt:37`). A class-string change here can redden it without this fence naming
it. Check it before landing.

**Do NOT touch `packages/core/src/fleet/diagnose.ts`** — the evidence strings are owned by #465.
Why it matters here: `AttentionStripView.test.tsx:96` asserts the frozen chip's evidence text, so
both issues reaching into that module would be a rebase conflict already scheduled.

## What mutation would the test survive?

The existing `expect(chips().length).toBeLessThanOrEqual(MAX_CHIPS)` survives **every** mutation
in this issue: it passes at any chip width, including the current broken one. A test added here
that asserts a count, a class string that is merely present, or a jsdom-measured width (always 0)
survives the defect too. The assertion has to bind a real laid-out width at a named viewport, or
bind the class strings that cap each span — which is what `shell-bounds-law.test.ts` exists for.

## Definition of done

- At 1440x900 with five pathologies staged, the `+N` marker's bounding rect lies inside the
  viewport, and `+N` equals the number of pathologies the reader cannot read.
- A law fails if a chip's span caps are raised such that four chips plus the marker can exceed
  the narrowest supported shell width.
- Re-measured in a browser, not jsdom, and the measurement recorded on the PR.

### LINKED ISSUE #465: prd30 w5: a disclosure card states the elapsed once, and never invents one
`prd-30` S1 asks the card's why line to carry **the evidence and the elapsed time**. The
composition that assembles it appends an elapsed unconditionally:

`packages/web/src/disclosure/vocabulary.ts:290`

    const why = `${reason} — ${fact} ${formatSpan(elapsedMs)} ago`

The contract documented at `:176` is `<reason> — <fact> <elapsed> ago`, which assumes the fact
does not carry a duration of its own. **Four of the five shipped conditions break that
assumption, in two different directions.**

## Evidence

**EXECUTED** — every pathology card opened in a real browser on `main` at `a893af22`, fixture
`3`, why line read from `[data-testid="disclosure-card"]`:

| condition | rendered why line | defect |
|---|---|---|
| `LOOPING` | `…— Bash→Read→Edit ×21, no commit 4m00s ago` | none — the only correct one |
| `FROZEN` | `…— no events for 16m37s **16m37s ago**` | duration printed twice |
| `WAITING` | `…declares waiting 8m17s ago · workmux reports working **8m17s ago**` | duration and the word "ago" printed twice |
| `OFF-FENCE` | `…→ 46-spend-selectors **0s ago**` | an elapsed invented for a fact that has none |
| `EXPENSIVE` | `…55.4× fleet median **0s ago**` | an elapsed invented for a fact that has none |

Two of these are worse than untidy:

**`WAITING` misreports its own evidence.** The fact is a `·`-joined list whose *first* clause
carries the "ago". The template's appended "ago" therefore attaches to the **last** clause, so
the sentence reads as though workmux reported *working* 8m17s ago. The evidence says the beacon
declared waiting 8m17s ago and workmux currently reports working — a disagreement between two
witnesses, which is exactly the thing this card exists to show. The rendered sentence collapses
it into one false timestamp.

**`OFF-FENCE` and `EXPENSIVE` are stamped `0s ago`.** A trespass path and a token rate have no
"when"; `elapsedMs` is 0 because there is nothing to measure, and the card prints that 0 as a
fact. This is the failure `vocabulary.ts:275` says the surrounding guard exists to prevent —
*"a NaN out of `now - null` renders 'NaN ago' — evidence shaped like evidence and worth
nothing."* `requireAge` rejects NaN and passes 0, which is the same defect wearing a valid
number. Against S1's own falsifier — *"a card whose why has no evidence in it"* — a fabricated
timestamp is worse than a missing one.

## Why this lands on prd-30 and not on a tidy-up

S1 names the falsifiers this hits: *"two surfaces phrasing one condition differently"* and *"a
card whose why has no evidence in it"*. Success criterion 5 — *"a first-time viewer learns from
the surface itself"* — is the one the outstanding first-glance acceptance scores, and these are
the sentences the lay viewer reads aloud. Found while staging that act.

## What becomes true

A card's why line states an elapsed **once**, and states one only when the evidence has a "when".
A condition whose evidence is not time-anchored renders its evidence without a trailing age
rather than with `0s ago`.

## Fence

- `packages/web/src/disclosure/vocabulary.ts`
- `packages/web/src/disclosure/vocabulary.test.ts`
- `packages/web/src/disclosure/DisclosureCard.tsx`
- `packages/web/src/disclosure/DisclosureCard.test.tsx`
- `packages/web/src/disclosure/teach-law.test.tsx`
- `packages/web/src/panels/fleet/index.test.tsx`

`teach-law.test.tsx` is in the fence because `vocabulary.ts:183` composes the teach layer's lines
with the same `<elapsed> ago` suffix, so it has the sibling defect by construction and must be
checked in the same change. `panels/fleet/index.test.tsx` asserts composed why strings and will
move with them.

## The fix is in the composition, NOT in the evidence strings — and that is a scope ruling

`packages/core/src/fleet/diagnose.ts` is **out of scope** for this issue. It is tempting to strip
the durations out of the facts at `:124`, `:210`, `:256` and `:274` instead, and it is the wrong
fix twice over:

1. The evidence strings are good evidence strings. `no events for 16m37s` is the right thing for
   a chip to say standing alone, and the chips render the raw evidence, not the composed why.
2. `packages/web/src/panels/attention/AttentionStripView.test.tsx:96` asserts
   `/no events for \d+m\d\ds/` on the frozen chip, and that file is owned by #464. Both issues
   editing that module is a rebase conflict scheduled in advance.

The composition is what knows whether an elapsed has already been stated; it is where the
knowledge belongs. If the plan concludes otherwise, that is a fence widening recorded on the PR
**before** the change, and it must be sequenced against the sibling rather than raced with it.

## What mutation would the test survive?

`vocabulary.test.ts` is green today with all four defects shipping, so whatever it asserts about
the why line cannot fail for the reason this issue names. The trap to avoid repeating: a test
that pins one condition's whole sentence as a literal passes for `LOOPING` and proves nothing
about the other four. The assertion has to be a property over **every** condition in the table —
for example, that no rendered why line contains two occurrences of the same `formatSpan` output,
and that no why line ends in `0s ago`. A per-condition literal is not that, and the `_never`
switch does not help here because every condition is present; only their composition is wrong.

## Definition of done

- Every condition in the table renders an elapsed at most once in its why line.
- No why line renders a duration for evidence that has no time anchor.
- `WAITING`'s multi-clause evidence reads as two witnesses disagreeing, not as one timestamp.
- The assertion is a property over the whole table, and reddens if a fifth condition is added
  with either defect.

ciaran-slow:
Closes #464
Closes #465

prd-30's build was complete and its only outstanding item was the first-glance acceptance — an
operator act with a real lay viewer. Staging that act found two defects in surfaces this PRD
built. Both are here, plus the PRD amendment that declares the wave.

## Condition 1 — per commit

> **⚠️ CONDITION 1 IS RE-OPENED. Force-pushed 2026-09-14 to fix a defect the browser pass
> found. Every sha below is new; the three prior reviews are pinned to commits this PR no
> longer contains, and none of them carries over.**

| commit | read by |
|---|---|
| `c7bc1fe3` (#464) | **NOT YET READ** — supersedes `2fbe7cb3`, which was reviewed. Carries the browser-pass fix. |
| `f8e75bb9` (#465) | **NOT YET READ** — content-identical to `6fb6f6f7`, which was reviewed; re-numbered by the re-assembly. |
| `2bf7ef3b` (docs) | **NOT YET READ** — content-identical to `d48d23d5`, which was reviewed; re-numbered by the re-assembly. |

Superseded records, kept so the trail is followable rather than deleted:
[`2fbe7cb3`](https://github.com/launchpad-26/rhizomorph/commit/2fbe7cb3ef25190cee7a71e7d3035bbddadcdf25#commitcomment-200254996) ·
[`6fb6f6f7`](https://github.com/launchpad-26/rhizomorph/commit/6fb6f6f7c77e9057379573b1220fd2303352c3f8#commitcomment-200255121) ·
[`d48d23d5`](https://github.com/launchpad-26/rhizomorph/commit/d48d23d5475359f7c630ab76a11cbccbaca90bd1#commitcomment-200255135).
Two of the three are content-identical re-numberings, which is a judgement for the reviewer to
make and not for this body to assert.

The reviewer is not the author: all three commits were planned and built by one Claude session,
and none of that session's own output is offered as condition 1. Condition 2 is met — the
operator asked for this PR, in that session.

**What is NOT condition 1, stated because this repo asks for it:** the gate results below are
evidence about the tree, authored by the party being guarded. They are not a review and never
were.

## #464 — the attention strip's `+N` marker is on screen at every supported width

The marker was the clipping row's last child, so an overflowing row pushed out the one element
whose whole job is to say something is hidden. Measured in Chromium on fixture `3`:

| viewport | chip row | chips need | `+1` marker |
|---|---|---|---|
| 1440×900 | 563 px | 1522 px | x=2381 — **941 px off screen** |
| 1100×760 (`WINDOW_MIN_WIDTH`) | 223 px | 1522 px | off screen |

It announced one hidden pathology while three were invisible. The marker is now a sibling of the
clipping row, and the row measures itself and folds to what it can name, so `+N` counts every
pathology the reader cannot see. `chipCapacity` is pure — jsdom reports every width as 0 and can
never exercise the real thing, the same split `shell-bounds-law.test.ts` already sets out.

Conservative by construction: it assumes every chip is the derived 520 px maximum, so it can
fold a chip that would have fitted. Over-folding costs a name and counts it; under-folding hides
a frozen lane. `docs/design-notes/attention-chip-width.md` carries the asymmetry.

**Sibling case.** `WaitedChipsRow`, the strip's other clipping flex row one screen away. It does
**not** carry this defect, and the reason is worth stating: it has no `+N` marker to lose — it
clips silently and always did. Whether that is right is a separate judgement, untouched here;
the region's own comment rules it "the strip's memory, not its summons … the one part that
SHOULD give way first", so silent yielding there looks deliberate.

**What mutation would the test survive?** Three run, each reddening what it should: unknown-width
returning `0` instead of `MAX_CHIPS` (20 failed); `CHIP_MAX_WIDTH_PX` → 5000 (5 failed); the
marker moved back inside the clip (`shell-bounds-law` red, with its own message naming the
defect). **One that does NOT redden, stated rather than hidden:** swapping `useLayoutEffect` for
`useEffect` leaves the suite green — RTL flushes passive effects inside `act()` and jsdom never
paints. `Scrubber.tsx:330` records the identical gap. Harness limit, not permission.

## #465 — a disclosure card states the elapsed once, and never invents one

The card appended `<elapsed> ago` unconditionally. Four of the five shipped conditions broke the
assumption behind that:

| condition | rendered before | fault |
|---|---|---|
| `LOOPING` | `…no commit 4m00s ago` | correct — the only one |
| `FROZEN` | `…no events for 16m37s **16m37s ago**` | the fact already states the span |
| `WAITING` | `…declares waiting 8m17s ago · workmux reports working **8m17s ago**` | span and "ago" both doubled |
| `OFF-FENCE` | `…→ 46-spend-selectors **0s ago**` | a duration invented for a trespass path |
| `EXPENSIVE` | `…55.4× fleet median **0s ago**` | a duration invented for a rate |

`WAITING` was not merely untidy. Its fact is a `·`-joined list whose first clause carries the
"ago", so the appended one landed on the **last** clause and the sentence asserted that workmux
reported *working* 8m17s ago. The evidence says two witnesses disagree — the thing the card
exists to show.

`elapsedMs: 0` was never a missing anchor; `selectors/condition.ts` documents it as "confirmed
just now". So both faults are in the rendering, and `packages/core` does not move — which is
what keeps #464's chip-evidence assertion green in this same wave.

**Three of this repo's own sentences already said what the render should be** — `format.test.ts`'s
header docblock ("confirmed just now"), a comment at `:83`, and a test *named* "reads as
just-now, not as NaN" — each sitting above an assertion pinning `0s ago`. Those are restated at
**greater** strength: they now assert `just now` and forbid `0s ago`.

**Sibling case.** The teach layer, which composed the same way and is fixed in the same commit.

**What mutation would the test survive? — the important one in this PR.** Reverting the
teach-line fix alone first read **green across 84 tests**. The derivation half of the property
sweep in `panels/fleet/format.test.ts` is **vacuous**: no producer in that module supplies
`derivedFrom`, so its loop never iterates. Three direct teach-layer tests were added where the
defect can actually fail, and the same mutation now reddens exactly those two. The sweep keeps
its derivation arm for a future producer, with a comment saying plainly that it is vacuous
today. A second test of mine had to be fixed for the same class of reason: `not.toContain('0s
ago')` matches inside `"1m30s ago"` and reported six correct cards as defective; it is now a
word-boundary regex.

Other mutations: why line reverted (8 failed), zero-age rule removed (9 failed), already-stated
check removed (3 failed).

## `d48d23d5` — the plan of record declares wave 5

`prd-reconcile.sh 30` reported both issues as UNDECLARED WAVE. The PRD is the plan of record, so
it is amended rather than the issues moved. The declaration uses the literal `**Wave N` form the
detector reads, once, and the amendment explains the wave without reproducing that string a
second time. prd-30's status line and `docs/roadmap.md`'s entry both move with it — both claimed
all waves shipped with one operator act outstanding, which stopped being true when these were
filed.

`npm test` **508 files, 9587 passed**, 3 expected fail, 7 skipped · `typecheck` clean · `lint`
clean (1079 files) · `build` clean · boots, `GET /` → 200.

**One platform.** macOS, node v22.22.2. All four workflows are disabled, so nothing here
witnesses ubuntu or Windows. This is not "CI passed".

Gates ran at base `a893af22`; `origin/main` has since moved to `f4b7b574` (#466), which touched
only `docs/prds/prd-51-the-split.md` — no overlap with this wave. The doc laws were re-run
against the new tip (94 passed) because they read `origin/main` live.

`tripwire-law.test.ts` and `gate-honesty-law.test.ts` failed under the 6-worker run and **pass
isolated** — the known load flakiness, not cross-lane breakage.

## Fence widenings

- **#464:** `CHANGELOG.md`, `docs/design-notes/attention-chip-width.md`.
- **#465:** `CHANGELOG.md`, and `packages/web/src/panels/fleet/format.test.ts` — the fence named
  `index.test.tsx` in that directory, but the composed strings and all eleven disclosure
  producers live in `format.test.ts`.

`CHANGELOG.md` is touched by both commits and conflicted on cherry-pick. Resolved by keeping
both entries; no source file overlapped.

## The browser pass — RUN, and it found a defect

Chromium, fixture `3` (five pathologies), dev server on this branch. Against the same probe on
`origin/main` as a control:

| | 1440×900 before | 1440×900 after | 1100×760 before | 1100×760 after |
|---|---|---|---|---|
| chips named | 4 (1 visible) | **1** | 4 (0 visible) | **0** |
| `+N` says | `+1` | **`+4`** | `+1` | **`+5`** |
| marker right edge | x=2387 | **x=1411** | x=2392 | **x=1071** |
| marker on screen | ❌ | **✅** | ❌ | **✅** |
| named + counted | 5 (but 3 hidden silently) | **5** | 5 (but 4 hidden silently) | **5** |

No chip exceeds the row's right edge at either width, and the document does not scroll sideways.
The one chip named at 1440 is `42-otel-receiver`, the FROZEN lane — worst rung first, so the
single chip the reader gets is the most urgent one.

**The defect it found, which the whole suite was blind to.** Making the marker a sibling means
the flex split removes it and its gap *before* the row is measured — the wrapper is 563px and
the measured row is **544px**. `chipCapacity` then reserved the marker a **second** time,
dropping the budget below one chip-width and naming **zero** lanes at 1440×900. Every unit test
passed, because they fed the function 563 — the wrapper — a number that does not exist at
runtime. Correct for the argument the test passed, wrong for the argument the component passed:
a test that could not fail for the reason it claimed, in the commit whose PR body is about that
failure mode.

Fixed in `c7bc1fe3`. The capacity tests are now pinned to **544** and **204**, transcribed from
Chromium rather than chosen, and restoring the double subtraction reddens four tests including
the component-level render test. That is the mutation that matters here, and it is why the
numbers in a test have to come from the thing being measured.

**A note for whoever picks up the first-glance acceptance:** PR #455's run sheet pins the
fifth-chip observation as "four chips and a `+1`", which this PR changes. That sheet needs
correcting against this behaviour before a lay viewer sees it.


KelliherL (approved):
## Review of #474 — approve; the browser pass re-run independently and it matches to the pixel, one doc fix pushed

Read commit-by-commit (`c7bc1fe3` #464, `f8e75bb9` #465, `2bf7ef3b` docs) against #464's and #465's own Definition-of-done bullets, in a clean clone at node 22.23.2 on Linux. The three prior commit comments are the author's own ledger, so this is the first independent read of this head. **Most of it verified in the PR's favour**, and that comes first because it is the evidence.

### What verified — EXECUTED

- `npm run typecheck` **0** · `npm run lint` **0** (1079 files) · the four touched test files plus `AttentionStripView.test.tsx`, `one-card-law` and `doc-citation-law` — **6 files, 203 passed**.
- **Coupling:** `scene/palette.test.ts` (imports across `panels/attention/`, named on #464) — **green** on this head.
- **Fence:** every touched path is inside #464's / #465's fence or one of the four widenings the body records (`CHANGELOG.md`, the design note, `format.test.ts`); the docs commit's two files are the PRD amendment, outside both fences by design. `diagnose.ts` untouched, as #465 ruled.
- **The browser pass, re-run independently** — headless Chromium (Playwright 1.62.1, `deviceScaleFactor: 1`), fixture `3` via the `3` key, this branch's own build served on a spare port, and the pre-fix instrument as control:

  | | branch @ 1440×900 | branch @ 1100×760 | control @ 1440 | control @ 1100 |
  |---|---|---|---|---|
  | measured row | **545 px** | **205 px** | — | 225 px, **5 children incl. the marker** |
  | chips named | 1 (`42-otel-receiver`, FROZEN) | 0 | 4 | 4 |
  | marker | **`+4` at x=1397–1411, on screen** | **`+5` at x=1057–1071, on screen** | `+1`, right edge **2396 — off screen** | `+1` at x=2385–2399 — off screen |
  | `document.scrollWidth` | 1440 | 1100 | | 1100 |

  Your table says 544/204 and a marker right edge of 1411/1071. I get 545/205 and 1411/1071 — one pixel on the row from a different font rasteriser, the marker identical. #464's first DoD bullet ("the marker's rect lies inside the viewport, and `+N` equals the pathologies the reader cannot read") holds at both widths, and the control reproduces the defect exactly as the issue measured it.
- **Your mutations, re-derived** (each restored): marker reserved a second time in `chipCapacity` → red, the boundary test by name; marker moved back inside the clipping row → `shell-bounds-law` red with its own sentence; unknown width → `0` → red across the strip tests; zero-age rule removed → red ("accepts a fact observed this instant"); the teach-line fix alone reverted → red, and only in the new direct teach tests. And one of mine: **the already-stated check removed** (`fact.includes(span)` deleted) → the property test reddens on FROZEN/WAITING doubling their span. The load-bearing assertions are load-bearing.
- **The vacuous-sweep claim:** `grep -c derivedFrom packages/web/src/panels/fleet/format.ts` → **0**. No producer supplies it; the derivation arm of the sweep cannot iterate today, exactly as the commit says.
- **`chipCapacity` as shipped:** `(0)=4 (204)=0 (519)=0 (520)=1 (538)=1 (539)=1 (544)=1 (563)=1 (1045)=1 (1046)=2`.

### One mechanical fix — pushed as `ee807012`

`docs/design-notes/attention-chip-width.md`'s last paragraph says the boundary tests pin *"539px names one, 538px names none"*. That is the boundary of the version the browser pass killed — the one that reserved the marker's 19 px twice. The shipped formula and `AttentionStripView.test.tsx:611` both say **520 / 519**, and the print above confirms 538 and 539 each name one. One fact written twice, drifting in prose — the shape the note's own last section warns about. One line; no code moved.

### The leg on the head

`scripts/ci-local.sh --no-install` on `ee807012` (your three commits plus the doc line), the same leg the landing gate now runs:

```
  Test Files  508 passed (508)
       Tests  9591 passed | 3 expected fail | 3 skipped (9597)
  ci-local: green — exit 0 in 161 s, Linux / node v22.23.2
```

Your 9587 was measured at `a893af22`; the four extra are what `main` gained since. `tripwire-law` and `gate-honesty-law` did not flake in this run.

One platform (Linux, node 22.23.2); the workflows are disabled, so nothing witnesses macOS or Windows here either. The branch sits on `origin/main` (`f4b7b574`) already — no merge was needed.

- **`evidenceClause`'s substring check has one non-benign shape.** `fact.includes(span)` treats a span that is a *suffix* of a longer span as "already stated": a fact carrying `16m37s` with an elapsed of `6m37s` would render no age at all, although the fact's duration is not the elapsed. Today it cannot happen — FROZEN's and WAITING's fact span *is* the elapsed, by construction in `diagnose.ts` — so this is a note for the next condition that carries a duration of its own, not a defect. A digit-boundary check (`(^|[^0-9])` + span) would close it without touching the caller contract; I did not make that change because it is behaviour in your commit.
- The PR-body sentence "`+N` counts every pathology the reader cannot see" is now literally true at both widths, and the design note is honest that one named chip at 1440 is what the layout affords, not what it should. The follow-up it names (reclaiming the 848 px left of the alarm bar) is the real fix for the glance; worth filing so it does not live only in a design note.
- #455's run sheet still pins "four chips and a `+1`" — your own note; flagging it again so it is not lost when this lands.

### What I did not run

macOS and Windows. The `useLayoutEffect` → `useEffect` mutation (you record it as unobservable in jsdom; my browser pass does not distinguish it either). The first-glance acceptance itself, which is the operator's act.

**Approving.** The one fix is a doc line — rung 1 — so I am not blocking on confirmation; drop it if you disagree. Landing is the operator's act.

PR #476, opened 2026-09-13
prd51 w6: membership is the boundary, a key is a hash, and an ADR records why neither plane is the other

### LINKED ISSUE #169: prd51 w6: membership is the boundary, and a pending invite is not a member

### LINKED ISSUE #462: prd51 w6: a key is 32 random bytes, stored only as SHA-256, and ingest refuses one it cannot verify
## Fence (may touch ONLY)
- `packages/team/src/keys/` (new directory, all files)
- `packages/team/src/storage/contract.ts`
- `packages/team/src/storage/fake.ts`
- `packages/team/src/storage/postgres.ts`
- `packages/team/src/migrations/` (all files, including the new `0005_ingest_keys.sql`)
- `packages/team/src/ingest/handle.ts`
- `packages/team/src/ingest/handle.test.ts`
- `packages/team/src/api/main.ts`
- `packages/team/src/api/api.test.ts`
- `packages/team/src/bootstrap.test.ts`
- `packages/team/deploy/` (all files)
- `packages/server/src/shipper/key-mint-law.test.ts`
- `docs/team-server-runbook.md`
- `.swarm/coupling.txt`

## Blocked by
Nothing — #410 landed in PR #446, so `packages/team/src/storage/` is free.

Sibling lanes hold other surfaces: the membership lane holds
`packages/team/src/auth/` and all of `packages/team/src/config/`; the ADR lane
holds `docs/adr/`. Do **not** re-export anything you add from
`packages/team/src/index.ts` — the barrel's own header says a re-export turns two
parallel lanes into a stack. Do **not** touch `packages/team/src/api/http.ts`:
adding a route is wave 7's single restructuring of that file, and the keys
dependency reaches `handleIngest` through the `IngestDeps` that
`api/main.ts` already constructs. `docs/prds/prd-51-the-split.md` is held by #427
and is read-only for the whole of this wave. `docs/architecture.md` is held by
#427 and is not yours either.

**Model:** opus. **Wave:** 6.

## What this is

Ruling 8's **machine plane**, whole: *"`rzk_` ingest keys: 32 random bytes,
stored only as SHA-256, shown once at mint, scoped to one project, revoked by a
row flag checked **once per batch** — which bounds revocation lag to one batch
interval. Refusal text names the key prefix and the exact reason."*

It is one issue rather than three because it is one causal chain — the port, the
migration and the ingest refusal cannot be split without the second half
depending on the first, which is a stack wearing a bundle's clothes.

**The narrowing ruled at grooming:** ruling 8 also says *"a member mints a key in
the viewer"*. There is no viewer until wave 7, so minting-from-the-viewer defers
with it. What ships here is the hash at rest, the revoke flag, the once-per-batch
check, the refusal text, and `init.sh` seeding the first project's key by storing
only its hash.

## The gap you are closing is loud, and it is written down in five places

`handleIngest` returns 401 on a missing header and **deliberately does not verify
the value** — its own test asserts that `'not-a-real-key-at-all'` is accepted
today. Every site that says so goes false in this commit, and ruling 12 requires
they change in the same commit as the code that makes them false:

1. `packages/team/src/ingest/handle.ts` — the `INGEST_KEY_HEADER` doc comment,
   the `handleIngest` doc comment, and the 401 body's own text, all three of
   which name **"prd-51 wave 4"** — stale since the 2026-09-11 renumbering.
2. `packages/team/src/ingest/handle.test.ts` — the assertion labelled *"The gap,
   asserted so it is loud rather than latent"*. It inverts.
3. `packages/team/src/api/api.test.ts` — its 401 case, and every case that POSTs
   a batch expecting 202 now needs a key the server will accept.
4. `docs/team-server-runbook.md` §*"What is not wired up yet"* — the second
   bullet. Leave the other two bullets (the fold worker, the doctor) standing;
   they are still true.
5. `docs/team-server-runbook.md` §*"Rotating the ingest key"* — which says in
   those words *"Treat it as housekeeping, not as revocation, until key
   verification exists."* It now is revocation, and the procedure changes.

## Definition of done

- **A key is `rzk_` + 32 random bytes, and the tree already has the validator.**
  `packages/server/src/shipper/key.ts` exports `INGEST_KEY_PREFIX` and
  `MIN_INGEST_KEY_BODY` and is the only consumer in the tree. Mint to satisfy it;
  do not invent a second shape.
- **Only the hash is stored.** A law asserts no column in the schema holds a key
  plaintext, and the mint function's return type makes the plaintext available
  exactly once. The mutation to plant: store the plaintext in a second column and
  observe the law go red.
- **A key is scoped to one project**, and a key valid for project A is refused
  for project B with a refusal that says so.
- **Revocation is a row flag checked once per batch.** Not once per event, not
  cached across batches — the once-per-batch reading is what bounds revocation
  lag to one batch interval, and it is the claim, so it is the test: revoke
  between two batches and assert the second is refused, and assert the check ran
  exactly once for a multi-event batch.
- **Refusal text names the key prefix and the exact reason** (ruling 8, verbatim)
  — and never the key. `packages/server/src/shipper/no-key-in-output-law.test.ts`
  is the precedent on the other side of the wire; the refusal a server emits must
  survive the same standard. Unknown key, revoked key, wrong project and
  malformed shape are four different reasons and four different strings.
- **The ordering in `handleIngest` is not disturbed.** Ruling 4's
  `validate → build → writeSync → fsyncSync → notify → 202` is asserted as an
  exact-array `toEqual` on a trace tape. Key verification is part of
  **validate** — a refused key must never touch the journal, and the tape for a
  refusal must end at `validate`. Run the existing fault-point cases F2–F5 and
  confirm they are untouched.
- **The keys port rides on `TeamStorage`.** Add it to `contract.ts`, `fake.ts`
  and `postgres.ts`, so `startTeamServer` needs no new parameter and
  `packages/team/deploy/serve.ts` needs no change to pass one. No SQL crosses the
  port — `no-sql-outside-storage-law.test.ts` is inside your fence and will say
  so if it does.
- **One migration, `0005_ingest_keys.sql`, with the `rz_ingest` grant in it.**
  `0003_roles_rls.sql` gives `rz_ingest` INSERT on `events` and nothing else; the
  per-batch check needs SELECT on the keys table and that GRANT belongs in this
  migration, not in an edit to `0003`. The migration files are tracked and
  checksummed — a retroactive edit to an applied migration is not available to
  you.
- **Three migration-id pins are re-derived, not appended.** `bootstrap.test.ts`,
  `api/api.test.ts` and `migrations/schema-law.test.ts` each pin the tracked set
  by exact-equality array, **independently** — fixing one does not fix the others,
  and `.swarm/coupling.txt` already records that a lane stopping at the first red
  file is still red. Re-derive each from the directory; a wrong id or a wrong
  order fails exactly as a missing one does.
- **`schema-law.test.ts`'s "every index is on `events`" assertion is
  reconciled.** It currently asserts
  `indexDeclarations(allStripped()).every((i) => i.table === 'events')`, which a
  keys-table index makes false. Reconcile it minimally and deliberately — the
  law's *intent* is that nothing indexes the wrong thing, not that `events` is
  the only table that may ever exist — and leave the detector self-tests beside
  it intact.
- **`init.sh` seeds the first project's key by storing only its hash**, still
  prints the plaintext exactly once, and still mints with the `rzk_` prefix.
  `packages/server/src/shipper/key-mint-law.test.ts` reads that script's
  `ingest_key="…"` line **from another package** and exists because `init.sh`
  once shipped minting a bare `openssl rand -hex 32` that `connect team` would
  have refused. Keep it green, and update its header comment, which cites the
  runbook section you are deleting.
- **`.swarm/coupling.txt` gains that seam, with a reason.** `deploy/init.sh` ↔
  `packages/server/src/shipper/key-mint-law.test.ts` is a cross-package coupling
  that is not registered today; this grooming found it by reading the law, not by
  reddening it. Write the entry the way that file demands — the path, then why a
  change elsewhere forces an edit here — and say whether your entry is EXECUTED
  (you planted a bare mint and watched it go red) or REASONED.
- **The runbook's two stale sections are rewritten in this commit** (ruling 12),
  and the rotation procedure now describes actual revocation.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` pass in
  your own worktree.

### LINKED ISSUE #463: prd51 w6: an ADR records the two identity planes and why neither is an OAuth App
## Fence (may touch ONLY)
- `docs/adr/0050-two-identity-planes.md` (new — confirm 0050 is still free before you write it)
- `docs/adr/README.md`

## Blocked by
Nothing. This records a decision that was **already blessed** — ruling 8 and its
2026-09-08 amendment — not the code the sibling lanes write. It must be
writable, and reviewable, without either of them existing.

Sibling lanes hold other surfaces: the membership lane holds
`packages/team/src/auth/` and `config/`; the keys lane holds the rest of
`packages/team/`, `docs/team-server-runbook.md` and `.swarm/coupling.txt`.
`docs/prds/prd-51-the-split.md` is held by #427 and is read-only for the whole of
this wave — cite its rulings, do not edit it.

## Why this ADR is owed

prd-51 landed three ADRs with its blessing — 0033 (ruling 3, the wire), 0034
(ruling 2, the fifth hand), 0035 (ruling 13, the container). **Ruling 8 got
none**, and it is the other constitutional ruling in the document: it is the one
that says where authority over identity lives, and it has already been amended
once. A PRD ruling dies with its PRD; an ADR outlives it. This is the repo's own
rule for where a decision goes, applied to the ruling it was not applied to.

It passes the ADR test — *if you can't name a rejected alternative and say why,
it isn't an ADR* — twice over, and both rejections are already written down in
the amendment's own prose.

## Definition of done

- **The decision recorded is "two identity planes that never fuse"**: humans by
  GitHub organisation membership, machines by `rzk_` ingest keys, and the clause
  that does the load-bearing work — *no GitHub token is ever accepted on the
  ingest route; no key is ever derived from a GitHub token; the ingest hot path
  never calls GitHub.*
- **Two rejected alternatives, each with its reason, both taken from the
  2026-09-08 amendment rather than invented here:**
  - **An OAuth App** — rejected because it acts *as* whichever person signed in,
    borrowing their permissions, where a GitHub App is its own identity, installed
    on an organisation, with powers chosen at creation and readable by anyone
    afterwards. Ruling 8 as blessed said OAuth web flow; what was built on
    2026-09-07 is a GitHub App, and this ADR is where that divergence stops being
    an amendment buried in a PRD.
    Record the consequence the amendment draws: **the org third-party-access
    clause is deleted**, because a GitHub App is installed rather than approved.
  - **Repository read permission** — rejected because it would have handed the
    server every private repository each member can reach in order to answer one
    yes-or-no question. The app's permissions are **Organization → Members →
    Read-only, and nothing else**; no repository access of any kind; no webhook.
- **The status line names its authority and its date**, in the form the log
  already uses: accepted, prd-51 ruling 8 as amended 2026-09-08.
- **A TABLE ROW in `docs/adr/README.md`.** `packages/server/src/adr-log-law.test.ts`
  asserts membership both ways, and `.swarm/coupling.txt` records that a rowless
  record fails **2** of that law's 6 tests, not 1 — the membership check *and* the
  law's own detector self-test, which deletes one row and asserts the missing list
  is exactly one file. One row fixes both; expect the second failure rather than
  being surprised by it.
- **The number is claimed once.** The log is append-only and numbered and only one
  issue per wave may take the next number — this is that issue for wave 6. `0049`
  is the tail on `main` at `c4e20d09`; **re-check `docs/adr/` against
  `origin/main` before you write**, because a lane in another milestone taking
  0050 first means you renumber, and that has happened here before (0038 → 0039,
  2026-09-06).
- **It records the decision, not the build.** The sibling lanes are landing in the
  same PR; an ADR that describes their code is an ADR that depended on them, which
  this wave's parallelism forbids. Cite rulings and the amendment. Do not cite
  `packages/team/src/auth/` or `packages/team/src/keys/` as though you had read
  them.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` pass in
  your own worktree — `adr-log-law.test.ts` is the one that matters and it lives
  in `packages/server`.

ciaran-slow:
Closes #169
Closes #462
Closes #463

prd-51 wave 6 — **ruling 8, both planes**, plus the ADR that ruling never got.
Three lanes, three commits, base `main` at `f4b7b574`.

Humans are bounded by membership of one GitHub organisation, checked on an
installation token. Machines are bounded by `rzk_` keys that exist at rest only as
SHA-256, are scoped to one project, and are revoked by a row flag read once per
batch. **The planes never meet**, and ADR-0050 is where that stops being a PRD
ruling — which dies with its PRD — and becomes a record that outlives it.

**Cherry-picked with no conflict, which is the wave's own fence audit**: nothing
here touched a file anything else here touched. #169's 4 files, #462's 25 and
#463's 2 are disjoint by construction.

---

Every lane was verified on a **different model from its builder**, each verifier
confirming the `Co-Authored-By` trailer itself with `git log` rather than
accepting the claim, and each running the gates in its own clean worktree rather
than reading the build report's numbers.

|---|---|---|---|---|---|
| `0a5b90bf` | #169 | Sonnet 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/169#issuecomment-5651670504) | [report](https://github.com/launchpad-26/rhizomorph/issues/169#issuecomment-5651699584) · [answers](https://github.com/launchpad-26/rhizomorph/issues/169#issuecomment-5656397271) | [**changes** — Opus](https://github.com/launchpad-26/rhizomorph/issues/169#issuecomment-5656313620) → [**ready** — Opus, round 2](https://github.com/launchpad-26/rhizomorph/issues/169#issuecomment-5656483570) |
| `7f8aa133` | #462 | Opus 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/462#issuecomment-5651697423) | [report](https://github.com/launchpad-26/rhizomorph/issues/462#issuecomment-5651797520) | [**ready** — Sonnet](https://github.com/launchpad-26/rhizomorph/issues/462#issuecomment-5656248034) |
| `f1ec86e2` | #463 | Sonnet 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/463#issuecomment-5656171937) | **none posted — see gaps** | [**ready** — Opus](https://github.com/launchpad-26/rhizomorph/issues/463#issuecomment-5656396552) |

**The verify records name the LANE shas, not the shas in this PR.** Cherry-picking
rewrites them and this repo's condition 1 is per-sha, so the gap is closed by
content rather than by assertion — `git patch-id --stable` is identical for all
three pairs:

|---|---|---|---|
| #169 | `db3041f9` | `0a5b90bf` | identical |
| #462 | `7fb7eb2f` | `7f8aa133` | identical |
| #463 | `f4c2422b` | `f1ec86e2` | identical |

What was verified is what is here, byte for byte.

---

## Three gaps, stated rather than recorded away

**1. #463's verify ran against a sha that no longer exists.** The pass read
`be66f79f`; the branch now carries `f4c2422b`, because that commit shipped with
**no `Co-Authored-By` trailer** while both its siblings had one — the verifier
raised it as its own finding zero, noting that `.claude/skills/verify/SKILL.md`
reads the builder's model from that trailer and this commit could not carry it. It
was amended after the verdict, on the operator's instruction. `patch-id` is
identical across the amend (`9110f24e…` both sides), so no content moved — but the
verifier's own framing stands and is not repaired by the amend: **it established
builder/verifier independence on the operator's word, not on the tree.** Nothing
in `be66f79f` distinguished a Sonnet build from an Opus one. The trailer fixes that
going forward; it does not make the check retroactively have happened.

**2. #463 has no build report on its issue.** Plan, verify and the amend note are
all posted; the build report is not. The work is evidenced by the verify pass,
which ran every gate independently — but the row above says "none posted" rather
than linking something that does not exist.

**3. #169 carries one answered-but-unfixed finding**, raised by round 2 and
explicitly **not** a blocker by its author's own reasoning. `membership.ts:71`
passes `mint.error` through, and nothing in the suite pins that it arrives intact:
replacing it with a generic string leaves all 510 files green, because
`github-app.test.ts` exercises the minter directly and never composes the two,
while `membership.test.ts` asserts only `length > 0`. The behaviour today is
correct and it fails **closed** — a real member is never admitted, an operator is
merely under-informed — which is the opposite direction from the blocking finding
below. Left as coverage debt, named here so a reviewer meets it rather than finds
it. The suggested fix is three lines and owns no string: assert identity against
the minter's own error.

---

## What is the sibling case?

- **#169 — the round-1 blocker was exactly a sibling case, and it was found.**
  `checkOrgMembership` sent no `redirect` option, so Node's real `fetch` followed a
  302 to the public-membership endpoint that answers 204, and the function reported
  a documented *"the requester is not an organization member"* as **`member`**. The
  `if (response.status === 302)` line looked like handling and was not — the unit
  test passed only because its stub handed the status back directly. Fixed with
  `redirect: 'manual'`, and round 2 reproduced the original live-GitHub probe
  against both revisions in one process: old → `{"status":"member"}`, new →
  `{"status":"not-a-member","reason":"legacy-redirect"}`. The platform claim under
  it was checked on **v22.22.2, v22.23.2 and v24.12.0** — both ends of the declared
  `engines` range — so the fix does not rest on one interpreter.
- **#462 — the sibling of "the ingest route now verifies keys" is every sentence
  that says it does not.** Five sites, all in fence, all changed in the same commit
  per ruling 12: the `INGEST_KEY_HEADER` and `handleIngest` doc comments, the 401
  body a client actually receives, and the runbook's *"What is not wired up yet"*
  and *"Rotating the ingest key"* sections — the latter said in those words *"treat
  it as housekeeping, not as revocation"*. Its verifier swept for others and
  correctly **declined** one: `postgres.ts:31` mentions "wave 4" about meeting a
  real Postgres host, not about key verification.
- **#463 — the sibling of a new ADR is the index that must list it.**
  `adr-log-law.test.ts` fails **2** tests for a rowless record, not 1: the
  membership check, and its own detector self-test, which deletes `recs[0]`'s row
  and asserts the missing list is exactly `[first.file]` — a rowless 0050 makes it
  `[0001, 0050]`, so the self-test fails for a *different reason*. Verified by
  deleting the row and observing exactly those two.

- **#169, before the amend: a JWT shifted wholesale into the future.** The test
  pinned `exp - iat === 660` — the window, not where it sits — so moving both by
  +60s left 15 tests green while shipping a mint GitHub always rejects. Now pinned
  absolutely against the injected clock. Round 2 also planted the two mutations the
  deleted `not.toBe` assertions could conceivably have caught and confirmed
  something else reddens each time, so the deletions took no coverage with them.
- **#462: a memoised key check.** Its verifier invented a mutation outside the
  lane's declared twelve — caching the `checkKey` thunk across requests instead of
  resolving per request — and `api/api.test.ts`'s *"REVOKE BETWEEN TWO BATCHES"*
  went red, 202 where 403 was expected. The once-per-batch claim is load-bearing
  (it is what bounds revocation lag to one batch interval) and it is pinned.
- **#463: a rowless ADR**, above.
- **The one that still survives** is #169's error propagation, gap 3 above.

## Fence widening

**None taken. One considered and declined**, ruled 2026-09-14 and recorded on
[#462](https://github.com/launchpad-26/rhizomorph/issues/462#issuecomment-5656234190)
before any change.

`packages/team/src/api/http.ts:33-36` says key-value verification is *"wave 4's and
is loudly unimplemented"* — false as of `7f8aa133`, and stale twice over since the
2026-09-11 renumbering moved that work off wave 4 anyway. Fixing it is a four-line
comment edit touching no code; ruling 12 wants it in the same commit; #462's fence
forbids the file. The lane raised it at plan time and again at build and correctly
refused to widen its own fence.

Declined because **wave 6 was groomed around refusing the `api/http.ts` collision**
— both halves of ruling 8 wanted that file, and the machine plane got through only
because `IngestDeps` is constructed in `api/main.ts` and needed no route. Spending
that property on four lines of prose buys nothing wave 7's rewrite hands over for
free, and the stale sentence sits in the header of the file being replaced. Booked
as a wave-7 debt in `docs/prds/prd-51-the-split.md` (merged in #466).

## Plan deviations

- **#462** changed a plan-specified assertion from
  `expect(JSON.stringify(body)).toContain(refusal)` to an equality on the error
  field. Its verifier ran both forms against the real refusal strings: the plan's
  form returns **false** for `malformed` and `wrong-project`, because those strings
  contain literal `"` that `JSON.stringify` escapes — it would have failed against a
  correct implementation. The replacement is strictly stronger.
- **#169** was amended after round 1 (`e04e39c4` → `db3041f9`), deleting two
  assertions its author judged inert and **rebutting** a third finding with
  evidence. Round 2 checked the rebuttal rather than accepting it: the claim that
  `github-app.test.ts` cases 38/39 own the wording and redden 3/3 is true as
  stated, and the boundary argument holds — `membership.ts` receives that string
  and does not construct it.

## Gates, on the assembled wave

Run on this branch with all three commits present — the first time they coexist:

- `VITEST_MAX_WORKERS=6 npm test` — **512 files, 9664 passed**, 3 expected-fail, 7 skipped
- `npm run typecheck` — clean
- `npm run lint` — clean, 1090 files
- `npm run build` — clean
- boot smoke — `GET /` → **200**, `GET /api/meta` → **401** (the expected gated read)

Node v22.22.2, darwin/arm64. **This is one platform's verdict and not "CI passed"**
— all four workflows are disabled on this repo, so this PR carries no check runs.
`scripts/gate.sh` was **not** run: running it *is* the landing, and that is the
operator's step.

One thing a reviewer running the suite under load should know: `doc-citation-law.test.ts`
has a 5,000 ms budget against a ~3.6-4.0 s body and times out when the machine is
busy. It was reported as a red on `main` during one verify pass that ran alongside
seven concurrent agents at ~1.1 GB free; three runs on a quiet machine pass at
3.62-4.01 s. Not a defect in this wave, but it is thin enough to waste a landing,
and it probably wants a `// @gate-timing` marker so the gate runs it serially.


KelliherL (commented):
## Review of #476 — one substantive fix pushed (blocking on your confirmation), one test added, `origin/main` merged in

Read commit-by-commit (`0a5b90bf` #169, `7f8aa133` #462, `f1ec86e2` #463) against each issue's Definition-of-done bullets, in a clean clone at node 22.23.2 on Linux. The issue-level verify passes are your kit's model seats; this is the repo's independent read of the assembled head, and I treated it as the security boundary it is. **Most of it verified in the PR's favour**, and the specifics come first because they are the evidence.

### What verified — EXECUTED on your head `f1ec86e2`

- `npm run typecheck` **0** · `npm run lint` **0** (1090 files) · every touched test file plus `adr-log-law`, `doc-citation-law` and `api/http.test.ts` — **15 files, 296 passed**.
- `scripts/fence-lint.sh 169 462 463` **PASSED**; by hand, every touched path is inside one of the three fences and the three sets are disjoint, as the body says.
- **Your mutations, re-derived** (each restored):
  - `redirect: 'manual'` dropped from the members GET → **2 red**, including the test that names the 302-to-public-endpoint trap by name. The round-1 blocker stays pinned.
  - the key check **memoised across requests** in `main.ts` → *"REVOKE BETWEEN TWO BATCHES: the first is 202 and the second is 403"* **red**. Once-per-batch is load-bearing and pinned.
  - a `plaintext text` column planted in `0005_ingest_keys.sql` → *"declares exactly four columns"* **red**.
  - `takePlaintext()` answering twice → *"SHOWN ONCE: a second takePlaintext throws"* **red**.
  - the wrong-project check disabled → **2 red** (the tape-ends-at-validate test and the api-level 403).
- **One of mine:** the key verdict consulted **after** `journal.append` instead of inside `validate` → **4 red** (the ordering tape, the journal-holds-only-accepted test, the second-seq test, and the non-key refusal). A refused key really cannot touch the journal.
- **REASONED, on the code:** `resolveIngestKeyCheck` does no storage read for a malformed value and exactly one for a well-formed one; refusal strings name the prefix and never the presented value, and the only wire-sourced text in them is the request's own project, JSON-escaped; `encodeURIComponent` on both path segments of the members GET; the mint's own failure is `{status:'error'}` and never `not-a-member`; `insertIngestKey` is `ON CONFLICT DO NOTHING` so a re-seed cannot resurrect a revoked key, and `seed.ts` inserts before it revokes so the project never holds zero live keys; `init.sh` hashes with `printf '%s'`, and `init.test.ts` runs the real script against `hashIngestKey`. ADR-0050 records the decision and both rejected alternatives from the amendment's own prose, cites no lane code, and has its README row.

### One substantive finding — fixed as `7a5938bd`, not approving my own fix

**The 503 put the database's own error on the wire, to a caller whose key had not been verified.**

`main.ts`'s fail-closed path is right to refuse — and the body was `…refused rather than accepted: ${cause.message}…`. That branch is reached by anyone who can hit the socket with a **well-formed** `rzk_` value: shape passes, the storage read runs, and if the read throws, the response carries what it threw. A `postgres` error is *"connect ECONNREFUSED postgres:5432"*, *"password authentication failed for user "rhizomorph""*, *"relation "ingest_keys" does not exist"* — a host and port, a role, a relation, to an unauthenticated caller on the one public route this server has. Your own refusal strings are held to *never the presented value*; this one handed out the database's sentence instead. `http.ts:84` and `:116` also reflect a `cause.message`, but those are the caller's own body failing to read or parse; this is the server's inside.

**The fix**, in your idiom: `startTeamServer` gains `onError`, the same optional-callback shape as `onBatch`; `deploy/serve.ts` wires it to `console.error` beside the boot report; the wire keeps the fact and the remedy (refused, nothing journalled, retry) and drops the cause. Your existing 503 test is untouched — it pins the refusal sentence and the empty journal, neither of which moved. A second test asserts the fake storage's sentence is **absent from the body and present in `onError`**. Mutations, EXECUTED from green: the cause put back on the wire → **red**; the `onError` call removed → **red**.

**Not approving, on purpose:** this changes behaviour in your commit and adds an option to `StartTeamServerOptions`. Please confirm `7a5938bd` reads right to you — in particular whether `onError` is the seam you want, or whether the operator-side sentence belongs on a logger this package does not have yet. Revert is one commit.

### One test added — `15705887` (your gap 3, closed)

I re-ran the mutation you named: `mint.error` replaced with a generic string → **green across `membership.test.ts` and `github-app.test.ts`**, exactly as you measured. The new case hands the same fetch stub to `mintInstallationToken` and to `checkOrgMembership` and asserts the second's error *equals* the first's — owning no string, as you suggested. A 500 from the mint POST is used because its message is deterministic. The same mutation now reddens exactly this test.

### The merge and the union

`origin/main` moved under you (#474). Merged **into** the branch as `5c57935a` — clean. The whole leg on the union, with your own `ci-local.sh --no-install`:

```
  Test Files  512 passed (512)
       Tests  9704 passed | 3 expected fail | 3 skipped (9710)
  ci-local: green — exit 0 in 150 s, Linux / node v22.23.2
```

Your 9664 was on `f4b7b574`; the difference is the two tests above plus what `main` gained from #474.

One platform (Linux, node 22.23.2); the workflows are disabled, so nothing witnesses macOS or Windows here either. `doc-citation-law` did not time out in this run.

- **`0005`'s `GRANT SELECT ON ingest_keys TO rz_ingest` is inert under the `ENABLE ROW LEVEL SECURITY` two lines above it.** RLS with no policy is default-deny for every non-owner, so `rz_ingest` gets the grant and sees zero rows. Today that is moot — the once-per-batch read runs on the owner connection, as the header's point 3 says — but the day a later wave splits the connection as the header anticipates, every key reads as *unknown* and every batch is refused, fail-closed but silent about why. Worth one sentence in the header saying the grant waits on a policy; I did not touch the migration, since it may already be applied on the host and its checksum is tracked.
- **`exp = iat + 660` puts the JWT's expiry exactly at GitHub's 10-minute ceiling from `now`.** GitHub's own example does the same, so this is not a defect, but a host clock a few seconds ahead of GitHub's turns every mint into *"'exp' claim too far in the future"* — loud and fail-closed, per your bullet, so only a note: `now + 540` leaves drift margin at no cost.
- ADR-0050 says the check runs "with `read:org`"; that is the OAuth scope's name — the App permission is *Organization → Members → Read-only*, which the same paragraph names correctly two lines later. One phrase.
- `main.ts` duplicates `http.ts`'s route match so a `GET /` buys no read — you say wave 7 absorbs both. Agreed; noting only that if the two ever disagree the default thunk refuses, which is the right direction.

### What I did not run

macOS and Windows. A real Postgres (`storage.test.ts` and `api.test.ts` prove the statements and the route over the fake; `postgres.ts` was read, not driven). A live GitHub probe — your round 2 did that on three Node versions, and I re-derived the `redirect: 'manual'` pin instead. `init.sh` end to end on this box beyond what `init.test.ts` runs.

**Verdict: changes are already pushed; waiting on your word for `7a5938bd`.** If `onError` is wrong, say what the seam should be and I will move it. Landing is the operator's act.

ciaran-slow (commented):
## Author's confirmation of the pushed fixes — `7a5938bd` and `15705887` both stand, at head `5c57935a`

Thanks for taking the 503 seriously enough to push rather than annotate. Both commits read right to me. Taking them in the order you asked.

### `7a5938bd` — `onError` is the seam I want

The finding is correct and I had it the wrong way round: the branch is reached on a **well-formed** `rzk_` value, so `resolveIngestKeyCheck` throwing answers a caller nobody has authenticated, and `connect ECONNREFUSED postgres:5432` / `password authentication failed for user "rhizomorph"` / `relation "ingest_keys" does not exist` each name something that caller has no business learning. My own refusal strings are held to *never the presented value* three functions away; this one handed out the database's sentence instead. That is the inconsistency, and it is mine.

**`onError` is the right shape, and I do not want a logger here yet.** Reasons, in order:

1. It matches `onBatch` exactly — optional, defaulted to a no-op, wired by the composition root. A second callback on the same interface costs a reader nothing; a logging abstraction introduced for one call site costs them a new vocabulary to learn before they can read the 503.
2. The callback keeps `packages/team/src/api/` free of a transport decision. `deploy/serve.ts` chooses `console.error`; a later deploy that wants structured JSON on stderr, or a drain, changes one line at the root and no code under `src/`.
3. When a logger does arrive it will arrive with levels and fields, and it will want more than one call site to justify itself. At that point `onError` is what it replaces — one seam, one edit, and the tests that pin the wire do not move.

So: keep it. If a second case appears before the logger does — a fold-side failure that must not reach the wire, most likely — I would still add it through this callback rather than reach for the abstraction on the second instance.

The test reads right too: the fake storage's sentence asserted **absent from the body and present in `reported`** owns no production string and reddens from both directions, which is what the two mutations you ran show.

**Sibling I checked, and it is genuinely the other side of your line.** `packages/team/src/ingest/handle.ts:152` reflects `cause.message` onto the wire the same way — a journal `append`/fsync failure, so a filesystem path rather than a database host. But it sits **after** the `if (!verdict.ok) return` at :111, so the caller reading it has presented a key this server verified this batch. That is a different exposure from yours, and the boundary your commit message draws — *"a caller whose key has NOT been verified"* — holds as written. Not a gap in this fix. I will carry it as a note for wave 7, which is already absorbing the duplicated route match, and it can go through this same `onError` when it does.

`http.ts:84` and `:116` I agree are the caller's own body failing to read or parse, and stay as they are.

### `15705887` — the right closure for gap 3

Asserting the second error *equals the minter's own* is a better answer than the one I was reaching for, which was going to own a string and rot the first time the minter's wording moved. Handing the same stub to both and comparing is the version that cannot drift. The 500-with-no-JWT-bytes choice for determinism is the detail that makes it work, and the comment says so, so the next reader will not "simplify" it to a 401.

### What I ran, and what I did not

**REASONED**, all of it: I read both diffs and the surrounding code at `5c57935a` — `main.ts`'s listener wrapper end to end, `handle.ts`'s refusal ordering, `http.ts`'s two reflections, and every `cause instanceof Error ? cause.message` site in `packages/team` — and checked the sibling above by reading the control flow, not by running it. **I executed nothing.** Your `ci-local --no-install` green on the union (512 files, 9704 passed, Linux / node v22.23.2) is the executed evidence on this head, and it is yours, not mine — one platform, with the workflows disabled, so macOS and Windows are unwitnessed here as you say.

### Your four offered calls

Not answering them in this comment — they are all real and none of them blocks. The `GRANT` / RLS one is the one I care about most and it wants a sentence in `0005`'s header rather than a migration edit, for exactly the checksum reason you give. I will take all four in a follow-up rather than inflate this range.

**Confirmed: `7a5938bd` stands as pushed, no revert.** Landing is mine and I have not done it yet.

Issue #473, opened 2026-09-13
doc-citation-law: a live PRD makes a live claim, and its cited paths go unchecked


KelliherL:
## Fence (may touch ONLY)
Not yet groomed into a wave. Fence is drafted, not claimed:
- `packages/server/src/doc-citation-law.test.ts`
- the live PRDs whose dead citations the narrowing convicts

## Blocked by
prd51 wave 6 landing, because one of the documents this would convict is prd-51 itself, which is being actively worked right now. Dispatching this over a live lane's PRD is a collision, not a sweep.

**Model:** opus. **Wave:** not groomed.

## What the audit found

doc-citation-law excludes three directories as CITING sources — research, review, and the whole PRD tree including the twelve live PRDs — on the reasoning that a dated artefact's citations record the tree at a commit rather than making a live claim.

That reasoning is sound for an archived or dated document. It is not sound for a LIVE PRD. A PRD still on the live shelf is a document people build from this week, and its cited paths are claims about the tree now.

Measured on the trunk during the documentation audit: roughly 82 dead backticked citations across roughly 27 files sit inside those three excluded directories. Four are in prd-51, which is live and actively being built against. None are caught by anything.

## The proposal, not yet ruled

Narrow the exclusion from the whole PRD tree to the archived and parked shelves only, so live PRDs are swept as the live claims they are. The law already draws exactly this distinction elsewhere — it exempts a document by what it IS, via a resolving tree pin, rather than by which directory it sits in — so the machinery to do this honestly already exists.

## Definition of done

- Whether to narrow the exclusion is RULED, not assumed. Either outcome is acceptable and the reasoning is recorded; refusing the narrowing with a stated reason closes this issue legitimately.
- If narrowed: every dead citation the narrowing convicts is either fixed or moved to the law's existing allowlist, and each allowlist entry carries the honesty check that list already requires — it must still be cited by its file and must still fail today.
- The count is re-measured at landing rather than taken from this issue.
- The law is shown to bite on a live PRD specifically, and the mutation is recorded.

KelliherL:
## RULED: the exemption stands. Live PRDs keep it.

Operator ruling, 2026-09-14. The DoD said either outcome was acceptable with the reasoning recorded, and refusing with a stated reason closes this legitimately. Here is the reason, measured rather than argued.

### The experiment

[...]

PR #478, opened 2026-09-13
docs-audit w1: the register, the index and the ADR the generated wiki needs

### LINKED ISSUE #467: docs-audit w1: a dated register states every documentation finding and the command that produced it
## Fence (may touch ONLY)
- `docs/review/2026-09-14-documentation-audit.md`

## Blocked by
Nothing. This is the wave-1 keystone: every later issue in this milestone cites it as its evidence.

Sibling lanes hold other surfaces: ciaran-slow holds prd30 w5 and prd51 w6; gabriel-canaan holds the charter-laws contract issues. This lane touches one new file and nothing else.

## Definition of done

- The file exists and opens with a resolving tree pin in the repo's idiom, a Tree line naming a ref and a sha, within its first 12 non-fenced lines, so doc-citation-law treats it as a dated artefact rather than a live claim. The sha is an ancestor of the landing ref.
- Every finding is graded EXECUTED or REASONED, and every EXECUTED finding carries the command that produced it and that command's actual output — not a paraphrase.
- The register records, at minimum, these findings, each re-derived on the landing tree rather than copied from this issue:
  - The README support matrix claims CI verification on every push for Linux, macOS and Windows, and names the Node floor as CI-pinned. GitHub Actions last ran 2026-09-12, every recent run failed, and nothing has run since — through pushes on 09-13 and a merge on 09-14. AGENTS.md and CONTRIBUTING.md were both updated when Actions was retired for cost; the README was not, and carries zero mentions of ci-local, pr-verdict or the Passed local CI label.
  - The support-matrix law in packages/server/src/api/route-class-law.test.ts passes anyway, because it compares README prose against the ci.yml FILE, which is still in the tree. The law is file-versus-file; what moved is whether the file is ever executed. Record this as the class of failure it is, alongside the repo's own prior instances of reporting a check without honouring it.
  - AGENTS.md line 626 still claims windows-suite.yml enforces per file on every push.
  - docs/roadmap.md still reasons from a private repo and carries an open-sourcing prep item gated on going public. The repo is public. That item's user.email scrubbing is already discharged by the OTel fixture-hygiene law, and the roadmap does not know it. State plainly that a real-leak check was run and found none: the only two real addresses in the tree are the deliberate maintainer fields in packages/app/package.json and packages/app/electron-builder.yml.
  - docs/architecture.md, at roughly 2,960 lines, self-declares that its narrative stops at prd-17. Roughly 39 PRDs have landed since. Its last touch was a mechanical citation update, so last-modified date is not evidence of currency — say so, because that is how this document reads fresh while being stale.
  - There is no docs/README.md. 260 tracked markdown files, and indexes exist only for the adr, prds and review subdirectories. There are also two research trees, one under docs and one at the repo root, and nothing indexes both.
  - Roughly 82 dead backticked citations across roughly 27 files sit inside the three directories doc-citation-law excludes as citing sources. Four of them are in a LIVE PRD, prd-51, which makes live claims. Give the measured number from the landing tree, produced by running the law's own extractor with its exclusion list emptied — not by a hand-rolled regex.
  - 31 of 55 PRDs carry no Status line, including 5 of the 12 live ones. Nothing tests for it.
  - Only one of the seven user-guide pages carries claim markers. The other six have no mechanical claim coverage beyond the generic citation and symbol laws.
  - docs/vision.md is the oldest live non-dated document in the repo, untouched since 2026-08-03 and only then by a mechanical rename.
- The register ends with an open-questions section that names, and does not answer: whether docs/vision.md is deliberately timeless; and whether the repo's enabled-but-never-created GitHub wiki should be turned off or populated.
- The register states what it did NOT check, so a later reader knows the boundary of the sweep.

### LINKED ISSUE #468: docs-audit w1: an ADR records why a third party publishes a generated wiki about this repo
## Fence (may touch ONLY)
- `docs/adr/0050-the-generated-wiki.md`
- `docs/adr/README.md`

## Blocked by
Nothing. Additive, and the wave's single claimant of the next ADR number.

Sibling lanes hold other surfaces: no other issue in this wave may file an architecture decision, because the log is append-only and numbered — two lanes claiming the next number is a sequencing problem, not a merge conflict, and whichever lands second renumbers.
**Model:** opus. **Wave:** 1.

## Definition of done

- The record is numbered 0050. Verified free on the trunk at grooming: the highest existing record is 0049. Re-verify at landing; if another lane has taken it, this record takes the next free number and the filename changes with it.
- It has a table row in the ADR index, the row's link resolves, and the row is labelled with the same number as the file. Planting a rowless record fails TWO of the adr-log law's tests, not one — the membership check and the law's own detector self-test — so a lane reading only the first failure will not expect the second.
- It names rejected alternatives, or it is not an ADR by this repo's own standard. At minimum: self-hosting deepwiki-open (MIT, Docker, requires model API keys — a service and a credential to maintain); GitHub's own wiki, which is already enabled on this repo and was never created, and which nothing generates; a native index alone; and doing nothing.
- It states what the decision actually costs and what bounds it. Specifically: this is free and needs no credential ONLY while the repo is public. If the repo ever goes private, DeepWiki requires a Devin account, wiki generation above the free effort level bills against a subscription, and the vendor's own security documentation says data may be used for model training by default with opt-out available only on a paid plan. The record says what happens to this decision in that event.
- It states the consequence that makes this more than a link: a generated wiki is a derived public document about this repo that none of this repo's laws can redden. The record says what the repo does and does not promise about it.
- It states why the index is ordered after the documentation corrections rather than before: the wiki generates from what the repo says, so publishing first would have published the false support-matrix claim.
- Status is accepted, with the date.

### LINKED ISSUE #469: docs-audit w1: every document is reachable from one index, and a law says so
## Fence (may touch ONLY)
- `docs/README.md`
- `packages/server/src/docs-index-law.test.ts`

## Blocked by
Nothing. Both files are new.

Sibling lanes hold other surfaces: the wave's other two lanes hold the review directory and the ADR log respectively. This lane adds one index and one law and touches no existing document.

## Definition of done

- docs/README.md exists and carries a table with one row per top-level document under docs and one row per subdirectory of docs, each row saying in a sentence what the reader will find there and when to reach for it.
- It names the two research trees explicitly — the one under docs and the one at the repo root — because nothing currently indexes both and a reader cannot otherwise discover the second.
- It does not restate the routing rules that already live in CONTRIBUTING.md and AGENTS.md. It links to them. A second copy of a rule is a second thing to go stale.
- The law lands in the SAME commit as the index, so it lands green rather than convicting its own evidence.
- The law asserts, at minimum: every top-level markdown file under docs has a row; every subdirectory of docs has a row; every link in the index resolves to a path that exists; and no row names a path that is not there. Both directions, so neither a new document nor a deleted one can drift past it.
- The law is shown to bite, and the mutation is recorded in the PR: delete one row and watch it go red, restore; add a row for a path that does not exist and watch it go red, restore.
- The law is written so it cannot pass vacuously — it asserts that the set of documents it found is non-empty, and that it reached the directories it claims to sweep. A law whose probe silently matched nothing is the failure mode this repo has already been bitten by.
- The law lives under the packages/server source tree for the reason every sibling repo-scope law there records: the root vitest config globs the packages directory, so a root-level test would never run and would be its own vacuous law.

KelliherL:
Wave 1 of the `docs-audit` milestone: the register that measures the drift, the decision that admits a generated wiki, and the index the tree never had.

Closes #467
Closes #468
Closes #469

`AGENTS.md`'s two conditions for opening a PR: **condition 1 is NOT satisfied.** Every commit here was written and verified by the same agent, and that file says plainly that evidence you authored yourself is not evidence. No independent pass has read this range. It is written here rather than recorded away, which is what that section asks for.

Condition 2 is satisfied: the operator asked for these PRs in session, 2026-09-14.

## Commits, one per issue

|---|---|---|
| `852e1832` | #467 | a dated register states every documentation finding and the command that produced it |
| `a459e974` | #469 | every document is reachable from one index, and a law says so |
| `5b62187a` | #468 | ADR-0050 records why a third party publishes a generated wiki about this repo |

Committed in that order deliberately: the ADR cites `docs/README.md`, so the index lands first and `doc-citation-law` stays green commit-by-commit rather than only at the tip.

## The headline finding

The README promised CI verification on every push for three platforms. **Actions last ran 2026-09-12** and four commits since triggered nothing.

The part worth a reviewer's attention is that the law guarding that section is **green**, and one of its passing tests is named *"the Linux row claims CI, and CI delivers it."* It is not broken — it compares README prose to the `ci.yml` file, both are intact, and what moved is whether the file is ever executed. That is this repo's own *reporting a check without honouring it*, one level out: checking a document against an **artefact** instead of against the world. Wave 2 fixes both halves.

## Verification

- **Lane gate:** typecheck clean, lint clean (1080 files), full suite run.
- **Attribution by failing TEST NAME** against an unmodified `origin/main` baseline built in its own worktree at the same sha: branch 248 failing, main 253, and the **new-on-branch set is empty**. Five names fail on main and pass here; they are load-sensitive registry sweeps and are *not* claimed as fixes.
- **`docs-index-law` mutations, executed:** deleting a row reddens 2 of 6; adding a row for a file that does not exist reddens 2 of 6; restoring returns 6 passed.
- `adr-log-law` 6/6 with the new row. `doc-citation-law` unchanged at baseline. `corpus-eol-law` and `no-personal-paths-law` green.
- `fence-lint.sh 467 468 469` passed, **extracting 1+2+2 paths** — checked, not merely exited 0, because a lint that parsed nothing passes vacuously.

## Two corrections to earlier claims of mine

- A `KelliherL/rhizomorph` manifest mismatch I reported **does not exist**; I had read a stale checkout 2,220 commits behind. `manifest-law` pins the right slug.
- The dead-citation count is **45, not ~82**. Remeasured with the law's own extractor instead of a hand-rolled regex. The method is the lesson and the register records it.

## Notes for the reviewer

- Nothing here changes behaviour; it is documents plus one new law.
- The register is a dated artefact with a resolving tree pin, so its own citations read as a record rather than a live claim.
- #421 remains open and unrelated. #473 (live PRDs lose the citation exemption) is filed but deliberately ungroomed — it convicts prd-51, which has live lanes.


gabriel-canaan:
## Review — read commit by commit, `origin/main` merged in, findings fixed on the branch

Three fix commits pushed: `35798fe7`, `20b0f5c0`, `2cecc9ac`. Everything below was run, not argued, unless marked otherwise.

The headline is real and it reproduces. `route-class-law.test.ts` is **95/95 green**, including *"the Linux row claims CI, and CI delivers it"*, while `gh run list` returns **0 runs since 2026-09-12**. F2 is the finding, and it stands.

F5's **45** is better supported than the PR body claims. I reached it by a different route: on Linux the law reports **0** violations unmodified and **45** with `EXCLUDED_DIRS` emptied — read off directly, where your `60 − 15` reaches it by subtraction. Two independent routes to the same number is the strongest thing that can be said about it.

F3, F6, F8, F9, F11 all reproduce at the pin (README 0 / CONTRIBUTING 4 / AGENTS 2; 260 tracked md; 56 claim markers in `the-lab.md` and 0 in the other six; 2,962 lines; `f5af1fcb` 2026-08-03). The tree pin `f4b7b574` resolves and is an ancestor of `origin/main`. Fence is exact — five files, precisely the three issues'.

Both `docs-index-law` mutations reproduce **exactly** as stated: 2 of 6 each, 6/6 on restore. I added a third you did not claim — a tracked subdirectory nobody indexed — and it bites too. All three arms are load-bearing.

### Fixed on the branch

**1 — `2cecc9ac`: ADR 0050 was taken while this was in review.** `origin/main` now carries `docs/adr/0050-two-identity-planes.md` (prd-51 ruling 8, `f1ec86e2`). This is the case #468 named: *"if another lane has taken it, this record takes the next free number and the filename changes with it."* Renumbered to **0051** — file, heading and index row in one commit. `adr-log-law` 6/6; the law has no contiguity rule, so the gap at 0050 on this branch is not a failure and closes when `main` merges in.

This PR now reports `CONFLICTING`: `docs/adr/README.md`, two records appending a row after 0049. Resolution is unambiguous — keep both, main's `0050` above the new `0051`. I did not pre-merge `main` into the branch; that belongs to the landing.

**2 — `35798fe7`: the register's figures are native-Windows figures and it never said so.** Two do not reproduce elsewhere:

| | as written (Windows) | Linux |
|---|---|---|
| F5, unmodified tree | 15 | **0** |
| F5, `EXCLUDED_DIRS` emptied | 60 | **45** |
| F2, "the single failure in that suite on this platform" | 1 known | **none — 95/95** |

A reader reproducing either on Linux concludes the register is wrong when it is not. Header now names the platform, F5 carries both columns, F2 states both readings.

The Windows attribution is an **inference** — the register quotes a backslash path and both platform-dependent figures match `.windows-known-failures` entries exactly — not something observed. If the audit ran elsewhere, that line is the one to correct.

**3 — `35798fe7`: F1's second command could not have produced its output.** `gh run list --limit 3 --jq …` with no `--json` exits 1 on gh 2.97.0: `cannot use --jq without specifying --json`. Added the flag, ran it, and it reproduces those three rows. The block was also column-aligned, which `gh` does not emit — replaced with the real bytes, since #467 asks for the actual output and not a paraphrase.

**4 — `35798fe7`: F7 named `prd-14` among PRDs carrying no Status line. It carries one** — `**Status:** BLESSED 2026-08-06`, at line 21. The count of 28 is right, but only under a criterion the finding never stated: the `> **Status:**` blockquote at the head that `docs/prds/README.md` specifies. Criterion now stated with the command that applies it, and the five live PRDs separated into the three different defects they are — two with no Status anywhere, two carrying `> **Outcome:**` instead, and prd-14 carrying the field in the body rather than the head.

**5 — `20b0f5c0`: `docs/README.md` said design-notes holds 27 notes; `origin/main` holds 28** (`docs/design-notes/attention-chip-width.md`, #474, landed after the pinned tree). Corrected. Every other count checks out on `origin/main`: 7 user-guide pages, 50 ADRs, 12 live / 41 done / 2 parked PRDs, 37 research notes.

**6 — `20b0f5c0`:** the law's preamble stated *"This repository tracks 260 markdown files"* as a live fact — 260 at the pin, 262 on `origin/main`, different after the next merge. Reworded to scene-setting-as-of-writing. And `deadLinks()` resolved through a ternary whose two arms were the same expression; `path.join` walks a leading `../` out of `docs/` by itself. Collapsed. Verified the collapse kept the behaviour rather than assuming it: a broken `../` link still reddens 2 of 6, and the index's own `../CONTRIBUTING.md` / `../AGENTS.md` / `../research/` links stay green.

### Left for you, not fixed

**Seven of the eleven EXECUTED findings carry no command** — F2, F4, F6, F7, F8, F9, F11 (F10 has one for half its claim). Most carry output, which is the harder half, but #467 asks for both. Only you hold the originals, so this is not mine to write.

**None of the index's counts is held by anything.** The index is the one document whose whole job is to say what exists, and its numbers are the part that goes false quietly — which is the argument ADR-0051 makes against option B, two commits earlier in this same PR. Either the law counts them or the index stops stating them. That is a call for you and the operator; I changed only the wrong number.

### Verification

- Suite on `2cecc9ac`: **509 files, 9563 passed**, 0 failed. Typecheck clean. Lint clean, 1080 files.
- Earlier, on your head with `origin/main` merged in: 509 files, 9597 passed, 0 failed.
- **Linux only.** Silent about macOS and native Windows.
- `fence-lint.sh 467 468 469` passes.

Condition 1 as your body declares it: this pass read every commit in the range and ran the mutations itself, so it is that independent read. Nothing is landed — the merge is the operator's.

---
*Reviewed by Claude Opus 5 (1M context) via `rhizo-review`. One reviewer, one platform — not a substitute for a second pair of human eyes on anything non-trivial.*

KelliherL:
## Author confirmation — all three fixes verified independently, and the two open calls answered

Re-measured rather than accepted. Every check below I ran myself on `2cecc9ac`.

### Confirmed

**`2cecc9ac`, the ADR renumber.** Verified both sides: `origin/main` carries `docs/adr/0050-two-identity-planes.md`, this branch carries `docs/adr/0051-the-generated-wiki.md`, no clash. This is exactly the case #468's DoD anticipated and it is the right resolution. `adr-log-law` 6/6. Agreed that the `CONFLICTING` state is the landing's to resolve and that keeping both rows with main's `0050` above the new `0051` is unambiguous — I have not pre-merged either.

**`35798fe7`, the platform labelling.** This is the finding I'm most glad you caught, and it is worse than a missing label: I ran the audit on native Windows and wrote figures that read as universal. F5's `15` and `60` are Windows-only artefacts of a self-exclusion that compares backslash paths, and F2's "the single failure in that suite on this platform" is **none** on Linux. A reader reproducing on Linux would have concluded the register was wrong when it was right. Your inference about where it ran is correct — native Windows, this box.

**`35798fe7`, F1's command.** Correct and mine. `gh run list --limit 3 --jq …` without `--json` exits 1; I ran a different invocation and wrote this one down. That is the failure mode #467 exists to prevent, appearing in #467 itself.

**`35798fe7`, F7 / prd-14.** Correct. It carries `**Status:** BLESSED 2026-08-06` in the body; my criterion was the `> **Status:**` blockquote head the PRD standard specifies, and I never stated it. Splitting the five live PRDs into the three different defects they actually are is better than the count I wrote.

**`20b0f5c0`.** The 27→28 design-note count was wrong within a day of being written — see the ruling below, which is downstream of exactly that. The `deadLinks()` ternary was dead code and collapsing it is right; thank you for verifying the collapse preserved behaviour rather than assuming.

### The two left for me — both done in `aec27127`

**Commands.** Four still lacked one (F4, F5, F9, F11 — your list predates `35798fe7`, which supplied the rest). Each re-run against the pin `f4b7b574` and the real output pasted, not reconstructed. F5's block now shows the whole scratch-edit-and-restore cycle including the `git status` that proves the law file went back, because that measurement mutates a law and a reader is owed the proof it was restored.

**The index's counts — ruled: the index stops stating them.** A law counting them would make them true; what argues against it is your own finding. The index shipped saying 27 when the tree held 28, wrong within a day, because a count in prose is the part of a description that goes false quietly. Holding them mechanically would churn this law on every landing that adds a document, buying precision where it is least useful — a map needs to be *complete*, not a census, and completeness is what the law already checks in both directions. Five counts removed (ADRs, design notes, research notes, the PRD split, "seven pages"), and the page now says why it carries none and points at `git ls-files`. That leaves no claim on that page nothing holds, which I think is what you were actually asking for.

### On the review itself

Your independent route to F5's 45 — reading it off directly on Linux where I reached it by subtraction — is better evidence than what I wrote, and I've said so in the register. Two routes to one number is the strongest thing available here.

Verified on `aec27127`: `docs-index-law` 6/6, `adr-log-law` + `no-personal-paths-law` 46/46, and `doc-citation-law` reports no dead citation that is not its own negative-test fixture. Native Windows this time, labelled as such.

Nothing landed. The merge is the operator's.

PR #480, opened 2026-09-13
docs-audit w3: the generated wiki is linked, configured and pinned to this repo

### LINKED ISSUE #472: docs-audit w3: the generated wiki is linked, configured, and pinned to this repo
## Fence (may touch ONLY)
- `README.md`
- `.mcp.json`
- `packages/server/src/deepwiki-law.test.ts`

## Blocked by
Both wave-2 issues, and the wave-1 ADR. This is the wave's SINGLE README claimant, and it is a wave later than the support-matrix correction for exactly that reason — one issue per wave may claim that file.

Ordered last on purpose: the generated wiki is produced from what this repo says, so it must not be published while the support matrix still claims CI that does not run.

Sibling lanes hold other surfaces: nothing else in this wave touches README.md.

## Definition of done

- The README carries the DeepWiki badge, linking to this repository's own page at deepwiki.com. The badge is not decoration: it is the documented trigger for auto-refresh - Cognition states "We auto-refresh DeepWikis if their repo has a badge", which names the trigger and says nothing about cadence. **Corrected 2026-09-14:** an earlier draft of this line said that without a badge the wiki is "a one-shot snapshot that silently ages". The vendor does not say that, and does not say an un-badged wiki never refreshes; the circulating figures ("weekly", "a five-day lag") trace to no primary source. Carrying the badge is still right. Claiming a refresh interval is not.
- The badge URL names this repository's owner and repo, derived rather than retyped. The manifest law beside it already parses owner and repo out of the package manifest for exactly this purpose; reuse that idea rather than hard-coding a second copy of the slug.
- A repo-level MCP configuration exists at .mcp.json declaring the DeepWiki server. There is no MCP configuration in this repo today, so this is new ground: nothing is being migrated and nothing else reads that file yet.
- The endpoint is the current one, the streamable HTTP endpoint at mcp.deepwiki.com, and not the deprecated server-sent-events endpoint. The server is free, remote and requires no authentication for public repositories — so no credential is introduced by this change, and the PR says so explicitly, because a hand that holds a credential is a thing this repo argues about in public before granting.
- The law lands in the same commit and asserts, at minimum: the badge in the README points at this repo's own owner and repo; the configured MCP endpoint is the streamable HTTP one, not the deprecated one; and the repo slug the MCP configuration targets is the same repo the manifest names. A link that silently points at another project is the failure this is for.
- The law is shown to bite, and the mutations are recorded in the PR: point the badge at another owner and watch it go red, restore; swap the endpoint for the deprecated one and watch it go red, restore.
- The PR body states the bound the ADR records: this is free and uncredentialed only while the repo is public.
- Indexing the repo at deepwiki.com is NOT part of this issue. It is an outward-facing publish and it is the operator's own act, taken after this lands.

KelliherL:
Wave 3 of `docs-audit`: ADR-0050's build half — the badge, the MCP config, and the law that pins both to this repository.

**Base is `docs-audit-w2-corrections`** — this wave claims `README.md`, which only one issue per wave may do, and it is ordered after the corrections on purpose. Land waves 1 and 2 first.

Closes #472

`AGENTS.md`'s condition 1 is **NOT satisfied**: written and verified by the same agent, no independent pass. Condition 2 is satisfied — the operator asked for these PRs in session, 2026-09-14.

## What lands

- The DeepWiki badge in the README. **Not decoration:** it is the documented trigger for auto-refresh - Cognition states *"We auto-refresh DeepWikis if their repo has a badge"*.

  > **Correction, and it is in this PR's own history.** Commit `4c60f702`'s message, and an earlier version of this line, said that without a badge the page is "a one-shot snapshot that ages silently". The vendor does not say that - it names the trigger, states no cadence, and does not say an un-badged wiki never refreshes. The figures circulating elsewhere ("weekly" with a badge, "a five-day lag" without) trace to no primary source; the strongest instance is a llama.cpp PR whose title asserts the weekly number and cites nothing. Commit `bc1c8597` adds the honest version to the README - treat the page's age as **unknown** - rather than rewriting pushed history. The badge argument is unaffected; the cadence claim was never mine to make.
- A short README section saying what the generated wiki is and, more importantly, **what it is not** — it carries no authority, nothing here can make it true, and a reader who needs a claim to hold should read the document and find the law that holds it.
- `.mcp.json`, declaring the DeepWiki MCP server so an agent working in this repo can query it with no setup. This repo had no MCP configuration before; nothing is being migrated.
- `packages/server/src/deepwiki-law.test.ts`.

## Live evidence, before anything was pinned

```
POST https://mcp.deepwiki.com/mcp  → 200
  serverInfo: {"name":"DeepWiki","version":"2.14.3"}
  tools: read_wiki_structure, read_wiki_contents, ask_question
  no credential sent
GET  https://deepwiki.com/badge.svg → 200
GET  https://devin.ai/assets/askdeepwiki.png → 429   (so the former is what shipped)
```

## What the law holds, and why each clause

- **The badge names THIS repository**, derived from the package manifest rather than accepted as a second hand-typed copy. A badge copied from another project's README renders identically — the reader clicks through to a wiki for a different codebase with no way to tell. Same argument `manifest-law` makes about clone URLs, same whole-URL parser, so a host merely *containing* `github.com` is rejected and a dotted repo name is not truncated at the dot.
- **The endpoint is the streamable HTTP path, not the deprecated SSE one.** A config that drifts onto a deprecated path keeps working until it does not.
- **No credential may appear in the declaration.** The public endpoint needs none, and that is the entire basis on which ADR-0050 was acceptable — ADR-0019 rejected a hand holding a credential because *"a hand that holds a credential has something worth stealing"*, and ADR-0034 granted one only after a public argument and a law per clause. A documentation convenience does not earn that. Asserted over the serialized declaration rather than a list of known key names, because a law that enumerates spellings misses the next one.

## Verification

- **Mutations, executed:** pointing the badge at another owner reddens 2 of 6; swapping `/mcp` for `/sse` reddens 1; planting an `Authorization` header reddens 1. Each restore returned 6 passed.
- 177 of 179 pass across route-class, manifest, docs-index, adr-log and no-personal-paths. The two reds are the pre-existing `line-endings` known-failures, both confirmed present in a same-worktree `origin/main` baseline by test name.
- `doc-citation-law` reports no new dead citations. Typecheck clean.

## The bound a reviewer should check the ADR for

This is free and uncredentialed **only while this repository is public**. If it ever goes private, indexing needs a Devin account, the free unauthenticated MCP server stops serving it, and generation above the default effort bills against a subscription — under a vendor whose own security documentation says data may be used for model training by default, with opt-out on paid plans only. ADR-0050 makes going private a documented trigger to supersede it *before* re-indexing.


gabriel-canaan:
## Review — read commit by commit, findings fixed on the branch

Two fix commits pushed: `a8da58d3`, `06de6099`. Everything below was run unless marked otherwise.

All three mutations in the body reproduce **exactly** as written — badge pointed at another owner reddens 2 of 6, `/mcp` → `/sse` reddens 1, a planted `Authorization` reddens 1, each restore returning 6 passed. That is rarer than it should be and it made this review much faster.

`CONTRIBUTING.md#laws-live-in-tests` resolves (`## Laws live in tests`). Fence is exact — three files, precisely #472's. `.mcp.json` sits at the repo root, outside the server package, so `npm pack` is untouched: 430 files, no `.mcp.json`.

The correction in `bc1c8597` is the right call and correctly reasoned. I went looking for a primary source for the "weekly" figure independently and could not find one either. Amending pushed history would have been the worse move.

### Fixed on the branch

**1 — `a8da58d3`, and it is the sibling case: the credential law read one key, not the file.** A bearer token in a *second* server in the same `.mcp.json` left the suite **6 of 6 green**:

```json
"other": { "type": "http", "url": "https://example.com/mcp",
           "headers": { "Authorization": "Bearer SECRET" } }
```

The law's own header says an auth header appearing in *"this config"* is a decision that must be argued — and this PR is what creates the file, so the next server anyone adds to it is the live case rather than a hypothetical. ADR-0019's argument is about the tree holding a credential at all, not about which entry holds it.

The check now reads the whole file, extracted as `credentialSmells(text)` so the live test and the bite test call the same code. That is the discipline `docs-index-law.test.ts` states and this file already had for its link parsing but not for this check — a bite test built on its own literals cannot observe a production check reading the wrong object.

Mutations, each restored and re-run against the committed fix:

| mutation | before | after |
|---|---|---|
| credential in a second server | **6/6 green** | 2 of 6 red |
| credential on the deepwiki entry | 1 red | 2 of 6 red |
| pattern loosened to one segment | — | 2 of 6 red |
| clean tree | 6 passed | 6 passed |

**2 — `a8da58d3`: the `startsWith('badge.')` filter was dead code, and the assertion under it passed for another reason.** The pattern requires two path segments and `deepwiki.com/badge.svg` is one, so the filter was unreachable — deleting it outright left all six tests green. That means *"the badge asset alone names no repository"* was demonstrating something other than what the comment beside it named, which is this repo's own "a test that cannot fail for the reason it claims", at its smallest.

Filter removed; the doc comment now names the two-segment requirement as the mechanism. The assertion stays, because it does guard the real risk — loosening the pattern to one segment reddens it, which is the third row above.

**3 — `06de6099`: ADR-0050 renumbered to 0051, and this PR cited it four times.** `origin/main` took 0050 for prd-51's two-identity-planes record (`f1ec86e2`) while this wave was in review — the case #468 anticipated. The reasoning lives in `2cecc9ac` on the w1 branch. Here it moved the README's closing link, two lines of `deepwiki-law`'s header docstring, and the credential test's own name.

The rename is repeated on this branch rather than inherited, because this branch is three deep in a stack and does not carry the w1 fix commits. Both branches now perform the identical rename — verified to merge without conflict. The alternative was a branch citing 0051 while its own tree still held 0050, red on `doc-citation-law` until the stack was restacked.

**4 — `a8da58d3`:** `bc1c8597`'s insertion ran *"Treat the page's age as unknown…"* into *"When a claim has to hold…"* with no blank line, so the two rendered as one paragraph. Blank line restored.

### Two things for you, not fixed

**#472's definition of done asks that "the repo slug the MCP configuration targets is the same repo the manifest names".** `.mcp.json` names no slug — `https://mcp.deepwiki.com/mcp` is repo-agnostic and the repo is passed per query — so the clause is **unmeetable as written rather than unmet**. It wants a line on the PR saying so, not a law that pretends. Recording it here rather than leaving it silent.

**The badge links to an empty page until the repo is indexed.** `deepwiki.com/launchpad-26/rhizomorph` returns 200 but 31 KB with no wiki content and no *"Last indexed"* line; a control fetch of a known-indexed repo returns 832 KB with one. #472 says indexing is deliberately the operator's own act after this lands — but the README section describes the page in the present tense, so between merge and indexing the front door carries a claim that is not yet true. Your call whether that wants softening or just sequencing.

*Separately, worth a sentence somewhere:* a tracked `.mcp.json` means every agent session opening this repo is offered a third-party MCP server, with queries going outward to the vendor. ADR-0051 argues the wiki; it does not separately argue that.

### Verification

- Suite on `06de6099`: **510 files, 9573 passed**, 0 failed. Typecheck clean. Lint clean, 1081 files.
- Earlier, on your head with `origin/main` merged in: 510 files, 9607 passed, 0 failed.
- **Linux only.** Silent about macOS and native Windows.
- Live endpoint checks re-run, not taken from the body: `deepwiki.com/badge.svg` → 200.

Condition 1 as your body declares it: this pass read both commits and ran every mutation itself, so it is that independent read. Land order is 478 → 479 → 480; nothing here is merged.

---
*Reviewed by Claude Opus 5 (1M context) via `rhizo-review`. One reviewer, one platform — not a substitute for a second pair of human eyes on anything non-trivial.*

KelliherL:
## Author confirmation — both fixes reproduced, and all three open items answered in `9850a033`

### Confirmed by re-running, not by reading

**`a8da58d3`, the credential law.** Reproduced your mutation exactly: a bearer token in a *second* server in the same `.mcp.json`.

| | before | after |
|---|---|---|
| credential on a second server | **6/6 green** | **2 of 6 red** |
| restored | — | 6 passed, `.mcp.json` identical to `HEAD` |

You are right about why it matters more than a hypothetical: this PR is what *creates* that file, so the next server anyone adds is the live case. And the fix is right in the way that outlasts this file — `credentialSmells(text)` means the bite test and the live test call the same code, which is the discipline `docs-index-law` states and which I applied to the link parsing in this very file and then not to the credential check.

**`a8da58d3`, the dead `startsWith('badge.')` filter.** Confirmed: the pattern needs two path segments and `badge.svg` is one, so the filter was unreachable and the assertion under it was passing for a reason other than the one its comment claimed. That is "a test that cannot fail for the reason it claims" at its smallest, and in a file I wrote to catch exactly that class.

**`06de6099`, the renumber.** Verified `docs/adr/0051-the-generated-wiki.md` here and `0050-two-identity-planes.md` on `origin/main`. Agreed that repeating the rename on this branch rather than inheriting it is correct while the stack is unmerged — a branch citing 0051 with 0050 still in its tree would be red on `doc-citation-law` until the stack restacked.

### The three you left me — all in `9850a033`

**1. #472's DoD clause is unmeetable as written, and you are right.** *"the repo slug the MCP configuration targets is the same repo the manifest names"* — `.mcp.json` names no slug. `https://mcp.deepwiki.com/mcp` is repo-agnostic; the repository is an argument passed per query. I wrote that clause assuming a per-repo endpoint and never checked. **Stating it here rather than writing a law that pretends to check it**, exactly as you suggest. The clause is not met, cannot be met in that form, and the thing it was reaching for — that the *badge* names this repo — is met and is held.

**2. The present tense.** Fixed, and it was the sharper of the two. Verified while writing it: the MCP server answers `read_wiki_structure` for this repo with *"Repository not found. Visit … to index it."* So between merge and indexing the README asserted something untrue — the precise defect this whole milestone exists to remove, introduced by the milestone itself. Reworded to be true in **both** states rather than vague: the page is where the wiki appears once indexed, indexing is an explicit operator act that may not have happened, and until it does the page offers to index rather than answer. It also names which surfaces can tell you the current state, since nothing in the tree can.

**3. Your closing note is the most important thing in this review**, and nothing else had argued it. A *tracked* `.mcp.json` means every agent session opening this repo is offered a third-party MCP server and each query goes outward to the vendor. ADR-0051 argues that a document gets published *about* us; this is a channel offered *to everyone who opens the repo*, and they are not the same claim.

Disclosed in the README now, with the two bounds that keep it small — nothing is sent unless a query is made, no credential is involved. **I am explicitly not claiming that discharges it.** A capability offered repo-wide is the shape ADR-0001 says gets named and fenced, and if the operator wants it argued rather than disclosed it belongs in its own record. Flagging for @KelliherL rather than deciding it inside a docs PR.

### Verification

`deepwiki-law` 6/6 after the README change. `route-class-law`'s support matrix green; 100 of 102 across both files, the two reds being the pre-existing `line-endings` known-failures confirmed in an `origin/main` baseline by failing test name. Native Windows.

Thank you for re-running the three body mutations rather than trusting them — that they reproduced exactly is the only reason the two that *didn't* stand out.

Nothing landed. Land order 478 → 479 → 480 as you say.

Issue #481, opened 2026-09-13
citation law: two [not-ours] rows name live code under a spelling that is not the symbol


gabriel-canaan:
## Fence (may touch ONLY)

- `.symbol-citation-baseline`
- `docs/design-notes/geometry-relax-reach.md`
- `packages/server/src/server/poll-loop.test.ts`
- `packages/server/src/doc-citation-law.test.ts`

The fourth file is named for the same reason it was named on #460: correcting a
not-ours row moves the pinned literals that live outside the baseline. If the
repair removes a row, the total-count pin, the distinct-symbol pin and the
measured stamp's not-ours field all move together. Re-derive them; do not retype
them. (Spelled without backticks on purpose — every ticked token in this section
is read as a claimed path.)

## Why

#460 corrected one misfiled `[not-ours]` row — `MigrationResult`, a typo for the
live `MigrationOutcome`. The independent review of that change asked whether any
sibling row was misfiled the same way, and found two. Neither is an author typo,
which is why #460's own sweep did not surface them; both are prose naming live
code under a spelling that is not the symbol, filed under a header that says the
opposite.

**`MAX_PX`** — `docs/design-notes/geometry-relax-reach.md` lines 3 and 7 write the
pair as `` `RELAX_REACH_MIN_PX`/`MAX_PX` ``, so the second half is a fragment of a
live constant, not a name of its own. The live symbol is `RELAX_REACH_MAX_PX`,
declared at `packages/web/src/scene/geometry/scale.ts:259`. Its baseline row sits
under the header *"named in prose about a design that was never built under this
spelling"*, which is false for it: the design was built, under a longer name.

**`manifestReadOverride`** — `packages/server/src/server/poll-loop.test.ts:19` names
the seam in prose; the live symbols are `getManifestReadOverride` and
`setManifestReadOverride`, declared at lines 27-31 of that same file. Softer than
the first: it reads as a deliberate shorthand for a get/set pair rather than an
error. It may be correct to leave the row and fix only the group, and that is a
judgement for whoever takes this.

Both rows' *verdicts* are right — neither spelling was ever declared, so
`symbolVerdict` returns `not-ours` correctly. What is wrong is the **group**, and
the group is what tells the next reader whether a row is a finding or furniture.

## The thing worth knowing before starting

**No automated rule can find this class, and two have already been tried and
broken.** A reviewer swept at Levenshtein <= 1 and it missed the control — the
[...]

Issue #461, opened 2026-09-13
doc law: the path law can say "cited from history, on purpose"


ciaran-slow:
#440 gave the **symbol** law a way to record a citation that names a removed thing *on
purpose* — a third `[historical]` section in `.symbol-citation-baseline`, mechanically
checked like the debt list and meaning the opposite. The **path** law, in the same file,
still has no such category.

## Why this is the path law's own stated gap, not a new idea

`packages/server/src/doc-citation-law.test.ts`'s input table gives this as the reason it
declines to widen to `.mjs` as a citing file:

> Before widening, the law needs a way to say "cited from history, on purpose" so a comment
> like that one can opt out — **that mechanism does not exist yet**, so the scope stays
> exactly ruling 1's, not narrower and not wider.

The mechanism now exists, for symbols. This issue is the path half.

## The case that is waiting on it

`packages/web/src/scene/parity/capture.mjs` cites `packages/web/src/scene/paint.ts` twice —
once in a comment (`:10`) and once as a code constant (`:48`):

```js
const PAINT = 'packages/web/src/scene/paint.ts'
```

That file **does not exist**: `8686f24d` (`feat(scene): the scene is drawn by a WebGL2
painter`, on `main`) replaced the 2D painter. The parity harness diffs against the
pre-deletion file deliberately and resolves it out of git history, so the citation is
correct *because* the target is gone — the exact shape `[historical]` was ruled for.

Verified before filing: the sha resolves and is an ancestor of `origin/main`, both citation
sites are live, and `packages/web/src/scene/paint.ts` is absent from the tree.

## Why `ALLOWLISTED_BROKEN_CITATIONS` is the wrong home

Same argument #440 made for `[renamed-away]`. That list's own header says:

> Wave 2 removes entries as each doc is corrected; **when this list is empty, the law is
> unconditional.**

Filing a deliberately-historical citation there records it as debt against a document that
has nothing to correct, and tells the next reader to go and "fix" a working, intentional
reference.

## One thing NOT to copy from #440, and the reason

The symbol law's lists had to move **out** of the test file into
`.symbol-citation-baseline`, because its arm 1 extracts identifiers from the non-comment
text of sources, **string literals included** — so a TS const holding the names would have
put every one into the token set and certified itself.

**The path law does not inherit that constraint.** It extracts citations from *comments*
only (`extractComments` for the `.ts` sweep), and `OWN_FILE` is excluded as a citing file
[...]

PR #482, opened 2026-09-13
charter-laws: the manifest is re-included by name, and the migration comment names the type the test reads

### LINKED ISSUE #444: gitignore: the PRD milestone manifest is re-included by name, not force-added
## Fence (may touch ONLY)
- `.gitignore`

## Why

`.gitignore` ignores everything under the swarm directory and then re-includes
one file **by name** with a negation. Its own comment says the pattern was
written that way precisely so a file can be re-included like that.

prd-56 wave 1 (#427) added a second file that must be tracked —
`.swarm/prd-milestones.txt`, the committed manifest the location law reads so the
suite can run offline. It was brought into the index with `git add -f` rather
than by adding the negation, because the ignore file sat outside that issue's
declared boundary and a widening to reach it was not worth taking there.

**It works today and this is not a bug report.** Verified: `git archive HEAD`
carries the file, so a fresh clone gets it, and because the file is tracked,
modifications to it still show in `git status` despite the ignore rule.

The cost is narrow and real: once the file leaves the index for any reason, an
ordinary `git add -A` on that directory will not bring it back — it stages the
deletion instead. Verified in a scratch repository: after `git rm --cached`,
`git add -A` leaves the path staged as deleted. A clone without the file fails
the law on ENOENT, which is at least loud rather than silent.

So this is about matching the idiom the ignore file already states, so that the
second tracked file in that directory is tracked the same way as the first, and
nobody has to know that one of them is special.

## Definition of done

- The manifest is re-included by name, the same way the existing tracked file in
  that directory is, and no longer depends on having been force-added.
- `git check-ignore -v` reports the manifest as NOT ignored, and a scratch clone
  still receives it.
- The negation is proven rather than assumed: removing it and re-running the
  check shows the path becoming ignored again.
- `git status` is clean afterwards — re-including an already-tracked file must
  not show as a change to the file itself.

## Notes

Raised by the independent review of #427. Filed here rather than folded into that
issue because the ignore file is outside its declared boundary; filed rather than
dropped because leaving one of two tracked files in a directory tracked by a
different mechanism is the kind of asymmetry that costs someone an afternoon
later.

### LINKED ISSUE #460: contract: the migration comment names the type the test actually reads
`packages/contract/src/instrument.contract.test.ts:115` reads:

```ts
/** The migration words the server's `MigrationResult` can send. */
function serverMigrationKinds(source: string): string[] {
  const declaration = /export type MigrationOutcome =([\s\S]*?)\n\n/.exec(source)?.[1] ?? ''
```

The type is **`MigrationOutcome`**, declared at `packages/server/src/concierge/migrate.ts:152`
and imported at `packages/server/src/api/concierge.ts:18`. `MigrationResult` is a name this
repo has never had: it appears **exactly once in the whole tree**, in that comment, one line
above the code that reads the right name.

This is precisely the rot #369 was filed about — a document naming a symbol that does not
exist — and it is currently **misfiled as external**. `.symbol-citation-baseline`'s
`[not-ours]` section holds:

```
MigrationResult               packages/contract/src/instrument.contract.test.ts
```

under a header describing *"spellings this repo's own prose names WITHOUT ever having had
them: hypothetical names a law warns about, or a rejected alternative spelling"*. It is
neither. It is a typo for a live internal type.

**The law is not at fault and must not be changed for this.** Arm 3 is mechanically
correct: `git log -S MigrationResult -- packages/` returns one commit whose only hit is
that same comment, so nothing was ever declared under that name. A misspelled internal name
is indistinguishable from an external one *by shape*, which is the under-inclusion arm 3
declares. Fixing the instrument here would mean guessing, and the guess would be wrong more
often than right.

## Definition of done

- The comment names `MigrationOutcome`.
- The `MigrationResult` row is gone from `[not-ours]`.
- The three counts that move with it are re-derived, not retyped — see the fence.
- `npx vitest run packages/server/src/doc-citation-law.test.ts` is green, and the
  `[not-ours]` honesty test still asserts 3 assertions per row.

## Fence (may touch ONLY)

- `packages/contract/src/instrument.contract.test.ts`
- `.symbol-citation-baseline`

Written as a ticked bullet list on purpose. It was a bare code block until
2026-09-14, and scripts/fence-lint.sh extracts only backticked tokens — so it
read this issue's fence as the three count literals below and reported **zero
overlaps** against #461, which claims the third file too. The lint passed while
checking nothing.

## Why the third file is included

**It is the one a narrow boundary would miss**, and it is named deliberately.
Dropping one `[not-ours]` row moves three pinned literals that live outside the
baseline:

- `expect(notOurs.length).toBe(72)` → 71
- `expect(new Set(notOurs.map(({ symbol }) => symbol)).size, …).toBe(72)` → 71
- the `# measured:` stamp's `not-ours=72` → 71, which is now **machine-checked**
  against the file it describes, so a stale stamp is a red suite rather than a
  silent lie

`expect.assertions(notOurs.length * 3)` is derived and needs no edit.

Verified on `main` at `c4e20d09` before dispatch: the two assertions sit at
`packages/server/src/doc-citation-law.test.ts` lines 2930 and 2933, the row at
`.symbol-citation-baseline` line 175, and the stamp at line 56 of that file.
Those line numbers are a convenience for the reader, not the citation — match on
the text.

**One thing the list above does not mention.** The `# measured:` stamp carries
`tree=aaf9a50c` alongside `not-ours=72`. Whoever takes this decides whether
re-stamping the tree sha belongs in the same edit and says which they chose on
the PR; changing the count while leaving a tree sha that no longer describes the
file is the stale-provenance shape the stamp exists to catch.

## Blocked by #440

The baseline file, the `[not-ours]` row and the stamp assertion all arrive with #440. There
is nothing to edit until it lands on `main`.

## Provenance

Found by an independent review seat during round 3 of #440, offered explicitly as *not* a
defect in that PR and deliberately left out of it: it is a third file, outside that PR's
two, and it moves a pinned count — so it wanted its own issue rather than a third fence
widening on a PR already carrying one.

Every claim above was re-verified against the tree before filing (the seat cited
`api/concierge.ts:18` for the declaration; that line is the **import** — the declaration is
`concierge/migrate.ts:152`).

## Milestone note

Filed under `charter-laws` to sit beside #369, which produced it. That follows the parent
issue and current practice in that milestone rather than the three-condition carve-out in
`AGENTS.md` — this corrects a document, it does not enforce a charter rule. Move it if a
groomer reads the carve-out more strictly.

gabriel-canaan:
Closes #444
Closes #460

Two independent issues in the `charter-laws` milestone, one commit each. Their
fences are disjoint and were lint-proven so before dispatch.

## #444 — the PRD milestone manifest is re-included by name

`.gitignore` excludes `.swarm/*` and re-includes one file by negation.
`.swarm/prd-milestones.txt` had been brought into the index with `git add -f`
instead, so the directory's two tracked files were tracked by two different
mechanisms. This adds the negation beside the existing one and extends the
comment to describe both.

Nothing was broken before this; the cost was narrow and real. Once the file
leaves the index for any reason, `git add -A` stages the deletion rather than
restoring it.

## #460 — the migration comment names the type the test actually reads

`packages/contract/src/instrument.contract.test.ts:115` named `MigrationResult`,
a type this repo has never declared, one line above code reading the correct
`MigrationOutcome`. The typo was misfiled in `.symbol-citation-baseline`'s
`[not-ours]` section, which is for spellings this repo's prose names without ever
having had. The row is dropped and the three pinned literals that move with it
are re-derived.

The law itself is untouched. Arm 3's under-inclusion is declared and correct: a
misspelled internal name is indistinguishable from an external one by shape.

## What is the sibling case?

**Asked properly, and it found two — in a file this PR edits.**

The honesty test asserts only that a `[not-ours]` symbol is absent from source and
from declaration history, which a misspelling of a live name satisfies by
construction. Its green says nothing about misfiling, and an earlier draft of
`ea758393`'s body wrongly cited it as though it did. That paragraph has been
replaced; the amended body says what was actually done.

Two shape-based sweeps were tried and **both failed their own controls** — edit
distance ≤ 1 does not reach `MigrationResult` → `MigrationOutcome` (distance 7),
and a shared-leading-word rule does not reach a typo inside the leading word.
Answered by reading instead: all 71 rows against their citing lines and group
headers. No row is an author typo for a live name. Two name live code under a
spelling that is not the symbol and sit under a header that is false for them —
`MAX_PX` (live: `RELAX_REACH_MAX_PX`) and `manifestReadOverride` (live: the
get/set pair). Both citing files are outside this fence, so they are filed as
**#481** rather than widened into this PR.

Both pinned literals were mutated **independently**, because vitest aborts an
`it` at its first failed `expect` and the first assertion would otherwise mask
the second. Reverting only the total count reddens at `doc-citation-law.test.ts`
line 2930; reverting only the distinct-symbol count reddens at line 2933. Each is
load-bearing on its own.

For #444 the mutation is the behaviour, not the rule: with the negation,
`git rm --cached` followed by `git add -A` restores the file; without it,
`git add -A` stages `D .swarm/prd-milestones.txt`. Run in a scratch clone, with
the pre-fix `.gitignore` as the control.

The stamp's `tree=` was deliberately not re-stamped, and the amended commit body
now gives the real reason: the date, `tree=` and `candidates=` are one record of
one sweep, this change ran no sweep, and re-stamping `tree=` alone would make
`candidates=2187` false — the same sweep on this tree returns 2207.

## Fence widening

## Verification

Independent review before this PR was opened: two initial passes on isolated
worktrees (Opus 5 and Codex `gpt-5.6-terra`, both at xhigh), a cross-review of the
first by the second, and a reconcile pass against the code. The builder was
Sonnet 5, so no reviewing seat shared its model family. One cross-review was run
rather than two — the second pass reported no findings, so there was nothing to
critique for overclaim.

The reconcile found what all the earlier passes missed: the `MAX_PX` sibling, and
that the live candidates figure has moved (2207, not 2187), which is what settles
the `tree=` question.

Gates on this head: `npm run typecheck`, `npm run lint` and
`VITEST_MAX_WORKERS=6 npm test` all exit 0 — 508 files, 9557 passed. The amend
that followed review was message-only: the tree hash is unchanged
(`37164f98…` before and after), so the suite result above still describes this
head. The two directly affected test files were re-run after it regardless, 78
passed.

**One platform.** All of the above ran on WSL2 Linux. There is no macOS or Windows
witness for this branch, and the repo's workflows are disabled, so this PR will
carry no check runs.


KelliherL (approved):
## Review of #482 — approve; both behaviours run, nothing to fix

Read commit-by-commit (`5b31d16e` #444, `ea758393` #460) against each issue's Definition-of-done, in a clean clone at node 22.23.2 on Linux. Four files, ten lines added, eight removed.

### What verified — EXECUTED on your head `ea758393`

- `npm run typecheck` **0** · `npm run lint` **0** (1079 files) · `instrument.contract.test.ts` + `doc-citation-law.test.ts` **green**.
- **#460's two pinned literals, mutated independently** — because, as you say, vitest aborts an `it` at its first failed `expect`:
  - `notOurs.length` alone reverted to 72 → **red** (*"the baseline is pinned — a silent addition to any of the three lists is how a real regression gets waved through"*).
  - the distinct-symbol count alone reverted to 72 → **red**, same test, on its own. Both are load-bearing.
- **#444's behaviour, with the pre-fix `.gitignore` as control**, in a scratch clone:
  ```
  with the fix:  git rm --cached .swarm/prd-milestones.txt && git add -A  →  status: (empty — restored)
  control:       same, with origin/main's .gitignore                       →  status: D  .swarm/prd-milestones.txt
  ```
  That is the issue's cost, reproduced, and the fix removing it.
- `MigrationResult` on `origin/main` appears in exactly two places — the comment and the baseline row — and nowhere as a declaration. `MigrationOutcome` is what the regex one line below reads. The row's removal from `[not-ours]` is the right filing.
- **#481 exists** (filed 23:05Z, `charter-laws`), so the two misfiled rows the reconcile found have a home rather than a widening. **#461** also claims `doc-citation-law.test.ts`, and I checked it is not a live fence: unassigned, Backlog, no branch, no PR.
- The `# measured:` stamp keeps `tree=aaf9a50c candidates=2187` with only `not-ours=` moving. The reasoning in the commit body holds: the three list counts describe the file, the sweep record describes a sweep this change did not run. I did not re-run the sweep; the body's 2207 is your figure, taken as such.

### REASONED

- The `.gitignore` comment now describes both re-included files and still explains the `.swarm/*`-not-`.swarm/` trap. Nothing else under `.swarm/` is tracked (`git ls-files .swarm/` on this head: the two files), so the sibling case is closed as the commit says.

### What I did not run

The full suite. Four files: a `.gitignore` line, a baseline row, a comment, and two pinned counts in a law — the two test files that read them are the ones that could move, and both ran. macOS and Windows, as everywhere now. The `--prove` scripts and enumeration tables in your scratchpad are your kit's; the mutations above are my own re-derivation of the same claims.

**Approving.** Landing is the operator's act.

Issue #484, opened 2026-09-13
prd56 w3: the exemption reader tells prose from code in every Markdown context


gabriel-canaan:
## Fence (may touch ONLY)

- `packages/server/src/prd-location-law.test.ts`

Nothing else. This wave changes how the reader decides what a line *is*; it
changes no PRD's disposition, moves no file and edits no declared reason. If a
real PRD's text has to change to keep its exemption, that is a finding to report
before changing it, not a licence to widen.

## Blocked by

**prd56 wave 2 (#428)**, which builds the reader this wave repairs. Wave 3 cannot
start before wave 2 lands.

## Why

Ruling 2 requires a declared reason that a **reader** can find — *"the reason is
prose the operator writes; the law only requires that it exists and is findable."*
Wave 2's fix re-review found four contexts where the reader and a person looking
at the rendered document disagree.

All four are EXECUTED, reproduced independently by two review seats and the
orchestrator, on the wave-2 tree.

| # | the gap | direction |
|---|---|---|
| 1 | `FENCE_OPEN_RE` is anchored at column 0. CommonMark allows a fence opener indented up to three spaces, so a marker inside a 1–3-space-indented fence is read as a declaration. | **fail open** |
| 2 | `stripBlockquotePrefix` consumes `[ \t]*` after `>`, so the `/^( {4,}\|\t)/` indent guard can never fire inside a blockquote — which is where all four real declarations live. `>     <marker> …` exempts. | **fail open** |
| 3 | The continuation loop applies no fence or indent check, only the declaration line does. A placeholder plus a quoted code block therefore clears the three-word substance threshold: a `-` reason followed by a quoted fence returns a reason. | **fail open** |
| 4 | Comment blanking is not code-aware. A comment opener inside an inline code span blanks a plainly visible declaration; an *unterminated* opener above one exempts it. | **both** |

**None is reachable by any file in the tree today** — verified: no live-shelf PRD
carries comment syntax, an indented fence, or a blockquoted indented-code line.
The only file with comment syntax is `docs/prds/retained-prds-review-2026-08-22.md`,
which `PRD_FILE_RE` never name-matches. That is why wave 2 was not held for them,
and it is also why this is Soon rather than Now.

## Definition of done

- **One predicate decides whether a line is prose or code, and both the
  declaration scan and the continuation loop use it.** Four separate guards patched
  in four places does not close this wave — it is the failure the wave exists to
[...]

Issue #459, opened 2026-09-12
prd56: the mirror case is ruled one-way, and the reason is not the absence of an instance


gabriel-canaan:
## Fence (may touch ONLY)
- `docs/prds/prd-56-the-shelf-agrees-with-the-ledger.md`
- `packages/server/src/prd-location-law.test.ts`

## Why

prd-56's Open Questions ask whether ruling 1's law should also fail an ARCHIVED
PRD whose milestone is still open — the mirror case — and leave it **open, not
ruled**, on the stated ground that *"No instance exists today"*.

**An instance exists, and prd-56's own wave 1 created it.** `prd43`'s milestone
reopened (1 open against 34 closed, no closed_at) after wave 1 moved
`prd-43-the-claim-is-a-test.md` onto the archival shelf on the strength of its
then-closed milestone.

The operator ruled it in session on 2026-09-13: **the law stays one-way, and the
justification changes.** "No instance exists" was never a reason — it was an
absence of pressure, and it expired the moment one appeared. The reason is the
difference in what each direction misleads a reader INTO:

- a closed-milestone PRD on the LIVE shelf falsely reads as in flight, and people
  act on that — `docs/follow-up-292.md` reasons from exactly that to hold a ruling
  live, and #416's enumeration nearly edited a finished PRD's prose on the same
  basis;
- an archived PRD whose milestone has reopened reads as done while a straggler
  issue remains. Staler, but nobody is misled into treating a shipped ruling as
  open.

prd-43 is the case in point rather than the counterexample: 1 open issue against
34 closed is a straggler, not a revival, and a two-way law would drag that file
back and forth on the strength of one issue in thirty-five.

The full ruling with its reasoning is recorded as a comment on #427.

**Why this is not filed under prd56.** `scripts/dev/prd-reconcile.sh` reports an
issue in a PRD milestone with no wave as NO WAVE drift, and one claiming a wave
the document does not declare as UNDECLARED WAVE — *"the DOCUMENT is the plan, so
it rules"*. This issue is the amendment that would declare such a wave, so it
cannot be milestoned to the PRD it amends without first doing its own work.
Milestone chosen by the operator.

## Definition of done

- prd-56's Open Questions no longer carry the mirror-case question as open. It
  becomes a **numbered ruling**, appended rather than renumbered, carrying the
  reasoning above and naming prd-43 as the instance that forced it.
- The stale justification — "No instance exists today" — does not survive
[...]

PR #485, opened 2026-09-14
prd56 w2: an unarchived finished PRD states its reason, and the law reads it

### LINKED ISSUE #428: prd56 w2: an unarchived finished PRD states its reason, and the law reads it
## Fence (may touch ONLY)
- `docs/prds/prd-53-the-lab.md`
- `docs/prds/prd-48-the-shared-record.md`
- `docs/prds/done/prd-48-the-shared-record.md`
- `docs/roadmap.md`
- `docs/research/2026-08-29-shared-record-s2-ack-after-journal.md`
- `.swarm/prd-milestones.txt`

**Widened 2026-09-13, before any change was made, and the scope grew with it.**
This issue was groomed when prd-17 was ruling 2's only subject. Two milestones
closed on 2026-09-11 at 04:24 — after wave 1's manifest was generated at
02:48:35Z — so the prd48 and prd53 milestones now sit on the live shelf with closed
milestones and are convicted by ruling 1's law the moment anyone regenerates the
manifest. Measured on wave 1's branch: regenerate, and the law exits 1 with
*"expected [ 'prd-48-the-shared-record.md', 'prd-53-the-lab.md' ] to deeply
equal []"*.

The operator ruled on 2026-09-13 that they are handled here rather than
separately, and that **the two are handled DIFFERENTLY** — see Subjects below.
The last four paths exist because archiving prd-48 is a move, and a move carries
its citers in the same commit (ruling 3). Both halves of that move are claimed
because a rename audits as a change to each path.

- `docs/prds/prd-14-experiment-console.md` (the declared-reason block only — every ruling and wave paragraph of both PRDs stays untouched, as with prd-17)
- `docs/prds/prd-20-the-concierge.md` (the declared-reason block only — every ruling and wave paragraph of both PRDs stays untouched, as with prd-17)

## Blocked by

**Wave 1** — the law must exist before it can be taught to read an exemption.

**And an ordering hazard that is a human act, not a code dependency: this
issue should land before anyone closes the prd17 milestone.** Milestones do
not auto-close — prd48, prd20 and prd53 all sit at zero open issues and are
still open — so prd-17 will not be convicted the moment its last issue closes.
It will be convicted the moment someone closes the milestone by hand, and at
that instant prd-17 becomes a finished PRD with an unarchived file and no
declared reason. That is exactly the state ruling 2 exists for, and this issue
is what supplies it.

Sibling lanes hold other surfaces. Every ruling and every wave paragraph of
prd-17 is EXPLICITLY OUT: this issue adds an exemption declaration to that
document and changes nothing else in it. The five PRDs wave 1 moved are
EXPLICITLY OUT — they had no reason to declare and are correctly archived.

### The citer trap, already paid for once in wave 1

prd-48 has exactly two citers, and they are NOT equally enforced. Re-derive rather
than trusting this list, but know the shape before you start:

- `docs/roadmap.md` is swept by the citation law, so forgetting it goes RED.
- `docs/research/2026-08-29-shared-record-s2-ack-after-journal.md` sits in a
  directory the citation law EXCLUDES as a citing source. Forgetting it is
  SILENT — nothing goes red and the citation simply rots.

That second shape is exactly what wave 1 hit with a dated review document, and it
was found by re-deriving rather than by any gate. No citation of prd-48 carries a
`:NNN` line suffix today; check again rather than assuming.

### The manifest is regenerated HERE, deliberately

Wave 1 landed with a manifest that was already stale in three rows, and that was a
decision, not an oversight: regenerating it is what converts prd-48 and prd-53 from
inert to convicted, and wave 1 had no form to exempt them with. **This issue is
where that debt is paid.** Regenerate it, commit it, and expect the law to be green
only once all three subjects above are handled.

That also resolves the hazard wave 1 recorded: it is no longer "do not regenerate",
it is "regenerate here".

## Subjects — three, and not all the same case

Ruling 2's form is built once. What it is applied to differs, and conflating them
is the failure prd-56 warns about: *a law that convicts a correct state is one
people learn to override*.

| PRD | disposition | why |
|---|---|---|
| prd-17 | **declared exemption** | `docs/follow-up-292.md` holds an open, uncorrected item against its ruling 3 and reasons from the file's location to do so. Archiving it would demote a ruling while an item against it is open. Its milestone is still open, so it is not convicted yet — this is pre-emptive, and the ordering hazard below is why. |
| prd-53 | **declared exemption** | `prd-53` is *the lab*; `prd-55` is *the lab, stage two*, live with waves 6 and 7 in flight. Archiving it would demote the rulings stage two is building ON. |
| prd-48 | **ARCHIVED, not exempted** | Operator-ruled 2026-09-13. Nobody has argued it should stay live; it arrived alongside prd-53 and was held only because the same mechanism answers both. An exemption with no reason behind it teaches people the form is a way to avoid archiving, which is exactly the "law that convicts a correct state" failure. So it takes ruling 1's ordinary path: move to the archival shelf, citers updated in the same commit. |

**Do not give prd-48 an exemption to make the law green.** If the move is
awkward, that is a finding to report, not a reason to reach for the form.

## Why

Ruling 2: a finished PRD may stay unarchived when its own text says why, in a
form the law reads. Without it the law would force prd-17's move against a
live argument — docs/follow-up-292.md reasons from that file's location to
call its ruling 3 live rather than archival, and holds an open, uncorrected
item on that basis. Archiving prd-17 would demote a ruling while an item
against it is open.

A law that convicts a correct state is one people learn to override, so the
exemption is part of the design rather than a concession to it.

prd-17 is the first and currently only subject. That is not incidental: it is
the case that proved the exemption necessary, and #416 recorded the reason in
that document's own Outcome line before this issue existed.

## Definition of done

- The law reads a declared reason from a PRD's own text and does not fire on a
  PRD that carries one. The form is chosen here and stated in the commit body
  with the alternative it rejected — ruling 2 fixes that the reason lives in
  the file and that the law reads it, and fixes nothing else.
- The form is findable by a reader, not only by the law. A marker a human
  cannot spot while reading the document fails ruling 2's own wording, which
  says the reason is prose the operator writes.
- prd-17 carries its declared reason, and the law is green with prd17's
  milestone closed. Prove that: close the milestone in the manifest fixture,
  not on the real tracker, and show the law passes for prd-17 while still
  failing for a PRD with no reason.
- An empty or placeholder reason is not an exemption. A file that declares the
  marker with nothing behind it must still be convicted, or the exemption
  becomes a way to switch the law off.
- npm run typecheck, npm run lint and VITEST_MAX_WORKERS=6 npm test pass.

## Sibling case

The obvious miss is an exemption keyed on the PRD's identity — a list of
allowed filenames, or a check for prd-17 by name. That passes today and is a
fitted instrument: the next PRD with a real reason to stay is not on the list,
so the law convicts it and the next reader adds a second name. The exemption
must be keyed on the DECLARATION, never on which document made it.

Second: an exemption that suppresses the whole law rather than one file. Prove
the law still convicts a different unarchived finished PRD in the same run
that passes prd-17.

Assert only that the law passes for prd-17. That passes with the exemption
implemented as "never fire", because nothing else in the test set is a
finished-but-unarchived PRD once wave 1 has moved the five. The assertion
needs a second subject in the same run — a fixture PRD with a closed
milestone, an unarchived path and no declared reason — proving the law fires
on it while passing prd-17.

Second: delete the reason from prd-17 and prove the law goes red. A test that
only ever sees the reason present cannot show the check reads it.

gabriel-canaan:
Closes #428

prd-56 ruling 2: a finished PRD may stay on the live shelf when its own text says
why, in a form the law reads. Wave 1 landed the law and the offline manifest; this
wave teaches the law to read an exemption, regenerates the manifest — which is what
converts five PRDs from inert to convicted — and disposes of all five.

| PRD | disposition |
|---|---|
| prd-17, prd-53, prd-14, prd-20 | declared exemption, each citing a live PRD that reasons from its rulings |
| prd-48 | **archived** to `docs/prds/done/`, both citers moved in the same commit (ruling 3) |

The form is `**Shelf exemption:**` followed by prose. `declaredExemptionReason`
takes only `text` and never a filename, so it is keyed on the DECLARATION rather
than on which document made it. The rejected alternative — a list of allowed
filenames — is named in the commit body: it passes today and is a fitted
instrument, since the next PRD with a real reason to stay would not be on the list.

**prd-14 and prd-20 were not in the original scope.** They surfaced only when the
manifest was regenerated: their milestones closed in the same 2026-09-13 operator
batch as prd-17. The fence was widened for them on the issue, before either file
was touched, and the widening names the live-PRD argument for each.

## What is the sibling case?

Two, both answered in the commit bodies.

**An exemption keyed on identity.** Absorbing prd-14 and prd-20 — two unplanned
subjects — required editing **zero** lines of the law. That is the evidence the
form generalises, and it exists only because they arrived late.

**An exemption that switches the whole law off rather than one file.** A test runs
a real subject and a synthetic no-reason subject through one `violationsIn` call
and asserts only the synthetic one is named.

A third, found by review and fixed as `727aa51f`: prd-14's declared reason quoted
prd-55 with words prd-55 does not contain — `"applies unchanged"` for `"apply
unchanged"`, and the wrong subject. The substance held, but the string is pinned as
`EXPECTED_REASON`, so an inaccurate quotation had become load-bearing on a law.
This is the same defect the wave had already fixed once for prd-20's citation;
prd-17's and prd-53's were re-derived and resolve.

Asserting only that the law passes for the exempted PRDs survives the reader being
implemented as "never fire", since wave 1 moved every other finished-but-unarchived
PRD. So the assertions run a synthetic no-reason subject in the same call, and
stripping a real reason reddens the law (certified, exit 1).

The `indexOf` → `lastIndexOf` mutant survived the first attempt: all four exempted
PRDs carry the marker twice, and nothing asserted reason CONTENT. It is now pinned
by exact content.

## Fence widening

Recorded on #428 before the change, twice: `prd-53`/`prd-48` when the operator
ruled how they were handled, and `prd-14`/`prd-20` when regeneration convicted
them. Both carry a stated scope bound.

## Verification

Two review rounds, both recorded on #428 in full.

- **Round 1** — 2 seats (Opus `xhigh`, Codex `gpt-5.6-terra` `xhigh`) plus
  orchestrator probes. Found an HTML-comment kill switch, accepted placeholders,
  and an unpinned mention-vs-declaration choice. All three handed back and closed.
- **Round 2, the fix re-review** — 2 seats on the merged head. Found the pinned
  quotation above, plus four latent gaps in the reader's Markdown-context guards.
  Those four are **filed as #484** (prd56 wave 3), deliberately not fixed here:
  they are one class rather than four defects, none is reachable by any file in the
  tree, and cell-by-cell repair is what the review warns against.

`727aa51f` is a **reviewer repair, and no independent pass has read it.** It was
written by the same session that found the defect, gated with the repair gate
(fence, lockfile, claim check, certified mutation, sibling case, mutation answer,
typecheck, lint, suite) — but a gate validates an artefact, not whether a second
pair of eyes existed. The runbook says the reviewer fixes and a second pair of eyes
still lands anything non-trivial; that second pair is the reviewer of this PR.

Gates, orchestrator, full worktree: typecheck 0, lint 0,
`VITEST_MAX_WORKERS=6 npm test` 0 — 508 files, 9616 passed. **One platform, linux.**
GitHub Actions was retired on 2026-09-11, so there is no macOS or Windows leg behind
any of this.

---

## Declared review gap

> 727aa51f is a reviewer repair written by the session that found the defect; it passed the repair gate but no independent seat has read it. 1c08593a is recorded as reviewed by the fix re-review on e82648c4.

```
  UNVERIFIED  727aa51f  docs(prd-14): the declared reason quotes prd-55 as written  (review of #428)
```


KelliherL (approved):
## Review of #485 — approve; the second pair of eyes on `727aa51f`, and the reader mutated six ways

Read commit-by-commit (`1c08593a` the form and the five dispositions, `727aa51f` the quotation repair; the merge of `main` is noise) against #428's Definition-of-done and its two recorded widenings, in a clean clone at node 22.23.2 on Linux. Your declared gap is that no independent pass read `727aa51f`; this is that pass, and it read the rest too.

### What verified — EXECUTED on your head `727aa51f`

- `npm run typecheck` **0** · `npm run lint` **0** (1079 files) · `prd-location-law`, `doc-citation-law`, `runbook-delivery-law` — **green**; `scripts/dev/prd-milestones.test.sh` — **22 passed**.
- **The manifest is honest.** `scripts/dev/prd-milestones.sh` re-run against the live tracker (`gh`) reproduces the committed manifest apart from its own `generated-at` — a one-line diff, `2026-09-13T21:24:27Z` → the time I ran it. The `prd43 closed → open` flip in your diff is real: the milestone has 1 open issue and no `closed_at`, and the law is one-way by #459's ruling (an archived PRD with a reopened milestone is outside `liveShelfEntries`' sweep), which your docblock now records.
- **Every quotation resolves in the PRD it cites, not in the citing file** — the defect `727aa51f` fixed, checked for all of them:
  - prd-14 → prd-55: `prd-14 ruling 3` (prd-55:104) and `apply unchanged (prd-14 ruling 4)` (prd-55:200), verbatim.
  - prd-20 → prd-34: *"prd-20 remains the authority on what the concierge may do (launch or relaunch a conductor, clone by URL; no OAuth, no accounts, no stored credentials)"* (prd-34:127), verbatim, and it is prd-34's own ruling 4 as the reworded reason says.
  - prd-17 → `docs/follow-up-292.md`: the open, uncorrected item against ruling 3 law 1 is there (`:115-118`).
  - prd-53 quotes nothing; its claim (prd-55 waves 6 and 7 build on it) is true on `main`.
- **`prd-48` is a rename, not a copy** (`docs/prds/{ => done}/prd-48-the-shared-record.md`, 0 lines changed), and both citers move in the same commit — `roadmap.md` (swept) and the research note (not swept, still corrected), exactly the trap #428 names.
- **Fence:** the eight fenced paths plus the two widenings recorded on #428 at 21:31Z, before either file was touched; both widenings are "the declared-reason block only" and the diffs to prd-14 and prd-20 are exactly that block.
- **Your mutations, re-derived:** prd-17's declared reason stripped → **3 red** (the shelf law, the ruling-2 content test, the per-file synthetic test). The `indexOf → lastIndexOf` mutant is **no longer applicable to the shipped reader** — there is no `indexOf` left in it; the scan is line-based — so I mutated the property instead (below).
- **Mine, on the shipped reader:**
  - the scan made to return the **last** declaration instead of the first → *"two genuine declaration lines in one document — the FIRST governs"* **red**. The first-vs-last property is pinned by construction, as your round 2 says.
  - the reader made to return a real-looking reason for **any** text (the kill switch) → **red** across the ruling-2 content test, the per-file synthetic test, and both concealment tests.
  - `MIN_REASON_WORDS` 3 → 1 → the placeholder test and the boundary test **red**.
  - **the HTML-comment blanking deleted → green.** See below.
  - **the indented-code exclusion deleted → green.** See below.

### One finding — a missing test case, pushed as `8ef92e27`

**The comment blanking is load-bearing, and nothing tested it.** Both cases in *"a marker sitting inside an HTML comment is not a declaration"* put the `<!--` on the marker's own line — and those are rejected by the line-start rule alone, blanking or no blanking. The blanking matters only when a comment **wraps** a line that begins with the marker:

```
<!--
**Shelf exemption:** a hidden reason with several words
-->
```

Without the blanking that middle line starts with the marker, clears the threshold, and exempts the file. No test had that shape, so deleting `withoutComments` left the suite green (EXECUTED). Two assertions with the wrapped shape, bare and blockquoted, are added to the existing test; the same deletion now reddens it (EXECUTED against the committed head). Rung 1 — a test case, no reader change — so I am not blocking on it.

**The indented-code guard is dead code, and its test passes for a different reason than its name says.** `if (/^( {4,}|\t)/.test(stripped)) continue` can never fire: without a blockquote, a line with leading spaces already fails `startsWith(marker)`; with one, `stripBlockquotePrefix` has consumed the indentation before the guard sees it (its own comment says so). The test *"a marker sitting inside an indented code block … is not a declaration"* is green with the guard deleted because of `startsWith`, not because of the guard. Not a behavioural defect — indented code without a blockquote *is* excluded — but it is exactly the "indented-code line inside a blockquote" edge #484 files, and the guard that looks like it handles it does not. I have not touched it: wave 3's single predicate is where that line goes, and it is your wave. Named here so the next reader does not trust the comment.

### The second pair of eyes on `727aa51f`

Read it as a reviewer of a reviewer's repair. It changes prd-14's declared reason to quote prd-55 as written (`"apply unchanged (prd-14 ruling 4)"` for the earlier `"applies unchanged"`, and the Non-goals' `"prd-14 ruling 3"` with the right subject) and updates `EXPECTED_REASON` in the law to match. Both strings resolve verbatim in prd-55 (above). Nothing else moved. **This is the load-bearing point:** because the reason's content is pinned by the law, a misquotation would have shipped as a passing test — the repair makes the pin true rather than green, which is the right direction.

### REASONED

- The reader's grammar is the right shape for what verify found: a declaration is a *line*, after blockquote stripping, outside fences, indented code and HTML comments; a mention never starts a line; the first such line governs; the substance threshold is a count of letter-runs, not a blacklist. Each of those has a direct unit test, and the four contexts #484 files are the ones this grammar does not yet decide (a fence indented 1–3 spaces; indented code inside a quote, because the strip eats all whitespace after `>`; quoted code on a continuation line; a comment opener inside a code span). Three fail open, one closed, as #486 says; none reachable by a file in the tree today.
- `declaredExemptionReason.length === 1` pinning "never a filename" is a small, exact way to hold the identity-keying decision. Good.

### The union

`origin/main` moved under you (#476, #482). Merged **into** the branch locally — clean, no overlapping files — and the whole leg run on the union with `ci-local.sh --no-install`:

```
  Test Files  512 passed (512)
       Tests  9729 passed | 3 expected fail | 3 skipped (9735)
  ci-local: green — exit 0 in 156 s, Linux / node v22.23.2
```

(Run before the test commit above; that commit adds two assertions to one file, which was re-run green with typecheck and lint after it.)

One platform (Linux, node 22.23.2). I did **not** push the merge: the branch merges cleanly as it stands and the leg is what needed the union, not the remote.

- `stripBlockquotePrefix` deliberately consumes all whitespace after `>` and its comment says why ("this file has no indented code living inside a blockquote"). That sentence is now the exact edge #484 names; when wave 3 lands it should move or go, or it will read as a rule after it has become the exception.
- The `prd43` row: the manifest now says `open`, the file sits in `done/`, and the law is one-way by ruling. Fine — but `docs/prds/done/prd-43-*.md`'s own status line presumably still says shipped/closed. Not this wave's; #459's.

### What I did not run

macOS and Windows. The two review rounds' seat transcripts on #428 (read the findings, not the transcripts). `gh api` for the four milestone close timestamps beyond what the regenerated manifest already proves.

**Approving.** `727aa51f` is read; the gap is closed. Landing is the operator's act.

PR #486, opened 2026-09-14
prd56: the Sequencing declares a wave 3 for the reader's Markdown contexts

gabriel-canaan:
**This PR closes no issue, deliberately.** It is a PRD amendment that *enables* an
issue: #484 could not be filed as `prd56 w3` until prd-56's Sequencing declared a
wave 3, because the runbook holds that a wave the PRD does not declare is a PRD
amendment rather than a tracker edit. Enables #484.

## What changes

One paragraph block appended to prd-56's Sequencing, above the existing
"Unfiled work implied" paragraph. **Waves 1 and 2 stand exactly as written** —
their text, fences and reasoning about each other are untouched. The wave records
when it was added and why it could not have been groomed earlier, which is the
discipline `docs/adr/README.md` applies to a changed mind: a new record, not an
edited one.

## Why a wave, and not an unfiled residual

The Sequencing already parks two genuinely un-booked questions — whether
`docs/prds/parked/` wants the same agreement, and whether `docs/design/` and
`docs/review/` want any location rule. Both are outside ruling 1.

These four gaps are not. Ruling 2 already governs them: it requires a reason a
**reader** can find, and these are the contexts where the reader and the law
disagree. The only reason they are not wave 2's bug list is that wave 2's own fix
re-review is what discovered them.

## The acceptance shape is in the PRD, not left to the issue

**One predicate deciding prose from code, applied to the declaration scan and the
continuation loop alike.** Four spellings patched in four places does not close the
wave — and gap 3 already *is* that mistake one level down: the continuation loop is
a second reader that was never given the first one's rules.

## One thing done on purpose

The amendment **never writes the exemption marker at the start of a line.**
prd-56's own milestone closes when wave 2 lands, at which point a line-start marker
in the document that *defines* ruling 2 would exempt it from its own law. Verified:
zero line-start occurrences in this file.

**No independent pass has read this commit.** It was written by the session that
ran wave 2's review, and its content derives from that review's findings. The
runbook's first condition for opening a PR is an independent review pass on the
record; that condition is not met here, and this paragraph is the disclosure rather
than a way around it. The reviewer of this PR is the first other pair of eyes.

## Gates

Run in its worktree: typecheck **0**, lint **0**, `VITEST_MAX_WORKERS=6 npm test`
**0** — 512 files, 9704 passed, 3 expected-fail, 3 skipped.

The typecheck needed `postgres` supplied first: it is declared in
`packages/team/package.json` and **absent from the main checkout's `node_modules`**,
which every worktree hardlinked from it inherits — including both review seats,
whose typecheck reds this session were caused by that and not by any change under
review. Worth fixing in the main checkout; unrelated to this commit.

**One platform, linux.** With Actions retired there is no macOS or Windows leg.


KelliherL (approved):
## Review of #486 — approve; the first other pair of eyes your declared gap asked for

One commit (`1e822803`), one file, twenty-two lines, no behaviour. Read against prd-56's existing Sequencing, against #484 as filed, and against what the detector reads. In a clean clone at node 22.23.2 on Linux.

### What verified — EXECUTED on `1e822803`

- `npm run typecheck` **0** · `npm run lint` **0** (1090 files) · `doc-citation-law`, `prd-location-law`, `runbook-delivery-law` — **green**.
- The declaration is in the form the detector reads: `grep '^\*\*Wave [0-9]'` on the branch's prd-56 →
  ```
  133:**Wave 1 — the keystone, and it is a STACK in one issue.** …
  142:**Wave 2 — the exemption.** …
  147:**Wave 3 — the reader tells prose from code.** …
  ```
  Waves 1 and 2 are byte-identical to `main` (the diff adds one block and touches nothing above or below it).
- **#484 exists** (`prd56 w3: the exemption reader tells prose from code in every Markdown context`, filed 23:55Z), and its title is the literal the wave block quotes.
- **Your "never at the start of a line" claim:** `grep -c '^\*\*Shelf exemption:\*\*'` on the branch's prd-56 → **0**. The file's only mentions are mid-line, so the document that defines ruling 2 does not exempt itself the day its milestone closes.
- `scripts/dev/prd-reconcile.sh 56` on this checkout reports **UNDECLARED WAVE #484** — and says why in its own header: it judges against `origin/main`, where wave 3 is not declared yet, not against the working copy. That is the drift this PR exists to close, seen from the side that has not merged it. Not a finding against the change; noting it so nobody reads the red as this branch's.

### REASONED

- "One repair, not four" is the right acceptance shape and it is in the PRD rather than left to the issue — the same argument your #485 body makes about fitted instruments, applied one level down. The four edges match the ones #485's commit body says the fix re-review found, and the "three fail open, one fails closed" split is consistent with the reader as shipped there (a fence indented one to three spaces is not caught by `FENCE_OPEN_RE` after the blockquote strip; a quoted-code continuation line counts toward the threshold; a comment opener inside a code span is blanked as if it were a comment — that one fails closed).
- The aside about `postgres` missing from the main checkout's `node_modules` is your machine's: this clone's `npm ci` installed it and typecheck ran clean without a manual step.

### What I did not run

The full suite — one Markdown file, and the three laws that read PRDs ran. macOS and Windows.

**Approving.** Landing is the operator's act; once it lands, `prd-reconcile.sh 56` goes clean by construction.

PR #492, opened 2026-09-14
charter-laws: the manifest coupling is in the registry, and two [not-ours] rows name the symbol they mean

### LINKED ISSUE #443: coupling: the PRD milestone manifest is a registry entry
## Fence (may touch ONLY)
- `.swarm/coupling.txt`

## Why

prd-56 wave 1 (#427) introduced a committed manifest, `.swarm/prd-milestones.txt`,
and a law that reads it — `packages/server/src/prd-location-law.test.ts`. Together
they create a coupling that nothing in the registry records:

**Closing any `prdNN` milestone on GitHub now reddens `packages/server/` for a
lane that never entered that directory.** The remedy — regenerate the manifest,
move the PRD onto the archival shelf, commit both — is written only in the
generator's own header, which that lane has no reason to open.

The same is true in the other direction: adding a new PRD file to the live shelf
reddens the law's floor until the manifest is regenerated.

This is the situation the registry exists for, and the one the wave lint warns
on. It is a registry entry, not PR prose — the same conclusion #353, #358, #381
and #388 each reached for their own cross-package pin.

Raised by the independent review of #427 and deferred rather than folded in:
the registry file sits outside that issue's declared boundary, and widening it
to reach a second concern is how a bundle becomes a grab-bag.

## Definition of done

- The registry names the manifest, the law that reads it, and the generator that
  writes it, so a lane that trips this finds the remedy where it already looks.
- The entry says what the coupling COSTS — which action reddens which package —
  rather than only listing the paths.
- It is cited by phrase or by sha, not by line number: the registry's own header
  records that a pinned line number for a file that moves rots by construction.
- The wave fence lint (scripts/fence-lint.sh, deliberately unticked here — see
  Notes) still passes for any wave that claims a path under the swarm directory.

## Notes

No code changes. This is a documentation-of-coupling issue; the behaviour it
describes is correct and deliberate.

**Why one path above is written without backticks.** scripts/fence-lint.sh
extracts a fence by taking every backticked token from the fence section *and*
from any line containing the word "fence" — so a ticked path in a Definition of
done sits in the linter's claim set exactly as if this issue had fenced it. It
did: the lint read this issue as claiming the linter's own source, a file it
must never touch. The path is spelled bare above so the claim set matches the
fence. The linter's over-extraction is a defect in the linter, recorded
separately; this issue is not the place to fix it.

### LINKED ISSUE #481: citation law: two [not-ours] rows name live code under a spelling that is not the symbol

gabriel-canaan:
Closes #443
Closes #481

Two independent issues in the `charter-laws` milestone, disjoint fences, bundled
into one PR so the queue toll is paid once. `scripts/fence-lint.sh 443 481`
reports no overlaps.

**No production code changes.** One registry entry, one design note, one baseline
data file, and three pinned count literals that move with it.

## #443 — the PRD milestone manifest is a registry entry

`.swarm/coupling.txt` gains three entries recording a coupling nothing held:
closing or opening a `prdNN` milestone on GitHub reddens `packages/server/` for a
lane that never entered that directory, because `prd-location-law.test.ts` reads
the committed manifest `scripts/dev/prd-milestones.sh` writes.

Written in terms of the law and the generator rather than the manifest's row
shape, deliberately — #428 was in flight and changing that shape. It landed
before this reached review and the entry needed no rewrite, which is the argument
rather than a coincidence.

## #481 — two `[not-ours]` rows named live code

Follow-up to #460. Two rows in `.symbol-citation-baseline` named live code under a
spelling that is not the symbol, grouped under a header that says the opposite.
They were treated differently, and the difference is the point:

- **`MAX_PX`** — the document was wrong, not the row. It now writes
  `RELAX_REACH_MAX_PX` in full, the row has nothing to pin, and it is gone.
  Count 71 → 70.
- **`manifestReadOverride`** — a deliberate shorthand for a real get/set pair,
  both halves declared seven lines below the citation. Row **kept**, with a group
  header that is true of it.

## What is the sibling case?

Answered by reading, not by sweeping — and the commit body says so, because no
automated rule here can answer it. Two have been tried and broken: edit distance
≤ 1 misses `MigrationResult`→`MigrationOutcome` (distance 7), and a
shared-leading-word rule misses a typo inside the leading word.

The same shape occurs in `packages/**/*.ts` comments, not only in `docs/`, and
244 distinct instances there are unread. That figure is distinct pairs over
**tracked** source; the body carries the command and the reason the framing
matters.

All three moved literals were mutated **independently**, because vitest aborts an
`it` at its first failed `expect` and the first assertion would otherwise mask the
others: `doc-citation-law.test.ts:2930`, `:2933`, and the `# measured:` stamp each
redden alone. Reintroducing `` `MAX_PX` `` into the doc reddens with an actionable
message, so dropping the row narrowed the law rather than widening it. Bending
`scale.ts`→`scaIe.ts` reddens the path law, so the new citations are load-bearing.

For #443: deleting the three entries drops `fence-lint.sh 443 428`'s `ok` lines
from 2 to 0.

The stamp's `tree=`, `candidates=` and date deliberately did **not** move. They are
one record of one sweep and this change ran no sweep.

## Fence widening

## Review history — three rounds, and two of the blockers were in the repairs

This is recorded because it is the useful part, not to pad the PR.

**Round 1** (two seats, independent worktrees) — one blocking finding in #443: the
registry entry told a tripped lane the remedy lived "exactly once, in the
generator's own header comment". False three ways. Plus three accuracy findings.

**Round 2** (two seats, fix re-review) — **the round-1 repair was defective.** The
false clause sat in *two* entries two lines apart; the repair corrected only the
one the review named and shipped the identical false pointer one line below the
paragraph describing it. Both seats caught it. The second repair was therefore
made at the level of the class: the remedy is now stated once and the other
entries defer to it, so no two can disagree.

**Round 3** (one seat — see the gap below) — three factual errors in commit
bodies, all now corrected. The most instructive: a figure of "245 distinct pairs"
shipped with a `grep -r packages` command to prove it. That command descends into
`packages/*/node_modules`, where exactly one dependency file carries the shape, so
the number counted a file that is not this repo's source. The reviewer ran the
shipped command, got 244, and controlled it by planting one synthetic pair to show
the harness was sensitive to exactly one. The command now uses `git ls-files`.

## Declared gaps — please read before approving

**The final commit `4e0c8aba` has not been reviewed by anyone.** It is the amend
that applies round 3's three corrections, and no seat has run against it. Its
changes are confined to commit-body prose; the tree it carries is identical to the
round-3 head apart from that. This is disclosed rather than recorded away, because
a review record the reviewed party writes for itself is not a record.

**Round 3 was one seat, not two.** The second seat returned an empty report,
consistent with its provider quota being at 92% of a 5-hour window at launch. Its
silence is not agreement.

**One platform.** Every gate here ran on WSL2 Linux. There is no macOS or Windows
witness, and this repo's workflows are disabled, so this PR will carry no check
runs. The `.gitignore` negation in the parent PR and this branch's path handling
are the kind of thing a case-insensitive filesystem judges differently.

## Gates

512 files, 9729 passed, 3 expected-fail, 3 skipped. `scripts/fence-lint.sh 443`
passes. `scripts/dev/coupling.test.sh` reports 10 passed / 1 failed; that failure
is the pre-existing `docs/adr/` entry and is byte-identical on `main`.

---

## Declared review gap

> 4e0c8aba is the amend applying round 3's corrections and no seat has run against it; its changes are commit-body prose only. Disclosed in the PR body rather than force-recorded.

```
  UNVERIFIED  4e0c8aba  fix(baseline,geometry-relax-reach): two [not-ours] rows are read against their citing text, not swept   (#481)
```


KelliherL (approved):
## Review of #492 — approve; the unreviewed commit read, five mutations run, nothing to fix

Read commit-by-commit (`d210b572` the registry entries, `4e0c8aba` the two `[not-ours]` rows) against #443's and #481's own Definition-of-done, in a clean clone at node 22.23.2 on Linux. Your declared gap is that `4e0c8aba` has had no independent pass; this is that pass, and it read `d210b572` too.

### What verified — EXECUTED on your head `4e0c8aba`

- `npm run typecheck` **0** · `npm run lint` **0** (1090 files) · `doc-citation-law` + `prd-location-law` — **green**.
- **The three moved literals, each mutated alone** (your point about vitest aborting at the first failed `expect` is right, so they were reverted one at a time): `notOurs.length` → **red**; the distinct-symbol count → **red**; the `# measured:` stamp's `not-ours=` → **red**, and it reddens a *different* test (*"the measured stamp parses, its tree sha landed, and its counts match the lists it describes"*). Three literals, three independent convictions.
- **Your `MAX_PX` mutation** — reintroducing `` `MAX_PX` `` into the design note → **red** on *"every symbol cited from an in-scope document or comment still exists, unless the baseline names it"*. Dropping the row narrowed the law rather than widening it, as claimed.
- **One of mine:** deleting the `manifestReadOverride` row outright → **3 red** (the stamp, the pin, and the citation sweep). The row that was *kept* is load-bearing too, which is the half a "we removed a row" change usually leaves unproven.
- **The doc's new symbol claims resolve:** `RELAX_REACH_MAX_PX` is declared at `scene/geometry/scale.ts:259` and re-exported through `geometry.ts:51`; `RETIRE_RELAX_PX` is a module-private `const` at `geometry/layout.ts:66` and appears nowhere as an export (the `scale.ts:254` hit is a `{@link}` in a doc comment). `manifestReadOverride` is a real get/set pair at `poll-loop.test.ts:27–31`, cited at `:19` — seven lines above, exactly as the row now says.
- **Your two greps reproduce:** the false-pointer clause is gone from `.swarm/coupling.txt` (0 hits), and `scripts/dev/prd-milestones.sh` has 0 hits for `done/` — so the split remedy the manifest entry documents is really split the way it says.
- `scripts/fence-lint.sh 443 481` → **PASSED**. `scripts/dev/coupling.test.sh` → **10 passed, 1 failed**, the `docs/adr/` entry, byte-identical on `main` as you say.

### On the round-1-repair-was-defective story

That is the part worth keeping, and the second repair is the right shape: stating the remedy once in the manifest entry and having the generator and law entries point at it means no two can disagree, which is a property rather than a correction. The entry now says the full remedy is *"regenerate the manifest, move the archived PRD(s) into `docs/prds/done/`, commit both together"* and that you will not find that sentence written anywhere else — verified, you will not.

### REASONED

- Writing the entry against the law and the generator rather than the manifest's row shape was the right call and it was tested by #428 landing underneath it. That is a prediction that was made and then paid off, not a coincidence, and the entry needed no rewrite.
- The `manifestReadOverride` row keeps its place with a header that is true of it, while `MAX_PX`'s row goes because the *document* was wrong. Treating those two differently is the whole point of #481, and the new group header says which case it is.

### What I did not run

The full suite — four files, none of them production code, and the two laws that read them ran. macOS and Windows. Your round-1/2/3 seat transcripts (read the findings, not the transcripts).

**Approving.** `4e0c8aba` is read; the declared gap is closed. Landing is the operator's act.

PR #493, opened 2026-09-14
prd56 w3: the exemption reader tells prose from code in every Markdown context

### LINKED ISSUE #484: prd56 w3: the exemption reader tells prose from code in every Markdown context

gabriel-canaan:
Closes #484

prd-56 wave 3. Ruling 2 requires a declared shelf exemption that a **reader** can
find. Wave 2 shipped a reader that could not tell prose from code; its fix
re-review found four contexts where the reader and a person looking at the
rendered document disagree. This closes them with one predicate rather than four
patches.

## The shape, which is the acceptance criterion

`classifyLines` is the single prose-or-code predicate, and **both scans read the
same classified array** — the declaration scan tests `line.isCode`, the
continuation loop tests `cont.isCode`. That is what closes gap 3 structurally: the
continuation loop was a second reader that had never been given the first one's
rules.

| gap | was | direction |
|---|---|---|
| 1 | `FENCE_OPEN_RE` column-0 anchored; CommonMark allows a fence opener indented up to 3 | fail open |
| 2 | `stripBlockquotePrefix` ate `[ \t]*` after `>`, so the indent guard could never fire inside a blockquote — where every real declaration lives | fail open |
| 3 | continuation loop applied no fence or indent check | fail open |
| 4 | comment blanking was not code-aware, in both directions | both |

The input class is written down as a **truth table in the file**, which the issue
required. It is the artefact the next reader will trust instead of probing, so it
is part of the deliverable rather than documentation.

## What is the sibling case?

**Fixing the three fail-open gaps and leaving the fail-closed one**, because
fail-closed feels safe. It is not: it reddens the suite for every lane, and ruling
2's whole design rests on a law that does not convict a correct state. Gap 4's
false-conviction half is fixed.

**Second, and it is the one that bit:** fixing the declaration scan and forgetting
the continuation loop. Gap 3 already *was* that mistake one level down. The
predicate is called from both places, and a test proves it.

A test per gap asserting only the exact inputs the issue named would survive a fix
that special-cased four strings. Review verified each gap test reddens against the
base tree — including by *content* for gap 4b — and each runs beside a control that
still clears a valid declaration in the same call.

Round 2 found the sharper version of this: the first attempt treated a tab as
exactly 4 columns, which is not the rule. It is now real column arithmetic — a tab
advances to the next multiple of 4, which depends on what precedes it. Verified
across the space/tab prefix class, including `>` + 3 spaces + tab (6 columns, code)
and `>` + space + tab (2 columns, prose), neither of which was named in the
findings that prompted the fix.

## Review

One round, recorded in full on #484.

**It was a REDUCED pass — one seat, not two.** Codex exited with
`workspace_member_credits_depleted`; the 5h window read 41% earlier the same day,
so this was genuine credit exhaustion rather than a rate limit, and there was no
second seat to run. The reconciler's job was therefore done by the orchestrator
against the code, re-executing every claim rather than accepting it.

That seat verified the corpus both ways — **56 PRDs across both shelves,
byte-identical verdicts under the base and new readers**, so nothing changed
disposition — and built a tab-expansion oracle pinned on three CommonMark spec
examples rather than arguing from intuition.

It returned two must-fix items, both now closed:

1. **Truth-table rows 9 and 10 stated a verdict the code does not produce.** The
   code was right — a 4+-column-indented backtick run is indented code, not a fence
   opener, so the unindented line below it is visible prose. The table was wrong,
   and nothing pinned it. Rows corrected, and they now say explicitly which of the
   two was at fault.
2. **A new fail-closed regression on `>` + space + tab**, from the missing tab
   expansion above. Fixed at the instrument, not at the four affected prefixes.
   Two claims that outran the code — a doc comment asserting CommonMark conformance
   and a table row reading *"a tab counts the same as 4 spaces"* — are gone or
   narrowed to what shipped.

Four further findings were recorded and **deliberately not fixed here**: missing
container tracking on fences, list-item declarations, lazy continuation, and
comments whose span crosses a fence delimiter. The last is #459's, whose scope was
widened for it before this PR opened. The rest are pre-existing and outside this
wave's four gaps.

**The amended head has not been through an independent pass.** The seat read
`3547d6ab`; round 2 amended it away, and the repair — the column arithmetic and the
table rewrite — has been read only by the session that requested it. The verify
ledger reads it UNVERIFIED, and that is accurate rather than a bookkeeping lag.

Stopping was a judgement, and here is the case against it as well as for it. For:
the round counter stood at 3 against a cap of 1 for test-only findings, both round-2
findings were *claims* rather than behaviour, and with Codex down any further round
is single-seat — which is where a blind spot is most expensive. Against: the
orchestrator got the tab rule wrong twice during this review and the lane got it
right both times, and tab arithmetic is exactly the code now going out unreviewed.

**The reviewer of this PR is the first independent read of this commit.**

## Gates

Orchestrator, full worktree, on the merged head: `typecheck` **0**, `lint` **0**,
`VITEST_MAX_WORKERS=6 npm test` **0** — 512 files, 9744 passed, 3 expected-fail, 3
skipped. **One platform, linux.** GitHub Actions was retired on this repo on
2026-09-11, so there is no macOS or Windows leg behind any of it.

---

## Declared review gap

> 7712c6a5 is an amended head: the verify seat read 3547d6ab, round 2 amended it away, and the repair (tab column arithmetic + truth-table rewrite) has been read only by the session that requested it. Stated in full in the PR body's declared-gap section.

```
  UNVERIFIED  7712c6a5  fix(prd-location-law): one predicate tells prose from code in every context  (#484)
```


KelliherL (approved):
## Review of #493 — approve; the amended head read, and one claim in it is false where you said to look

This is the independent pass your declared gap asked for, and it went at the part you named: *"the orchestrator got the tab rule wrong twice during this review and the lane got it right both times, and tab arithmetic is exactly the code now going out unreviewed."* The arithmetic is right. The **sentence describing it** is not, and that is the finding.

Read in a clean clone at node 22.23.2 on Linux, against #484 and against the four gaps as stated.

### What verified — EXECUTED on your head `446b44c7`

- `npm run typecheck` **0** · `npm run lint` **0** (1090 files) · `prd-location-law` green · `doc-citation-law` + `runbook-delivery-law` green.
- **The corpus claim, re-derived through the real law file rather than a copy of it.** I appended a dump to a copy of `prd-location-law.test.ts` at this head and to `origin/main`'s version, ran both, and diffed the verdicts: **0 PRDs change disposition**, 4 carry a declared reason. My count is **55** rather than 56 — I excluded `README.md`; the shelves hold 56 `.md` files, three of them the non-PRDs your law's own glob-hazard note names. Either way the claim holds: nothing changed disposition.
- **Five mutations, each restored, every one red:**

  | mutation | result |
  |---|---|
  | gap 1 — `FENCE_OPEN_RE` back onto the raw line | **red**, incl. the gap-1 test by name |
  | gap 2 — the `credit >= 4` indent guard disabled | **red** ×4, incl. "an indented code block (4 spaces or a tab)" |
  | gap 4a — the comment regex requires a closing `-->` again | **red**, the gap-4a test by name |
  | gap 4b — comments blanked in the raw text, code spans unmasked | **red**, the gap-4b test by name |
  | round 2 — `nextTabStop` flattened to `col + 4` | **red** ×2, incl. the continuation-loop case |

  (My first attempt at the gap-4a mutation used `sed` and silently did not apply; the grep I printed beside it showed the unmutated line, which is the only reason I noticed. Re-run properly, it reddens.)
- **Your disclosure about `if (cont.isCode) break` is accurate, and I checked it rather than taking it.** Removing that line leaves the suite green, because every code line carries `content: ''` and the next check ends the paragraph anyway. Gap 3 is still pinned — both the gap-1 and gap-2 mutations redden the gap-3 test through `classifyLines`, which is the structural claim the wave actually rests on.

### The finding — a false claim about CommonMark, in the tab rule. Pushed as `acf283f2` (prose only)

`stripBlockquoteMarkers` consumes a tab following `>` **whole** and credits the marker with the whole advance. Its doc comment calls that *"the entire rule"*. It is not CommonMark's rule: CommonMark takes **one column** of that tab and leaves the rest as content indentation. The spec's own example is the proof — `>\t\tfoo` renders as indented code carrying **two leading spaces**, which is only arithmetic if the marker took one column of the first tab and left two behind.

Measured through your shipped reader, not a re-implementation:

| prefix before the marker | this reader | CommonMark |
|---|---|---|
| `>\t` | prose (credit 0) | prose (credit 2) |
| `>\t ` | prose (1) | prose (3) |
| **`>\t  `** | **prose (2)** | **indented code (4)** |
| **`>\t   `** | **prose (3)** | **indented code (5)** |
| `> \t`, `>  \t`, `> \t `, `>  \t ` | prose | prose ✓ |
| `>\t\t`, `> \t\t` | code | code ✓ |

So rows 30–36 are right, and the class they generalise to has two members they do not cover: a tab **immediately** after `>` followed by two or more spaces. It **fails open** — hidden from a reader, read as a declaration here — in the wave that closes three other fail-open gaps.

**I narrowed the claim rather than changing the code**, and wrote the case into the truth table as **row 37**, because the table is the artefact this wave argues the next reader will trust instead of probing. Three reasons for not fixing the behaviour here, in order: the divergence is **unreachable in this corpus** — EXECUTED, no PRD on either shelf contains a tab at all — partial-tab consumption is a structural change to the instrument rather than a cell, and this wave already books exactly that class (container tracking, lazy continuation) out to a later one. Round 2 corrected two overclaims of CommonMark fidelity in this file; this is the third, one level down.

The commit touches **no line outside a doc comment** (asserted mechanically before committing), and the law is green before and after. Drop it if you would rather fix the arithmetic in wave 4 and say so there instead.

### The union

`origin/main` moved under you. Merged **into** the branch locally — clean — and the whole leg run on the union with `ci-local.sh --no-install`:

```
  Test Files  514 passed (514)
       Tests  9754 passed | 3 expected fail | 3 skipped (9760)
```

One platform (Linux, node 22.23.2). I did not push the merge; the branch merges cleanly as it stands.

- **The fence-close guard is unpinned.** Dropping `indent.credit <= 3` from the close check in `classifyLines` (so a 4+-indented ``` ``` ``` line closes an open fence) leaves the suite green. The guard is correct — CommonMark lets a closing fence be indented up to 3 — and it is the same shape as gap 1 on the other side of the fence. One case away from being held.
- Row 30's "a bare `>` + tab is a paragraph, and always was" is true under both readings (CommonMark gives it credit 2), so the correction to the record stands unaffected by row 37.

### What I did not run

macOS and Windows. The seat's own transcript on #484 (read the findings, not the transcript). A renderer — the CommonMark arithmetic above is the spec's, worked by hand and cross-checked against the shipped reader's verdicts, not taken from a live renderer.

**Approving.** The four gaps are closed and each is pinned; the tab arithmetic is right and now says so honestly. Landing is the operator's act.

Issue #27, opened 2026-08-23
prd39 w2: the NUL guard fails when it cannot run


ciaran-slow:
## Fence (may touch ONLY)
- `scripts/gate.sh`

## Blocked by
Blocked by prd39 w1 (#1). Same file, so this is a **stack, not a bundle** — prd-39's Sequencing says so by name, and the working agreement forbids bundling across a live fence. Do not open this beside #1.

Also downstream of prd-39 **wave 0, an operator act**: wave 1 is landed by hand, without `gate.sh`, because the gate cannot be trusted to land its own fix. The same caution applies to this issue's own landing.

Sibling lane owns `AGENTS.md` (#2) and never reads this script's line numbers, which is the point of that change.

**Model:** sonnet. **Wave:** 2.

## Definition of done
- `python3` is asserted **once, up front** — near the other preflight checks around `gate.sh:27-29`, not inside the loop — and `fail`s when absent. Same posture the timing pass already takes toward its own file count at `:104`.
- The `|| echo 0` fallback at `:62` is **deleted**. An unreadable file is a failure, not a clean verdict.
- `2>/dev/null` on that line no longer hides why the probe failed. A guard that cannot say why it could not run is the same defect one layer down.
- `fail()` at `:25` is reused, never reimplemented — it already prints `GATE FAILED`, prints `>>> HOLDING $H`, and exits 1. prd-39's "what already exists" says both of this PRD's fixes route through it.
- The binary-asset exemption at `:59-61` is untouched. A docs lane committing PNGs was falsely held before that exemption existed; narrowing it is not this issue.

## Proving it — a lane cannot run the gate

`gate.sh` **is** the landing: it merges to local `main` and pushes. A lane must never run it, so "I ran the gate" is not available as evidence here, and neither is a unit test — `grep -rln "gate.sh" packages/**/*.test.ts` finds only `// @gate-timing` marker comments, so **nothing in the suite asserts this script's behaviour and there is no green test to lean on.**

Prove it by executing the guard's logic in isolation, and show both directions:

- **It bites when the dependency is missing.** Run the NUL block with `python3` masked from `PATH` (a scratch dir first on `PATH` containing no `python3`, or a stub exiting non-zero) against a file with **no** NUL bytes. Before: prints `no NUL bytes` and continues. After: `GATE FAILED`. Paste both.
[...]

PR #499, opened 2026-09-14
prd43 w11: a PRD's declared open issues are reconciled against the tracker

### LINKED ISSUE #448: prd43 w11: a PRD's declared open issues are reconciled against the tracker
prd-43's thesis is that a document's factual claim fails the build when it goes false. **A PRD's own status header is exempt from that today**, and it went false twice in one week with a green suite throughout.

## Evidence

`docs/prds/prd-30-the-open-hand.md`'s header listed `#389 (wave 4)` under "what is left" and quoted `NOT_YET_SWEPT`'s four carve-outs and twelve surviving tooltips as live — after #389 merged (`b33da639`, in `cea5a361`) and **deleted** that identifier. `docs/roadmap.md`'s prd30 entry was staler still, naming three items as held by a law that names none of them.

The mechanism is worth recording because it generalises: the stale block was authored in `c243c28f`, a commit sitting **directly on top of** the wave-4 work, inside the same PR that landed it. *A status line authored inside the PR that lands the thing it calls outstanding is stale at merge by construction.*

Both were corrected by hand in PR #431 (`715f0a21`, `7befc492`, `f5038239`). Nothing went red at any point.

## Why nothing caught it

Two gaps, and the second is the larger:

1. `packages/server/src/doc-citation-law.test.ts` checks that backticked **paths** resolve. It does not check issue state, and it does not check identifiers — a renamed constant already leaves a green suite and a lying doc.
2. **It does not sweep PRDs at all.** `EXCLUDED_DIRS` at `doc-citation-law.test.ts:493` excludes `docs/prds/` as a *citing* source. Proven by mutation during #431's review: a fabricated path inside a PRD leaves the law green (49/49, exit 0); the same fabrication in `docs/roadmap.md` turns it **red**, naming the file. So `docs/roadmap.md` is covered and `docs/prds/` is not.

## What becomes true

A PRD status claim that names an issue as outstanding names an issue that is actually open, and the build says so when it stops being true.

## Rulings this needs before it is dispatchable

It is filed as **not ready** on purpose — three decisions have to be made first, and each one can make the law worse if guessed:

- **Which text is a status claim.** A whole-file sweep for `#NNN` is wrong. AGENTS.md's citation note records that this repo's tracker was deleted and rebuilt from `#1` on 2026-08-21, that the two sequences overlap, and that prior-tracker citations are *legitimate historical provenance* left in place deliberately. A naive law fires on all of them. Scope it to a declared syntax or to the status header block.
- **Whether `docs/prds/` leaves `EXCLUDED_DIRS`.** That exclusion is deliberate — PRDs are dated artefacts and their *Evidence* sections are explicitly records of a tree at a named sha, not live claims (prd-30's header says so in as many words). A status header is the opposite: it is read as current. The law may need to distinguish the two rather than the directory.
- **What a correct-but-stale claim costs.** prd-30's block was true when committed and false on merge. A law that only fires at authoring time would not have caught it.

## Constraints learned the hard way

- **CI is a depth-1 clone.** History-reading laws pass locally and invert on CI; the suite leg needs `fetch-depth: 0` and the law should assert it is not shallow.
- **Scope the guard by what a file *is*, not what it is named.** AGENTS.md records a fixture-hygiene law that matched on a filename prefix, missed three older files, and then asserted truthfully and uselessly that nothing was wrong.
- `doc-citation-law.test.ts` already derives the live tracker maximum from the tracker rather than from prose, so there is precedent in that file for a law that reads tracker state.

## Known customers

- prd-30's header and `docs/roadmap.md`'s prd30 entry (fixed by hand in #431).
- #447 — prd-14-experiment-console.md line 317 named #220 as an open blocker for prd-14 w2; it closed at 8ee1b498. Same class, found by the same review. Use it as a test case; do not block on it. (Path and sha deliberately unticked — the lint reads a backticked token on such a line as a claim.)

Found while reconciling prd-30 (PR #431). The law was deliberately kept out of that PR: a prose fix should not smuggle a new gate in beside itself.

---

## Fence (may touch ONLY)
- `scripts/dev/prd-reconcile.sh`
- `scripts/dev/prd-reconcile.test.sh`
- `docs/prds/done/prd-43-the-claim-is-a-test.md`

## Definition of done

The three rulings this issue was filed pending are **made** — see the ruling comment
dated 2026-09-14. In short:

1. A status claim is a **declared field** in the head blockquote, not a region of it.
2. `docs/prds/` **stays** in the citation law's excluded directories; nothing about that
   law changes.
3. The check is a **fourth check in the reconciler**, reported next to DOUBLE-DECLARED /
   VACANT / NO WAVE. **It does not fail the build** — both suite mechanisms were measured
   dead, and the evidence is in the ruling comment.

So this issue delivers:

- a `STALE STATUS` check in the reconciler, reading only the declared field, and reporting
  in the established shape;
- coverage for it in the reconciler's existing standalone harness, which already builds a
  throwaway git remote+clone pair and puts a mock `gh` on PATH;
- an amendment to prd-43 declaring **wave 11**, so the reconciler stops reporting this very
  issue as NO WAVE, and marking the **wave 5 DOUBLE-DECLARED** drift the same run reports
  (mark in place — citations must keep resolving; do not delete).

Verify by running the reconciler against prd-20, whose head declares #216 outstanding while
the tracker closed it on 2026-09-08, and against prd-49, whose head declares #190 and is
**correct** — the check must stay silent there. Those two are the bite and the control.

gabriel-canaan:
Closes #448

A fourth check, `STALE STATUS`, in `scripts/dev/prd-reconcile.sh`, reported beside
DOUBLE-DECLARED / VACANT / NO WAVE: a PRD's head declares an issue outstanding and the
tracker disagrees. Operator-run, **reporting only — it never fails a build.**

prd-30's header named an issue as outstanding after that issue's own wave had merged and
deleted the thing it was open about, with a green suite throughout, because nothing read a
PRD's header against the tracker at all.

## ⚠️ Read before merging — one commit in this range is NOT independently reviewed

The two-seat pass (Opus `xhigh` + Codex `gpt-5.6-terra` `xhigh`) reviewed **`4b620c50`**.
A later round then **inverted the classifier's default** to non-claim and added a new
corpus law. That change is larger than the five repairs the seats did review, and **no
independent pass has read it.**

The reviewer's own review-record tooling holds a row for this head, but it was accepted on
file timestamp alone — it cannot tell whether that record was about this commit, and it is
not evidence that this content was reviewed. That row's text now says so. **Any check that
consumes such a record will still find one and pass.** Do not read it as verification.

This is written here rather than recorded away, per AGENTS.md: *"Evidence you authored
yourself is not evidence… the honest move is to write the gap into the PR body, where a
reviewer sees it."*

## The three rulings behind it

Recorded in full on #448, dated 2026-09-14.

1. **A status claim is a declared field, default non-claim.** Only `#N` inside a dedicated
   `**Open:**` field in the head blockquote is a claim. `**Status:**` and `**Outcome:**`
   are prose and their contents are never claims. **Opt-in, and no backfill** — other
   programmes adopt the field when they choose to.
2. **`docs/prds/` stays in the citation law's `EXCLUDED_DIRS`**, untouched. This check
   reads only its own field and never consults that list.
3. **Reporting only.** A `gh`-backed suite check reddens on a tree nobody touched the
   moment someone else closes an issue; a git-log closure signal is unsound (`#190` is
   open with a landed commit naming it, and `Closes #N` appears zero times in the last 20
   merges). Both were measured and rejected.

## The accepted cost, stated rather than hidden

**No PRD carries an `**Open:**` field today, so this check currently catches nothing** —
including prd-20's `#216`, which is a genuine stale claim. It becomes live as programmes
opt in. An earlier draft read every bare `#N` in the head paragraph as a claim and fired
**17 times across the corpus, ~12 of them false** — against heads that say *"milestone
closed with issues #1, #2 and #27"* and *"twenty-eight issues, all closed"* — instructing
the operator to amend documents that were already correct. A law that is honestly inert
beats one that is confidently wrong twelve times.

Coverage is reported either way, so an unchecked head is never mistaken for a verified one:

```
head field: NOT CHECKED — no **Open:** field in the head blockquote
```

## What is the sibling case?

The fix for the silent-pass path created one, and it was caught: making an unreadable
tracker set `STALE=1` gave `INCONCLUSIVE` a second cause, while the banner still named
only the first — telling the operator to fix the git remote when the GitHub API had
failed. The banner now branches on cause, with negative controls asserting each does not
name the other.

The wider sibling is the one this change is built around: an unread claim and an unfetched
trunk are the same epistemic situation, and the trunk case already had `STALE=1` → exit 3
one screen away. The tracker case now does the same.

Before this branch, none of the corpus behaviour was tested at all — five rounds of review
ran entirely against invented fixtures, and the 12-false-positive design survived every
one of them. The guard for that is new here: a corpus law that extracts the field
extractor and classifier **from the live script** rather than reimplementing them, sweeps
all 55 real PRD files, and asserts that no head without an **independently grepped**
`**Open:**` field yields a claim. Independently grepped, so it cannot agree with the
classifier by sharing its bug.

The suite also retires the assertions that encoded the old default, by name — e.g.
*"prd-20 shape (Outcome:, no Open:): now silent, exit 0 (was: fired, rounds 1-5)"* — so
the change of behaviour is recorded rather than silently replaced.

## Fence

`scripts/dev/prd-reconcile.sh`, `scripts/dev/prd-reconcile.test.sh`,
`docs/prds/done/prd-43-the-claim-is-a-test.md`. **No widening.** No other programme's
document is touched: prd-14, prd-17, prd-20, prd-37 and `docs/roadmap.md` were read and
none edited. The prd-43 edit is append-only — 0 deletions — declaring wave 11 and marking
a double-declared wave-5 paragraph superseded in place.

Two defects in other documents were found and deliberately **left alone** for their owners:
prd-20's `#216` is genuinely stale, and the `(PR #462)` cited during review belongs to
prd-37, not prd-34.

## Gates

Run locally on this platform, exit statuses read directly. **This is one platform, not CI
— all workflows are disabled on this repo, so nothing here is evidence about macOS or
Windows.**

```
bash scripts/dev/prd-reconcile.test.sh   exit 0   135 passed, 0 failed
npm run typecheck                        exit 0
npm run lint                             exit 0
VITEST_MAX_WORKERS=6 npm test            exit 0   514 files, 9754 passed
corpus sweep, all 53 PRDs, live tracker  0 firings
```


KelliherL (approved):
## Review of #499 — approve; the unreviewed head read, the truth table driven through the shipped code, and one hole in the corpus law closed

This is the independent pass your warning asked for: *"a later round inverted the classifier's default to non-claim and added a new corpus law … no independent pass has read it."* I read that round, and I drove the documented input-class table through the **shipped** extractor and classifier rather than re-reasoning about them — extracted from `prd-reconcile.sh` by marker, the same way your corpus law extracts them.

Clean clone, node 22.23.2, Linux.

### What verified — EXECUTED on your head `7da6bdda`

- `bash scripts/dev/prd-reconcile.test.sh` → **135 passed, 0 failed** · `bash -n` clean on both scripts · `npm run typecheck` **0** · `npm run lint` **0**.
- **The input-class table, row for row, through the real code:**

  | input | field read? | claims |
  |---|---|---|
  | `**Open:** #216, #235` | yes | 216, 235 |
  | `**Status:** … #216 outstanding` | **no** | — |
  | `**Outcome:** milestone closed with issues #1, #2 and #27` | **no** | — |
  | `**Outcome:**` line then `**Open:** #216` | yes | 216 |
  | `**Open:** #216` then `**Note:** #999` | yes | 216 only |
  | `**Open:** #216`, blank `>`, `**Open:** #999` | yes | 216 only |
  | a second blockquote / a body field | **no** | — |
  | blank lines before the H1 | yes | 216 |

  The default really is non-claim, and `**Status:**`/`**Outcome:**` prose really is never read — including prd-39's *"closed with issues #1, #2 and #27"* shape, which is the one that would have told an operator to amend an append-only corpus.
- **The five round-6 boundary findings, each through the shipped classifier:** `ACPR #811` → **claim** (no false PR-reference); `PR #811` → not a claim; `#812-#813-#814` → **no claims**, the whole chain; `#216 — #235` spaced → **two independent claims**; `landed in PR #431, and #216 is still open.` → **#216 is a claim**; `[#225](https://x/a_(b)_c). and #226` → **both claims**, so the balanced-paren consumption holds. (a) through (e) all behave as the table says.
- **The corpus law is what it claims to be.** It extracts both awk programs from `$SCRIPT` by literal marker and asserts each came out non-empty, so it cannot silently sweep nothing — the failure mode you record having already hit. It sweeps **53** real `prd-*.md` files and conditions on an independent grep rather than a hardcoded zero.
- **The INCONCLUSIVE split is real and tested both ways.** Three branches, each with its own cause sentence and remedy, and the suite asserts the fetch-failure banner does **not** carry the tracker's wording or tell the operator to check `gh`. Your round-5 note — that asserting the bare substring `INCONCLUSIVE` is what let a wrong-cause banner ship — is the reason those negative controls exist, and they are the right shape.
- **`docs/prds/done/prd-43-the-claim-is-a-test.md` is append-only**: 47 insertions, 0 deletions.
- **"No PRD carries an `**Open:**` field today" is true of heads.** The only file in the corpus containing the marker at all is prd-43, at lines 626 and 639 — body prose describing the ruling, not a declared field. The check is inert, as stated.

### One finding — fixed and pushed as `3a645c7a`

**The corpus law's independent grep reads the whole file, while its own comment states the invariant over the head — so prd-43 exempts itself from the law, and it does so because of prose this PR adds.**

```
# THE INVARIANT: any PRD whose head does NOT independently grep-match `**Open:**` …
  grep -q '\*\*Open:\*\*' "$f" && has_open=1        # ← the whole file
```

`has_open=1` short-circuits both regression branches, so the one file whose body now quotes the marker twice is outside the law that guards this ruling. EXECUTED, before the change: plant `#99999` in prd-43's head `**Status:**` prose, revert the extractor to the rounds-1-5 bold-label sweep, and the law names **52 files and stays silent about prd-43**. With the presence check scoped to the head (`sed -n '1,/^## /p'` — a different tool and a different boundary rule from the extractor's awk, so it stays independent of what it checks) the same plant yields **53, prd-43 among them**.

The law still worked — it caught 52 — so this is a hole rather than a failure, and it is the shape this PR is about: a sentence and a check that had come apart. Suite stays at 135 passed. Test-only; drop it if you would rather narrow the comment than widen the grep.

### The union

`origin/main` merged in locally — clean. The shell suite on the union: **135 passed, 0 failed**. The whole leg:

```
  Test Files  514 passed (514)
       Tests  9754 passed | 3 expected fail | 3 skipped (9760)
```

One platform (Linux, node 22.23.2), and with the workflows disabled that is all there is. I did not push the merge.

- **The corpus count is stated two ways.** The body says the law *"sweeps all 55 real PRD files"* and the gates block says *"corpus sweep, all 53 PRDs"*. The law's own `find … -name 'prd-*.md'` returns **53**; the 55 figure is the `.md` count once `docs/prds/`'s two non-PRD dated documents are included. 53 is the number the law actually sweeps.
- **A tab after the quote marker hides a declared field.** `>\t**Open:** #216` extracts an *empty* field, so the claim is silently reported NOT CHECKED — the extractor's `sub(/^> ?/, "", line)` takes one optional *space*. Unreachable today (no PRD on either shelf contains a tab, EXECUTED at the review of #493), and worth knowing it is the **sibling of the tab gap in `prd-location-law.test.ts`** that #493 just recorded as row 37: two readers of the same corpus, both fine with a space and both blind to a tab in the same position. Whoever closes one should look at the other.
- The `**Open:**` field is a real improvement on the sweep, and the "do not backfill" ruling is the right call — but it does mean the check's first genuine catch is whenever a programme opts in. prd-20's `#216` staying uncovered is stated in the PR, in the script's doc comment and in prd-43; that is as visible as an accepted gap gets.

### What I did not run

macOS and Windows. `shellcheck` (not installed on this box) — `bash -n` only. The round 1-6 seat transcripts on #448: I read the findings and re-derived the behaviour, not the transcripts.

**Approving.** The inverted default is right, the corpus law is honest about what it sweeps, and the classifier does what its table says on every row I could feed it. Landing is the operator's act.

PR #503, opened 2026-09-14
doc law: the path law can say "cited from history, on purpose"

### LINKED ISSUE #461: doc law: the path law can say "cited from history, on purpose"

gabriel-canaan:
Closes #461

The **path** law gains a `[historical]` equivalent: a way to record a citation that
names a removed file **on purpose**, distinct from the debt list, mechanically
checked, and meaning the opposite.

The symbol law got this in #440. The path law's own input table has named the gap
as blocking ever since — at its `.tsx`/`.mjs` row it declines to widen and says
*"the law needs a way to say 'cited from history, on purpose' … that mechanism
does not exist yet."* This builds it.

**One file, one commit.** No production code: this is all inside
`packages/server/src/doc-citation-law.test.ts`.

## The case it exists for

`packages/web/src/scene/parity/capture.mjs` cites
`packages/web/src/scene/paint.ts` twice. That file is gone — `8686f24d` replaced
the 2D painter with a WebGL2 one — and the parity harness diffs against the
pre-deletion file deliberately, resolving it out of git history. **The citation is
correct because the target is absent.** Filing that as debt would have told the
next reader to go and "fix" a working reference.

## Explicitly NOT in scope

Widening the sweep to `.mjs`. That is a separate ruling with its own measurement.
This makes the widening possible without false-positiving on a live citation; it
does not make it, and no sweep pathspec changed here.

## What is checked

- **Both directions.** The citing file must still exist and still carry the
  citation, and the target must still be absent. Each of the three arms reddens
  independently for its own reason — proven by mutation, separately, because
  vitest aborts an `it` at its first failed `expect`.
- **The two categories cannot hold the same pair.** Disjointness is asserted.
- **The list is pinned**, so a silent addition cannot slip in.

## Operator ruling — recorded on the issue before dispatch

`ALLOWLISTED_BROKEN_CITATIONS` stays a TypeScript const rather than moving to a
data file the way #440's lists did.

**The ruling's first stated reason was wrong**, and the correction is on the issue
and in the docblock. It claimed a TS const would enrol its own contents into the
symbol law's token set and self-certify. It would not: `buildSourceTokens` carves
the law file out to its declarations only. The conclusion survives on simpler
ground — this law extracts citations from comments only and excludes the law file
as a citing file, so the data-file machinery buys it nothing. Two review seats
verified that independently.

## Review history — two rounds, and every finding after round 1 was in a repair

Recorded because the pattern is the useful part.

**Round 1** (two seats) — one blocking: the category was **inert**. It was built,
honesty-checked and pinned, and the violations check consulted only the debt list,
so a historical entry could never exempt anything. Recording an in-scope citation
would have required also listing it as debt, which the disjointness test forbids —
the category contradicted itself for every file the law sweeps. Invisible only
because its one entry names an unswept `.mjs` file. Plus three accuracy findings.

**Round 2** (two seats) — **three defects in round 1's repairs.** The docblock
still told readers the list was not consulted, 765 lines from the wiring that made
it consulted. The correction to the ruling's false reason was itself false in the
same way — asserting a mechanism without probing it, twice on one passage. And the
`reason.trim()` fix was applied to the new list and not to the identical check one
screen up.

All fixed, each with a mutation and a control.

## Declared gaps — please read before approving

**The head has not been reviewed.** Round 2 reviewed the pre-fix head; this commit
is the fix for what round 2 found, and no seat has run against it. Its changes are
one `.trim()` and two prose corrections.

**A known limitation, left in deliberately.** An entry can record a target that
never existed, not merely one that was removed on purpose. The honesty checks prove
the citation is still made and the target is absent now; they do not prove it was
ever there. Both seats demonstrated this with a fabricated pair and **split on
severity**. It is left because it is outside what #461 specified, the pinned list
makes every addition a visible reviewed edit, and the symbol law's `[historical]`
section has exactly the same property — tightening only this half would split two
categories meant to mirror each other. `git log --diff-filter=D -1 -- <cite>` is
the one-call check if it is ruled in later.

**One platform.** WSL2 Linux only. No macOS or Windows witness, and with this
repo's workflows disabled this PR will carry no check runs — the absence of a green
bar is not a failure. Note `.windows-known-failures` already lists this file, so a
new Windows-only failure inside it would not be separately visible.

## Gates

Rebased onto current `main` first — four PRDs had just moved to the done shelf,
which this law sweeps, so the pre-rebase result would not have described the tree
that lands.

514 files, 9757 passed. The law alone: 75/75. The two corpus counts the commit
re-derives (158 `.tsx`, 6 `.mjs`) re-checked on the rebased tree.


KelliherL (approved):
## Review of #503 — approve; the unreviewed head read, the three arms mutated independently

One commit, one file, no production code. Read against #461, in a clean clone at node 22.23.2 on Linux. Your declared gap is that round 2's *repairs* have had no seat on them; this is that pass.

### What verified — EXECUTED on your head `0388120e`

- `npm run typecheck` **0** · `npm run lint` **0** · `doc-citation-law.test.ts` alone — **green**.
- **The case the category exists for is real, on this tree.** `packages/web/src/scene/parity/capture.mjs` cites `packages/web/src/scene/paint.ts` **twice**; `paint.ts` is **absent**; and it was removed deliberately — `git log --diff-filter=D` names `8686f24d feat(scene): the scene is drawn by a WebGL2 painter (#578)`. So the citation is correct *because* the target is gone, and filing it as debt would have sent the next reader to repair a working reference.
- **Round 1's blocking finding is the interesting one and it is genuinely closed.** A category that was built, honesty-checked and pinned while the violations check consulted only the debt list is a category that could never exempt anything — and it would have been invisible for as long as its only entry named an unswept `.mjs` file. Two rounds later the wiring is there and the disjointness assertion no longer contradicts it.

### REASONED

- The three arms — citing file exists, citation still present, target still absent — each reddening on its own is the right shape, and your reason for testing them separately (vitest aborts an `it` at the first failed `expect`) is the same one I have had to apply to three PRs this week.
- Keeping `ALLOWLISTED_BROKEN_CITATIONS` a TypeScript const is right, and the *corrected* reason is the sound one: this law extracts citations from comments only and excludes its own file as a citing source, so the data-file machinery buys nothing. That the first stated reason was wrong — and that the correction to it was itself wrong in the same way, asserting a mechanism without probing it — is the most useful thing in the PR body.

- **The known limitation you left in — an entry can name a target that never existed — is the right call, and the asymmetry you cite is the reason.** Tightening only this half would split two categories built to mirror each other. `git log --diff-filter=D -1 -- <path>` is one call if it is ever ruled in, and it would close the same hole in `[historical]` at the same time.
- `.windows-known-failures` already lists this file, so a Windows-only failure inside it would not be separately visible — your note, repeated only because it is the kind of thing that gets forgotten when the workflows come back.

### What I did not run

The full suite — one file, no production code, and the law it changes ran. macOS and Windows. The round 1-2 seat transcripts.

**Approving.** Landing is the operator's act.

PR #504, opened 2026-09-14
prd56: ruling 4 rules the mirror case one-way, and says why

gabriel-canaan:
Part of #459 — **does not close it.** This is Part 1 of a branch that was split on
review; Part 2 (the parser fix) is held and #459 stays open for it.

## What this is

prd-56's Open Questions asked whether ruling 1's law should also fail an **archived**
PRD whose milestone is still **open** — the mirror case — and left it open on the
stated ground *"No instance exists today."*

**An instance exists, and prd-56's own wave 1 created it.** `prd43`'s milestone
reopened (1 open / 34 closed, `closed_at` null) after wave 1 archived
`prd-43-the-claim-is-a-test.md` on the strength of it being closed. So the reason the
question was parked has been false on `main` since then.

This makes it **Ruling 4**: the law stays one-way, and the justification changes. The
asymmetry is in what each direction misleads a reader INTO — a closed-milestone PRD
on the live shelf falsely reads as in flight and people act on it (that is
`docs/follow-up-292.md`'s reasoning, and #416's near-miss); an archived PRD whose
milestone reopened reads as done while a straggler remains, which is staler but
misleads nobody into treating a shipped ruling as open.

The ruling also states what the asymmetry does **not** license: the manifest row
stays accurate, and an archived PRD with a reopened milestone stays visible as such.
**The asymmetry is in what the law CONVICTS, not in what the repo RECORDS.**

## What is the sibling case?

Ruling one direction and silently leaving the other undocumented — which is how the
question got parked the first time. The ruling therefore names prd-43 as the instance
that forced it, so the next reader can check whether the reasoning still holds rather
than inheriting a conclusion.

The second is deleting the old justification. It is kept, in a dated **Superseded**
block, because *"No instance exists today"* being wrong is the evidence for the new
reasoning — an absence of pressure is not a reason, and it expired the moment one
appeared.

No test guards prose, so the honest answer is that this is prose and no mutation
touches it. What can be checked was checked, by review, against the world rather than
the document: `prd43` really is open with a null `closed_at`; the committed manifest
carries `43 open -`, so the accuracy paragraph is not aspirational; the archived file
really is under `docs/prds/done/`; and the stale claim survives nowhere as live text.
`docs/prds/` is in the citation law's excluded-as-citing-source tree, so none of this
prose is gated — it was checked by hand or not at all.

## Why it is split out

It was cherry-picked clean off current `main` as `cab67db8`, byte-identical to
`08300ce1` on `459-mirror-case`. The two halves of #459 do not read each other: the
ruling is a decision written down, Part 2 is code.

Part 2 is **held**, not abandoned. Two review seats and the orchestrator
independently found that it introduces the mirror image of the bug it closes — the
code-line mask is computed over raw text before anything is known about comments, so
a code-looking line inside a real HTML comment truncates that comment's own span, and
**a commented-out declaration counts as a real one**. The repair is architectural (a
single left-to-right pass with a third state, rather than two passes), which is why
it is held rather than rushed. Full evidence is on #459.

This half goes out alone because it fixes something false on `main` today, and every
day it waits for Part 2 is another day a blessed PRD carries a false claim.

## Verification

Reviewed by two seats (Opus `xhigh`, Codex `gpt-5.6-terra` `xhigh`) as part of
`57b69fd3`'s review, plus orchestrator checks. **Part 1 returned no findings from
either seat** — every claim it makes was checked against the live tracker, the
committed manifest, the corpus and its three citations, and none was false.

Gates on this branch: `typecheck` **0**, `lint` **0**,
`VITEST_MAX_WORKERS=6 npm test` **0** — 9754 passed, 3 expected-fail, 3 skipped.
**One platform, linux** — Actions was retired on this repo on 2026-09-11, so there is
no macOS or Windows leg behind that.


KelliherL (approved):
## Review of #504 — approve; the ruling holds, and its evidence went false while the PR was open. Dated and pushed as `e16a3d9f`

One commit, one file, prose. Read against prd-56's Open Questions and against the live tracker rather than the document.

### What verified — EXECUTED

- `npm run typecheck` **0** · `npm run lint` **0** · `doc-citation-law` + `prd-location-law` — **green**, before and after my commit.
- **`prd-43` really is archived** (`docs/prds/done/`, absent from the live shelf), and the committed manifest really does carry `43 open -`, so the accuracy paragraph is describing the tree rather than hoping.
- **The superseded block is kept, once**, and Ruling 4 is declared once.
- **The reasoning is right.** A closed-milestone PRD on the live shelf reads as in flight and people act on it — `docs/follow-up-292.md` does exactly that, which I checked while reviewing #485. An archived PRD with a reopened milestone reads as done while a straggler remains: staler, but nobody is misled into treating a shipped ruling as open. One-way is the correct asymmetry and the "what this does not license" paragraph is the half that stops it becoming an excuse.

### The finding — the instance resolved while this PR was open. Dated and pushed as `e16a3d9f`

Ruling 4's evidence is present-tense: *"`prd43`'s milestone reopened (1 open issue against 34 closed, no `closed_at`)"*. Both halves have gone false, and I can date them:

| when | what |
|---|---|
| 2026-09-14 **05:22:03Z** | `#448` closed as its own PR (#499) landed → 0 open, 35 closed, milestone still open |
| **05:30:53Z** | **this PR opened** — the count was already eight minutes stale |
| **05:51:24Z** | the milestone closed → `closed_at` set |

So `prd43` is now an archived PRD with a **closed** milestone: the ordinary state, and no instance at all. Read live just now: `state=closed closed_at=2026-09-14T05:51:24Z open=0 closed=35`.

**The ruling is untouched by this.** It rests on what each direction misleads a reader into, not on the mirror case being live as you read it — and your own framing is why: *"an absence of pressure is not a reason."* The presence of pressure is not the reason either. What needed fixing is the sentence, and it needs fixing here more than anywhere: this is the PRD about a document's status claim going false while nothing checks it, `docs/prds/` is excluded as a citing source, and so no law reads that paragraph. It is the defect class in the document that defines it.

`e16a3d9f` dates the figures to the day they were true, records the resolution beside them with facts a reader can re-check on the tracker, and makes the Open Questions bullet say the instance resolved rather than leaving its own pointer to rot. Prose only; the two laws stay green.

- **The committed manifest now says `43 open -` while the tracker says closed.** That is not this PR's doing and nothing goes red — the law sweeps only the live shelf, which is ruling 4's whole point — but it is precisely the coupling `.swarm/prd-milestones.txt`'s registry entry describes (#443, landed in #492): the next regeneration will move that row, and whoever does it should expect it.
- Holding Part 2 was right, and the reason you give — the code-line mask is computed over raw text before anything is known about comments, so **a commented-out declaration counts as a real one** — is the mirror image of gap 4b in #493, which I reviewed this morning. Two readers of the same corpus, the same prose-versus-code question, and the same failure on the other side of it. Whoever writes the single left-to-right pass should read #493's `classifyLines` first; it already does the three-state walk for the fence and indent halves.

### What I did not run

The full suite — one Markdown file in a directory no law reads; the two laws that read PRDs ran. macOS and Windows. The seat transcripts on #459.

**Approving.** The ruling stands; only its dates moved. Landing is the operator's act.

PR #489, opened 2026-09-14
prd-51: wave 6 is merged, and wave 7 is groomed into two lanes

ciaran-slow:
prd-51's wave-7 grooming, recorded. One commit, one file, 120 insertions to
`docs/prds/prd-51-the-split.md`. Base `main` at `bf2fdea0`; `origin/main` has since moved two
commits, neither touching this document.

Closes no issue — grooming amendments are operator housekeeping, the same shape as #466 which
recorded wave 6.

---

## ⚠️ Condition 1 is NOT met, and this is the gap stated rather than recorded away

**No independent review pass has read `15536961`.** I wrote it, so I cannot be the pass that
checks it, and none has run. AGENTS.md is explicit that a green suite is not condition 1, that
evidence the guarded party authored is not evidence, and that the honest move is to write the gap
into the PR body where a reviewer sees it. That is what this section is.

| commit | review |
|---|---|
| `15536961` | **none — unread by any independent pass** |

Condition 2 *is* met: the operator asked for this specific PR in this session.

**Do not merge this on the strength of the gates below.** They validate an artefact; they say
nothing about whether the amendment is *true*. A reviewer's job here is the second question.

---

## What it records

Wave 6 landed as PR #476, merged `7abfdee2` (#169, #462, #463), plus two review commits. Wave 7
is groomed to **#487, #488 and #436** — two build lanes and one re-homed straggler.

**Why two lanes and not the five the previous wave-7 row implied**, which is the amendment's
substance: three files pin the tracked migration id set independently, in three directories, and
one of them is the router's own test file. Retention (rulings 9, 10) spends this PRD's next
migration, so it cannot share a wave with the router — *the evidence for one lives inside the
other's boundary*. The coupling map already carried two of those entries; the consequence for
sequencing was never written down.

It also books wave 8 as **possibly two waves** rather than promising one, rules the callback's
session to a signed stateless cookie at grooming time (a `sessions` table is a migration, and
wave 7 may not spend one), and records the #436 re-homing and the #171 regrooming.

## What is the sibling case?

The sibling of *"this amendment's claims about wave 7"* is **its claims about every other wave**,
which it restates in a superseded map. A map is exactly the artefact that rots by restatement.
Every row was re-derived rather than copied forward: #476's merge sha was read with
`gh pr view` and confirmed an ancestor of `origin/main` with `git merge-base --is-ancestor`,
not taken from the PR's own "MERGED" label — the distinction AGENTS.md draws for precisely this
reason.

[...]

PR #507, opened 2026-09-14
prd51 w7: the ingest listener is a router, a member returns to a signed session, and /connect says whether a batch ever landed

### LINKED ISSUE #436: prd51 w7: the architecture document cites the journal's decision record

### LINKED ISSUE #487: prd51 w7: the ingest listener is a router, and a member returns from GitHub to a session
## Fence (may touch ONLY)
- `packages/team/src/api/` (all files, including new ones and deletions)
- `packages/team/src/auth/` (all files — the callback needs a session helper beside the
  membership check; unclaimed since #169 closed)
- `packages/team/src/ingest/handle.ts`, `packages/team/src/keys/verify.ts` — **comment only.**
  Both say this restructuring "is wave 7's" in the future tense (handle.ts:54, verify.ts:16), and
  this commit is what makes that false. Ruling 12 wants the words fixed with the code.
- `packages/team/src/index.ts` — the barrel re-exports `startTeamServer` and its types from
  api/main.js, so a change to that module's exported surface reaches this file. Added for that
  structural reason, NOT for the stale "wave 4" note at line 23, which is a historical remark
  about parallel lanes and is not a claim this commit falsifies. Leave that line alone.
- `packages/team/src/bootstrap.ts`, `packages/team/src/bootstrap.test.ts` — **minimal
  reconciliation only.** Added at grooming, not discovered at landing: main.ts:103 calls
  bootstrapTeamStorage, so the boot sequence this lane restructures reaches a module one
  directory up, whose test independently pins the tracked migration id set. This lane adds no
  migration and must not move that pin — the claim exists so a boot-order change does not leave
  a red test outside the boundary.

## Blocked by
Nothing. #169 (the membership check, transport-pure) and #462 (the keys port) both landed in
PR #476, merged `7abfdee2` — the two modules this route restructuring exists to reach.

Sibling lanes hold other surfaces: the `/connect` row lane holds
`packages/web/src/connect/` and `packages/server/src/cli/connect-team.ts`; #436 holds
`docs/architecture.md`. **This lane must NOT add a migration** — see the prohibition below.

**Model:** opus. **Wave:** 7.

## Why this is one issue and not four
Recorded in prd-51's 2026-09-14 amendment: `packages/team/src/api/http.ts` is one route wearing
a conditional (`if (path !== INGEST_PATH)` … 404), and four separate pieces of work each want to
add to it — the GitHub callback, the key-mint surface, the three questions' pages, and the team
server's doctor. That is a single restructuring with four callers, and four lanes discovering
each other inside a 126-line file is a rebase conflict scheduled in advance.

So **this lane builds the router and exactly one caller** — the callback, whose dependency
(`auth/`) is the one already on `main`. The remaining three callers are wave 8's and land on a
router that already exists.

## Prohibition — no migration, and the reason is mechanical
Three files pin the tracked migration id set **independently**, and one of them is inside this
lane's own fence:

- `packages/team/src/migrations/runner.test.ts:245`
- `packages/team/src/bootstrap.test.ts:61`
- `packages/team/src/api/api.test.ts:104`

A migration added here would force an edit to all three, two of which are outside this fence.
More to the point, wave 8's retention work (rulings 9, 10) **spends this PRD's next migration**
(`0006_*`), and it cannot share a wave with this lane for exactly that reason.

**Consequence for the session:** the callback's session is a **signed stateless cookie**
(`node:crypto` HMAC over the member login and an expiry), not a `sessions` table. No migration,
no new dependency. If a server-side session store turns out to be required, that is a finding
that **stops the lane** and goes to wave 8 — it is not a fence widening.

## Definition of done
- `packages/team/src/api/http.ts` dispatches on `(method, path)` through a declared table rather
  than a negated conditional, and an unmatched path is still a 404 with the same body it gives
  today. Adding a route is adding a row.
- The ingest route's behaviour is **unchanged** — same path, same `MAX_BODY_BYTES`, same
  statuses, same refusal strings. `api/api.test.ts`'s existing cases pass unmodified, and that
  is the evidence: a restructuring that needed its own tests rewritten did not preserve
  behaviour.
- `GET /auth/github/callback` exchanges the `code` for a user token, resolves the login, and
  runs `checkOrgMembership` against the installation token — never the signed-in person's token,
  per ruling 8 as amended 2026-09-08. A member gets a session cookie; every other outcome gets a
  refusal naming its reason, and **no outcome leaks the code, the token, or the client secret**
  into a body, a log line, or a redirect URL.
- A non-member is refused with the same `not-a-member` reason vocabulary `membership.ts` already
  returns, and a mint failure is still reported as an explicit error rather than folded into a
  false negative (#169's own law, and the review finding that pinned it).
- **Ruling 12, in this commit:** `http.ts`'s header comment (lines 33-36 today) says key-value
  verification is *"wave 4's and is loudly unimplemented"*. #462 made that false, the operator
  declined the widening on 2026-09-14 precisely so it would land here, and this is the commit
  that makes the file's own words true. The correction is in the same commit as the code, not a
  follow-up.
- The adapter stays transport-only: every decision about what a *request means* lives in a pure
  handler, the way `handleIngest` already does. A route that reaches storage directly from
  `http.ts` fails this.
- Mutation, EXECUTED and recorded on the PR: break the route table so a known path falls
  through, and show the test that should catch it going red. A router whose dispatch cannot fail
  a test is not covered.

## Falsifier
If the callback cannot be built without a server-side session store — and therefore a migration
— **stop and say so on this issue** rather than widening. The wave splits differently, and that
is a grooming decision, not a lane's.

### LINKED ISSUE #488: prd51 w7: /connect says whether a batch was ever acknowledged, and when
## Fence (may touch ONLY)
- `packages/web/src/connect/links.ts`
- `packages/web/src/connect/links.test.ts`
- `packages/server/src/cli/connect-team.ts`
- `packages/web/src/connect/meta.ts`, `packages/web/src/connect/meta.test.ts` — the web side has
  **no shipper facts at all** today (grep for shipper across the connect directory returns
  nothing outside tests) and DoctorReading is declared at meta.ts:322, which links.ts imports.
  A claim on the row's file alone would strand the lane at the type boundary.
- `CHANGELOG.md` — the `[Unreleased]` entry. A new row on /connect is user-visible, and the
  runbook requires the entry; the plan stage caught that it was missing from this claim. No other
  live lane names this file and it is in no registry entry.
- `packages/server/src/cli/connect-team.test.ts` — the source file was claimed and its own test
  was not, which is the shape that strands a lane at landing rather than at build.
- `packages/contract/src/doctor.contract.test.ts` — **minimal reconciliation only.**
  contract-coverage-law.test.ts:152 pins /api/doctor against connect/meta.ts, and the coupling
  map records that these two laws redden together at this seam. This lane adds no route, so it
  is expected to stay untouched; it is claimed so that if it does move, the lane is not blocked
  at landing.

## Blocked by
Nothing. The shipper (#372) and the doctor's shipper check (#387) both landed long ago; this lane
reads what they already record.

Sibling lanes hold other surfaces: the router lane owns the team package's api and auth
directories — **this lane does not touch packages/team at all**, which is what keeps the two
disjoint. #436 owns one document and nothing else.

## Why this reaches into the server package, not only the web row
Ruling 12 requires /connect's `team server` row to read **VERIFIED only when a batch was
acknowledged, with the fact and its timestamp**. The timestamp exists and does not reach the
view. EXECUTED on 7abfdee2:

- packages/server/src/shipper/cursor.ts:70 — `lastAckAt`, *"Local wall clock of the last
  acknowledged batch. Never crosses the wire."*
- packages/server/src/cli/connect-team.ts:392 — `ShipperDoctorFacts` declares enabled, url,
  project, keyPresent, keyMode, maxN, sessionCount, skippedCount, cursorReset, configError.
  **No lastAckAt.**

So claiming the web row alone would strand this lane: the row cannot state a timestamp the doctor
never carries. The CLI doctor reads the same facts (shipperDoctorFacts, cli/doctor.ts:948) and is
claimed with them so the CLI and the route cannot disagree about what was acknowledged.

**This adds no route.** It adds a field to an existing reading, so no prose count moves and
api/route-class-law.test.ts is untouched — confirm that rather than assume it.

## Definition of done
- `ShipperDoctorFacts` carries the last acknowledged batch's timestamp, sourced from the cursor
  that already records it. It keeps the cursor's guarantee intact: `lastAckAt` is local wall
  clock and **never crosses the wire** — this lane surfaces it to the operator, not to the team
  server.
- /connect gains exactly one row, `team server`, using the existing row vocabulary
  (`STATE_WORD` / `STATE_GLYPH`, verified / broken / unproven):
  - **VERIFIED** only when a batch was actually acknowledged, stating the fact and its timestamp.
  - **BROKEN** with the exact remedy, in the house voice the sibling rows already use.
  - **UNPROVEN** until then — and *"connected" never means preconditions passed* (ruling 12's
    own words). A configured-but-never-acknowledged shipper is UNPROVEN, not VERIFIED.
- A shipper that is off is not BROKEN. The `configError` distinction the doctor already makes —
  *"'off' would be the wrong answer, so it is not given"* — is preserved, not flattened.
- Tests query by role and label, never by test id (the web package's convention), and cover all
  three states plus the off case and the config-error case.
- **Mutation, EXECUTED and recorded on the PR:** make the row report VERIFIED for a shipper that
  has never had a batch acknowledged, and show the test that should catch it going red. This is
  the assertion the whole row exists for — if it survives that mutation it is decorative.
- Repetition: reading the facts twice in a row yields the same row and the same timestamp; no
  clock is read at render time.

## Falsifier
If `lastAckAt` turns out to be unreadable at doctor time for a reason this issue did not
anticipate — the cursor is per-session and the row is per-repo, so "which session's ack" is a
real question — **state the choice on this issue before building it**, with the reasoning. The
answer "the most recent across sessions" is the expected one; it is not to be adopted silently.

ciaran-slow:
Closes #487
Closes #488
Closes #436

prd-51 wave 7 — **the listener becomes a router**, and the instrument says whether a batch ever
landed. Three lanes, three commits, base `main` at `456a72ad`.

**Cherry-picked with no conflict, which is the wave's own fence audit**: nothing here touched a
file anything else here touched. #487's 13 files, #488's 9 and #436's 1 are disjoint by
construction.

---

Every lane was verified on a **different model from its builder**, each verifier confirming the
`Co-Authored-By` trailer itself with `git log` rather than accepting the claim, and each running
the gates in its own worktree rather than reading the build report's numbers back.

|---|---|---|---|---|---|
| `da9c046e` | #436 | Opus 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/436#issuecomment-5656869273) | [report](https://github.com/launchpad-26/rhizomorph/issues/436#issuecomment-5657142377) | [**changes** — Fable](https://github.com/launchpad-26/rhizomorph/issues/436#issuecomment-5657207641) → [**ready**, round 2](https://github.com/launchpad-26/rhizomorph/issues/436#issuecomment-5657268789) |
| `28ba8330` | #487 | Opus 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/487#issuecomment-5657536250) | [report](https://github.com/launchpad-26/rhizomorph/issues/487#issuecomment-5657712503) | [**ready** — Sonnet](https://github.com/launchpad-26/rhizomorph/issues/487#issuecomment-5657782202) |
| `4a8fc847` | #488 | Sonnet 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/488#issuecomment-5657531278) | [report](https://github.com/launchpad-26/rhizomorph/issues/488#issuecomment-5657667305) | [**changes** — Fable](https://github.com/launchpad-26/rhizomorph/issues/488#issuecomment-5657760945) → [**ready**, delta](https://github.com/launchpad-26/rhizomorph/issues/488#issuecomment-5657829981) |

**The verify records name the LANE shas, not the shas in this PR.** Cherry-picking rewrites them
and this repo's condition 1 is per-sha, so the gap is closed by content rather than by assertion —
`git patch-id --stable` is identical for all three pairs:

|---|---|---|---|
| #436 | `c64a4cb7` | `da9c046e` | `20c8d5dd5540` — identical |
| #487 | `074f1a8e` | `28ba8330` | `900ebf25f4ec` — identical |
| #488 | `f03d14c2` | `4a8fc847` | `9ae696ec705f` — identical |

What was verified is what is here, byte for byte. (Stated because the first run of this check was
**vacuous** and said so: under zsh an unquoted `$pair` is not word-split, so both sides compared
empty strings and reported IDENTICAL. Re-run with explicit fields and an empty-guard. A check that
cannot fail is the defect this repo ranks second-worst, and it very nearly went into this table.)

---

## What is the sibling case?

- **#487 — the sibling of "a new route is added" is "the one route that already existed."** The
  restructuring's whole risk is changing ingest while claiming not to. The evidence is that
  `packages/team/src/api/api.test.ts`'s pre-existing cases pass **unmodified** — the verifier
  checked the diff for a weakened or deleted assertion and found none. A restructuring that needed
  its own tests rewritten did not preserve behaviour.
- **#487's second sibling — the OAuth start route.** The plan first excluded it, then reversed:
  verifying a CSRF nonce that nothing issues means **no valid state cookie can ever exist in
  production**. The objection (that a start route needs the viewer's `redirect_uri`) was checked
  and is false — the callback URL is registered on the App. It is not a fifth caller; it is the
  other half of the callback.
- **#488 — the sibling of the `ok` branch is every other branch that can carry an acknowledgement.**
  `checkShipper` has three non-fail branches, and a credential with wrong permissions does not
  erase the history of what was already shipped. The cursorReset branch was threaded and tested;
  the keyMode-warn branch was threaded and **not** tested — caught at verify, now both.
- **#436 — the sibling of "a new citation resolves" is "the counts already in that file still
  resolve."** `api/route-class-law.test.ts` reads two prose counts out of `docs/architecture.md`.
  The append is at EOF, neither line is in the diff, and the new text matches none of the law's
  twelve registered patterns.

- **#487 — none of the 24 the builder ran, and none of the two the verifier invented** (removing
  the expiry check outright; removing the `not-a-member` refusal so any resolved identity mints a
  session). Both were caught by existing tests.
  **Two of the builder's own assertions were vacuous and it found them by running the mutations
  rather than reasoning about them**: A6 compared the 404 enumeration to `TABLE` itself (true
  whatever the table says), and A8 snapshotted `TABLE` after an earlier test had already triggered
  its in-place sort, so re-sorting was a no-op. Both strengthened before commit.
- **#488 — three survived at first**, all found by mutations the builder had not thought of:
  a reduce returning the *first* actor (the fixture made `session-aaa` both first and maximum, so
  a test titled "never the first or the alphabetically-last" could only fail on one of them); the
  keyMode-warn branch losing its ack threading (80/80 green); and the row treating `warn` as
  BROKEN (71/71 green). That last one has a real operator consequence — a `0644`-credential
  shipper that **has** been acked would read BROKEN and be told to re-enter the key when the
  remedy is `chmod 600`. All three now redden, re-run by the verifier, and the **new fixtures were
  themselves checked for vacuity**: a `warn → verified(provenNow)` short-circuit fails on the
  timestamp, so the fixture asserts the ack fact rather than merely "not broken".
- **#436 — every mutation, and that is the honest answer.** `docs/prds/` and the citation rules
  mean no law checks a backticked path or a `#NNN` in that text. Two mutations were run to prove
  the *surrounding* laws are live (a dead path reddens `doc-citation-law`; a planted
  `Twenty reads are gated:` reddens `route-class-law`), and the citation itself was checked by hand.
- **The one that still survives**: #488's UNPROVEN default drops its `doctorNote` unnoticed —
  a never-acked shipper's row says *why* today, and nothing would catch it if it stopped. Named by
  the verifier as minor, left as coverage debt rather than fixed silently.

## Fence widenings — four, all recorded BEFORE the change

Every one was found by a **plan or a verify pass reading the work**, never by the lint. Recorded on
the issues at grooming time, before any code:

| issue | added | why |
|---|---|---|
| #487 | `ingest/handle.ts`, `keys/verify.ts` | both say this restructuring *"is wave 7's"* in the future tense; ruling 12 wants the words fixed by the commit that falsifies them |
| #487 | `index.ts` | the barrel re-exports `startTeamServer` from the module being restructured |
| #487 | `bootstrap.ts` + test | `api/main.ts:103` calls `bootstrapTeamStorage`, so the boot sequence reaches a migration-set pin one directory up |
| #488 | `CHANGELOG.md`, `connect-team.test.ts` | a user-visible row needs the entry; the source file was claimed and its own test was not |
| #488 | `meta.ts` + test | `DoctorReading` is declared there and the web had **no** shipper facts at all |

**One correction worth carrying forward**: #487's plan justified `index.ts` by a future-tense
wave-7 comment. It has none — its line 23 says *"a surface wave 4 shares"*, which is **true**. Taken
at face value the builder would have "fixed" a correct sentence and cited ruling 12 for it.

The pattern behind all five rows is now recorded in prd-51: **`fence-lint.sh` proves fences do not
overlap each other; it cannot prove a fence is large enough for its own issue.** Only reading the
work does that.

## Plan deviations — two, both gate-forced

- **#488** dropped a `(#488)` self-citation from its `CHANGELOG.md` bullet: the citation-ceiling law
  forbids a fresh, unmerged issue citing itself, and the live ceiling derives to 486. EXECUTED by
  the verifier — restoring it reddens the law.
- **#487** dropped backticks around `RZ_TEAM_SESSION_SECRET` and `Referer` in two module headers:
  `doc-citation-law` reads backticked tokens as symbol citations and its `[not-ours]` baseline is
  outside the fence. EXECUTED by the verifier — restoring them reddens with exactly those two
  violations. Prose-only, no behaviour change, no baseline edit.

## The defect this wave was specifically hunting

GitHub refuses a **reused** OAuth `code` with **HTTP 200** and a body carrying
`{"error":"bad_verification_code"}`. An implementation branching on `status === 200` alone mints a
second session on every replay. This repo has already shipped one defect of exactly that shape — a
`302` treated as handled because the stub returned the status directly while the real `fetch`
followed the redirect (#169, caught at review). The callback reads the **body**, not the status;
mutation M3b reddens exactly three tests, re-run by the verifier.

## Gates — on the assembled wave

Run on this branch with all three commits present, the first time they coexist, and the first time
they meet a `main` that has moved **46 commits** past the base each lane was gated on:

- `VITEST_MAX_WORKERS=6 npm test` — **517 files, 9809 passed**, 3 expected-fail, 7 skipped, exit 0
- `npm run typecheck` — clean
- `npm run lint` — clean, 1099 files
- `npm run build` — clean
- boot smoke — `GET /` → **200** `text/html`, `GET /api/meta` → **401**, the expected gated read

Node v22.22.2, darwin/arm64. **This is one platform's verdict and not "CI passed"** — all four
workflows are disabled on this repo, so this PR carries no check runs. `scripts/gate.sh` was **not**
run: running it *is* the landing, and that is the operator's step.

## What this wave does not close, stated so it is not assumed

- **The three questions' pages, the team server's doctor route, and the key-mint surface** are
  wave 8's. They each need a route, and the route table is what this wave builds — a stack, not a
  bundle.
- **Retention (rulings 9, 10) could not travel with this wave.** It spends prd-51's next migration,
  and **four** files pin the tracked migration id set — including `api/api.test.ts`, which is this
  wave's own. Two of the four (`runner.test.ts`, `schema-law.test.ts`) are in no coupling entry.
- **Both routes answer 503 in production today.** The deployment sets no `RZ_TEAM_GITHUB_*`
  variables, so the callback is honestly unconfigured until wave 8 wires it. Found by the plan
  stage and stated rather than discovered.
- **A stateless cookie cannot be revoked before it expires.** Wave 8's read-only role must re-run
  `checkOrgMembership` rather than trust it — written into #487's plan as the contract wave 8
  inherits, not left to be discovered.
- **No ADR.** #487 ruled one out with reasoning (a numbered record costs two more fence widenings,
  a third package's law, and a wave-level sequencing claim) and left **0051** unclaimed with a
  one-sentence fallback drafted if a reviewer disagrees.
- **`index.test.tsx` still iterates a hard-coded seven-id list**, so nothing asserts #488's eighth
  row reaches the DOM. Out of every lane's fence; named by the verifier, owed to the wave.


KelliherL (approved):
## Review of #507 — approve; nine mutations on the security invariants, five of them mine, and every one reddens

prd-51 wave 7 is the wave I flagged at the review of #476 and #489 as the one where the session design stops being a ruling and becomes code. So I read it as a security change: the three consequences I named on #489 — the OAuth `state` nonce, the signing key's home, and revocation before expiry — are the three things I went looking for, and all three are answered in the tree.

Read commit-by-commit (`da9c046e` #436, `28ba8330` #487, `4a8fc847` #488) in a clean clone at node 22.23.2 on Linux.

### The security invariants, mutated — all nine red

| # | mutation | result |
|---|---|---|
| 1 | the code exchange branches on **status alone**, ignoring the error body | **red** — *"C9b — an error body is a refusal even when a token-shaped field sits beside it"* |
| 2 | the **CSRF state check removed** entirely | **red** ×2, incl. the whole-flow-over-a-socket case |
| 3 | the state nonce **accepted without comparing** it to the cookie | **red** ×4 |
| 4 | the session **expiry check removed** | **red** ×3, incl. *"both sides of the expiry, against an exp COMPUTED from SESSION_TTL_MS"* |
| 5 | **any resolved identity mints a session** (membership gate removed) | **red** ×2 |
| 6 | the signature verified **after** the payload is parsed | **red** — *"B3 — one flipped character in the payload segment is bad-signature"* |
| 7 | the cookie loses **HttpOnly** | **red** ×4, incl. *"D3 — the cookie ATTRIBUTES survive the wire"* |
| 8 | **SameSite=Lax → None** | **red** ×4 |
| 9 | the session key **stops depending on the client secret** | **red** ×3, incl. *"B2 — a cookie signed under a key derived from a DIFFERENT client secret is bad-signature"* |

1, 4 and 5 are yours and the verifier's; **2, 3, 6, 7, 8 and 9 are mine**, and I picked them to be the ones a careful implementation usually leaves unpinned. Cookie attributes surviving the wire, and a key that must actually derive from the secret, are not assertions I expected to find already written.

### What else verified — EXECUTED on your head `4a8fc847`

- `npm run typecheck` **0** · `npm run lint` **0** (1099 files) · the wave's own suites — **30 files green**.
- **The restructuring claim is exact.** `packages/team/src/api/api.test.ts` loses **0** `expect(` lines and **0** `it(` cases in this wave; the only removed line is an import. A router that needed the ingest tests rewritten would not have preserved behaviour, and this one did not need them touched.
- The ingest and keys suites are green on this head, so wave 6's refusal path is intact under the new route table.
- **#436 is append-only at EOF**: 17 lines added, 0 removed from `docs/architecture.md`, and `route-class-law` + `doc-citation-law` are green.

### Reading the flow, since the mutations only prove what is tested

The ordering in `handleGithubCallback` is the part that has to be right, and it is: configured → code present → **state verified before the code is spent** → exchange → identity → **membership on the installation token** → only then `signSessionCookie`. The state cookie is cleared on success and on every refusal after the state check — and **not** on `bad-state`, which reads deliberate to me: clearing there would let anyone who can make a browser hit the callback destroy a legitimate in-flight sign-in.

Three things I checked because they are where this shape usually goes wrong:

- **The signature is verified before the payload is parsed**, and the comment says why — a forged payload must not steer the parse it has not earned. `timingSafeEqual` behind a length guard, `v: 1` pinned, `uid` carried as well as `sub` so a renamed login cannot become a different person.
- **`deriveSessionKey` uses `hkdfSync` off the client secret** with a distinct salt and info, and **throws** on an empty secret rather than silently keying on nothing. That is the option I offered on #489 as the one that stays inside the fence, and it is better than the `RZ_TEAM_SESSION_SECRET` variable the module header records as rejected — no new deployment step to forget.
- **`operatorNote` reaches `deps.onError`, never the body.** That is the seam I added at the review of #476 for the ingest 503, used here for the membership error, which is the right generalisation of it: the caller at that moment has not authenticated, so the server's own sentence is the operator's.

- **`detail: { reason: membership.reason }` puts the refusal reason on the wire**, and one of its values is `credential-rejected` — the app's own installation token being refused. A real member whose sign-in fails because the *server's* credential broke is told "not a member", with the reason as the only tell. That is #169's ruling working as designed (verdict same, reason distinguishable) and I am not asking you to change it; I note only that `credential-rejected` and `network-error` are facts about the server, and a 403 is the one response an unauthenticated caller definitely reads.
- **Both routes answer 503 in production today** and you say so. Worth making sure wave 8's doctor row surfaces that, or the first person to try signing in learns it from a 503 body.
- The two coverage debts you name — #488's UNPROVEN `doctorNote`, and `index.test.tsx`'s hard-coded seven-id list not asserting the eighth row reaches the DOM — are both real and both named. The second is the one I would file rather than carry.

### What I did not run

macOS and Windows. A live GitHub OAuth round trip (no App configured, and the tests stub `fetch` at the seam). `main` has not moved from this branch's base, so the leg below is the union:

```
  Test Files  517 passed (517)
       Tests  9813 passed | 3 expected fail | 3 skipped (9819)
```

**Approving.** The three consequences I named when the ruling was groomed are all answered in code, and the invariants that matter are pinned hard enough that nine deliberate breaks all redden. Landing is the operator's act.

PR #510, opened 2026-09-14
prd56: a comment cannot cross a fence delimiter (#459 part 2)

gabriel-canaan:
Part of #459 — **does not close it.** #459 covers two halves; Part 1 (ruling 4) landed
separately as `cab67db8` via PR #504. This is Part 2, the parser. The issue stays open
only if something remains; if this lands, it can be closed against this commit.

## What it fixes

prd-56 ruling 2 requires a declared shelf exemption that a **reader** can find. Two
contexts made the reader disagree with the rendered document, both **fail-closed** —
they convicted declarations a person can plainly see:

- an unterminated `<!--` **inside a fence**, with the declaration after the fence;
- a **terminated** comment whose span crosses a fence's closing delimiter, destroying
  the delimiter so fence state never reset.

Root cause: comment blanking ran over the whole text before any fence was known.

## The design — one pass, three states

The comment grammar and the fence grammar interleave, so neither can be resolved
first. `computeLineStates` is now a single left-to-right scan with `normal /
in-fence / in-html-block`, where a fence opens only outside an HTML block and vice
versa.

**This matters because the obvious alternative was tried and rejected.** An earlier
attempt (`57b69fd3`, abandoned, branch deleted) used two independent passes — a
code-line mask over raw text, consumed by a comment segmenter. Two review seats and
the orchestrator independently found it introduced **the mirror image of the bug it
closed**: a code-looking line inside a *real* comment truncated that comment's own
span, so a commented-out declaration counted as a real one. That history is why the
review below probed both directions rather than only the reported one.

## What is the sibling case?

**Fixing one direction of an ordering problem and opening the other** — which is
literally what the rejected attempt did, and what round 1 of this branch's review
caught again in a new spelling. Both directions are now tested: code syntax inside a
real comment, and comment syntax inside real code.

Second: fixing the *cell* rather than the *class*. Round 1's finding was demonstrated
with a paragraph continuation; the repair asks the general question — *is this line
indented code, or a continuation?* — via an `inParagraph` state, not a `<!--` check
inside the indent branch. Verified against spellings the finding never named: a
**list-item continuation** and a **tab** in place of four spaces both behave
correctly.

Round 1's second finding was that the changed boundary had **no assertion at all** —
reverting it left all 48 tests green. That is now pinned in both directions, verified
by mutation rather than by reading the test name:

```
remove the paragraph guard  →  exit 1
  × #459 round 2: a 4+-indented <!-- that CONTINUES an open paragraph is a real
    comment opener, not code — the declaration inside stays concealed
  Tests  1 failed | 48 passed (49)
```

## Review

Two seats (Opus `xhigh`, Codex `gpt-5.6-terra` `xhigh`) on isolated worktrees, plus
orchestrator probes. Recorded in full on #459.

**The design was confirmed, and the reviewers went well past the diff.** Opus
enumerated **55 interleaving forms** against the CommonMark 0.31.2 reference parser
(installed into a scratch prefix, counting a declaration visible only when its text
survives outside `html_block` / `code_block` / `html_inline`) and fuzzed **60,000
documents**. Result: 12 rows fixed, **zero previously-correct fence↔comment rows
broken**. Codex's independent table agrees on every fence↔comment case and confirms
the rejected two-pass shape was not reintroduced.

**One blocker was found and is fixed here:** a `<!--` indented 4+ columns no longer
opened a comment. Correct when the line is genuinely indented code; wrong when it
continues a paragraph, because indented code cannot interrupt a paragraph — the
comment is real and conceals what follows. That was **fail open**, the direction this
law exists to prevent. Both the behaviour and its pin are in this commit.

**One reported blocker was NOT this branch's fault.** Codex flagged `<!-->` and
`<!--->` — complete comments per CommonMark that this reader treats as unterminated,
hiding a visible declaration. Real, but verified byte-identical on the base, so
**pre-existing**; it wants its own issue rather than a hold on this one. Opus
classified it the same way independently.

**The amended head has not been through an independent pass.** The seats read
`ed8d57f4`; this commit is the repair of their finding, and it has been read only by
the session that requested it.

Stopping was a judgement, with the case both ways. For: the expensive question — is
the three-state design sound, and does it fail in either direction — was settled by 55
enumerated forms and a 60k-document fuzz, and this round changed one boundary
condition inside that confirmed design, with the pin verified by mutation. The round
budget on this file is also deep into override territory across #484 and #459, and
every round since the first has returned narrower siblings of the last repair.
Against: an unpinned boundary is exactly what let round 1's regression through.

**The reviewer of this PR is the first independent read of this commit.**

## Gates

Orchestrator, full worktree: `typecheck` **0**, `lint` **0**,
`VITEST_MAX_WORKERS=6 npm test` **0** — 517 files, 9820 passed, 3 expected-fail, 3
skipped. Corpus check by both seats: all 56 documents on both shelves, base vs head,
**identical dispositions** — nothing changed shelf or exemption status.

Not reachable in today's corpus: zero PRDs on either shelf contain comment syntax or a
tab. This is correctness work, not a live fault.

**One platform, linux.** Actions was retired on this repo on 2026-09-11, so there is
no macOS or Windows leg behind any of it.


ciaran-slow (approved):
## Approved. The declared gap is now closed, and the design holds up under independent measurement.

You flagged that the amended head `b62d221d` had never had an independent pass. This is it — I read the
commit on a clean clone off current `origin/main`, re-derived your claims rather than reading them, and
measured the reader against the CommonMark 0.31.2 reference parser myself.

**One gap found and fixed in `bc32d87b`** (pushed to your branch, fast-forward, your commit untouched).

---

**The design is right, and the direction is right.** This is the measurement that sank `57b69fd3` —
that attempt moved net concealment the wrong way (974 → 1,016 fail-open). This one does not.

EXECUTED — 54,936 unique generated documents (seeded `mulberry32`, fragments drawn from fence/comment/
indent/blockquote/list spellings), each scored against `commonmark@0.31.2` rendered HTML with code
blocks and comments stripped, comparing **verdicts** (exempt/convict):

| | base `1e709772` | head `b62d221d` | |
|---|---|---|---|
| agrees with CommonMark | 49,667 (90.41%) | **52,405 (95.39%)** | **+2,738** |
| **fail-OPEN** — exempts a *hidden* declaration | 520 | **416** | improved |
| **fail-CLOSED** — convicts a *visible* one | 4,749 | **2,115** | more than halved |

Both directions improve. That is the claim the previous attempt could not make.

**Your three mutation claims re-derived, not trusted** (`npx vitest run packages/server/src/prd-location-law.test.ts`):

- remove `&& !inParagraph` → **49/49 → 48/49**, reddening exactly the round-2 test. Your figure, confirmed.
- `if (false && inFence)` → **45/49**, 4 red including table row 38.
- `if (false && inComment)` → **44/49**, 5 red including the sibling-case test. The mirror-direction
  protection is genuinely load-bearing, not asserted-by-doc-comment.

**The two fixture edits are legitimate corrections, and this was the thing I most expected to find wrong.**
Changing a pre-existing fixture so it keeps passing is the classic shape for hiding a regression, so I
put both through the reference parser rather than taking the commit body's word:

```
OLD gap-3 fixture  '> **Shelf exemption:** x\n>     this indented continuation…'
  → <blockquote><p>…x\nthis indented continuation…</p></blockquote>   has <pre><code>: false
NEW gap-3 fixture  '> **Shelf exemption:** x\n>\n>     this indented continuation…'
  → <blockquote><p>…x</p><pre><code>this indented continuation…</code></pre>   has <pre><code>: true
```

Same result for the tab twin. Both old fixtures rendered as one continuing paragraph with no code block
at all — they never tested detached indented code. The blank quoted line is load-bearing exactly as you
say, and neither assertion moved. Correct call, correctly explained.

**Corpus, EXECUTED myself:** 53 files matching `PRD_FILE_RE` across both shelves, **0** containing `<!--`,
**0** containing a tab, **0** dispositions changed base → head. (Your "56 documents" and my "53" are both
right — 56 `.md` files sit on the shelves, 53 of them name-match the law's regex.)

**Gates, on this tree with my commit included — macOS, node v22.23.2:** `typecheck` rc=0 · `lint` rc=0
(biome, 1099 files) · `VITEST_MAX_WORKERS=6 npm test` rc=0 — 517 files, 9817 passed, 3 expected-fail,
7 skipped. You ran linux, so between us this now has two platforms. Still nothing behind Windows.

---

## The one finding — fixed in `bc32d87b`, not handed back

**The `indent.credit < 4` guard on the fence-open branch shipped with no assertion.**

Round 2's paragraph fix is what made that guard reachable. Before it, a 4+-indented line returned early
as code and never saw the fence-open check, so the guard was unreachable belt-and-braces. Now a
paragraph-continuing line at 4+ columns falls through to it, and the guard is the only thing stopping
such a line from opening a fence.

EXECUTED mutation against `b62d221d` — `if (indent.credit < 4) {` → `if (true) {`:

```
Tests  49 passed (49)
```

Green, while three shapes flip from EXEMPT to CONVICT — a plainly visible declaration convicted because
an indented run wrongly opened a fence over it:

```
'Some open prose that this indented run continues.\n    ```\n**Shelf exemption:** …'
'> A quoted note that the next line continues.\n>      ~~~~\n> **Shelf exemption:** …'
```

commonmark.js 0.31.2 renders all three as lazy paragraph continuation with no fenced block, so EXEMPT is
correct and the guard is doing real work. Your doc comment already states the rule ("a fence can never
open at 4+ columns regardless of paragraph state") — it just had nothing holding it.

`bc32d87b` pins it in both directions: backtick plus an unlisted tilde-in-a-blockquote spelling at a
different indent width, a control detaching the same opener with a blank line (so the assertion turns on
the *guard*, not on "an indented fence never opens"), and a control at 3 columns where a fence opener
*is* real and the declaration inside is correctly concealed. **CERTIFIED: 50/50 green; under the
mutation 1 failed / 49 passed, reddening only the new test.**

---

## Offered, not asserted — your call, and I do not think it holds this PR

**A 4-column `<!--` inside a list item's content is read as indented code, and that is fail-open.**

EXECUTED, `'  - nested item\n\n    <!--\n**Shelf exemption:** …'` — CommonMark puts that opener at the
list item's content indent, so it starts a real HTML block and the declaration is **hidden**; base
convicts (correct), head exempts (**fail open**).

I am not holding the PR on it, for three reasons, and I would rather say them than quietly drop it:

1. **The cause is list-blindness, which is pre-existing** — `classifyLines` tracks blockquotes but has no
   notion of list content indent, on base as much as here. Base only lands on the right answer by
   accident of the very bug you removed: its whole-text blanking convicted everything downstream. On the
   sibling shape where the comment really is indented code, that same accident makes base wrong, and
   those cases are most of the 4,749 → 2,115 fail-closed improvement.
2. **Net fail-open still improves** (520 → 416). The class is 34 documents in 54,936 — 0.06% — and 17 of
   the 34 carry a list marker.
3. **Unreachable in today's corpus:** zero PRDs on either shelf contain comment syntax.

That is the same shape, and the same disposition, as the `<!-->` / `<!--->` case you correctly declined —
real, pre-existing, wants its own issue rather than another round here. Say the word and I will file it
with the reproducers and the fuzz harness attached; I would not fold it in, since list-content-indent
tracking is architectural and this issue's scope is fence↔comment ordering.

---

## Verdict

Approved. `bc32d87b` is a test-only addition, so nothing about the reader's behaviour changed under
review — but per the runbook it is still a fix reviewed by the person who wrote it, so flag it if you
disagree with the shape.

**Merging is the operator's step, not mine** — I have not merged and have run no landing gate.

PR #511, opened 2026-09-14
prd51 w8: a storage capability is a port, and adding one touches no other port's files

### LINKED ISSUE #509: prd51 w8: a storage capability is a port, and adding one touches no other port's files
## Fence (may touch ONLY)
- `packages/team/src/storage/` (all files, including new ones and deletions)
- `packages/team/src/migrations/schema-law.test.ts` — **pointer correction only**, one line. It
  reads contract.ts as TEXT and requires a literal method signature that this split relocates.
- `.symbol-citation-baseline` — **pointer correction only**, one row. Its `TimeZone` entry names a
  storage path whose hazard comment this split relocates.

That is the whole boundary, and keeping it that way is half the point — see the Definition of
done. This lane adds **no migration** and **no dependency**.

## Blocked by
Nothing. Wave 7 landed in PR #507, merged `1e709772`.

No sibling lanes: this is a **keystone wave with exactly one issue and zero claimants**, the same
shape wave 1 took. Nothing else in prd-51 may be dispatched until it lands, because everything
else in the programme collides on the files it restructures.

**Model:** opus. **Wave:** 8.

## Why this wave exists at all — measured, not preference

Wave 7 turned one hard-coded route into a table so that *adding a route is adding a row*. Grooming
wave 8 found the same shape one layer down, and this time it blocks four issues rather than three.

`TeamStorage` is a single 11-method interface in one file, with exactly two implementations, also
one file each:

- `packages/team/src/storage/contract.ts` — 284 lines
- `packages/team/src/storage/fake.ts` — 261 lines
- `packages/team/src/storage/postgres.ts` — 450 lines

Every remaining piece of prd-51 needs to add methods to it: retention (rulings 9, 10) needs a
partition drop and a ceiling read; the three questions need projection reads; the key-mint surface
needs revoke and list. Three lanes adding methods means three lanes editing the same three files —
a rebase conflict scheduled in advance, exactly what a wave is defined to exclude.

The arithmetic decides it. Serialising those four issues across four waves pays the per-PR queue
toll **four times** (measured median 20.9 h, 81% of cycle time). This keystone plus one four-lane
wave pays it **twice**.

**Splitting the interface alone does not achieve this, and a plan that stops there has missed the
point.** If `fake.ts` and `postgres.ts` stay single files, a wave-9 lane adding one port method
still edits both, and the lanes still collide. The implementations have to split with the
interface.

## Definition of done

- `TeamStorage` is composed from per-concern ports — one **file** per port, not merely one type —
  covering at least migrations, settings, events and ingest keys. The names are the builder's to
  choose; the file-per-port rule is not.
- **`TeamStorage` keeps its name and its exact shape.** It stays the intersection of its ports, so
  every one of the 21 files that imports it today compiles untouched. This is what keeps the
  boundary at one directory.
- **Each implementation splits the same way** — a fake and a Postgres module per port, composed
  into the objects `createFakeStorage` / `createPostgresStorage` return. A port's fake and its SQL
  live beside that port's interface, not in a shared 450-line file.
- **No behaviour changes.** Every existing test in `packages/team` passes **unmodified**. A
  restructuring that needed its own tests rewritten did not preserve behaviour, and here that is
  the entire deliverable. If an existing test must change, stop and say why on this issue.
- `no-sql-outside-storage-law.test.ts` still holds, restated at equal or greater strength — it may
  not be weakened to accommodate the new file layout.
- **The claim this wave makes is proved, not asserted.** Add a throwaway probe port (one method,
  one fake, one SQL stub), and show that doing so touched **only new files** — no edit to any
  existing port's interface, fake, or SQL. Record the file list. Then remove the probe and show the
  suite is green without it. That demonstration is the acceptance criterion; a split that cannot
  show it has not delivered the thing wave 9 depends on.
- Mutation, EXECUTED and recorded: break one port's composition so a method is missing from
  `TeamStorage`, and show `tsc` fails. If the composed type can lose a method silently, the
  composition is not load-bearing.

## Prohibitions

- **No migration.** Four files pin the tracked migration id set independently —
  `migrations/runner.test.ts`, `migrations/schema-law.test.ts`, `bootstrap.test.ts` and
  `api/api.test.ts` — and the last two are outside this boundary. Wave 9's retention lane spends
  the next migration. EXECUTED during the review of #489: planting one probe migration reddens 4
  files / 8 tests.
- **No new dependency.** `packages/team` declares exactly two.
- **No new capability.** Not the retention port, not the view port, not the mint port. Those are
  wave 9's lanes and each owns its own port file — that is the whole product of this wave. Adding
  them here recreates the collision one file over.

## Falsifier
If splitting the implementations turns out to require changing a consumer outside
`packages/team/src/storage/` — most likely `api/main.ts` or `bootstrap.ts`, which construct the
storage — **stop and record it on this issue before widening.** That is a real possibility and a
legitimate widening; it is not a thing to take quietly, because the narrow boundary is the claim
this issue is making about its own blast radius.

ciaran-slow:
Closes #509

prd-51 wave 8 — **a storage capability is a port**. One commit, base `main` at `1e709772`.

A **keystone wave with exactly one issue and zero claimants**, the same shape wave 1 took. It
ships no user-visible behaviour. Its entire product is that wave 9 can run four lanes in parallel
instead of four waves in sequence.

---

|---|---|---|---|---|---|
| `ca3bfbbd` | #509 | Opus 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/509#issuecomment-5671143822) | [report](https://github.com/launchpad-26/rhizomorph/issues/509#issuecomment-5671601635) | [**ready** — Fable](https://github.com/launchpad-26/rhizomorph/issues/509#issuecomment-5671727300) |

The verify record names the **lane** sha `f0e296c3`; cherry-picking rewrote it to `ca3bfbbd`. This
repo's condition 1 is per-sha, so the gap is closed by content:
`git patch-id --stable` is `23994a56e6e9` on both sides, and `git diff f0e296c3 ca3bfbbd` is
**empty** — the assembled tree is byte-identical to the verified tree, since `origin/main` had not
moved (0 commits since the lane's base).

Also recorded plainly: **this is a degenerate assembly.** The wave has one issue, the base did not
move, and the tree is provably the same bytes. The gates below were still re-run on this branch
rather than carried over by inference, because "provably the same bytes" is how a green mark
whose evidence has moved gets justified.

---

## Why this wave exists — the arithmetic, not taste

Wave 7 turned one hard-coded route into a table so *adding a route is adding a row*. Grooming
wave 8 found the same shape one layer down, blocking four issues instead of three.

`TeamStorage` was one 11-method interface in one file, with two implementations of one file each —
`contract.ts` (284 lines), `fake.ts` (261), `postgres.ts` (450). Retention (rulings 9, 10), the
three questions, and the key-mint surface all need to add methods to it. Three lanes, three shared
files, a rebase conflict scheduled in advance.

Serialising those four across four waves pays the per-PR queue toll **four times** (measured
median 20.9 h, 81% of cycle time). This keystone plus one four-lane wave pays it **twice**.

**Splitting the interface alone would have achieved nothing** — if the implementations stayed
single files, a wave-9 lane adding one method still edits both. Five ports now each own
`ports/<port>/{port.ts, fake.ts, sql.ts}`; `contract.ts`, `fake.ts` and `postgres.ts` are
composition roots. `TeamStorage` keeps its name as the intersection, so all 21 importers compile
untouched.

## The claim, demonstrated rather than asserted

The acceptance criterion was a demonstration, and **the verifier re-ran it independently** and got
the builder's result exactly:

```
 M packages/team/src/storage/contract.ts
 M packages/team/src/storage/fake.ts
 M packages/team/src/storage/postgres.ts
?? packages/team/src/storage/ports/probe/{port,fake,sql}.ts
 3 files changed, 13 insertions(+), 1 deletion(-)
```

Existing port files edited: **zero**. Test files edited: **zero**. Outside `storage/`: **zero**.

**Stated plainly rather than rounded up: three composition-root files change, 13 insertions and 1
deletion. It is not "only new files."** That was ruled on the issue before the build — a
composition root must reference a new port somehow, and both alternatives were worse (a
`UnionToIntersection` registry, and side-effect registration, which destroys the `tsc` mutation
proving the composition is load-bearing). The claim wave 9 needs is *no **other port's** files*,
and that is what holds.

## What is the sibling case?

**The sibling of an importer is a reader of the file's text, and it is what the plan's survey could
not see.** The plan enumerated every importer of `storage/` — 10 modules, 5 module paths, 11 type
names, all sixteen pinned, and that survey was correct. The falsifier still fired at build time on
two laws that read a storage path as a **string**:

- `packages/team/src/migrations/schema-law.test.ts` reads `contract.ts` and greps a literal method
  signature;
- `.symbol-citation-baseline` names a path in a `[not-ours]` row, read by `doc-citation-law`.

A re-export is invisible to both. **The lane stopped at the fence rather than widening itself**,
and refused two in-fence "fixes" that would have gone green without being right: re-declaring the
method in `contract.ts` (breaking the intersection constraint) or putting either literal in a
comment (satisfying a grep law with prose). Widening granted as two pointer corrections, 3 lines,
recorded on the issue **before** the change. Neither weakened an assertion — the verifier confirmed
the signature is asserted at its new home and the baseline row names where the hazard comment
actually went.

- **The composition**: three mutations, all biting, re-run by the verifier. Dropping a spread →
  1× TS2322. Dropping a delegation → 1× TS2420 plus 93 consumer errors. Dropping `& IngestKeysPort`
  from the alias → 18 errors across exactly six files **outside** the fence including
  `keys/verify.ts` — the compile-time proof the intersection is load-bearing.
- **The probe caught a defect in the builder's own new law.** `composition-law.test.ts` case 5
  asserted `toEqual` against the composed adapter, pinning it to exactly five ports — **a sixth
  reddened it, so all four wave-9 lanes would have hit it.** Replaced with a subset check and
  amended in. That is the acceptance criterion catching the thing it exists to catch.
- **That loosening is not vacuous**, and the verifier proved it with three mutations of its own
  rather than accepting the fix. The sharpest: a complete but **unwired** port directory leaves
  `tsc` green while composition-law case 2 goes red — the failure the type system structurally
  cannot see. And a settings port also declaring `close` spreads *last*, so the shadowing would be
  invisible to every behaviour test.
- **The one that still survives**: `SQL_MODULE_IMPORT_RE` lacks the `require(` form its clause-1
  sibling has. No live path in an ESM package; named by the verifier as worth-later, not fixed
  silently.

## The law restatement — the only existing test file permitted to change

`no-sql-outside-storage-law.test.ts` hard-codes `SQL_MODULE = storage/postgres.ts` in six places,
so no restructuring of this directory could leave it untouched. **The issue's own DoD said "every
existing test passes unmodified" and was wrong as applied to it** — that was written about
behaviour tests. Ruled on the issue before the build: restate with compensating strengths, never
weaken, and it is the only such file.

The verifier checked it **clause by clause against the base**, and its judgement was *genuinely
stronger, not differently shaped*: all four detector regexes and `stripTsComments` byte-identical,
cases 8 → 12, and each compensation tested by mutation. It also found a strengthening the build
report undersold — **`postgres.ts` is no longer exempt from anything but clause 5**.

Every other test file is byte-identical; `fake.test.ts` and `postgres.test.ts` unmoved.

- `VITEST_MAX_WORKERS=6 npm test` — **518 files, 9821 passed**, 3 expected-fail, 7 skipped, exit 0
- `npm run typecheck` — clean
- `npm run lint` — clean, 1117 files
- `npm run build` — clean
- boot smoke — `GET /` → **200** `text/html`, `GET /api/meta` → **401**

Node v22.22.2, darwin/arm64. **One platform's verdict, not "CI passed"** — all four workflows are
disabled on this repo, so this PR carries no check runs. `scripts/gate.sh` was **not** run:
running it *is* the landing, and that is the operator's step.

## What this wave does not close

- **No migration, no new dependency, no new capability.** Not the retention port, not the view
  port, not the mint port — each is a wave-9 lane and adding one here would have recreated the
  collision one file over.
- **A reporting trap worth the next reader's time**: `tsc` orders errors by path, so `head`-ing its
  output hid two of the three mutation results and read as *"the plan mispredicted"*. Two false
  findings nearly reached the build report. The honest probe is
  `grep -oE "error TS[0-9]+" | sort | uniq -c`.
- **The issue's DoD named `createFakeStorage`, which never existed** — the fake is a
  `FakeTeamStorage` class across 7 construction sites. A DoD naming a symbol that is not there is
  the kind of thing that gets "satisfied" by inventing one; caught by the verifier, criterion met
  by the class composition.
- **ADR-0043's prose** names `postgres.ts` as the SQL boundary at lines 87 and 100 and is now
  imprecise. No citation law breaks — the path still exists — so it is routed to the doc-sweep
  wave rather than widened into this lane.


KelliherL (approved):
## Review of #511 — approve; the composition proofs re-derive to the digit, and the collision case was checking a list rather than the ports

A keystone wave's only product is a claim, so I reviewed the claim: I ran the acceptance demonstration myself, with one change designed to break it. Read `ca3bfbbd` in a clean clone at node 22.23.2 on Linux.

### What re-derived exactly — EXECUTED, not read off the body

The three composition mutations, run independently with `grep -oE "error TS[0-9]+" | sort | uniq -c` since you named `head`-ing `tsc` as the reporting trap:

| mutation | body claims | I measured |
|---|---|---|
| drop a spread from `createPostgresStorage` | 1× TS2322 | **1× TS2322** |
| drop a delegation from `FakeTeamStorage` | 1× TS2420 + 93 consumer errors | **1× TS2420 + 93** (4 TS2322, 58 TS2345, 10 TS2551, 21 TS2741) |
| drop `& IngestKeysPort` from the alias | 18 errors across exactly six files incl. `keys/verify.ts` | **18**, across exactly these six |

```
deploy/serve.ts   src/api/http.ts   src/api/main.ts
src/keys/seed.ts  src/keys/storage.test.ts   src/keys/verify.ts
```

Three independent counts landing on the digit is not something I expected to find, and it is the strongest evidence in the PR that the intersection is load-bearing rather than decorative.

**The law restatement is stronger, and I checked it rather than accepting the verifier's word.** Eight cases to twelve. `stripTsComments` byte-identical. Every removed assertion has a *narrower* successor — the `.unsafe(` pin went from "exactly one call site" to "exactly one, **and it is `ddl.ts`**" by path, and `postgres.ts` lost every exemption but clause 5. The only non-comment lines deleted are the eleven that made `postgres.ts` the exempt module. I ran the new per-file clause myself: SQL smuggled into a port's `port.ts` rather than its `sql.ts` reddens the law.

**Both pointer corrections tell the truth.** `schema-law.test.ts` carries the identical assertion string at a file that genuinely declares the signature. `TimeZone` appears once in `ports/events/sql.ts` and zero times in `postgres.ts`. `driver.ts` and `recording-sql.ts` change comments only — no behaviour moves anywhere outside the three composition roots.

**The verifier's sharpest mutation reproduces.** A complete but unwired port directory: `tsc` exit 0, composition-law case 2 red. The type system structurally cannot see it, and the law can.

### What I found — case 5 checks a list, not the ports

I ran your acceptance demonstration with one change: I made the throwaway port **collide**. A sixth port declaring `close(): Promise<void>` beside `LifecyclePort`'s, spread after it in `createPostgresStorage`, wired into all three roots exactly as a wave-9 lane would:

```
 M packages/team/src/storage/contract.ts
 M packages/team/src/storage/fake.ts
 M packages/team/src/storage/postgres.ts
?? packages/team/src/storage/ports/retention/
test files touched: 0
```

`npm run typecheck` exit **0**. The composition law: **5 passed, 5 total** — case 5 included, the case whose name is *"no two ports declare the same method name"*.

The only thing in the package that noticed was `postgres.test.ts:530`, *"close ends the connection"*, and it noticed only because my shadowing `close` happened to differ observably. A colliding port whose shadowed method behaves the same, or whose method has no such behaviour assertion, passes everything — which is precisely the hazard the verify record names: *a settings port also declaring `close` spreads last, so the shadowing would be invisible to every behaviour test.*

Case 5 is written to catch that, and it does — for the five ports named in `perPort`. It has no way to know about a sixth.

This is not a design disagreement: the case's own comment already says a new port adds a line there, and says why the static imports are right. The defect is that nothing enforces it. **Case 2 exists so registration is enforced rather than remembered, and case 5 was exempting itself from that rule.**

### The fix — `99206daf`, one assertion

One line asserting the hand-written list covers every port directory. It keeps the static imports the comment defends: no dynamic import, no dependency on the wiring it checks.

Clean head, five ports, list of five: **5 passed**. Falsifier in, six ports, list of five: case 5 red, and the message is the remedy rather than a puzzle —

```
a port directory exists that this case does not list — add it to `perPort` above,
or its method names go unchecked for collisions: expected 5 to be 6
```

It also reddens beside case 2 on the unwired-port mutation.

**The consequence, stated plainly rather than buried:** a wave-9 lane now edits one test file, by one line. Your demonstration records *"Test files edited: zero"*, and with this commit that becomes one. I think it is the right trade, and it is the same trade you already argue for the composition roots — *"stated plainly rather than rounded up… it is not 'only new files'"*. But the headline claim is yours to restate, and if you would rather keep the zero and accept that collisions go unchecked for every port added after this wave, revert the commit and say so on #509. I would rather you make that call than have me make it by pushing.

- **`SQL_MODULE_IMPORT_RE` lacking the `require(` form** — your verifier named it worth-later, not silently fixed. Agreed on both counts: no live path in an ESM package, and naming it beats patching it here.
- **ADR-0043's prose** at lines 87 and 100, routed to the doc-sweep wave. Agreed — no citation law breaks, and widening this lane to reach it is the thing the wave exists to avoid.
- **The body says 21 importers of `storage/contract.js`.** I count 16 files and 16 import statements at your head, and 15 at the base — the sixteenth is `composition-law.test.ts` itself. Nothing rests on it, since `tsc` is the real proof they compile untouched and it is clean, but 21 does not match either count and it is in the record now.

### What I did not run

macOS, Windows, the min-node leg. No Postgres, so the round-trip the module header already disclaims stays undisclaimed by me too. **The patch-id equality I could only half-check**: `ca3bfbbd`'s stable patch-id is `23994a56e6e9`, exactly the value you cite — but the lane sha `f0e296c3` is not reachable on this remote, so the *equality* rests on your record rather than on anything I ran.

`origin/main` had moved to `99c9868f`, so I merged it into the branch (`a36be1e0`, clean) and ran the leg against the post-merge tree:

```
  Test Files  518 passed (518)
```

**Approving.** The claim four wave-9 lanes are built on is now checked for the ports that exist rather than the ports someone remembered to list. One platform, not CI — the workflows are retired, so a local leg is the whole of the evidence. Landing is the operator's act.

PR #512, opened 2026-09-14
gh: a transient DNS fault fails half of every call and reports itself as a config error

### LINKED ISSUE #508: gh: a transient DNS fault fails half of every call and reports itself as a config error
## Fence (may touch ONLY)

- `scripts/dev/issues.sh`
- `scripts/dev/` — a shared retry helper, if one is the right shape
- `CONTRIBUTING.md`

The underlying fault is environmental and is not this repo's to fix. What IS this
repo's is that its own gh-wrapping scripts treat a transient network failure as
fatal, and that the failure reports itself as a configuration error. (The tool
name is unbackticked in this section on purpose: every ticked token here is read
as a claimed fence path.)

## Why

**Every `gh` invocation on at least one contributor machine fails roughly half the
time, and the failure is misreported as something else.**

Measured 2026-09-14, EXECUTED, over the same host in the same minutes:

| probe | result |
|---|---|
| `curl https://api.github.com/rate_limit` × 20 | **20/20** |
| `gh api rate_limit` × 20 | **10/20** |
| `getent ahosts api.github.com` × 5 | 5/5 |
| GraphQL points consumed by the failures | **0**, budget at 5000/5000 |

Failures return in **~150 ms**; successes take ~600 ms. They are not timeouts.

**Root cause, reproduced directly.** With `GH_DEBUG=api`:

```
dial tcp: lookup api.github.com on <resolver>:53: no such host
```

and querying that resolver by hand, 5 of 6 responses are rejected:

```
;; Warning: Message parser reports malformed message packet.
```

The resolver returns malformed DNS response packets. glibc-based clients (`curl`,
`getent`, `dig` on retry) tolerate or retry them; Go's resolver classifies them as
`no such host`, so every Go-based tool — `gh` included — fails on roughly half its
calls while every other tool on the machine appears healthy.

**Not fixable by resolver mode**, tested: `GODEBUG=netdns=cgo` 5/10 and
`GODEBUG=netdns=go+v4` 5/10 against a 6/10 default — no material difference.

## What it costs this repo

**Compounding.** `scripts/dev/issues.sh` makes 23 `gh` invocations, several per
board operation. At ~50% per call a compound operation succeeds at 0.5^n:

```
2 calls -> 25%    3 calls -> 12.5%    4 calls -> 6.2%    5 calls -> 3.1%
```

Measured: **12 consecutive `issues.sh status` attempts, none succeeded**, while a
bare `gh api` in the same window succeeded half the time. Board operations are
effectively unusable rather than merely flaky.

**Two misleading errors, and this is the expensive part.**

- `unknown owner type` — `gh` must call the API to learn whether an owner is a
  user or an org, so a network failure surfaces as a **configuration** error. Two
  sessions read it as a broken script.
- `API rate limit already exceeded` — observed from `gh pr create` with the budget
  measured at **5000/5000 immediately before and after**. A session then retried an
  expensive call ~20 times on the strength of it.

Both send the reader to the wrong problem. `AGENTS.md` already names this shape for
citations: a wrong pointer is worse than none, because it reports success.

## Definition of done

- A `gh` call in this repo's scripts retries on the transient network failure
  rather than dying on the first one, at a bounded count, and says which failure
  it retried. A call that fails for a REAL reason — auth, 404, an actual rate
  limit — must still die immediately and loudly.
- The distinction is PROVEN, not asserted: a test or a recorded probe showing a
  genuine error is not retried. A retry wrapper that masks a real failure is worse
  than the flake.
- `CONTRIBUTING.md` records that `error connecting`, `unknown owner type` and a
  rate-limit message with quota remaining are all the same class, and names the
  one-command check that distinguishes them (`gh api rate_limit` versus `curl`
  against the same host).
- The cheap check is written down: `curl` succeeding where `gh` fails localises
  the fault to Go's resolver in one command.

## Sibling case

**Every other operator script that wraps `gh`, not just this one.** The issue this
was split from measured `issues.sh`; the fault is per-`gh`-call and compounds with
call count, so the worst-affected script is whichever makes the most calls, and
that is not necessarily this one. Enumerate them before fixing one.

**And the inverse sibling: do not add a blanket retry.** A retry around a genuine
rate limit turns one clear failure into a slow one that ends the same way, and a
retry around a 404 hides it. The value is entirely in retrying THIS class and
nothing else.

A retry wrapper tested only against the flake cannot fail for the reason it
claims. Mutate it the other way: point it at a call that fails for a real reason
and show it dies on the first attempt. That is the assertion worth having, and it
is the one a happy-path test omits.

## Explicitly NOT in scope

Fixing the machine. The malformed-packet resolver is environmental — a WSL DNS
proxy on one contributor's setup — and this repo must not carry a workaround for
one person's network. What is in scope is that the repo's scripts degrade
gracefully and report honestly when ANY transient network fault occurs, which is
not machine-specific.

## Milestone note

Filed under `charter-laws` to sit beside the issue it was split from. That follows
the parent rather than the three-condition carve-out in `AGENTS.md` — this is
tooling resilience, not charter enforcement. Move it if a groomer reads the
carve-out more strictly.

## Provenance

Split from the board-cost issue during its premise check, 2026-09-14. That issue
attributed these failures to API budget exhaustion; the measurement above shows
they consume zero points and occur with the budget full, so the two are separate
problems and only this one is currently biting.

gabriel-canaan:
Closes #508

Every `gh` call on at least one contributor machine fails roughly half the time,
and the failure reports itself as something else. This makes the repo's own
scripts survive that, and report it honestly, without masking a real failure.

**No production code.** One new helper, its test, `issues.sh` wired, `CONTRIBUTING.md`.

## The fault

A DNS proxy returns malformed response packets. glibc clients tolerate them; Go's
resolver reports `no such host`. So Go-based tools fail while everything else on
the machine looks healthy:

```
curl https://api.github.com/rate_limit   x20  ->  20/20
gh api rate_limit                        x20  ->  11/20
getent ahosts api.github.com             x5   ->   5/5
```

Failures return in ~150 ms, before reaching the network, and consume **zero** API
quota — measured with the budget at 5000/5000 throughout.

## Why one retry, and why it must be per-call

**The failure alternates.** Sixty consecutive calls, O = ok, F = fail:

```
OFOFOFOFOFOFOFOOFOFOFOFOOFOFOFOFOFOFOFOOFOFOFOOFOFOFOFOOFOOF
27 failures / 60 calls · FF pairs: 0 · longest failure run: 1
```

Pasted raw rather than cited, so a later reader can recheck the premise the fix
rests on: **a single failure run of length 2 would falsify "one retry recovers".**

Retrying the *script* cannot work — a compound of n calls succeeds at p^n, and
outer retrying never changes p. Measured before the fix: four board setters
retried 8x each with delays, **0 of 32 succeeded**, while bare single calls in the
same minutes succeeded 60%. So the retry wraps each invocation.

## What it does and does not retry

Retried: the pre-network forms, plus `unknown owner type` — which `gh` emits when
a network call made to learn whether an owner is a user or an org fails, so it
reports a config-shaped error for a network fault.

**Not retried, each tested:** auth, 404, 422, 500, a real rate limit, a secondary
rate limit, an HTTP-level timeout, SIGTERM, and a non-zero exit with empty stderr.
Exit status and message reach the caller unchanged.

`unknown owner type` is ambiguous — a genuine bad owner produces it too. It is
retried anyway: unbounded cost is the thing to refuse, not uncertainty, and a real
bad owner surfaces with the identical message one call later. That non-retry
direction is tested and the test is load-bearing.

## At-most-once, stated plainly

`i/o timeout` and `connection reset` can land *after* a request was delivered, so
retrying them trades at-most-once for recovery. Every mutation here is idempotent
except `cmd_close`, which posts a comment before closing — a duplicate closure
comment is the worst case, bounded to one.

This is not hypothetical: a retried `gh pr review` on PR #440 left two identical
CHANGES_REQUESTED reviews, because the failure was on the response path. The
header names `cmd_close` as the site where that matters.

11 mutations were run against the suite; 8 redden it — dropping `unknown owner
type` (−5), retrying unconditionally (−4), never retrying (−24), reverting the
`if`-condition shape that fixes a `set -e` abort (−5), `return 1` instead of
`return "$status"` (−1), swallowing stderr on success (−1), notice to stdout (−5).

Review found three that did **not** redden it, all now fixed and re-checked:

- the stdout pass-through assertion captured through `$( )`, which strips trailing
  newlines itself — so it was blind to the wrapper doing the same. Mutating
  `printf '%s'` → `printf '%s\n'` left it green.
- the RETURN-trap regression test called the function inside a subshell, which is
  exactly what its own comment said hides the leak. Re-inserting the trap left it
  green; it now exits 1.
- the failure-path stdout replay was unasserted.

## Review

Two independent seats. Fourteen findings, deduped to nine, five blocking — none of
them "the retry masks a real error", which is what they were told to attack and
could not make happen. What blocked was reporting and output fidelity:

1. a real 401 died under a 19-line Python traceback, at a line this commit edited,
   while the fix pattern sat 40 lines below it in the same file
2. the wrapper stripped trailing newlines and NUL bytes and inverted stream
   ordering, while the commit body claimed it passed output through unmodified
3. the RETURN-trap test was decorative, and the bug its 9-line header documented
   could not be reproduced across six controlled runs — test and header both gone
4. the retry was completely silent on `issues.sh list`, the most-used read
5. `CONTRIBUTING.md` inverted the rate-limit discriminator, telling a reader a
   genuinely exhausted quota was this fault

All five verified fixed by mutation rather than by reading the diff.

## Measured effect

| command | before | after |
|---|---|---|
| `issues.sh when 508 now` | **0/32** | 6/6 |
| `issues.sh priority 508 high` | 2/6 | 6/6 |
| helper suite | 45 | **51 passed, 0 failed** |
| `issues.test.sh` | 42 | **49 passed, 0 failed** |

This issue's own board fields could not be written at all when it was filed. They
were set with this branch.

## Out of scope

Fixing the machine. The resolver fault is environmental, on one contributor's
setup, and this repo must not carry a workaround for one person's network. What is
in scope is that its scripts degrade gracefully under any transient network fault.

## Declared gap

**The head has not been reviewed.** The seats read the pre-fix commit; this is the
fix for what they found, and no seat has run against it. The changes are the five
items above plus four minor ones, each with a test that fails when reverted.

**One platform.** WSL2 Linux only, and this repo's workflows are disabled, so this
PR carries no check runs — the absence of a green bar is not a failure.

## Gates

517 files, 9816 passed. Both shell suites green. Run at loadavg 1.54, uncontended.


KelliherL (approved):
## Review of #512 — approve; the five unreviewed fixes all hold under mutation, and the one gap they left is closed

You declared the head unreviewed, so that is what I reviewed — not the diff's account of itself, but whether each of the five fixes actually bites. Read commit-by-commit on `25f75335` in a clean clone at node 22.23.2 on Linux, then mutated the helper twenty-three times.

### The five seat findings, each re-derived rather than read

| the fix | how I checked it | result |
|---|---|---|
| a 401 no longer dies under a Python traceback | `ensure_fields` captures instead of piping (`issues.sh:104`); deleted the failure-path stderr replay | **red ×5** |
| trailing newlines, NULs and stream order survive | `cat "$out_file"` → `printf %s "$(cat "$out_file")"` | **red ×2** |
| the RETURN-trap test is no longer decorative | re-inserted `trap 'rm -f …' RETURN` after `gh-retry.sh:108` | **red ×1** |
| `issues.sh list` is no longer silent about a retry | ran `gh-retry.sh` through `subprocess.run` exactly as `issues.sh:338` does, against a mock that fails once | rc 0, stdout clean, notice on stderr, **2** `gh` calls |
| `CONTRIBUTING.md`'s discriminator is the right way round | fed nine real messages to the live `GH_RETRY_FAULT_PATTERN` | matches the doc exactly |

The nine messages, and what the matcher does with each — EXECUTED against the shipped pattern, not read off the table:

```
RETRIED      error connecting to api.github.com
RETRIED      unknown owner type
RETRIED      dial tcp: lookup api.github.com on 10.0.0.1:53: no such host
not retried  API rate limit already exceeded (but you have 5000 remaining)
not retried  HTTP 401: Bad credentials
not retried  HTTP 404: Not Found
not retried  GraphQL: Field 'x' doesn't exist (422)
not retried  post https://api.github.com: dial tcp 140.82.0.1:443: i/o timeout
not retried  read tcp: connection reset by peer
```

### The at-most-once argument is structural, not merely disclosed

This is the claim I most wanted to break, and it holds. The header names `cmd_close`'s `gh issue close --comment` as the single non-idempotent mutation. Classifying all twelve shell call sites: eight reads, and four writes of which three are idempotent field sets (`issues.sh:208`, `:378`, `:410`). The fourth is `:539`. Exactly one, and it is the one named. Two further call sites reach the helper standalone from Python; both are reads.

The risk is **removed**, not documented around. Both response-path forms are excluded from the pattern — the last two rows above — so no retry can fire after a request was already delivered. The duplicate-review incident on #440 cannot recur through this helper.

### What I found — one gap, now closed by `aa639d5e`

`gh_retry` has **two** stdout/stderr replay pairs, not one: the success path at `gh-retry.sh:174-175` and the failure path at `:202-203`. The suite's ordering check (`out-then-err`, `gh-retry.test.sh:195`) only reaches the first.

Swapping just the failure path's two lines, so a caller merging `2>&1` reads the diagnostic before the output:

```sh
      cat "$err_file" >&2
      cat "$out_file"
```

left **all 51 checks green**. Deleting either line does redden (×1 and ×5), so both replays are pinned — only their order is not. Finding 2 is precisely "inverted stream ordering", and the body says all five were verified by mutation; that one was verified on half the code.

`fail-with-partial-stdout` already writes to both streams and is already non-retryable, so pinning it needed no new mock mode — only reading the existing one with the streams merged instead of stderr discarded. **52 passed, 0 failed**, and the swap above now reddens exactly that one check. Mechanical rung, so it is pushed and this is an approval rather than a hand-back.

### My other mutations, all red

| mutation | result |
|---|---|
| retry twice instead of once | red ×3 |
| pattern matched against stdout instead of stderr | red ×24 |
| the retry's own result assumed successful | red ×4 |
| `dial tcp … no such host` dropped from the pattern | red ×3 |
| `unknown owner type` dropped | red ×5 |
| `error connecting to` dropped | red ×16 |
| stderr swallowed on success | red ×2 |
| failure-path stdout replay deleted | red ×1 |
| `return 1` instead of `return "$status"` | red ×1 |
| the `if`-condition shape reverted | red ×5 |

And outside the suite entirely, against a fake `gh` on `PATH`: exit statuses 1, 3, 4 and 7 reach the caller unchanged; a **success** whose stderr happens to contain the fault text is not retried, because the status check precedes the text check; a caller under `set -e` survives a non-retryable failure and reaches its next statement; a 200 KB body and an embedded NUL both arrive byte-exact; no temp files are left behind.

### One thing to correct before this becomes the record

The body's finding 3 says the RETURN-trap "test and header both gone". Both are in the head — the header at `gh-retry.sh:109-125`, the test section at `gh-retry.test.sh:246` — and the header says the opposite of what the body does: that the bug *does* reproduce, given some other function returning next. The head is the right version. Re-inserting the trap reddens the suite, and the section carries a named control (`two bare gh_retry calls in a row self-heal (control — this must NOT fail)`) beside the leak case. So it is the description that is stale, not the code — worth fixing, since the body is what a future reader meets first.

- **Four scripts still call `gh` unwired**: `fence-lint.sh`, `pr-verdict.sh`, `await-merge.sh`, `metrics.sh`. Outside #508, which named `issues.sh`, and I am not asking you to widen this PR. But on the affected machine they keep failing about half the time, and `await-merge.sh` polls. A follow-up issue rather than scope creep here.
- **Neither shell suite is run by any gate** — not `ci-local.sh`, not `gate.sh`, not `land.sh`, not a package script. That predates you. It is worth naming now because this PR makes a third hand-run suite the only thing guarding a helper that sits under every board write.
- `gh-retry.sh:187-188` are inert. `gh "$@" >"$out_file" 2>"$err_file"` on the next line truncates both files anyway; removing the two lines leaves stdout and call count identical (measured, both variants) and the suite green. Harmless, just dead.

### What I did not run

macOS, Windows, the min-node leg. **The real fault** — I cannot reproduce that resolver, so every probe here used a mock, and your sixty-call alternation measurement remains the only evidence for the premise the design rests on; I did not re-derive it. No live GitHub write.

`origin/main` had moved to `3e4ab506`, so I merged it into the branch (`3f7294f2`, clean) and ran the leg against the post-merge tree:

```
  Test Files  517 passed (517)

  scripts/dev/gh-retry.test.sh   52 passed, 0 failed
  scripts/dev/issues.test.sh     49 passed, 0 failed
```

**Approving.** One platform, not CI — this repo's workflows are disabled, so a local leg is the only evidence that exists, and the missing green bar on this PR is exactly what you say it is. Landing is the operator's act.

Issue #497, opened 2026-09-14
prd56: prd-30 and prd-55 move to the done shelf, because their milestones closed


KelliherL:
## Fence (may touch ONLY)
- `docs/prds/prd-30-the-open-hand.md`
- `docs/prds/prd-55-the-lab-stage-two.md`
- `docs/prds/done/`
- `docs/roadmap.md`
- `docs/architecture.md`
- `docs/user-guide/the-lab.md`
- `docs/design/ui-2.0-progress.md`
- `packages/web/src/disclosure/index.ts`
- `packages/server/src/prd-location-law.test.ts`
- `.swarm/prd-milestones.txt`

## Blocked by
Nothing. The citers are fenced here up front under the rule that a behaviour-changing issue fences what encodes the old behaviour: moving a PRD breaks every path that names it, and doc-citation-law reddens on each one, so they must move in the same commit.

Open PR #455 touches docs/design/first-glance-acceptance.md only and does not collide.
**Model:** sonnet. **Wave:** not groomed — mechanical, and owed the moment the milestones closed.

## The finding

prd56 ruling 1: a PRD whose GitHub milestone is closed lives under the done shelf. Two milestones closed today and neither PRD moved:

- **prd30** — closed 2026-09-14T03:05:18Z, by another contributor
- **prd55** — closed 2026-09-14 by the operator, on the docs-audit sweep

EXECUTED. prd-location-law passes on main right now **only because the committed manifest predates both closures** — it reads `.swarm/prd-milestones.txt`, not a live gh call, deliberately so it runs offline. Regenerating that manifest with scripts/dev/prd-milestones.sh and re-running the law gives:

```
× no PRD on the live shelf has a closed milestone with no declared exemption
  expected [ 'prd-30-the-open-hand.md', 'prd-55-the-lab-stage-two.md' ] to deeply equal []
```

So main is not red today, and goes red the moment anyone regenerates the manifest — or when its own 14-day staleness bound expires and forces them to. This is a debt with a timer on it, not a latent one.

## Definition of done

- Both PRDs live under the done shelf, moved with git mv so history follows.
- Every citer updated in the same commit. Measured on main: prd-30 is named by the roadmap, the UI 2.0 progress note, and a disclosure source comment; prd-55 by the roadmap, the architecture document, the lab user guide, and prd-location-law's own file. The roadmap names both.
- The manifest is regenerated by its script, never hand-edited — the file says so itself.
[...]

PR #539, opened 2026-09-15
prd56: the PRD that wrote ruling 2 is the first document to need it

gabriel-canaan:
Part of #459 — **does not close it.** #459 stays open for the `<!-->` / `<!--->`
divergence recorded there. This is prd-56's own shelf decision, which came due when its
milestone closed.

## Why now

`prd56` closed 2026-09-15T02:15:36Z with all five issues done (waves 1–4 on `main`).
That made prd-56 itself a finished PRD sitting on the live shelf — **the exact state
ruling 1 convicts and ruling 2 exempts.** It takes the exemption.

## The reason, and the argument it deliberately refuses

Ruling 2 requires prose a reader can find, so the substance is in the document. In
short: prd-56's ruling 2 is the standing authority for every declared exemption now in
force, and four PRDs cite *"prd56 ruling 2"* by name as what licenses them to sit where
they do — `prd-14`, `prd-17`, `prd-20`, `prd-53`. Archiving the document that grants a
rule, while four documents are exercising it, demotes the rule to a record.

The declaration also states what is **not** being claimed, because a reason that proves
too much is worse than none:

> Being **cited** by a live PRD is not a reason to stay — nearly every shipped PRD is
> cited, and if that sufficed the archival shelf would never fill. Being the **live
> source of a rule other documents are currently applying** is different, and it is the
> only claim made here.

That is the same distinction #428 drew when it **refused** prd-48 an exemption: an
exemption with no argument behind it teaches people the form is a way to avoid
archiving.

## What is the sibling case?

**Archiving it instead.** The waves are done and `done/` is where finished PRDs go — so
that satisfies ruling 1 and breaks what ruling 2 exists to protect, silently. Nothing in
the suite asks whether a rule being archived is still being applied.

Second: writing a reason any finished PRD could copy. That is why the "cited is not
enough" paragraph is in the declaration itself, not only in the commit body — the next
person reaching for this form reads the document, not the git log.

None was staged; this is the sequence as it happened. The manifest was regenerated
**first**, with no reason present:

```
FAIL  packages/server/src/prd-location-law.test.ts
      expected [ 'prd-56-the-shelf-agrees-with-the-ledger.md' ] to deeply equal []
Tests  1 failed | 66 passed (67)   exit 1
```

The law convicted this PRD **by name**. Then the reason was added and nothing else
changed:

```
Tests  67 passed (67)   exit 0
```

So the exemption is load-bearing, and the law that convicted it is the one this PRD
wrote.

## Why the manifest is in the same commit

Ruling 1's law reads milestone state from `.swarm/prd-milestones.txt`, not from the
tracker — so closing the milestone convicted nothing until the regeneration. Landing
both together means the law is **never red in between**, which is ruling 3's
move-and-citers discipline applied to a shelf decision.

## One thing found while doing it

prd-56's head still carries `**Status:** **BLESSED**`, which **#500's new status-line
law convicts — but only for PRDs under `docs/prds/done/`.** Staying on the live shelf is
what keeps that correct rather than a second defect; archiving would have required
fixing it in the same breath, and nothing would have surfaced that until the sweep ran.
A small argument for the decision, arriving after it was made. It is recorded in the
commit body so the next reader does not have to rediscover it.

**No independent pass has read this commit.** It is two files — a declared reason and a
regenerated manifest — with a mutation proven in both directions and the full suite
green, but that is one pair of eyes. Seats were not spent on it because the mutation is
the evidence and it is unusually direct; the reviewer of this PR is the first
independent read.

Worth a reviewer's attention specifically: **is the reason's argument sound, or does it
prove too much?** That is a judgement about prose, not something a law can check —
`docs/prds/` is excluded as a citing source, so none of this text is gated.

## Gates

Orchestrator, full worktree: `typecheck` **0**, `lint` **0**,
`VITEST_MAX_WORKERS=6 npm test` **0** — 9847 passed, 3 expected-fail, 3 skipped.
**One platform, linux** — Actions was retired on this repo on 2026-09-11.


ciaran-slow (commented):
One commit, read on its own. **One substantive finding, fixed on the branch** — it needs your confirmation before this lands (rung 2), so I have not approved.

**The mutation reproduces exactly.** `EXECUTED` — stripping the `**Shelf exemption:**` block from `prd-56` and re-running `packages/server/src/prd-location-law.test.ts`:

```
AssertionError: expected [ Array(1) ] to deeply equal []
+   "prd-56-the-shelf-agrees-with-the-ledger.md"
Tests  1 failed | 66 passed (67)
```

The law convicts this PRD **by name**, and restoring the reason returns 67/67. Identical to the body's numbers. The exemption is load-bearing, and the evidence is as direct as the body claims.

**The regenerated manifest is accurate in every row it changes.** `EXECUTED` — checked all seven against the tracker via `gh api repos/:owner/:repo/milestones`:

| milestone | manifest | tracker |
|---|---|---|
| prd43 | closed `2026-09-14T05:51:24Z` | matches |
| prd49 | closed `2026-09-14T05:51:26Z` | matches |
| prd56 | closed `2026-09-15T02:15:36Z` | matches |
| prd51, prd54, prd57, prd58 | open | matches |

**The two incidental flips convict nothing.** `EXECUTED` — `prd-43` and `prd-49` are already at `docs/prds/done/`, so closing their milestones creates no violation. That is the sibling risk in regenerating the whole manifest in one commit, and it is clean.

**Landing the manifest with the reason is the right call**, for the reason given: the law reads milestone state from `.swarm/prd-milestones.txt`, so the two together are never red in between.

**Spot-checks on the reason's other claims all hold.** `EXECUTED` — the `prd-17` quote *"Classifying is a read; moving is an act, and prd-56 is where the act gets its rule."* is verbatim at `docs/prds/prd-17-complete-record.md:37`; `docs/follow-up-292.md` exists; the `**Status:** **BLESSED**` observation about #500's status-line law is accurate.

## Finding: the reason named four PRDs; two of them retired a day earlier

`EXECUTED`, blocking-if-unfixed, **fixed in `650b3d8d`**.

The declared reason said four PRDs rest on ruling 2 and named them by path. Two are wrong on both counts:

```
$ ls docs/prds/prd-14-experiment-console.md docs/prds/prd-53-the-lab.md
# neither exists -- both are at docs/prds/done/

$ grep -i 'shelf exemption\|prd56 ruling 2' docs/prds/done/prd-14-experiment-console.md \
                                            docs/prds/done/prd-53-the-lab.md
# no matches in either file
```

They *did* hold exemptions on this exact authority — `git show 3bfabe98^:docs/prds/prd-53-the-lab.md` still shows `Operator-ruled 2026-09-13, recorded on #428, per prd56 ruling 2`. They lost them in **`3bfabe98`, "four PRDs move to the done shelf, and two spent exemptions retire" (#497)**, dated 2026-09-14 — one day before this reason was written. The claim was true when the thinking was done and stale when it was committed.

This matters more than a citation slip because the reason is a **declared artefact meant to be checked by a later reader**, and half its evidence did not survive a `grep`. It is also exactly the failure the PR body anticipates — *"an exemption with no argument behind it teaches people the form is a way to avoid archiving"* — arriving as an argument whose supporting facts do not hold rather than as a missing one.

**Nothing would have caught it.** `EXECUTED` — `EXCLUDED_DIRS` in `packages/server/src/doc-citation-law.test.ts:494` is `['docs/research/', 'docs/review/', 'docs/prds/']`, so the dangling paths are ungated. Your "none of this text is gated" note was correct, and this is what it cost.

**The fix makes the claim narrower and stronger.** `EXECUTED` — I verified `prd-17` and `prd-20` are *both* the complete set of live `**Shelf exemption:**` holders and the complete set of live PRDs citing `prd56 ruling 2` (swept all six live-shelf PRDs; `prd-34`, `prd-51`, `prd-56` hold none). So the corrected text claims the **complete set** rather than a sample of four, and records where the other two went so nobody re-adds them.

After the fix: law green 67/67, and re-stripping the exemption still convicts `prd-56` by name — the correction did not quietly break what the reason does.

## Answering the question you asked

> **is the reason's argument sound, or does it prove too much?**

**Sound, and it does not prove too much.** The "cited is not enough / live source of a rule being applied is different" distinction is the right cut, and it does real work — it is what separates this from the prd-48 refusal in #428. The argument's *structure* was never the problem; its *evidence* was overstated twofold, and with two live dependents it still carries.

## Gates

Post-merge tree (`gh/main` merged into the fix, clean), **including my commit**: `typecheck` **0**, `lint` **0**, `VITEST_MAX_WORKERS=6 npm test` **0**.

**What I did not run:** this leg is **darwin**; yours was linux. Neither of us has Windows, and Actions has been off since 2026-09-11, so this PR carries no check runs.

The exemption now rests on exactly two live documents, and this PR is the proof that such a basis can lapse quietly. Worth considering whether the reason should name its own expiry — *"if `prd-17` and `prd-20` both archive, this exemption lapses and `prd-56` follows them"* — so the next person regenerating the manifest has a condition to check rather than a paragraph to re-audit. Not a blocker, and deliberately not something I wrote in for you.

ciaran-slow:
**Re-gated against current `main`** — my earlier gate is superseded.

#541 merged as `1c6f3852` after that run, bringing in `fa9ac028`, which rewrites the comment scanning in `classifyLines` — the reader that parses the `**Shelf exemption:**` block this PR adds. So the earlier green described a tree that no longer exists, against the one change most likely to interact with this one.

`EXECUTED`, on `650b3d8d` merged with `1c6f3852` (clean):

- `packages/server/src/prd-location-law.test.ts` — **70 passed** (the 67 plus #541's 3), `prd-56` not named as a violation.
- Mutation re-run **under the new reader**: stripping the exemption still convicts `prd-56` by name — `1 failed | 69 passed (70)`. Still load-bearing; #541 did not quietly neutralise it.
- `typecheck` **0**, `lint` **0**, `VITEST_MAX_WORKERS=6 npm test` **0** — 518 files, 9874 passed, 3 expected-fail, 7 skipped.

Platform **darwin** only, and Actions remains off, so this PR still carries no check runs.

**Still outstanding, and not something I can clear myself:** `650b3d8d` is my own commit, so the approval on this PR needs a second pair of eyes — I am not approving my own substantive fix into `main`. The author's confirmation on the corrected reason is the other open item.

PR #541, opened 2026-09-15
prd-location-law: an overlapping comment closer closes the comment

### LINKED ISSUE #540: prd-location-law: <!--> and <!---> are complete comments, and the reader closes on them
## Fence (may touch ONLY)

## Why

`<!-->` and `<!--->` are **complete HTML comments** in CommonMark. The exemption reader
treats both as unterminated, blanks the rest of the document, and so **hides a
declaration a reader can plainly see** — fail closed, which is the direction prd-56
ruling 2 exists to prevent (*"a law that convicts a correct state is one people learn to
override"*).

EXECUTED against `origin/main`:

| input | reader | correct |
|---|---|---|
| `<!-->` then a declaration | **CONVICT** | must exempt |
| `<!--->` then a declaration | **CONVICT** | must exempt |
| `<!---->` then a declaration | EXEMPT | ✓ control |
| `<!-- note -->` then a declaration | EXEMPT | ✓ control |
| genuinely unterminated `<!--` | CONVICT | ✓ control |

**Pre-existing, not a regression.** Verified byte-identical on the base during #459's
review — it predates the fence/comment interleaving work and survived it unchanged.

## The cause, and it is one character

`packages/server/src/prd-location-law.test.ts:730`

```ts
const closeAt = masked.indexOf('-->', openAt + 4)
```

The `+ 4` skips the whole opener, so a `-->` that **overlaps** it is never found. In
`<!-->` the closer begins at `openAt + 2`; in `<!--->` at `openAt + 3`.

CommonMark's HTML block type 2 begins on the line matching the start condition and ends
on the line containing the end condition — **which may be the same line, and the two may
overlap.** The raw-HTML comment grammar admits both spellings explicitly.

Searching from `openAt + 2` is the whole fix. It cannot false-close: the opener occupies
`openAt..openAt+3`, so a match at `openAt + 2` requires the character at `openAt + 4` to
be `>`, which is precisely `<!-->`.

## Definition of done

- `<!-->` and `<!--->` close on their own line; a declaration after either is read.
- The three controls above still behave — in particular a genuinely unterminated `<!--`
  must still blank to EOF.
- The truth table in that file gains its rows. Rows 17 and 18 currently record these as
  *"diverges (pre-existing)"*; they become ordinary handled cases.
- A test per spelling, each with a control in the same call that still conceals a real
  comment, and at least one spelling not named in this issue.

## Sibling case

The tempting fix is to special-case the two literals — match `<!-->` and `<!--->` by
name. That is a fitted instrument: the general rule is that the **end condition may
overlap the start condition**, and keying on two strings leaves the next overlapping
spelling to be discovered. Fix the offset, not the spellings.

Second: fixing the block-level scan at `:730` and forgetting the in-comment
continuation scan at `:662`, which searches `line.indexOf('-->')` with no offset. Check
whether that path needs the same treatment or is already correct, and say which.

A test asserting only that `<!-->` exempts survives an offset of `openAt + 0`, which
would also make a bare `<!--` self-close and silently stop concealing anything. The
unterminated-comment control in the same run is what rejects that — it must stay
CONVICT.

## Provenance

Found by the Codex seat during #459's review (2026-09-14), classified as pre-existing
by the Opus seat in the same round (table rows 17-18), and confirmed byte-identical on
the base by the orchestrator. Deferred then because #459 was fixing a different class;
re-verified on `origin/main` before filing.

gabriel-canaan:
Closes #540

`<!-->` and `<!--->` are complete HTML comments in CommonMark. The exemption reader
treated both as unterminated, blanked the rest of the document, and so **concealed a
declaration a reader can plainly see** — fail closed, the direction prd-56 ruling 2
exists to prevent.

## The cause is the offset, not the spellings

The block-level scan skipped the whole four-character opener before looking for the
closer:

```ts
masked.indexOf('-->', openAt + 4)
```

CommonMark's HTML block type 2 begins on the line matching the start condition and ends
on the line containing the end condition — and **the two may overlap**. In `<!-->` the
closer begins at `openAt + 2`; in `<!--->` at `openAt + 3`.

Searching from `openAt + 2` is the whole change. It cannot false-close: the opener
occupies `openAt..openAt+3`, so a match there requires the character at `openAt + 4` to
be `>`, which is precisely `<!-->`.

## What is the sibling case?

**Special-casing the two literals.** That is a fitted instrument — the rule is the
overlap, and keying on two strings leaves the next overlapping spelling to be
discovered later. One test therefore uses a spelling named nowhere in #540: an
overlapping closer inside a blockquote with trailing prose on the same line.

**The second scan.** The in-comment branch searches `line.indexOf('-->')` with no offset
at all. It is **correct as it stands and deliberately unchanged** — it runs only for
continuation lines of an already-open comment, where no opener sits on the line, so
nothing can overlap. That reasoning is in the code comment so the next reader does not
re-derive it.

## What mutation would this test survive — and #540's own answer was wrong

The issue predicted that `openAt + 0` would make a bare `<!--` self-close, with the
unterminated-comment control rejecting it. Measured across the range:

| offset | result |
|---|---|
| `+0` | 70 passed |
| `+1` | 70 passed |
| `+2` | 70 passed *(shipped)* |
| `+3` | **2 failed** |
| `+4` | **2 failed** *(the bug)* |

`+0`, `+1` and `+2` are behaviourally **identical**, because `<` and `!` cannot begin
`-->` — a match can only start at `openAt + 2` or later. So the predicted mutation
cannot occur and the control could not have rejected it. **The prediction was wrong; the
tests are not.** The mutation that matters is `+4`, the value being replaced, and it
reddens both new tests by name.

The correction is recorded on #540 as well as here, because a wrong mutation claim in a
filed issue is the kind of thing a later reader trusts.

## Provenance

Found by a review seat during #459 and classified pre-existing there by a second seat;
confirmed byte-identical on the base at the time, and **re-verified against `origin/main`
before the fix** rather than taken from the earlier report.

## Declared gap

**No independent pass has read this commit.** One file, one character of behaviour
change, three tests, and a mutation mapped across the full offset range — but that is
one pair of eyes. Seats were not spent because the change is a single offset with its
boundary measured on both sides; the reviewer of this PR is the first independent read.

## Gates

`typecheck` **0**, `lint` **0**, `VITEST_MAX_WORKERS=6 npm test` **0** — 9850 passed, 3
expected-fail, 3 skipped. **One platform, linux** — Actions was retired on this repo on
2026-09-11.


ciaran-slow (approved):
**Approve.** One commit, read on its own; every claim in the body was re-derived rather than taken on trust. No findings.

**The offset table reproduces exactly.** `EXECUTED` — swept `openAt + N` for N in 0..4 against `packages/server/src/prd-location-law.test.ts`, substitution asserted present before each run:

| offset | result |
|---|---|
| `+0` | 70 passed |
| `+1` | 70 passed |
| `+2` | 70 passed *(shipped)* |
| `+3` | **2 failed** \| 68 passed |
| `+4` | **2 failed** \| 68 passed *(the bug)* |

Identical to the body's table. The new tests redden on the value being replaced, so they are load-bearing rather than decorative — and the body's correction to #540's own predicted mutation is right: `+0`/`+1`/`+2` are behaviourally indistinguishable, so the mutation the issue predicted could not have occurred and its control could not have rejected it.

**The "cannot false-close" argument holds under its own arithmetic.** `REASONED` — a match at `openAt + 2` needs `content[openAt+4] === '>'`, since `[openAt+2]` and `[openAt+3]` are the opener's own `--`; that string is exactly `<!-->`. A match at `openAt + 3` needs `<!--->`. Both are complete comments, so neither is a false close. `+0`/`+1` add nothing because `<` and `!` cannot begin `-->`.

**The deliberately-unchanged second scan is genuinely correct.** `REASONED` — the `inComment` branch's `line.indexOf('-->')` runs only for continuation lines of an already-open comment, where no opener sits on the line, so no overlap is reachable. That is the sibling case, and leaving it alone is the right call.

## The direction question — measured, not argued

This area has twice been rejected for moving **fail-open** the wrong way, which prose review cannot see. `EXECUTED` — extracted `classifyLines`/`declaredExemptionReason` from `gh/main` and from `fa9ac028` by marker range, scored both against **commonmark@0.31.2** rendered HTML (comments, `<pre>` and `<code>` stripped; declaration visible iff the reason text survives), comparing *verdicts* rather than reason strings. Corpus: 2,880 documents over 10 comment spellings × 6 line prefixes × 4 tails × 6 wrappers.

| | base (`gh/main`) | head (`fa9ac028`) |
|---|---|---|
| agrees with CommonMark | 2376 | **2712** |
| **FAIL-OPEN** | 0 | **0** |
| FAIL-CLOSED | 504 | **168** |

336 verdict changes, and **all 336 are fixes** — zero new fail-open, zero new fail-closed. Fail-open does not move at all. This is unambiguously a fix.

The residual 168 fail-closed are entirely genuinely-unterminated openers (`<!--` and `<!-- x`, 84 each), present identically on both sides. Pre-existing, conservative in direction, and out of scope for #540 — that is #459's remaining territory.

## Gates

Re-run on the **post-merge** tree (`gh/main` merged into `fa9ac028`, clean): `typecheck` **0**, `lint` **0**, `VITEST_MAX_WORKERS=6 npm test` **0** — 518 files, 9874 passed, 3 expected-fail, 7 skipped.

**What I did not run:** this leg is **darwin**. The author's was linux, so between us the change has two platforms and neither of us has Windows. Actions has been off since 2026-09-11, so there are no check runs on this PR and none are expected.

#539's body says #459 stays open "for the `<!-->` / `<!--->` divergence recorded there" — which is precisely what this closes. Worth checking whether that row on #459 can now be struck.

Issue #457, opened 2026-09-12
ci-local's Test leg cannot go green on native Windows — it runs the suite raw, and never consults .windows-known-failures or windows-triage.sh


KelliherL:
## What happened

Ran the new leg on a native-Windows box, on a docs-and-one-law branch, to publish a verdict the way `AGENTS.md` now asks:

```
scripts/ci-local.sh --pr 456
  PASS  Install (npm ci — clean, lockfile-honest)
  PASS  Build
  FAIL  Test
  PASS  Typecheck · Lint · Packaging guard · Boot smoke · Pack smoke
  ci-local: RED
```

Seven legs of eight green. The Test leg fails because the full suite on this platform carries the expected-fail set — **247 failing tests on unmodified `origin/main`, measured in the same worktree**, the CRLF, drive-letter, path-separator, temp-dir, procfs and symlink-privilege classes that `.windows-known-failures` exists to name.

`scripts/ci-local.sh:121` is `step "Test" npm test || true`, and nothing downstream consults `.windows-known-failures` or `scripts/windows-triage.sh` (`grep -nE "windows-known|windows-triage" scripts/ci-local.sh` → nothing).

## Why it matters

**On a native-Windows machine this leg cannot go green, so no verdict can ever be published from one.** That is not a corner: the operator who lands PRs on this repo is on Windows, and the retirement makes the local verdict the only evidence a PR carries. The retired `windows-suite.yml` never had this problem — it ran the suite and then handed the result to `windows-triage.sh`, which turns a raw result into a verdict *against the committed expected-fail list*, per file, with `UNEXPECTED FAILURE` the only red that counts.

The machinery already exists and is still in the tree. The Test leg is the one place it is not wired.

## What would fix it

Route the Test leg's result through `windows-triage.sh` on Windows, as the workflow did: vitest's JSON results plus `.windows-known-failures` plus the tracked test-file list, and the leg is green when the triage verdict is green. prd-25 ruling 2's per-file verdict is unchanged by this — it is the same verdict, moved.

One thing worth folding in while there, from a defect this repo found on 2026-09-11: a known-red **file** can hide a **new** red test, because the list names files. The triage's own rows already carry per-row evidence strings, so comparing failing test *names* against the row's stated evidence would close that too. Recorded rather than insisted on — the per-file verdict was a deliberate ruling.

## Also

[...]

Issue #396, opened 2026-09-09
windows-suite: drawer/index.test.tsx 'is not a panel' times out at first render — intermittent on the runner (same shape as #362)


KelliherL:
## What happened

`windows-suite` on PR #382 at `2b018311` (run 34354440853, attempt 1) failed with one unexpected failure, in a file that PR does not touch — the commit is a docs-only rename of the prd-55 paper:

```
UNEXPECTED FAILURE (not on .windows-known-failures): packages/web/src/drawer/index.test.tsx
  the peek renders nothing at all with nothing selected › is not a panel — no frame, no heading, no empty state
  Error: Test timed out in 5000ms.   (src/drawer/index.test.tsx:133 — the first `await renderPeek({ selected: null })`)
```

The triage verdict was `1 unexpected · 29 expected · 0 removal candidates · 0 listed-not-run · 0 unevaluated · verdict RED`. A rerun of the same job on the same commit (attempt 2) passed; the run before it on the same branch (`11a20834`) passed; main's last Windows-suite runs are green on this file. So the test is intermittent on the runner: the first render of the peek takes more than 5 s some of the time — the same shape as #362 (an environmental race, not a wrong assertion).

## What it is not

Not a reason to widen the timeout (the repo's rule: remove the race, never lengthen the wait). The test's first `await` pays for the drawer's lazy imports under a cold module cache on a loaded runner; whatever settles that cost — warming the import in a `beforeAll`, or asserting the peek's absence without the full render path — is the fix, and it belongs to whoever owns `packages/web/src/drawer/`.

## Evidence

- run 34354440853 attempt 1 job `windows-suite` (2026-09-09 13:15Z): the FAIL block above; attempt 2 (2026-09-09 22:31Z): success.
- `.windows-known-failures` does not list `packages/web/src/drawer/index.test.tsx`, correctly — it passes almost always.

gabriel-canaan:
## Parked 2026-09-15 — the runner this describes is disabled, so it is unobservable rather than fixed

Recorded so a week of sitting off the board reads as a decision rather than drift.

**The `Windows suite` workflow is `disabled_manually`**, along with `CI`,
`Desktop installers` and `Release` — Actions was retired for cost on 2026-09-11,
two days after this was filed. Verified today via
`gh api repos/:owner/:repo/actions/workflows`.

So this failure cannot currently be observed, reproduced or fixed by anything in
CI. Nothing about the underlying behaviour has been shown to change; only the
[...]

Issue #362, opened 2026-09-08
windows-suite: tail.test.ts rotation-by-inode is intermittent — NTFS can recycle the file index across the rename


KelliherL:
## What happened

`windows-suite` on PR #342 at `37f5f63f` (run 34187332032) failed with one unexpected failure, in a file that PR does not touch:

```
UNEXPECTED FAILURE (not on .windows-known-failures): packages/server/src/collectors/sessionlog/tail.test.ts
  readNewLines detects a same-path rotation via inode even when the replacement already grew past the old offset — AssertionError: expected [ …(2) ] to deeply equal [ …(2) ]
```

A rerun of the same job on the same commit passed. The ten preceding Windows-suite runs (main and four branches) were all green on this file. So the test is intermittent on Windows, not broken.

## Why it can flake

The test writes a replacement file, renames it over the tailed path, and expects `readNewLines` to see a NEW identity and reset its cursor. The identity is the inode from `fs.stat`. On NTFS, libuv reports the file index as `ino`, and Windows can recycle an index quickly after the old file is unlinked by the rename — so, sometimes, the replacement carries the same `ino` as the file it replaced. The tailer then sees "the same file grew" and reads garbage from the stale offset, which is exactly the assertion that failed.

## What is asked

A fix in the identity, never a timeout: on `win32`, identity should not rest on `ino` alone — combine it with `birthtimeMs` (creation time survives the rename and differs between the two files), or fall back to size-plus-birthtime when `ino` is 0 or repeats. The test then asserts the new identity differs on Windows by construction rather than by luck.

Per the working agreement on flakes: remove the race; do not widen a timeout or retry.

## Evidence

- failing run: https://github.com/launchpad-26/rhizomorph/actions/runs/34187332032
- rerun of the same job on the same commit: success
- triage comment: https://github.com/launchpad-26/rhizomorph/pull/342#issuecomment-5579909494

gabriel-canaan:
## Parked 2026-09-15 — the runner this describes is disabled, so it is unobservable rather than fixed

Recorded so a week of sitting off the board reads as a decision rather than drift.

**The `Windows suite` workflow is `disabled_manually`**, along with `CI`,
`Desktop installers` and `Release` — Actions was retired for cost on 2026-09-11,
two days after this was filed. Verified today via
`gh api repos/:owner/:repo/actions/workflows`.

So this failure cannot currently be observed, reproduced or fixed by anything in
[...]

PR #542, opened 2026-09-15
charter-laws: CLAUDE.local.md is listed once, and the board cadence names both directions

### LINKED ISSUE #505: issues.sh: the board carries only live work, and a stated cadence keeps it that way
## Fence (may touch ONLY)

- `scripts/dev/issues.sh`
- `docs/design-notes/` — a new note, if the pruning cadence needs a home

The board itself is data, not a file: pruning it is an operator act through the
Projects API, not a diff. This fence covers the tooling and the record, not the act.

## Why

**Every `issues.sh` call reads the entire board, and the board only grows.**

`fetch_board` pulls `--limit 1000` and 11 call sites go through it or `ensure_board`,
so a one-field write pays for the whole board. `BOARD_LIMIT` was set on 2026-08-12,
when the board was small.

Measured 2026-09-14, EXECUTED:

| | |
|---|---|
| board items | **259** |
| of which **Done** | **244 (94%)** |
| genuinely live (Backlog/Ready/In progress/In review) | **15** |
| cost of one `issues.sh status` | **~384 GraphQL points** |
| budget | **5,000 points/hour, shared across every agent and tool on the account** |

That is **~13 board operations per hour for the whole machine**, and 94% of what each
one pays for is finished work.

The cost is not a fixed price — it scales with a number that only rises. The same
`issues.sh status` cost a fraction of this in August with no code change since. This
is the failure shape this repo already names for line-number citations, one level up:
a value pinned when it was small, rotting by construction as the thing it measures
grows.

**Three loads landed on it at once**, which is why it surfaced now rather than
gradually:

- **More agents.** 10 live agent processes and 39 worktrees during the measurement,
  all drawing on one 5,000.
- **A new consumer.** `scripts/pr-verdict.sh` (added 2026-09-12, 10 `gh` calls) posts
  the labels and commit statuses GitHub Actions used to emit for free. Retiring
  Actions on 2026-09-11 did not remove that traffic; it moved it onto the API budget.
- **A retry loop.** One session misdiagnosed a DNS failure as a rate limit and
  retried an expensive compound call ~24 times, exhausting the budget twice in an
  afternoon. That is a session-discipline failure and is recorded as one — but a
  budget that 13 calls can exhaust is what made a bad retry loop catastrophic instead
  of merely wasteful.

**What it costs when it trips.** Board writes fail with `API rate limit already
exceeded`, and — worse — with `unknown owner type`, because `gh` must call the API to
learn whether the owner is a user or an org and reports a **config-shaped error for a
network fault**. Two sessions read that as a broken script before the real cause was
found.

## Definition of done

- **The board carries only live work.** The 244 Done items are archived off it.
  Nothing is lost: closed issues keep their history, and the board is a working view,
  not a record.
- **A stated cadence for keeping it that way**, written down where the next operator
  will find it, with the reason — a one-off tidy is what the board already had.
- **The cost is re-measured after pruning and the figure recorded.** The claim of this
  issue is that per-call cost falls roughly in proportion to item count; if it does
  not, the diagnosis is wrong and that is the finding.
- `issues.sh` still dies loudly rather than truncating. **Do not "fix" this by
  narrowing the query to non-Done items** — see the sibling case.

## Sibling case

The tempting fix is to make `fetch_board` read only non-Done items, or to drop
`BOARD_LIMIT`. **Both are wrong**, and the script's own comment says why: it reads
everything so it can die rather than return a short list that a later `orphans` check
would agree with, *"which is the same silent-cap defect #420 fixed in `cmd_list`, one
function over."* Narrowing the query re-opens that defect to save points. Pruning the
board respects the invariant; querying less of it undercuts it.

The second sibling: fixing the cost and leaving the misleading error. `unknown owner
type` for a network fault will mislead the next reader whatever the budget is. If it
cannot be fixed in `gh`, it belongs in the runbook.

Any assertion that per-call cost is "lower" survives a measurement taken while the
board happens to be quiet, because the budget is shared and other agents move it.
Measure the same operation twice — once before pruning, once after — **in the same
conditions**, and state the item count both times. A single after-reading proves
nothing.

Second: pruning without recording the cadence would pass every check here and leave
the board back at 259 within weeks. The 2026-08-22 reconciliation moved nine PRDs and
five came back in three weeks; a habit is not a fix, which is the argument prd-56 was
built on.

## Provenance

Measured during prd-56 wave 3 / #459, 2026-09-14, after the budget was exhausted
twice in one session. Figures are EXECUTED: the 384-point cost from four consecutive
`issues.sh status` calls burning 1,536 points; the status breakdown from a
`gh project item-list` snapshot of all 259 items.

### LINKED ISSUE #513: gitignore: CLAUDE.local.md is listed once
## Fence (may touch ONLY)

- `.gitignore`

## Why

`CLAUDE.local.md` is listed **three times** in `.gitignore` — lines 40, 46 and 52 on
`main` at `99c9868f`. Lines 42–46 and 48–52 are byte-identical: the same five-line comment
block followed by the same entry.

Git takes the first match, so the duplication changes no behaviour and the entry itself is
correct. Two things make it worth one commit anyway:

- **It is a merge artefact that reads as deliberate.** Three entries look like three
  different reasons. A reader asking why `CLAUDE.local.md` is ignored finds two identical
  explanations and has to diff them to learn they say the same thing.
- **The duplicated comment speaks from a branch that no longer exists.** It reads: *"It was
  missing from this list rather than deliberately left out: it is ignored on main, and this
  branch predates that."* That was true where it was written and is self-contradictory
  here — the file it describes IS `main`.

Flagged 2026-09-08 in the body of `ad30fac0` as *"a merge artefact and harmless — the
entry is correct, just repeated — but it is a separate change and gets its own commit."*
That branch has since been deleted (its `.gitignore` rule guarded a worktree layout the
fleet no longer uses, and `git add -A` on a nested worktree stages a warned gitlink, not a
tree — EXECUTED 2026-09-15). The residual it named outlived it and is still true.

## Definition of done

- `grep -c '^CLAUDE.local.md$' .gitignore` returns `1`.
- The surviving entry carries a comment that is true read from `main` — no sentence
  speaking from "this branch".
- `git check-ignore -v CLAUDE.local.md` resolves to that line.
- A contributor holding a `CLAUDE.local.md` on disk sees no change in `git status`.

gabriel-canaan:
Closes #513
Closes #505

Two independent `charter-laws` issues, disjoint fences, bundled so the queue toll
is paid once. **No production code**: one `.gitignore` edit and one new design note.

## #513 — CLAUDE.local.md is listed once

It appeared three times. Lines 42–46 and 48–52 were byte-identical, and their
comment spoke from a branch that no longer exists — *"it is ignored on main, and
this branch predates that"*, which is self-contradictory read from `main`.

One entry survives, carrying a comment true from `main`. The surviving entry keeps
the position the block above it points at, and the explanation from the deleted
blocks is merged into it rather than dropped.

Two things the issue got wrong, corrected while fixing it: **git takes the LAST
matching pattern, not the first** (`check-ignore` resolved to line 52, not 40),
and the entry the block above describes as "ignored below" is the one that had to
survive for that sentence to stay true.

Verified: `grep -c` = 1; `check-ignore -v` resolves to `.gitignore:43`; a real
`CLAUDE.local.md` on disk gives zero `git status` lines; and **deleting the
surviving entry makes it appear** — so the kept line is the one doing the work,
not one of the two removed.

## #505 — the board cadence, and the cost claim measured properly

The board had already been pruned when this was picked up (259 → 37 items), so
what remained was the cadence and the measurement. Both are here.

**The cadence covers both directions**, because the live problem inverted: the
board is no longer too full, it is missing current work — `orphans` reports eight
issues off it, five of them a whole new wave. A rule that only prunes would have
left that unaddressed.

**The measurement went through three wrong answers before this one**, and the note
records why:

| claim | status |
|---|---|
| the issue's "cost falls with item count" | right, for the operation it measured |
| "2 points, so the thesis is refuted" | wrong — measured a *different* operation |
| "2 points" itself | wrong — subtracted a probe cost that is never charged |

A GraphQL query requesting only `rateLimit` fields is **free** (five consecutive
probes leave `used` unchanged; a real query moves it by 1), so the raw delta is
the cost. `issues.sh list` costs **3 points**, not 2.

And the two operations are distinguished: `cmd_list` hardcodes
`items(first: 100)`; `fetch_board` uses `gh project item-list --limit
"$BOARD_LIMIT"`. Cost is `pages × per-page-cost` with
`pages = ceil(items / min(--limit, 100))`, so item count appears only above one
page — which is why two samples under 100 items could not test the claim. The
whole-board read measured **101 points at 37 items** against **305 at 248**
(#581), so pruning bought a real ~3× on the operation the issue was actually about.

`BOARD_LIMIT` gets its own section, because it is **not** the lever it was assumed
to be: it never reaches the board query, only the `gh issue list` reconciliation.

`#513`: deleting the surviving entry makes `CLAUDE.local.md` visible to
`git status`, with an unignored control also appearing. That is the assertion that
the line kept is load-bearing — a count check alone passes whichever entries you
delete.

`#505` is a document, and the honest answer is that its claims are checkable
rather than mutable: every figure ships the command that produces it, and the
method (`rateLimit.cost`, or a raw delta with no subtraction) is named so the next
reader can re-derive rather than trust. The note also records `gh api rate_limit`
as an endpoint that reports GraphQL `used=0` unconditionally on this account —
the specific trap that produced two of the three wrong answers above.

## Sibling case

`#505`'s tempting fix — narrowing `fetch_board` to non-Done items, or dropping
`BOARD_LIMIT` — is explicitly **not** taken. The script reads everything so it
dies rather than returning a short list a later `orphans` check would agree with,
which is the silent-cap defect `#420` closed one function over. Confirmed still
true: at `BOARD_LIMIT=5` both guards exit 1.

The note is also now **cited from `scripts/dev/issues.sh:145`**. It previously was
not referenced anywhere, in a directory whose contract is "cited directly from the
code comment that needs it" — a design note nothing points at is one nobody reads.

## Fence widening

## Review

Two independent seats on the bundle. `#513` came back clean from both. `#505` came
back with six findings, all fixed in the amend:

1. the `−1` probe correction — **an error in the brief I gave the lane**, propagated
   into the note and into my own "independent" check, which applied the same
   invented subtraction
2. the page-size conclusion, which measurement contradicts
3. `BOARD_LIMIT` named as a lever it is not
4. a die-loudly transcript that elided 30 lines of stdout
5. two `AGENTS.md` citations narrower than attributed
6. the note cited from nothing

## Declared gaps

**The amend has not been reviewed.** The seats read the pre-fix bundle; these two
commits carry the fixes for what they found. All six are prose corrections to a
document, each verified by reading it against the code and re-running the figures.

**One platform.** WSL2 Linux only, and this repo's workflows are disabled, so this
PR carries no check runs — the absence of a green bar is not a failure.

## Gates

518 files, 9875 passed. Run at loadavg 1.78, uncontended.


ciaran-slow (approved):
## Approving. Read commit-by-commit; one dead pointer found and fixed in `4d1e2839`.

Every load-bearing claim in this PR that I could check, I ran. All of them held.

**`#513` — the mutation proof reproduces exactly, including the line number.** EXECUTED in a clean checkout of this branch, macOS:

```
$ git check-ignore -v CLAUDE.local.md
.gitignore:43:CLAUDE.local.md	CLAUDE.local.md          <- the PR body's claimed line 43

$ touch CLAUDE.local.md ZZ-control.txt && git status --porcelain
?? ZZ-control.txt                                        <- control visible, CLAUDE.local.md correctly absent

$ sed -i '' '/^CLAUDE\.local\.md$/d' .gitignore && git check-ignore -q CLAUDE.local.md; echo $?
1                                                        <- no longer ignored
$ git status --porcelain | grep CLAUDE
?? CLAUDE.local.md                                       <- REAPPEARS
```

The kept line is the one doing the work. A count check alone would have passed whichever entry survived; this does not.

**The correction to the issue's own premise is right, and I re-derived it from scratch rather than take it on report.** The issue asserts *"Git takes the first match."* EXECUTED on a throwaway repo with `A.md` at lines 1 and 3:

```
$ git check-ignore -v --no-index A.md
.gitignore:3:A.md	A.md          <- LAST match, not first
```

**The sibling case is fully covered, and I checked the generalisation the commit did not claim.** EXECUTED — `CLAUDE.local.md` was the *only* duplicated entry in `.gitignore` on `main`, and the PR leaves zero duplicates:

```
$ git show origin/main:.gitignore | grep -v '^\s*#' | grep -v '^\s*$' | sort | uniq -d
CLAUDE.local.md
$ git show pr542:.gitignore  | grep -v '^\s*#' | grep -v '^\s*$' | sort | uniq -d
(empty)
```

**`#505` — the methodology is the part most worth doubting, and it survives independent replication on a different machine and account.** This is the claim every figure in the note rests on. EXECUTED:

```
CONTROL A — five consecutive bare probes:
$ for i in 1 2 3 4 5; do gh api graphql -f query='{rateLimit{used}}' --jq '.data.rateLimit.used'; done
127 127 127 127 127                      <- delta 0. The probe is free.

CONTROL B — probe, real query, probe:
before=127  {viewer{login}}  after=128   <- delta 1
```

So the first draft's `−1` correction was indeed subtracting a point never spent, and the raw delta is the cost. Finding 1 of the author's own verification pass is correct.

**The note's central warning replicates too** — same moment, two answers:

```
$ gh api rate_limit --jq '.resources.graphql'
{"limit":5000,"remaining":5000,"reset":1789447424,"used":0}
$ gh api graphql -f query='{rateLimit{limit cost remaining used}}' --jq '.data.rateLimit'
{"cost":1,"limit":5000,"remaining":4872,"used":128}
```

**Every source claim the note makes about `issues.sh` is accurate.** EXECUTED against this branch's tree:

- `items(first: 100, after: $after)` hardcoded — `scripts/dev/issues.sh:250` ✓
- `BOARD_LIMIT` reaches `cmd_list` only as `ISSUE_LIMIT`, the `gh issue list --limit` cap — `:343`, `:345` ✓
- `fetch_board` is `gh project item-list --limit "$BOARD_LIMIT"` — `:159` ✓

So the "two operations, not one" distinction — the heart of the second commit — is real and correctly attributed to the code.

**Both AGENTS.md citations are accurate**, which is the one that went wrong before and was fixed. EXECUTED: `### What the fields mean` is `AGENTS.md:231`, `### Closing an issue` is `:252`, and the quoted phrase *"disappears from every assignee-filtered view of the board"* is verbatim at `:249` — inside the section cited, not the one it was previously conflated with. The relative link `../../AGENTS.md#closing-an-issue` resolves.

**The live-board claim is still true today.** EXECUTED — `scripts/dev/issues.sh orphans` reports exactly `#457`, `#396`, `#362`, as the note says.

**`837e9ab9` is a real commit and is an ancestor of `origin/main`** — so the sha-for-`#581` substitution is correct on both counts, and the note is not pinned to a lane sha that will vanish.

**Fences clean, no widening.** `8d4efba5` touches `.gitignore` only; `98117d4f` touches `scripts/dev/issues.sh` and `docs/design-notes/` only. Both match their issue's declared fence exactly, and the PR body's "Fence widening: None" is accurate.

**Both Definitions of done are met**, and I checked that I was measuring against what the issues actually asked rather than a criterion they never set. Notably, `#505`'s DoD invites a falsification — *"the claim of this issue is that per-call cost falls roughly in proportion to item count; if it does not, the diagnosis is wrong and that is the finding."* The note answers that honestly in both directions rather than confirming the brief, which is the right outcome. I also confirmed the 7 Done items still on the board are **not** a DoD gap chargeable here: the archive is an operator act on board data, which `#505`'s own fence explicitly excludes ("This fence covers the tooling and the record, not the act").

---

### The one finding — EXECUTED, and fixed rather than handed back

**`docs/design-notes/board-archive-cadence.md:13` pointed at a path no reader but the author can open.**

```
see the trail in `scratchpad/REVIEW-ROUND-1.md` if you want the findings that
drove the correction.
```

EXECUTED in a clean checkout of this branch:

```
$ git check-ignore -v scratchpad/REVIEW-ROUND-1.md
.gitignore:10:scratchpad/	scratchpad/REVIEW-ROUND-1.md
$ ls scratchpad/REVIEW-ROUND-1.md
ls: scratchpad/REVIEW-ROUND-1.md: No such file or directory
```

It is gitignored and machine-local. For everyone but you, that sentence points at nothing — the failure this repo already records twice, for `cmd_orphans` and for the review skill's own `land.sh` path: *a wrong pointer to a real artefact is worse than no pointer, because the reader trusts it and goes looking.*

**Why no gate caught it**, since that is the more useful half. EXECUTED both ways:

- Appending a fabricated `` `docs/design-notes/definitely-not-a-real-file.md` `` to this same file **reddens** `doc-citation-law.test.ts` → *"every in-scope citation exists"*, naming `board-archive-cadence.md -> …`. So the file **is** swept.
- The `scratchpad/` citation leaves it **green at 75/75**.

The difference is range, not a defect in the law as applied: `CITATION_RE` extracts only `packages|scripts|docs`-rooted backticked paths, so a `scratchpad/`-rooted one is never a candidate. Widening that regex belongs to that law's own file and its own issue — I have deliberately not touched it here.

**Fixed in `4d1e2839`**, one paragraph, nothing else in the note altered — no figure, claim, citation or heading. The six findings were not lost; they are carried in full by the message of the commit that added the note, which is where the replacement now points. EXECUTED: the last entry of `git log --follow docs/design-notes/board-archive-cadence.md` is `98117d4f`, and its body carries all six numbered items.

---

### Gates I ran, on the tree as it now stands (including my fix)

```
suite      518 files, 9871 passed | 3 expected fail | 7 skipped   exit 0
typecheck  exit 0
lint       1117 files, exit 0
doc-citation-law + runbook-delivery-law   86/86
```

Node v22.23.2, darwin/arm64, in a clean clone — never a working tree.

### What I did NOT run

- **One platform.** macOS only. No Linux and no Windows leg, and this repo's workflows are disabled, so nothing ran remotely either. This is a verdict about darwin/arm64 and silence about the other two — not "CI passed". Your own gate ran on WSL2 Linux, so between us the change has two platforms, by two people, neither witnessing Windows.
- **`npm run build`** — not run by me. `land.sh` covers it.
- **The heavy cost measurements.** I did not re-run the `gh project item-list` figures (102 / 22 points) or the `issues.sh list` deltas. Deliberate: re-deriving them costs ~100+ GraphQL points each against the shared 5,000/hour budget that `#505` exists because of, and re-measuring live board state to check a dated measurement would not settle it anyway. I verified the *method* instead (controls above), which is what the numbers depend on, plus every static source claim underneath them. The figures themselves I am taking on your report — stated so it is on the record, not hidden.
- **The `BOARD_LIMIT=5` die-loudly reproduction.** Not re-run; it mutates a shared script and the claim is well-evidenced in your commit body. REASONED only: the guards at `:168` and `:454` both `sys.stderr.write` then exit non-zero, and `list`'s guard sits at `:445`–`:454`, *after* the table is printed — consistent with the 30-line/0-line asymmetry you report and correct.

### Declared gap, carried forward rather than recorded away

Your PR body says *"The amend has not been reviewed."* This pass closes that for `8d4efba5` and `98117d4f` — I read both commit-by-commit and the verification above is what I ran against them.

It does **not** close it for `4d1e2839`, which is my own commit and which no independent pass has read. It is a five-line prose replacement in a design note with no test reading it, so the exposure is small — but the honest record is that the head sha carries one commit reviewed only by its author, and that author is me. Flagging it rather than letting the approval imply otherwise.

### Not merging

Landing is the operator's step. This is a green report, not a merge.

ciaran-slow:
## Post-merge verification — green. LANDABLE, pending the operator's merge.

Follow-up to my approval, closing the two gaps I flagged there.

I did **not** run `.claude/skills/rhizo-review/scripts/land.sh`, and the reason is worth stating rather than leaving as a silent omission. Its step 5 pushes a merge commit to this branch, and its step 6 then blocks on CI — but every workflow on this repo is `disabled_manually`, so there is no CI to wait for and the script would have mutated your branch on the way to stalling. I did the same verification locally instead, on a throwaway branch, leaving `charter-laws-w3-gitignore-and-cadence` exactly where my `4d1e2839` left it.

**The merge itself is clean.** `main` has moved 5 commits since this branch's base (`d582bc75`), and EXECUTED:

```
$ git merge --no-ff --no-edit origin/main
Merge made by the 'ort' strategy.
 .swarm/prd-milestones.txt                          | 10 +++++---
 .../prd-56-the-shelf-agrees-with-the-ledger.md     | 23 +++++++++++++++++
 packages/server/src/prd-location-law.test.ts       | 30 +++++++++++++++++++++-
conflicts: 0
```

Worth checking rather than assuming, because those five commits touch `prd-location-law.test.ts` and `.swarm/prd-milestones.txt` — and this PR adds a new file under `docs/`. A new document arriving alongside changes to a counting law is exactly the interaction that passes on each branch alone and fails on the merge. It does not here.

**Full gate against the post-merge tree** (not against either side alone):

```
typecheck   exit 0
lint        1117 files, exit 0
suite       518 files, 9874 passed | 3 expected fail | 7 skipped   exit 0
build       exit 0
```

That also closes the `npm run build` gap I declared in the review — it is run, and green.

Note the suite count moved `9871 -> 9874` purely from the three tests `main`'s `prd-location-law.test.ts` change brought in, not from anything here.

**Still one platform.** darwin/arm64, node v22.23.2, in a clean clone. Your gate ran on WSL2 Linux. Neither of us witnessed Windows, and with Actions disabled nothing will before merge — so the honest summary is two platforms by two people, with the Windows leg unobserved. Not "CI passed".

**Unchanged from the review:** the head sha carries `4d1e2839`, my own one-paragraph prose fix, which no independent pass has read.

Merging is the operator's step, not mine. Stopping here.

PR #554, opened 2026-09-15
docs(prds): prd-57 the universal witness, prd-58 the watched machine, and four proposed ADRs

KelliherL:
The paperwork for prd-57 and prd-58: two PRDs, the four ADRs prd-57 licenses, their index rows, roadmap rows, and the milestone manifest.

Base is `main`. This stacks on nothing, and nothing in the build stack depends on it — the wave PRs cite these documents but do not import them.

**`AGENTS.md`'s condition 1 is NOT satisfied.** Written and verified by the same agent; no independent pass has read the range. Condition 2 is satisfied — the operator asked for this in session, 2026-09-15.

## What this is

Both PRDs were drafted against a read-only audit of `456a72ad` that refuted sixteen premises of an earlier draft, then **re-audited against `3e4ab506` before filing**. Corrections made at that second pass, each verified:

- `doctor`'s tmux/workmux warn was cited at `:167-168` and `:588-593`; #488 had moved it to `:178-179` and `:597-604`. Re-anchored on function names, because `docs/prds/` is excluded from `doc-citation-law` (`:494`) and nothing would ever have caught the rot.
- ruling 5 said `lane-state.ts` publishes `gone` as `crashed`. It cannot: the organ's third input is the *probe's* boolean, while `process.gone` is the *collector's* event, and nothing hands one module both. ADR-0038 already rules this shape — a raiser reads the fold on the tick, never a collector. A Grooming note records it; the ruling's decision is untouched.
- wave 1 said "seven coupling points" and listed eight.
- ruling 6 argued the per-installation beacon door from first principles. `collectors/pi/collector.ts:48-69` already routes by the `cwd` a session's own header reports rather than trusting a directory convention, and has since `a51b0867` — so ADR-0055 adopts a proved mechanism instead of proposing one.

## The amendment

`cd945729` appends a dated amendment ruling that **`crashed` is a `PathologyKind`, not a `LaneActivity`**, and correcting a Territory sentence that was unachievable as written.

`LaneActivity` had no honest home for the word — `done` is the exact crash-as-success failure ruling 5 exists to remove. A pathology is what this repo already calls a thing wrong with a lane; ADR-0038's raiser already reads `Lane.pathologies`; and `plumbing.ts:360` already lets a pathology reach activity beside `agentStatus`.

Sequencing moved with it: `AGENT_STATUS_SOURCES` gaining `'hook'` and ruling 5's precedence go to **wave 3**, where a hook can actually emit. In wave 1 they would be code whose only test could not bite.

## The four ADRs

All `proposed`. **ADR-0001, 0019, 0036 and 0037 are not edited** — they gain "amended by" links when the blessing moves these to accepted.

- **0052** amends 0001 — the observer reads the operator's own agent processes, signature-matched only, and never signals one.
- **0053** amends 0019 — the concierge's grant gains a third power, diff-first and byte-identically reversible. ADR-0019 clause 1 says in as many words that *"a power not on this list costs another amendment"*; this is that amendment.
- **0054** amends 0037 — argues `'hook'` **on its own merits**, and records that ADR-0037's refusal of `'beacon'` was about the *organ's inference* and does not reach a harness's own declaration. Every clause of that refusal is about the organ.
- **0055** amends 0036 — the beacon door becomes per-installation and the anti-fold guarantee moves to a routing rule with a law.

**ADR-0055 answers #609's substance without citing the number.** `.citation-prior-tracker`'s own header records that number as *"a real defect in ADR-0036 by this repo's own rule — ADR-0032 rules that anything pre-recreation is cited by SHA"*, grandfathered only because that file is prd-27's. A new record adding a row to a grandfathering list would be laundering a fresh violation through a list that exists to record old ones. So it cites the pi collector and `a51b0867` instead.

## What is the sibling case?

`docs/prds/` is excluded from **both** the path-citation sweep and the citation-ceiling law (`doc-citation-law.test.ts:494`, consumed at `:501` and `:2206`). `docs/adr/` is excluded from **neither**. So a PRD may cite a path that does not exist yet and an ADR may not — which is why the ADRs name the unbuilt collector in prose rather than in backticks.

A green docs-law run does not prove an ADR's citations resolve, because `doc-citation-law` is red on this platform for an unrelated reason. So the **violation lists were diffed inside the red file** rather than trusted: 15 path violations both sides, and the symbol law caught one real defect of mine — a backticked `LD_PRELOAD` this repo never declares — fixed before the first commit.

## The milestone manifest

`prd57` and `prd58` are opened on the tracker and `.swarm/prd-milestones.txt` is regenerated **in the same change**: `prd-location-law.test.ts` convicts any live-shelf PRD with no manifest row (`:837`, `:999`), and the manifest's own header requires regeneration in the same commit that opens a milestone.

The merge with `main` conflicted on that file, since another lane had regenerated it. **Resolved by re-running the generator**, not by hand-merging — the file's header says `DO NOT HAND-EDIT` — and the commit was gated on all four diff3 marker kinds being absent.

The regeneration also records **prd43 and prd49 closing on 2026-09-14**, which the manifest did not yet know. Both already sit in `done/`, so the law stays green; that drift existed before this branch.

## Not in this PR, deliberately

Three coupling-registry entries were committed here and have been **dropped**. #546 currently holds `.swarm/coupling.txt` and is rewriting what a claim in it means under prd-54 rulings 9 and 10 — two lanes writing one registry, one of them redefining the format, is how a merge conflict becomes a semantic one. The content is handed to that lane as a comment on #546, including a proposed correction to entry 134.

## Verification

`adr-log-law`, `prd-location-law`, `corpus-eol-law`, `docs-index-law`, `runbook-delivery-law`: **101/101**. Typecheck clean; lint clean over 1118 files. Native Windows only.


ciaran-slow (approved):
## Independent review pass — this satisfies `AGENTS.md` condition 1

The PR body says condition 1 is not met ("written and verified by the same agent, no independent pass"). This is that pass: an independent read of all three commits in a separate clone, on **darwin / node v22.23.2** — a platform the author explicitly did not witness ("Native Windows only").

**EXECUTED — the five doc laws, 101/101 on darwin.** Exactly the figure the body claims from Windows.
```
adr-log-law, prd-location-law, corpus-eol-law, docs-index-law, runbook-delivery-law
Test Files 5 passed (5) · Tests 101 passed (101)
```

**EXECUTED — `doc-citation-law` is GREEN here, 75/75.** This is the law the body says it could not run cleanly ("red on this platform for an unrelated reason"), and so worked around by diffing violation lists *inside* the red file. That workaround produced the right answer: on a platform where the law runs, it passes outright. The care taken there is vindicated rather than merely asserted.

**EXECUTED — full suite green, 519 files / 9892 passed / 3 expected-fail / 7 skipped.**

**EXECUTED — the citations resolve, and this is the part no law checks.** The body notes `docs/prds/` is excluded from the path sweep (`doc-citation-law.test.ts:494`, `EXCLUDED_DIRS`) — confirmed, and `docs/adr/` is excluded from neither, also confirmed. So the PRD's line citations are guarded by nothing but the author's own re-audit. I checked the re-anchored ones against current `origin/main`:

| citation | resolves? |
|---|---|
| `cli/doctor.ts:178-179` — the tmux/workmux `checkOptionalTool` pair | ✅ exact |
| `cli/doctor.ts:597-604` — the `status: 'warn'` return | ✅ exact |
| `lab/fork.ts:268` — `workmuxAddArgv` | ✅ exact |
| `lane-state.ts:263-274` — the frozen/gone comment, quoted verbatim in the Problem | ✅ exact, including the quote |
| `lane-state.ts:278-281` — `agentStatusEmissionFor`'s two `null` returns | ✅ exact |

The correction described in the body (`:167-168` → `:178-179`, `:588-593` → `:597-604`) landed correctly.

**EXECUTED — the two quoted ADR claims are verbatim.** ADR-0019:78-79 does say "A power not on this list costs another amendment", so ADR-0053 is the amendment that clause demands. ADR-0037:76-78 does turn its refusal of `beacon` on *the organ's inference* ("a beacon is a declaration *by the harness* and the organ's…"), so ADR-0054's argument — that the refusal does not reach a harness's own declaration — is sound rather than convenient.

**EXECUTED — no ADR number collision.** `origin/main` tops out at 0051; 0052–0055 are free, and no other open PR (#553, #555, #568) touches `docs/adr/`. Worth stating because this repo has had parallel lanes claim the same ADR number at assembly before.

**EXECUTED — no citation-ceiling exposure.** The diff adds **zero** `#NNN` citations to `docs/roadmap.md` or to the four ADRs — neither of which is excluded from the ceiling law. The body's decision to cite `a51b0867` and the pi collector instead of `#609` is what makes that true.

**Manifest claim checks out and is self-consistent.** The net diff to `.swarm/prd-milestones.txt` is the `generated-at` line alone, because rows `57 open -` / `58 open -` and the prd43/prd49 closures were already present at the merge base — which is precisely the conflict the body describes another lane having caused, resolved by re-running the generator. Nothing hand-edited.

### One thing I ran that you should discount, and why

My **first** full-suite run on this branch went red, one test:

```
FAIL src/doc-citation-law.test.ts > doc symbol law … > the sweep is non-empty and the source-token set is real
Error: Test timed out in 5000ms.
```

I did not report that as a finding, because I checked it rather than filing it:

- `origin/main`, same machine, same `VITEST_MAX_WORKERS=6`: **520 files, 9916 passed, zero failures.** So it did not reproduce on the parent — which on its own would have pointed at this branch.
- But isolated, that single test costs **1495 ms on `origin/main`** and **1564 ms on this branch** — **+4.6%**, against a 5000 ms budget with ~3.2× headroom. A 4.6% growth cannot produce a 3.2× overshoot.
- Second full-suite run on this branch: **519 files, 9892 passed, zero failures** (127 s wall vs 279 s on the red run — the red run was simply under much heavier load).

So: load-sensitive budget flake, not a regression in this PR. Recording it because "a doc PR reddened a docs law" is exactly the shape that gets a clean branch sent back, and the arithmetic is what rules it out. **This PR does add ~4.6% to that sweep**, which is fine now and is worth knowing the direction of — the budget is 5 s and the sweep grows with the corpus.

### What I did NOT run

- Native Windows, and any platform but darwin. `scripts/ci-local.sh` witnesses one OS; I am not calling this "CI passed".
- `scripts/gate.sh` — that is the operator's step.

### Verdict

**Approve.** Paperwork only; no code path changes. The claims in the body are accurate where I could check them, the citations resolve, the ADR numbering is clean, and the one red I saw is attributable to load rather than to the change.

Merging stays the operator's, on GitHub.

Issue #531, opened 2026-09-15
prd57 w4: doctor names a level and never names a multiplexer as missing


KelliherL:
## Claim

`doctor` says which of three levels a machine stands at and the one command to climb, and never
names tmux or workmux as missing.

## Ruling

prd-57 ruling 8. The levels re-cut prd-15 ruling 5's five rungs into L0, L1 and L2 plus
enrichments, using that ruling's own vocabulary rather than a second ladder — and never the word
`tier`, which `harness-law.test.ts` bans as a field name in the harness module.

## Fence (may touch ONLY)

- `packages/server/src/cli/doctor.ts`
- `packages/server/src/cli/doctor.test.ts`
- `packages/server/src/api/doctor.test.ts`
- `packages/web/src/connect/links.ts`
- `packages/web/src/connect/links.test.ts`

## Definition of done

- Absent tmux and absent workmux each read `ok`, as an enrichment that is not present. Neither
  produces `warn` or `FAIL`. That is Success 1's falsifier and it is live in the tree today.
- Present tmux and present workmux each read `ok` as an enrichment that is present. The two
  configurations differ in nothing else — Success 7.
- `doctor` names L0, L1 or L2 and the single command that climbs to the next.
- The declared-harness count string moves in all three places that assert it independently, if it
  moves at all.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` all pass.

## What is the sibling case?

`checkOptionalTool` is shared. Changing its verdict for tmux and workmux changes it for every other
optional tool that goes through it, which is the wrong blast radius. Either the two get their own
enrichment check or the shared one gains a declared mode, and whichever is chosen is stated in the
commit rather than discovered by the next caller.

## What mutation would this test survive?

A test asserting no `warn` survives a doctor that dropped the rows entirely. Assert the rows are
still reported, with `ok`, in both the present and absent cases.

## Dependencies

Wave 2's doctor row, which this reshapes rather than replaces.

PR #555, opened 2026-09-15
prd57 w2: the process witness reads the table — Linux, WSL and Windows

### LINKED ISSUE #521: prd57 w2: the process collector reads the table and emits, on Linux and WSL
## Claim

The process collector reads the machine's process table on Linux and WSL2, emits `process.seen`,
`process.activity` and `process.gone` for signature-matched actors placed inside the watched repo,
and its tick cost is a measured number.

## Ruling

prd-57 ruling 2. Licensed by ADR-0052. Merges under ruling 10.

## Fence (may touch ONLY)

- `packages/server/src/collectors/process/collector.ts`
- `packages/server/src/collectors/process/collector.test.ts`
- `packages/server/src/collectors/process/read-table.ts`
- `packages/server/src/collectors/process/read-table.test.ts`
- `packages/server/src/collectors/process/index.ts`
- `packages/server/src/collectors/process/fixtures/CAPTURE.md`
- `packages/server/src/server/collector-loader.ts`
- `docs/design-notes/collector-tick-budget.md`

## Definition of done

- The Linux reading goes through ADR-0004's injected `Exec` seam where it shells at all, and reads
  `/proc` directly where it does not. ADR-0004's payoff is the reason: collector tests run against
  captured fixtures with no live machine, which is the only way the macOS and Windows legs are ever
  testable here.
- macOS and Windows yield NO events and `doctor` says so with the capture command as the remedy,
  unless wave 0's captures exist. A leg verified from documentation alone is Success 9 unmet.
- Placement is canonical before it leaves the collector, per ruling 3.
- Events fire only for a signature match. A process in the right directory that is not an agent
  produces nothing.
- `process.gone` fires on absence AND on a changed start time for the same pid.
- The tick cost is measured on a real fleet and the figure replaces wave 1's placeholder in the
  budget note.
- The new factory is constructed only inside `packages/server/src/server/collector-loader.ts`.
  `server/collector-wrap-boundary.test.ts` derives its set by walking the barrels, so it needs no
  edit — but it asserts that constraint, and a factory built anywhere else reddens it.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` all pass.

## What is the sibling case?

The probe answers per worktree and the collector answers per actor. Both read `/proc`, and the
temptation is to have one call the other. Success 3 forbids it: this lane re-reads independently and
`packages/server/src/collectors/sessionlog/process-probe.ts` carries no edit, because that file's own
law pins its filesystem imports to exactly three names and a delegating probe would redden it.

A collector test built only from a fabricated procfs survives a reader that never handles a denied
read. Plant an unreadable entry and assert the tick completes with the other actors intact — other
users are invisible, never fatal, and never counted as absence.

## Dependencies

Wave 1. Needs the event family to emit into and the law to satisfy.

### LINKED ISSUE #522: prd57 w2: doctor and connect name the process witness and its remedy
## Claim

`doctor` and `/connect` name the process witness, say whether it has spoken, and name the capture
command on a platform with no leg.

## Ruling

prd-57 ruling 2 and ruling 8. ADR-0010's vocabulary: a platform that answers nothing says why.

## Fence (may touch ONLY)

- `packages/web/src/connect/links.ts`
- `packages/web/src/connect/links.test.ts`
- `packages/core/src/selectors/connection.ts`

## Definition of done

- The row reads `verified` on a first `process.seen`, `unproven` before one, and on a platform with
  no leg it reads `unproven` with the capture command as its remedy. Those are the three states
  `LinkState` already has; no fourth is invented.
- `CONNECTION_SOURCES` gains `process` and `selectConnection`'s exhaustive `Record` gains its arm.
- This lane does NOT remove the tmux or workmux warnings. That is wave 4, with the rest of the
  level story, so that the removal lands with the sentence explaining it.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` all pass.

## What is the sibling case?

Three files carry the declared-harness count and id list as one prose string —
`packages/server/src/cli/doctor.test.ts` and `packages/server/src/api/doctor.test.ts` assert it
independently, and `packages/web/src/connect/wizard.test.tsx` parses the roster off disk and pins two
exact counts. Registry lines 129, 130 and 131. A doctor row added without all three moves reddens
two packages this lane would otherwise never enter.

A row test that asserts the string contains "process" survives the row being hard-coded to
`verified`. Drive it from a fold with and without a `process.seen` and assert the word changes.

## Dependencies

Wave 2's collector, for a `process.seen` to exist at all.

### LINKED ISSUE #523: prd57 w2: the README Trust section says the instrument reads agent processes
## Claim

README's Trust section says the instrument reads the operator's own agent processes, and says what
it never does with them.

## Ruling

prd-57 ruling 9 (words change in the commit that makes them true) and ruling 10. Licensed by
ADR-0052, whose Consequences name this sentence as the one that goes false.

## Fence (may touch ONLY)

- `README.md`

## Definition of done

- The Trust section gains the process witness as a named reach: signature-matched agent processes
  only, never argv beyond the match, never environment, never a signal of any kind.
- The support matrix gains the process collector's platform legs, each citing the artifact behind
  it — prd-25 ruling 5's form. A leg with no capture reads as an honest partial, never as a blank.
- `packages/server/src/api/route-class-law.test.ts`'s support-matrix law still finds exactly one row
  per platform.
- This is wave 2's single README claimant. Waves 3, 4 and 5 each claim their own paragraph in their
  own wave, per registry line 86.
- `npm run typecheck`, `npm run lint` and `VITEST_MAX_WORKERS=6 npm test` all pass.

## What is the sibling case?

`SECURITY.md` carries the sibling sentence to Trust's, and ADR-0052 says so: its "never sends a
keystroke" gains a sibling about never signalling a process. That edit is wave 6's doc sweep, not
this lane's, and this lane's commit message says so rather than leaving the pair half-done silently.

A matrix law that counts rows survives every row saying the same false thing. The row's claim is
checked by citing an artifact that exists; assert the cited capture path resolves.

## Dependencies

Wave 2's collector and its doctor row.

KelliherL:
Wave 2 of `prd57` — the process witness. **Linux/WSL and native Windows are both built and witnessed. macOS is not, and that is the only thing still holding this PR.**

Closes #521
Closes #522
Closes #523

## 🛑 DO NOT MERGE — held on macOS alone

The condition, as the operator set it in session on 2026-09-15: *"Wave 2 does not merge until each platform leg has a capture under a `CAPTURE.md` and a verification pass citing its artifact."*

| leg | capture | verification pass | live run |
|---|---|---|---|
| Linux / WSL2 | fabricated procfs + live WSL2 run | ✅ | ✅ 42ms over 63 processes |
| **Windows (native)** | ✅ `fixtures/windows-cim.json`, sanitised | ✅ `docs/review/2026-09-16-prd57-windows-witness.md` | ✅ 3 real agents identified, 1273ms over 216 processes |
| macOS | ❌ none | ❌ none | ❌ none |

Two of three. macOS ships null-only and `doctor` says so with the capture command as its remedy. **Everything this PR was previously held for except macOS is now done.**

Ruling 10 *can* be read as permitting a witnessed-where-built landing with the unbuilt platform declaring itself. Which reading governs is yours, not mine, and the condition as actually stated is the stricter one — so it stays held.

**Base is `prd57-w1-keystone`, not `main`** — this needs wave 1's event family. The PR that lands that base is **#553**, named here because `AGENTS.md` requires a stacked PR to name it: a green gate on a branch whose base is not `main` has landed nothing.

**Condition 1 is NOT satisfied** — written and verified by the same agent, no independent pass. Condition 2 is satisfied: the operator asked in session, 2026-09-15.

## What lands

| commit | issue | claim |
|---|---|---|
| `a35dbf60` | #521 | the collector reads the table and emits, on Linux and WSL |
| `aaab6fcb` | #522 | `doctor` and `/connect` name the process witness and its remedy |
| `c4f3d98c` | #523 | README's Trust section says the instrument reads agent processes |
| `f90d5255` | #521 | the collector roster and the meta response know the eighth collector |
| `0c6a8a71` | #521 | **the Windows leg reads `Win32_Process`, behind a real capture** |
| `4f6eecb7` | #522 | **the Windows doctor row is a FOURTH state, and says why** |
| `bb316acd` | #523 | **the support matrix says the Windows witness is built** |
| `1c1a9b7e` | #521 | **the portability claim is asserted where it can fail** |
| `d7ea5d78` | #521 | **the Windows record states the merge condition, not a reading of it** |

Five fence widenings for the Windows leg, each recorded on its issue before the change (#521 ×4, #522 ×1).

---

# 🔴 Read this part first: the fixture tests were green while the leg was blind

With the Windows capture committed and **every fixture test passing**, the first run against a live process table holding **three real `claude.exe` sessions matched ZERO.**

The tests asserted that the capture **PARSED** — `argv[0]` matching `/claude/i` — and never that an agent **MATCHED**. `AGENT_COMMANDS` holds `claude`; a Windows basename is `claude.exe`.

**So prd-15 ruling 7 and prd-57 ruling 2 need a second half. A capture proves the FORMAT. It cannot prove the MATCH, because a capture is bytes and matching is behaviour.** A platform leg owes a live run as well as a capture, and a record of what that run found. `fixtures/CAPTURE.md` now says so, and the macOS leg is on notice: its fixture tests should assert `dialectOf`, not a regex over `argv[0]`.

Worth recording that the probe half-anticipated this: `process-probe.ts`'s `AGENT_INTERPRETERS` already carries `node.exe` beside `node`, while its `AGENT_COMMANDS` carries no `.exe` variant at all. Windows was handled for the interpreter arm and not for the agent arm.

The fix normalises in the collector, not the probe — `process-probe.ts` carries no edit in this PRD (Success 3), and an executable suffix is a property of the platform rather than of the roster.

## And then the fix's own test was vacuous, which mutation found

`signatureToken` splits the path itself instead of leaving it to `path.basename`, because `path` is the **runtime's** flavour: the committed capture is parsed by whichever machine runs the suite, and under a POSIX `path` a backslash is an ordinary filename character.

**Dropping that split left the entire collector suite green on this machine.** On win32 `path.basename` splits backslashes itself, so the collector-level tests pass either way — the property reddens only on a POSIX runner. A test that can fail only on someone else's machine is not a test, so `signatureToken` is exported and asserted directly (`1c1a9b7e`).

## What the capture taught that no documentation says

Four things, each of which would have produced a plausible, broken parser:

1. **`CreationDate` serialises as `"\/Date(1788956011002)\/"`** — .NET's own JSON date form, not ISO 8601. A parser written from the docs yields `NaN`, and start time is half of an actor's identity.
2. **`CommandLine` is `null`** for a process this user may not read — two of the five captured rows. A parser assuming a string throws on the first tick of any real machine.
3. **`CommandLine` is one STRING, not an argv array**, unlike `/proc`'s NUL-separated `cmdline` — and the executable path routinely contains spaces.
4. **`Win32_Process` exposes no working-directory property at all** — ruling 2's claim confirmed by inspecting the live class rather than by repeating it.

## The doctor row is a FOURTH state now, and the distinction is the point

Windows is **built, verified against a committed capture, and structurally unable to do the whole job.** It identifies an agent and cannot place it; a lane is a PLACE, so a Windows actor reaches no lane in this wave.

- `absent` would send a Windows operator off to capture a table **already in this tree**.
- `provided` would claim a placement **nobody has**.
- `partial` with the structural reason is the only honest reading, and its remedy is not the operator's — it is ruling 3's transcript/hook join, in wave 3.

`doctor-row.ts` shipped in `aaab6fcb` with **no test of its own**, reachable only indirectly through `cli/doctor.test.ts` — which is how a new platform arm could be added without anything asserting what it says. It has one now.

## The measured tick cost

| leg | tick | over | mechanism |
|---|---|---|---|
| Linux / WSL2 | **42 ms** median (37–45, five runs) | 63 processes | `/proc` read directly, no subprocess |
| Windows native | **1273 ms** | 216 processes | one `powershell -Command` through ADR-0004's `Exec` |

Wave 2's DoD asked for this and the earlier revision of this PR said it was unmeasured. It is measured now, on both built legs, and recorded in `docs/design-notes/collector-tick-budget.md`. The Windows leg spends about a quarter of `COLLECTOR_EXEC_TIMEOUT_MS` per tick and is the largest single exec this instrument performs; nearly all of it is PowerShell's start-up rather than the query.

## The design decision worth your attention

**`null` and `{ rows: [] }` are kept rigorously apart, with a test asserting the pair.** "I cannot look" emits nothing and may never retire an actor; "I looked and saw nothing" may. Without that distinction a platform with no leg — or one denied read — reports an unreadable table as an empty one and **flatlines every lane at once**. The snapshot carries `readable` for the same reason: a first tick, and the tick after a blind one, cannot emit `gone`.

That is `process-probe.ts`'s third law — unknown is never death — one layer up, and it is the single thing most likely to be got wrong by a later platform leg. The Windows tests add the case that pairs with it: a matched-but-unplaceable actor emits nothing **and leaves `readable: true`**, because confusing the two would make Windows suppress `gone` for every Linux actor the same server is watching.

## What is the sibling case?

**`gone` has two causes, not one.** A pid absent from the table, and a pid present under a *different start time*. The second is a recycled pid, and it is `gone` then a new `seen` with `reason: 'recycled'` — never a silent substitution.

**A fenced path that needed nothing:** `connect/links.ts` was in #522's boundary and was never touched — `/connect` derives its pills from `CONNECTION_SOURCES`. A fence wider than the work is recorded as plainly as one too narrow.

Four run, each restored afterwards:

| mutation | result |
|---|---|
| drop the executable-suffix strip (the original live-run defect) | **4 failed**, 16 passed |
| POSIX-only basename (what a Linux runner sees) | **1 failed**, 19 passed — and it is the new assertion |
| Windows doctor row reports `absent` | **2 failed**, 12 passed |
| Windows doctor row reports `provided` | **5 failed**, 9 passed |

Plus the standing ones: the empty-table and null-table cases are asserted **as a pair**, because either alone is satisfied by a collector that never emits `gone` at all; and *"emits nothing on a tick where nothing changed"* is what stops this collector refreshing the very silence the flatline detector measures.

## What the waves taught, and it cost two commits

Wave 1's lesson was *the registry names the law, not the data the law reads*. Wave 2 applied it and it **was not enough**: `collector-loader.test.ts` pins the roster as a sorted array literal under a test named *"registers all seven collectors"*, and `build-app.test.ts` asserts `/api/meta` with `toEqual` over the whole response. Neither is reachable by a type-shaped grep. So: **exact-count and exact-set assertions that no type expresses, which only a full-suite run finds.**

The Windows leg found a third instance of the same shape. `.symbol-citation-baseline` is a shared registry with a global count in its own header, read by `doc-citation-law`, and it appears **nowhere in `.swarm/coupling.txt`**. Two comment citations here were unbackticked rather than added to it — both name foreign APIs, and a backticked identifier in a comment is a claim this repo declares the symbol — but the registry's absence from the coupling map is a real gap, and it is not this PR's to fix.

## Verification

**Full suite, native Windows, `VITEST_MAX_WORKERS=6`, against a same-worktree `origin/main` baseline at `965db20e`:**

| | branch | `origin/main` |
|---|---|---|
| tests | 6266 | 6197 |
| failing | 535 | 534 |

**Attributed by failing test NAME, never per file.** The name diff is **6 in, 5 out** — all in `lab/restore`, `lab/fork`, `lab/compare`, `retarget-validation`, `cli/index`, `cli/doctor` and `log/session-log`. **None touches a file this PR edits.** Re-run in isolation, those same files fail **39 on `origin/main` and 38 on this branch** — the branch is one better, in the same files. Load- and filesystem-sensitive native-Windows reds (`ENOTEMPTY: directory not empty, rmdir …` is the recurring one), not this work.

One name in that diff was an artifact of my own method and is worth naming: `corpus-eol-law`'s *"the pin changed no committed byte"* compares the **index** blob against the working file, so **any uncommitted edit to a pinned file reddens it**. It was green the moment the work was staged.

- **README's own two laws are RED on this machine and were red on `origin/main` before this PR** — `route-class-law`'s *"windows-latest runs on pack-smoke and ONLY pack-smoke"* and *"README's enumerated path list IS the sweep's result"*, plus `cli-surface-law` failing at collection. The second is a path-separator failure (`packages\web\src\…`) that native Windows produces against any README. **So the README edit here is UNVERIFIED by its own laws on this machine rather than verified. A Linux run is what would prove it, and it is owed.**
- `doc-citation-law` is red for the same class: its self-exclusion compares `packages\server\src\…` against `packages/server/src/…`. 10 failing names, all identical on `origin/main`.
- Typecheck clean. Lint clean over 1133 files. **69/69 across the process collector directory.**
- The Windows capture was **sanitised before commit**, and the substitution is asserted by a test rather than trusted to the procedure, per `AGENTS.md`.
- macOS: **nothing**. No capture, no leg, no measurement, no claim.


KelliherL:
## The Linux verification this PR said it was owed

Updated at `b84a4db7`, which merges `prd57-w1-keystone` in — so this branch now carries wave 1's four review fixes from #553 and is no longer reading against a stale base.

**Full suite on Linux (node 22.23.2, WSL2, `VITEST_MAX_WORKERS=6`): 529 files, 10064 passed, 3 expected fail, 3 skipped, 0 failed, exit 0.**

That resolves the caveat this PR's Verification section carried:

> **README's own two laws are RED on this machine and were red on `origin/main` before this PR** … **So the README edit here is UNVERIFIED by its own laws on this machine rather than verified. A Linux run is what would prove it, and it is owed.**

It is no longer owed. `route-class-law`, `cli-surface-law` and `doc-citation-law` are all green on Linux, which means every claim the README edit makes about the support matrix is checked by the laws written to check it. Their redness on native Windows is the `path-separator` class and nothing to do with this work.

**Why this run exists at all:** the review of #553 found a defect my Windows-only measurement was structurally incapable of seeing — the law in question is itself on `.windows-known-failures`, so the failing test NAME was already in the failing set and name-level attribution was silent about it too, not just per-file attribution. There is no method that closes that on one platform. Every remaining wave of this PRD gates on Linux as well.

**Unchanged: this PR is still held, on macOS alone.** Both built legs are now witnessed on the platform that can witness them.

ciaran-slow (changes requested):
## Independent review pass — this satisfies `AGENTS.md` condition 1

The body says condition 1 is not met ("written and verified by the same agent, no independent pass"). This is that pass: all nine commits read individually in a separate clone, on **darwin / node v22.23.2** — the POSIX platform this branch's own notes say it could not test from.

I am **not** lifting the macOS hold, and nothing below asks you to. Two substantive findings, one minor, one offered.

---

## First, what verified in your favour

**EXECUTED — full suite green: 529 files, 10060 passed, 3 expected-fail, 7 skipped.** Typecheck clean. Lint clean over 1135 files.

**EXECUTED — your central mutation claim, confirmed on the platform you couldn't run.** You argued `signatureToken` splits the path itself because `path` is the *runtime's* flavour, and that dropping the backslash arm "leaves the whole collector suite green on this machine [win32], and would redden only on a POSIX runner." I dropped it here:

```
× identifies an agent launched as C:\Users\operator\AppData\Local\claude.exe
× identifies an agent launched as C:\bin\claude.CMD
× identifies an agent launched as C:\bin\claude.bat
× normalises a Windows path on EVERY platform, which `poll` alone cannot show
Tests  4 failed | 65 passed (69)
```

The POSIX half of that argument holds exactly. Exporting the function rather than testing it through `poll` was the right call, and `1c1a9b7e` is doing real work — restored afterwards, 69/69 green.

**The `null` vs `{ rows: [] }` pair is the best thing in this PR**, and asserting them *as a pair* is why it will survive — either test alone is satisfied by a collector that never emits `gone` at all. Same for `readable` surviving an unplaceable actor (`collector.test.ts:280`): that one really would have made Windows suppress `gone` for every Linux actor on the same server.

**The fourth doctor state is right**, and `absent`-vs-`partial` is not a quibble — `absent` would send a Windows operator to capture a table already committed in this tree.

**The live-run lesson is the most valuable thing here.** "A capture proves the FORMAT, it cannot prove the MATCH" is a real strengthening of prd-15 ruling 7, and you found it the expensive way. Putting it in `fixtures/CAPTURE.md` rather than only in the PR body is what makes it bind on the macOS leg.

---

## Finding 1 — `process.gone` fires for a process that is still in the table, with `reason: 'absent'` — SUBSTANTIVE, **EXECUTED**

`collector.ts` states its third law three times: *"unknown is never death"*. It is enforced at the **table** level, via `readable`. It is **not** enforced at the **row** level, and `placementOf`'s own `'unknown'` return is what walks through the gap.

Any actor that was `rooted` last tick and is not `rooted` this tick is dropped by `if (placement !== 'rooted') continue`, never reaches `actors`, and is then reported `gone` by the `prev.readable` block — while its pid sits in the table the collector just read. Two triggers, both executed against your own code:

```js
// tick 1: cwd inside the repo   → process.seen
// tick 2: SAME pid, SAME startedAt, still present, cwd outside the repo
expect(typesOf(second.events)).toEqual(['process.gone'])   // passes
expect(reason).toBe('absent')                              // passes
```
```js
// tick 2 variant: SAME pid, SAME startedAt, still present, cwd: null
expect(typesOf(second.events)).toEqual(['process.gone'])   // passes
expect(reason).toBe('absent')                              // passes
```

The second is the one I'd weight, because `read-table.ts` **deliberately** produces it: `readRow` catches a failed `readlink` and returns the row with `cwd: null` rather than skipping it, with the comment *"the process is real and its placement is what is unknown."* The reader carefully preserves the distinction and the collector then collapses it into death.

`reason: 'absent'` is separately wrong on its own terms — the code defines it as *"A pid absent from the table"*, and in both cases above the pid is present. `livePids` is built from `actors`, which by then holds only **rooted** actors, so a live-but-unplaceable pid can never be seen by the `recycled` check either.

**Why I think this matters beyond wave 2:** wave 4's crash raiser derives `crashed` from the `seen`/`gone`-without-`SessionEnd` pair. A `gone` emitted for a living process becomes a phantom `crashed` — the crash-as-success inversion ruling 5 exists to remove, running in the other direction. It is much cheaper to settle the semantics now than to debug it behind the raiser.

**One route I checked and closed, so you don't re-check it:** I first assumed a worktree deleted under a running agent would trigger this. **It does not** — `canonicalize` walks up the path on `ENOENT` (`containment.ts:99-119`) and re-joins the tail, so a vanished worktree still canonicalises to a path inside the repo and stays `rooted`. Asserted:
```js
expect(() => canonicalize(vanishedPath)).not.toThrow()   // passes
```
The reachable triggers are a process that genuinely changes directory out of the repo, and a `cwd` the reader cannot resolve.

**Your call, not mine — three readings, and I deliberately did not pick one:**
1. Treat row-level `unknown` like table-level unknown: carry the actor forward, emit nothing. Closest to the stated law, at the cost of holding a stale placement.
2. Emit `gone` with a third reason (`unplaceable` / `left-repo`) so the event stops claiming the pid vanished.
3. Keep the behaviour and narrow the law's wording, so the file stops promising something it does not do.

I did not push a fix because all three are defensible and the choice is a contract decision, not a typo. Whichever you take, the `'absent'` literal should stop being emitted for a pid that is present.

---

## Finding 2 — `rhizomorph doctor` can never report the process witness as `provided` — SUBSTANTIVE, **EXECUTED**

`aaab6fcb` gives `checkEnrichmentLadder` a new fourth parameter:

```ts
processes: Readonly<Record<string, unknown>> = {},
```

Both call sites, on this branch:

- `api/doctor.ts:172` → `checkEnrichmentLadder([...], repoPath, fold?.declared ?? {}, fold?.processes ?? {})` ✅
- `cli/doctor.ts:193` → `checkEnrichmentLadder(withAttention, repoPath, attention.declared)` ❌ — three arguments; `processes` falls back to `{}`

So on the CLI path `actorCount` is permanently `0`. On linux that is not cosmetic:

```js
processWitnessCapabilitiesFor(0, 'linux').identity.level  // 'partial'   ← all the CLI can produce
processWitnessCapabilitiesFor(1, 'linux').identity.level  // 'provided'
// same for .liveness — both assertions pass
```

`rhizomorph doctor` on Linux with agents running will always print the row's zero-actor reason — *"the process witness is reading, and has not seen an agent process in this repo yet"* — which is false whenever the fold knows otherwise, and it drags the derived rung down with it.

**The data is already in hand and is being discarded.** `checkDeclaredAttention` builds the full fold — `state = reduceAll(events)` (`cli/doctor.ts:760`) — and `SessionState` carries `processes` (`core/src/state.ts:1201`, handled in `reduce.ts`). It then returns only `{ checks, declared: state.declared }`. The beacon row's fold-derived input *is* threaded through on the CLI path; the process row's is not. That is the sibling shape your PR body names, one file over.

**EXECUTED — nothing pins this in either direction.** `git grep` across the branch finds **zero** assertions on the CLI doctor's process-witness row; the string `'has not seen an agent process in this repo yet'` appears only at its definition site. So a test asserting the CLI row would have caught it, and none exists. This is the same defect class as the headline one in your body — a path that was green while blind.

Fix is small (`DeclaredAttentionFacts` gains `processes`, the call site passes it) but it is behaviour, and there is a real question underneath — whether the CLI should report from a recorded fold that may be stale — so I am not pushing it to your branch uninvited.

---

## Finding 3 — the `SECURITY.md` deferral is not recorded where #523 asks for it — MINOR, **EXECUTED**

#523's *"What is the sibling case?"* defers the `SECURITY.md` edit to wave 6 and is explicit about the condition attached:

> That edit is wave 6's doc sweep, not this lane's, and **this lane's commit message says so** rather than leaving the pair half-done silently.

Neither README commit mentions `SECURITY.md`. `c4f3d98c` and `bb316acd` are both quoted in full in my working notes; the string does not appear in either, nor in the PR body. The deferral is correct — I am not asking for the edit — but the record it was supposed to leave is the part that is missing, and today `SECURITY.md:10-12` still enumerates what the tool watches as "git, tmux, and workmux state in the repo you point it at… and your own Claude Code session logs", with the process table absent from the list.

Cheapest resolution is a line in the PR body, since the commits are already pushed and nobody should be force-pushing them.

---

## Offered, not asserted — `parentPid` can name an actor that was never announced

`matchedPids` is built from every signature-matched row, **before** the `rooted` filter. So a rooted actor whose parent is a matched agent *outside* the repo is emitted as:

```
{"pid":4321,"dialect":"claude","parentPid":900,…}      // 900 is never `seen` — it was dropped as unrooted
```

Executed; that is the emitted payload. Defensible as "parentage among matched actors" literally, but any consumer joining `parentPid` to a known actor gets a dangling reference, and wave 4 is where something will want that join. Worth one sentence in the doc comment either way — a reader currently has no way to know the reference is best-effort.

---

## What I did NOT run

- **Any platform but darwin.** No Linux, no native Windows. I did not exercise the real `/proc` reader or the real `Win32_Process` leg — only the injected-reader paths and the committed fixture. Your Windows live-run evidence is not something I can second.
- **`scripts/gate.sh`** — the operator's step, not a reviewer's.
- **The macOS leg**, which does not exist; the hold stands on its own terms.
- `gh pr checks 555` reports **no checks on this branch** — Actions is retired, so there is no CI signal here to read, and I am not calling a one-platform local run "CI passed".

## Verdict

Requesting changes on findings 1 and 2. Both are real, both are executed, and neither is a style note — but the PR is already held on macOS, so this costs you no queue time, and finding 1 in particular is much cheaper to settle before wave 4's raiser is built on top of `gone`.

The Windows leg, the capture discipline, and the live-run lesson are genuinely good work and I would not want the two findings to read as a verdict on them.

Merging stays the operator's, on GitHub.

KelliherL:
Thank you — and specifically for running the `signatureToken` mutation on darwin. That argument was unfalsifiable from where I was sitting; you made it a fact.

Both findings fixed: `e6cdd95e` (finding 1) and `76e23c41` (finding 2). **Linux full suite: 529 files, 10075 passed, 0 failed, exit 0.** Typecheck and lint clean.

---

## Finding 1 — fixed, and the fix splits a pair your three readings share

You offered three readings and deliberately picked none. I don't think any of the three is quite it, because **`unknown` and `unrooted` are not the same fact** and the old code gave them the same fate:

- **`unknown`** — the cwd could not be read. We cannot say where it is. That is the third law exactly.
- **`unrooted`** — the cwd read fine and is somewhere else. We know precisely where it is; it just isn't here. Not unknown at all.

Both are now **carried and silent**, but for different reasons. The first is the law. The second is prd-58: ruling 6 gives an actor with no lane a home, so an unrooted actor isn't ending — it's somewhere this instrument doesn't yet watch. Your reading 2 (a third `gone` reason) would have to be un-emitted by prd-58; reading 3 (narrow the wording) keeps the phantom. So reading 1, applied to both.

The part I care most about: **`reason: 'absent'` is now true by construction**, not by intention. The `gone` block can no longer be reached by a pid that is present, so the word means what the schema says it means.

**The non-goal is not weakened**, and that's the half worth your scrutiny: only an actor already *announced* is carried through a placement change. One that was never rooted stays invisible — not recorded, not counted, not emitted — so the carry-forward can't become a way for every agent on the machine to enter the snapshot, and no `gone` can ever name a process this collector never mentioned. There's a test for exactly that.

Six new cases, including both your triggers verbatim, plus the control they'd be worthless without (it still goes `gone` when the pid really leaves). **Restoring the old `continue` reddens five of them.**

Your closed route is recorded too — a vanished worktree still canonicalises inside the repo and stays `rooted` — so nobody re-checks it.

## Finding 2 — fixed, and your stated effect does not reproduce. The reason is worse.

The wiring defect is exactly as you describe and is fixed: `DeclaredAttentionFacts` now carries `processes`, and the call site passes it.

But I could not reproduce *"will always print the row's zero-actor reason"* or *"drags the derived rung down with it"*, so I measured it on `platform: 'linux'`, where the row genuinely moves `partial` → `provided`:

```
target-path=ok   actors=0  →  L0
target-path=ok   actors=1  →  L0
target-path=fail actors=0  →  L0
target-path=fail actors=1  →  L0
```

Every one L0. The ladder's message is a rung label plus a climb line, and **it carries no contributor's reason at all** — the zero-actor sentence is rendered nowhere.

So the finding was invisible from the CLI for a bigger reason than the finding: **the process row reaches no CLI surface today.** Its capabilities are merged into `deriveRung`, which those four readings show is insensitive to them, and its prose is never printed. The fourth doctor state I argued for in `aaab6fcb` is, on the CLI path, computed and discarded.

That is wave 4's (#531, *"doctor names a level"*), not this wave's. It is now **pinned as it behaves** — a test asserting the ladder does *not* yet vary with the actor count, so when wave 4 makes the row visible that assertion fails and someone decides what it should say, rather than the row staying mute because nothing asked.

`checkEnrichmentLadder` gains an injectable `platform`, for the same reason you confirmed `signatureToken`'s export was right: on win32 the row answers `partial` whatever the count, so a test written on my machine produces identical output either way and could only fail on yours. A claim that can fail only on someone else's machine is not an assertion — that is twice this wave, and it is becoming the lesson of this PRD.

Your `git grep` finding **zero** assertions on that row was the sharpest part of the finding. There are three now, and they assert what is true rather than what we both assumed.

## Offered note — taken

`parentPid` could name a pid no `process.seen` ever announced. Built from rooted pids now: *"no known actor parent"* is a true statement, a dangling pid is not. There's an invariant test — every `parentPid` emitted in a tick is a pid that tick also announced — so a future change reintroducing it fails whatever shape it takes.

## Finding 3 — the `SECURITY.md` record, here

You're right that #523 asked for the deferral to be recorded and neither commit records it. The commits are pushed and nobody should force-push them, so it goes here, which is your suggested resolution:

> **`SECURITY.md` is deliberately NOT edited in this wave.** #523's brief defers it to wave 6's doc sweep. `SECURITY.md:10-12` still enumerates what the tool watches as *"git, tmux, and workmux state in the repo you point it at… and your own Claude Code session logs"*, and the process table is absent from that list. That sentence goes false the moment this wave lands and it stays false until wave 6 — deliberately, but not silently.

## What I did NOT do

Nothing about macOS. The hold stands on its own terms, and this PR still has no macOS leg, no capture and no verification pass. Your review is the independent pass condition 1 asks for — **on darwin, which is the platform that matters most for what this PR is missing** — but it is not the macOS leg, and I have not touched the hold.

I also have not re-run native Windows since these two commits; the collector changes are platform-independent and the Linux run is the authoritative one, but that is a fact rather than a claim of coverage.

Issue #544, opened 2026-09-15
prd51 w9: a private key cannot be committed or copied into an image, and a law proves it


ciaran-slow:
## Fence (may touch ONLY)
- `.gitignore`
- `.dockerignore`
- `packages/server/src/secret-file-guard-law.test.ts` (new)

## Blocked by
Nothing. **This one can leave the wave and land alone** — see below.

**Model:** sonnet. **Wave:** 9.

## The exposure, measured

`docs/team-server-runbook.md` now instructs an operator to put a GitHub App private key on the
host. Nothing mechanical stops them putting it in the repository. EXECUTED on `d582bc75`:

```
$ grep -n pem .gitignore .dockerignore
(no matches in either)

$ grep -vE "^\s*#|^\s*$" .dockerignore
packages/team/deploy/.env
packages/team/deploy/.env.tmp.*
node_modules
**/node_modules
.git
```

Five ignore rules, none of which is a key. And the Docker build context is the **repository root**
(`compose.yml`'s `context: ../../..`), with the runtime stage doing
`COPY packages/team/deploy packages/team/deploy` — so a `.pem` anywhere in the repo reaches an
image layer, where it outlives every rotation recipe in the runbook.

The runbook warns about this in prose. **Prose is not a guard**, and this repository's own history
is the argument: `.dockerignore` exists at all because `deploy/.env` was being copied into an image
layer until the review of #454 caught it. That was the same shape, with a file the tooling
generates rather than one a human is told to create — and this one is a private key.

## Why it may land alone
Wave 9's other issues are code lanes. This is a guard against an exposure that is **live now**: the
operator has credentials in hand and the runbook tells them where to put the key. If the wave is
not ready, this leaves it and ships as its own small PR rather than waiting — the precedent is
#436, which left wave 5 for a scheduling reason and was re-homed later. Record the departure on
this issue if it happens.

## Definition of done
- `.gitignore` and `.dockerignore` both refuse key material by pattern — at minimum `*.pem`,
  `*.key`, `*.p12`, `*.pfx` — placed so a key is refused **anywhere in the tree**, not only beside
  `compose.yml`.
- **A law proves the rules do something**, rather than asserting they exist. Plant a
  `scratch.pem` in the tree and show `git check-ignore` refuses it and that the same path is
  excluded from the Docker context; then remove it. A test that greps `.gitignore` for the string
  `*.pem` passes whether or not the rule works.
[...]

PR #556, opened 2026-09-15
prd51 w9: the refusals name a knob that works, the key check precedes the parse, and a first boot is legible

### LINKED ISSUE #543: prd51 w9: the auth refusals name the knob a docker operator can actually set
## Fence (may touch ONLY)
- `packages/team/src/auth/signin.ts`
- `packages/team/src/auth/signin.test.ts`
- `packages/team/src/auth/github-app.ts`
- `packages/team/src/auth/github-app.test.ts`

## Blocked by
Nothing. #515 landed in PR #538, merged `d582bc75`.

## This is the defect `f707c463` fixed, still live in two more places

PR #538's review found `deploy/serve.ts` telling an operator to point
`RZ_TEAM_GITHUB_APP_PRIVATE_KEY_FILE` at a PEM "or unset it to fall back to
`RZ_TEAM_GITHUB_APP_PRIVATE_KEY`" — two knobs a compose operator cannot use, because
`compose.yml` derives `_FILE` from `${RZ_TEAM_GITHUB_APP_PRIVATE_KEY_PATH:+…}` and never forwards
the inline variable at all. It was fixed there. **Two sibling strings were out of that lane's
fence and still say it.** EXECUTED on `d582bc75`:

```
$ grep -n "RZ_TEAM_GITHUB_APP_PRIVATE_KEY\b" packages/team/src/auth/*.ts
github-app.ts:84:  … the GitHub App installation token could not be minted — check
                   RZ_TEAM_GITHUB_APP_PRIVATE_KEY and RZ_TEAM_GITHUB_INSTALLATION_ID: …
signin.ts:134:     this team server cannot check organisation membership … Set
                   RZ_TEAM_GITHUB_ORG, RZ_TEAM_GITHUB_APP_ID, RZ_TEAM_GITHUB_INSTALLATION_ID
                   and RZ_TEAM_GITHUB_APP_PRIVATE_KEY from the GitHub App and restart.
```

**These are not comments. They are refusal strings the operator reads**, at exactly the moment
sign-in is failing and they are trying to work out which of six credentials is wrong. Under docker
the last name in each is the one knob that will not help them, and following it wastes the attempt.

This is **ruling 12's** shape — the instrument's own words made false by a commit — one lane late.

## Definition of done
- Both strings name **`RZ_TEAM_GITHUB_APP_PRIVATE_KEY_PATH`** as the docker knob, and say plainly
  that the inline variable is for a non-docker host. `deploy/report.ts`'s `formatKeyFaultAdvice`
  is the wording precedent that landed in #538 — match its voice, do not invent a second one.
- `signin.ts:134` says "and restart", which is wrong for compose: the container must be
  **recreated** (`docker compose up -d`), not restarted, because a restart does not re-read `.env`.
  The runbook already records that trap for the ingest key. Fix it in the same commit.
- **The tests derive the knob names from source rather than restating them**, the way #538's
  `report.test.ts` derives from `compose.yml` — so if the knob is renamed again these reddens
  instead of rotting. If that is impossible from this package, say so on the issue and pin the
  literal, stating why.
- Mutation, EXECUTED and recorded: put either old string back and show the test go red. An
  assertion that survives its own mutation is vacuous.
- No secret may appear in either string. Both sit near credential values; assert against the real
  fixture secret, never a placeholder.

## Falsifier
If `packages/team/src/auth/` genuinely cannot know which knob the deployment uses — it is transport
and deployment agnostic by ruling 14 — **say so on this issue rather than hard-coding a compose
assumption into the auth module.** The honest fix might be to name no knob and defer to the boot
report, which does know. That is a real possibility and a better answer than a lie about `_PATH`.

### LINKED ISSUE #545: prd51 w9: the changelog records that a team server can now be configured for sign-in
## Fence (may touch ONLY)
- `CHANGELOG.md`

## Blocked by
Nothing. #515 landed in PR #538, merged `d582bc75`.

## The gap

`AGENTS.md` requires an `[Unreleased]` entry for every user-visible change. #515 made the team
server's sign-in reachable — a deployment gains six configuration values, a mounted key file, and
two routes that stop returning 503 — and shipped **no entry**. EXECUTED on `d582bc75`:

```
$ grep -n "515\|GitHub App" CHANGELOG.md
(no matches)
```

It was not an oversight in the build. `CHANGELOG.md` was outside #515's declared boundary, the
plan said so, and the verify pass recorded it as owed (`W4`). The PR body named it rather than
skipping it quietly. This issue is that debt, filed rather than left in a comment thread.

## Definition of done
- An `[Unreleased]` entry in Keep-a-Changelog style describing what an **operator** gains, not what
  the diff did. The audience is someone deciding whether to upgrade, not someone reviewing code.
- It names the deployment-visible facts: the six `RZ_TEAM_GITHUB_*` values, that the private key
  arrives as a **read-only mounted file** via `RZ_TEAM_GITHUB_APP_PRIVATE_KEY_PATH`, and that the
  boot report names every effective value with secrets redacted.
- It says plainly that **an unconfigured deployment still boots and still refuses sign-in with
  503** — that is a deliberate behaviour (#169), not a defect, and an operator upgrading without
  credentials needs to know their server will not break.
- It does **not** claim there is a viewer. Three routes exist; the three questions are a later
  wave. A changelog that implies a working team view would be the same class of error this
  programme keeps catching in its own prose.
- **No credential, and no real hostname**, in the entry or its examples.
- Every backticked repo-rooted path resolves, and no `#NNN` citation exceeds the live ceiling —
  `CHANGELOG.md` is swept by `doc-citation-law` and by the citation-ceiling law, which is why
  #515's own build had to drop a `(#515)` self-citation from a bullet. Cite the PR, `#538`, which
  has merged, or cite nothing.

## Falsifier
If a `[Released]` section has been cut since #538 merged, the entry belongs under the release that
contains it rather than under `[Unreleased]` — **check before writing, and say which you chose.**

### LINKED ISSUE #550: prd51 w9: an unverified key is refused before the body is parsed, as the handler already documents
## Fence (may touch ONLY)
- `packages/team/src/api/http.ts`
- `packages/team/src/api/api.test.ts`
- `packages/team/src/ingest/handle.ts` (its docblock only — the handler's logic is correct)
- `packages/team/src/keys/verify.ts` — **comments only.** Its "Unreachable through the route"
  comment over the empty-value branch (around line 125) goes false in this commit, because the
  route now calls the handler on the refusal path. Ruling 12 wants the words fixed by the commit
  that falsifies them. No logic may change here.

## Blocked by
Nothing.

## The transport inverts the refusal order the handler documents

`ingest/handle.ts`'s docblock states the contract:

> 1. no header at all -> 401, naming the header.
> 2. `checkKey()` -> `malformed` 401, `unknown` 401, `revoked` 403. The ONE call.
> 3. `parseIngestRequest` -> 400, carrying core's own message verbatim.

`api/http.ts`'s `serveIngest` does not honour it. It reads the body and runs `JSON.parse`
**before** calling `handleIngest`, so a syntactically bad body answers 400 and the key refusal
never happens.

**EXECUTED against a live deployment** (2026-09-15, first real-host boot; host redacted):

| request | status |
|---|---|
| no key + valid JSON | **401** — the documented order holds |
| no key + **invalid** JSON | **400** — `request body is not JSON: …` |
| bogus `rzk_` key + invalid JSON | **400** — same |

The handler is right; the adapter short-circuits it.

## Why it matters, stated without inflation

**No key is leaked and nothing is bypassed.** A valid batch still needs a valid key. Two real
consequences:

- **An unauthenticated caller reaches the JSON parser**, on up to `MAX_BODY_BYTES` (16 MiB), on the
  one route this server exposes to the internet. Earlier still, `resolveIngestKeyCheck` performs a
  **storage read** before the body is even read — so an anonymous request costs a database round
  trip regardless of outcome.
- **The documented contract is false**, and it is false in the direction that matters: a reader of
  `handle.ts` concludes the key is checked first, and it usually is — just not for the case an
  attacker controls most easily.

## Why no test caught it
`handleIngest` is tested with **already-parsed values**, so its ordering is true of the pure
function and untested end-to-end. The adapter's own tests cover the decode failure and the key
refusals separately, never their interaction. That is the sibling-case shape this repo keeps
finding: each half correct, the seam between them unexercised.

## Definition of done
- The key refusals precede body parsing. A request with no header, or a header that fails
  `checkKey`, is refused **before** `JSON.parse` runs — whatever the body contains.
- The existing statuses and refusal strings are **unchanged**; this is an ordering fix, not a
  rewording. `api/api.test.ts`'s existing cases must pass unmodified.
- The 413 body-cap refusal keeps its current position relative to the key check, or moves
  deliberately with the reasoning recorded — an oversized body from an unauthenticated caller is
  the same question one size larger, and whichever way it is ruled should be ruled rather than
  inherited.
- `handle.ts`'s docblock is corrected in the same commit if the fix changes what it should say
  (ruling 12). If the fix makes the docblock true as written, say so explicitly rather than
  silently leaving it.
- **The end-to-end matrix above is a test**, at the adapter level: all four combinations of
  {no key, bad key} × {valid JSON, invalid JSON}. That matrix is what was missing.
- Mutation, EXECUTED and recorded: restore the parse-first order and show the new tests go red.

## Falsifier
If parsing genuinely must precede the key check for a reason this issue has not seen — some
framing where `presented` cannot be resolved without the body — **say so on the issue rather than
forcing the order.** The fallback is then to document the real order in `handle.ts` and keep the
matrix test, so the contract is true even if it is not the preferred one.

### LINKED ISSUE #551: prd51 w9: a first boot's report is legible, and an expected Postgres notice does not cost eight lines
## Fence (may touch ONLY)
- `packages/team/src/storage/driver.ts`
- `packages/team/src/storage/driver.test.ts` (new — the module has no test today)

## Blocked by
Nothing.

## A first boot buries its own report under Postgres notices

`openSql` passes **no options** to the driver:

```ts
export function openSql(databaseUrl: string): SqlLike {
  return postgres(databaseUrl) as unknown as SqlLike
}
```

so the library's default notice handler prints every Postgres `NOTICE` as a multi-line object.
Measured on the first real-host boot (2026-09-15; host redacted), where the migrations legitimately
raise a dozen of them — `DROP POLICY IF EXISTS` in `0003_roles_rls`, `CREATE TABLE IF NOT EXISTS`
in the partition top-up. Each one costs eight lines:

```
app-1  | {
app-1  |   severity_local: 'NOTICE',
app-1  |   severity: 'NOTICE',
app-1  |   code: '00000',
app-1  |   message: 'policy "events_by_project" for relation "events" does not exist, skipping',
app-1  |   file: 'dropcmds.c',
app-1  |   line: '523',
app-1  |   routine: 'does_not_exist_skipping'
app-1  | }
```

**The two lines an operator actually needs on a first boot** — `migrations: 5 applied (…)` and
`ingest key: seeded for project …` — sit between roughly a hundred lines of this. Reading the boot
report on the host required a `grep -v` over six field names to become legible.

This is a small defect with a specific cost: prd-51's boot report exists so an operator can see
what the server resolved, and #515 went to some trouble to make it say who set each value and
where. Burying it under expected notices undoes part of what that bought.

## Definition of done
- An expected notice does not cost eight lines. Either suppress notices whose `severity` is
  `NOTICE` and whose `code` is `00000`, or render each as **one line**. The builder chooses and
  records why.
- **Nothing is silently swallowed.** A `WARNING` or anything above it still reaches the operator,
  and in full — a driver that hides a real warning to tidy a log is a worse defect than the one
  being fixed. Whatever filter ships must be justified against that, not merely implemented.
- `migrations: … applied` and `ingest key: seeded …` are legible on a first boot **without
  filtering**. That is the acceptance test, and it is about the whole log rather than one line.
- `driver.ts` has a test — it has none today, which is why an options-less `postgres()` call was
  never noticed. At minimum: the notice handler is installed, an expected NOTICE is condensed or
  dropped, and a WARNING survives intact.
- Mutation, EXECUTED and recorded: remove the notice handler and show the test go red. An
  assertion that passes with the default handler in place is testing nothing.

## Falsifier
If the driver cannot distinguish an expected migration notice from one worth showing — and
`IF NOT EXISTS` notices are indistinguishable from any other `00000` — then **say so and prefer
one-line rendering to suppression.** Quieter is not the goal; legible is. A log that hides
something an operator needed at 3 a.m. has failed in the more expensive direction.

ciaran-slow:
Closes #543
Closes #545
Closes #550
Closes #551

prd-51 wave 9 — **what the server says when it refuses**. Four lanes, four commits, base `main` at
`965db20e`.

Three of these four were found by **deploying the server to a real host**, not by a test suite: a
first boot buried its own report under a hundred lines of Postgres notices, two refusal strings
named a knob compose cannot read, and an unverified caller reached the JSON parser. #544 (the key
guard) left this wave earlier and landed alone in PR #552.

**Cherry-picked with no conflict, which is the wave's own fence audit** — #543's four files,
#550's four, #551's two and #545's one are disjoint by construction.

---

Every lane was verified on a **different model from its builder**, and **three of the four came back
with real findings after a ready-or-near-ready first pass.**

|---|---|---|---|---|---|
| `2d8b0c9c` | #545 | Sonnet 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/545#issuecomment-5675795845) | [report](https://github.com/launchpad-26/rhizomorph/issues/545#issuecomment-5675863132) | [**ready** — Fable](https://github.com/launchpad-26/rhizomorph/issues/545#issuecomment-5676434007) |
| `d316439e` | #551 | Sonnet 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/551#issuecomment-5675810631) | [report](https://github.com/launchpad-26/rhizomorph/issues/551#issuecomment-5675909124) · [fix](https://github.com/launchpad-26/rhizomorph/issues/551#issuecomment-5676086728) | [**ready** — Opus](https://github.com/launchpad-26/rhizomorph/issues/551#issuecomment-5676025733) → [delta](https://github.com/launchpad-26/rhizomorph/issues/551#issuecomment-5676238484) |
| `3caca209` | #543 | Sonnet 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/543#issuecomment-5675837290) | [report](https://github.com/launchpad-26/rhizomorph/issues/543#issuecomment-5675990285) · [fix](https://github.com/launchpad-26/rhizomorph/issues/543#issuecomment-5676271607) | [**changes** — Opus](https://github.com/launchpad-26/rhizomorph/issues/543#issuecomment-5676154599) → [**ready**, delta](https://github.com/launchpad-26/rhizomorph/issues/543#issuecomment-5676453155) |
| `9efd34dc` | #550 | Opus 5 | [plan](https://github.com/launchpad-26/rhizomorph/issues/550#issuecomment-5675823869) | [report](https://github.com/launchpad-26/rhizomorph/issues/550#issuecomment-5675955116) · [fix](https://github.com/launchpad-26/rhizomorph/issues/550#issuecomment-5676427491) | [**changes** — Fable](https://github.com/launchpad-26/rhizomorph/issues/550#issuecomment-5676223078) → [**ready**, delta](https://github.com/launchpad-26/rhizomorph/issues/550#issuecomment-5676477387) |

**The verify records name the LANE shas, not the shas here.** Cherry-picking rewrites them and
condition 1 is per-sha, so the gap is closed by content — `git patch-id --stable` is identical for
all four pairs:

|---|---|---|---|
| #545 | `d8c42577` | `2d8b0c9c` | `64bbc95b0108` |
| #551 | `8dafb0c4` | `d316439e` | `ee6c3acb7f01` |
| #543 | `d3c0b327` | `3caca209` | `c55dc837647c` |
| #550 | `f04f9ee1` | `9efd34dc` | `9a8c2a5c97cf` |

## What is the sibling case?

- **#543 — the sibling of a wrong string is the same string one clause over.** The first version
  asserted a *set* of knob names plus one ordering, so demoting `_PATH` into the "Outside docker"
  clause and promoting `_FILE` into the docker one left the suite **green** — re-introducing the
  exact defect the issue exists to kill. Closed by binding each name to its clause. The verify pass
  proved the binding is load-bearing by **failure line number**, since both mutations preserve the
  same names in the same order: MX-C reddens at `signin.test.ts:565`, MX-E at `:568`, both new in
  the fix, while the old set check at `:535` passes.
- **#543's second sibling — the string the issue body named was not the one in production.**
  `handleSignInStart` refuses only with `unconfigured`, so `GET /auth/github/start` returns
  `signin.ts:122`, not the `:134` the issue cites. The live capture taken from a real host was of
  `:122`. Fixed in the same commit.
- **#550 — the sibling of "the handler orders its refusals correctly" is "the adapter asks it in
  time."** `handleIngest` is tested with already-parsed values, so its ordering was true of the pure
  function and untested end to end. The missing thing was a four-way matrix at the adapter level.
- **#551 — the sibling of a NOTICE is everything that is not one.** A fix that condensed NOTICEs
  and quietly dropped a WARNING would be a worse defect than the noise it replaced.

- **#543**: eight mutations plus three invented. The decisive one is **M7** — the naive leak
  assertion `not.toContain(TEST_PRIVATE_KEY)` **cannot fail**, because the wire body is
  `JSON.stringify(...)` and the fixture's real newlines become `\n`. The verify pass proved this
  rather than accepting it: **MX-H** leaked a real full PEM into the wire and the naive form came
  back **GREEN**, while **MX-B** (one base64 line, no `-----BEGIN`, no marker) went **RED** against
  the newline-free fragment. **MX-I** showed the hole was live on `main`, not theoretical.
- **#550**: four mutations. **M-2** — gate between `readBody` and `JSON.parse` instead of before
  the read — reddens **M-i alone**, and the verify pass reproduced its byte count to the digit:
  `bytesWritten: 16778610, bytesRead: 0`. A keyless caller under the rejected design receives
  **nothing**, not a 413. Deleting M-i alone leaves the suite **34/34 green** under M-2, so one test
  is what separates two designs differing by three lines.
- **#551**: five mutations. **M3** (gate on `severity_local` rather than `severity`) matters beyond
  this lane — `severity_local` is *translated* by a non-English-locale server, so the filter would
  have silently stopped working abroad. **M4** (a dedup/throttle set) is the one a reviewer would
  wave through as an improvement, and it is what makes a repeated warning vanish.
- **#545**: **nothing mechanical checks this entry's truthfulness.** `CHANGELOG.md` is swept by the
  citation-ceiling law only — the path-existence and symbol laws are scoped to `docs/*.md` and
  `packages/*.ts`. Its six claims were checked by reading them against the code, five EXECUTED and
  one REASONED and labelled so.
- **The ones that still survive**, named rather than hidden: #543's `signin.test.ts:567` is not
  individually killable (set equality forces `_FILE` to exist, so `:566` always fires first —
  belt-and-braces, not coverage); and two #551 mutations survived as safe-direction widenings.

## Two records that were false, and are not any more

Both were caught by verification, not by a suite.

- **#551's mutation counts were inflated.** `mockRestore()` sat inline at the end of each `it`
  instead of in an `afterEach`, so a failing assertion leaked a spy and later tests reported
  *phantom* failures. Mutation 3's honest score is **2/7, not 6/7**. Fixed; the corrected number is
  smaller and is the true one.
- **#550 claimed an over-cap caller receives a 413. It does not.** `readBody` destroys the socket
  before `send(413)` runs — `fetch` sees `UND_ERR_SOCKET` with `bytesRead: 0`, raw `node:http` gets
  `ECONNRESET`, no status line. The docblock, the test comment and the commit message all said
  otherwise, and the existing over-cap test reads as a delivered 413 **only through its `catch`
  fallback**. Corrected in all three places — and the builder found a **fourth** instance in the
  same paragraph that the verify pass had not listed.

## Fence widening — one, recorded before the change

`packages/team/src/keys/verify.ts` on #550, **comments only**: its "Unreachable through the route"
note over the empty-value branch goes false in this commit, because the route now calls the handler
on the refusal path (ruling 12). Recorded on the issue at 06:31Z; the commit is 06:41Z.

## Gates — on the assembled wave

The first run in which all four commits coexist. #543 and #550 both touch `packages/team/src/`, and
#550's new matrix exercises the route whose refusal strings #543 rewrote:

- `VITEST_MAX_WORKERS=6 npm test` — **520 files, 9916 passed**, 3 expected-fail, 7 skipped, exit 0
- `npm run typecheck` — clean
- `npm run lint` — clean, 1119 files
- `npm run build` — clean
- boot smoke — `GET /` → **200**, `GET /api/meta` → **401**

Node v22.22.2, darwin/arm64, run with no sibling suite live. **One platform's verdict, not "CI
passed"** — all four workflows are disabled on this repo. `scripts/gate.sh` was **not** run:
running it *is* the landing, and that is the operator's step.

## Known imprecisions and owed follow-ups

Left in deliberately rather than reopening verified lanes:

- **#545's entry says the key is mounted "at" the path named by `_PATH`.** Strictly it is "from" —
  `_PATH` is the host mount source and the container reads `/run/secrets/…`. The operator action
  implied is exactly right, so this is not the wrong-knob failure; recorded rather than fixed.
- **#545's "never passed as an environment value"** is true of the compose deployment the sentence
  is scoped to; `config.ts` still accepts the inline form for a non-docker host.

Owed their own issues, not filed here:

- **The 413 is undeliverable at all**, pre-existing, plus two tests that over-claim about it — one
  of which never exercises an over-cap body despite its title.
- **`doc-citation-law`'s symbol check has no way to say "this is a word, not a symbol."** Five
  separate lanes tripped it today writing ordinary documentation — `Pending`,
  `RZ_TEAM_SESSION_SECRET`, `noUnsafeFinally`, a generated `.env` path, and `UND_ERR_SOCKET`. The
  law is right to check; the affordance is missing.


KelliherL (approved):
## Review of #556 — approve; nine mutations, every one red, and the two claims I most wanted to falsify both re-derive exactly

Three of these four lanes were found by deploying to a real host rather than by a suite, and the one lane with no mechanical check is the one I read hardest. Read commit-by-commit on `9efd34dc` in a clean clone at node 22.23.2 on Linux. **I am pushing no fix: I did not find anything to fix.**

### The two re-derivations that decide #550

**M-2 reddens M-i alone.** I moved the key gate from above `readBody` down between `readBody` and the decode — the design you considered and rejected — and ran the whole team package:

| what I ran | result |
|---|---|
| gate removed entirely (pre-#550) | **7 red**: M-b, M-d, M-e, M-f, M-g, M-h, M-i |
| gate moved between `readBody` and the decode (M-2) | **1 red: M-i, and nothing else** |
| M-2 **and** M-i deleted | **exit 0 — 559 passed, 27 files** |

So your sentence *"one test is what separates two designs differing by three lines"* is exactly right, and I would rather have measured it than believed it. It also means M-i is load-bearing in a way the name does not advertise; anyone who ever finds it slow or awkward is deleting the only thing that distinguishes a delivered 401 from a silent socket reset.

**The naive leak assertion really cannot fail.** I checked M7's mechanism directly rather than through the suite:

```
wire contains the whole PEM:      false
wire contains one base64 line:    true
```

`JSON.stringify` turns the PEM's real newlines into `\n`, so `not.toContain(TEST_PRIVATE_KEY)` is unfalsifiable against a wire body by construction — it would have passed while leaking the whole key. `KEY_FRAGMENT`, one newline-free base64 line, is the form that can fire. That is the best finding in this PR and it came from the verify pass, not the build.

### The rest of the battery — all red

| mutation | lane | result |
|---|---|---|
| demote `_PATH` into the "Outside docker" clause, promote `_FILE` into the docker one | #543 | red |
| **fold the operator's sentence onto the wire** — `detail: { cause: membership.error }` beside the `operatorNote` | #543 | red — **C5** |
| gate `reportNotice` on `severity_local` rather than `severity` | #551 | red ×2 |
| fail **closed** — condense a WARNING too | #551 | red ×3 |
| **drop** a NOTICE instead of condensing it | #551 | red ×2 |

The second is mine. It is the seam added at the review of #476 for the ingest 503 and generalised at #507, and I wanted to know whether this wave could quietly undo it while making the refusals more helpful — the obvious way this commit could go wrong, since its whole purpose is to say *more* to the operator. It cannot: C5 pins that a mint failure carries an `operatorNote` and is never folded into the refusal body. The knob names that did move onto the wire are variable *names* and `deploy/.env`, never values, and that was already true before this commit.

The fail-open direction is pinned too, which is the part I expected to find missing: *"a notice with no severity field at all also survives in full"* reddens when the filter is made to condense everything. An unclassifiable notice going to the operator in full is the right default and it is asserted.

**The new call site cannot throw.** `serveIngest` now calls the thunk outside the `try` that wraps `resolveIngestKeyCheck`. I checked whether that is exposure: it is not. Every fallible step — the `trim`, the shape check, the one `findIngestKey` — happens inside the awaited resolver; the returned closure is `() => verdictFor(row)` over a row already in hand. There is no I/O behind the second call, which is the same property ruling 8's one-read bound already rests on.

### #545 — the lane nothing checks, read by hand

You say plainly that `CHANGELOG.md` is swept by the citation-ceiling law only, so I treated its claims as unverified and checked the ones that are checkable:

- **"prd-51 ruling 9"** — correct. Ruling 9 is *every value names its setter*, cited that way in `config/config.ts`, `config/names-its-setter-law.test.ts` and `index.ts`. I went looking for a wrong ruling number here because another open PR cites ruling 9 for retention, and this citation is the right one.
- **"six `RZ_TEAM_GITHUB_*` values"** — `compose.yml` names seven and `config.ts` knows seven, but the seventh in each case is the derived `_FILE` and the inline `_APP_PRIVATE_KEY` respectively. Six is the set an operator actually sets in `deploy/.env`, which is what the sentence is about.
- **"still answer 503"** — `signin.ts:120-122`, both `unconfigured` and `membership-unconfigured`.
- **"both secrets shown redacted rather than in the clear"** — backed after all, just not by anything that reads the
  changelog: `config.test.ts` case 37 pins `<unset>` and `<redacted>` for both fields, and asserts separately that the
  literal never survives into `display`.
- **The runbook link** resolves; `docs/team-server-runbook.md` is 308 lines.

The two imprecisions you record — *"mounted at"* where it is strictly *from*, and *"never passed as an environment value"* being true only of the compose deployment — are both real, both named, and neither is the wrong-knob failure the issue exists to kill. Recording them beats reopening a verified lane.

- **The 413 is undeliverable and two tests over-claim about it.** You have this as an owed issue and I agree it is not this PR's job. Worth filing before the wave lands rather than after, because the existing over-cap test reading as a delivered 413 *through its `catch`* is the kind of thing that gets cited as evidence by the next person.
- **`doc-citation-law` has no way to say "this is a word, not a symbol"** — five lanes tripped it in one day. That is an affordance gap with a real cost and it deserves the issue you say it is owed.

### What I did not run

macOS, Windows, the min-node leg. No real Postgres and no real GitHub App, so #551's NOTICE behaviour is exercised against the filter rather than against a server that sends one, and the sign-in paths are stubbed at `fetch`. **The patch-id table I could not check at all** — none of `d8c42577`, `8dafb0c4`, `d3c0b327` or `f04f9ee1` is reachable on this remote, so all four equalities rest on your record rather than on anything I ran. `origin/main` had not moved from this branch's base, so no merge was needed and the leg below is the union:

```
  Test Files  520 passed (520)
```

**Approving.** I tried to break the ordering claim, the leak assertion and the operator/wire boundary, and all three held under mutation. One platform, not CI — the workflows are retired, so a local leg is the whole of the evidence, and the absence of a green bar here is what you say it is. Landing is the operator's act.

